PR #43, opened 2011-06-23
Please review and pull

s0be:
My only concern is in fc94f90155b2ca88d07a, lines 127, 128. I couldn't figure out a way to hit this branch and test if (Ship *) NULL works as expected.


onpon4:
Closing as this seems to be irrelevant to the current code base.

PR #51, opened 2011-11-20
Targeting Upgrade

MPink:
Defines an accurate linear targeting function and then implements it in multiple places.
aim_turret,seeker_think& ai_aim
If it ever needs updating again it will save lots of effort tracking down the code locations.
The function can also be used to find ship to ship intercept angles but for the moment is only used with weapons.

Fixed loads of launcher related problems.
The turret_ammo type is now pointless as all checks are done on the spawning outfit.
launchers could use a swivel setting but for the moment i use arc.
I done this due to none lockon rockets needing some help in there aiming.

Implemented an angular targeting system and linked it to the targeting array upgrade.
Although this is far from perfect it dose provide a starting place to increase from.
Iv got thoughts on a nerb tracking that would make this look poo.

### DIFF
--- a/dat/outfit.xml
+++ b/dat/outfit.xml
@@ -1774,7 +1774,7 @@
  <outfit name="Unicorp Caesar IV Launcher">
   <general>
    <slot>weapon</slot>
-   <size>medium</size>
+   <size>heavy</size>
    <license>Heavy Weapon License</license>
    <mass>30</mass>
    <price>120000</price>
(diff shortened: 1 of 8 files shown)


onpon4:
This PR has been untouched for 9 years now, longer than I've been a contributor! 😅 I don't know if this is applicable to the current Naev source code (any input on that front, @bobbens?), but if so merge conflicts would need to be fixed (and considering how long it's been ideally we would check to make sure it actually works).

🕷️

onpon4:
Closing this as the current Naev source code seems to have drifted far away from this for it to be usable.

🕵️

PR #94, opened 2013-04-23
Economy

BariumBlue:
Hail!

The way it works is that every system's price is the weighted average of it's neighbor's and it's own preferred price (if it has one).

So, if you have system A, B, and C, where A has a jump to B has a jump to C, and A has a preferred price of 2, B doesn't prefer any price, and C has a preferred price of 3, here's what you get:
http://imgur.com/lRwEgxr
(where "g" is the preferred price, and "r" is the real price)

How much a system likes it's own values is it's weight; a weight of 0 means that it's real price is simply the average of it's neighbors. A system with a weight of inf prefers it's own values infinitely more than it's neighbors, and has a real price basically equivalent to it's own  preferred price.

This setup means that if you give values to two systems, all the systems in between will have a steady gradient of price, from the price in the first system (for systems close to it), to the price in the second system (for systems close to that one), making it convenient to handle for the most part.

The preferred price is set in the system xml, and the commodity base price is simply a price multiplier. If you want prices to be higher throughout the galaxy, you can simply change the base price.

I have a version in python and a replica in C if you want to examine the setup more closely.

---

Also, there's dat/events/neutral/econ_events.lua, which creates economic events that change the prices for a short period of time (and are highly lucrative!), and appear in the news.

I haven't found any memleaks, i've checked and everything seems be good.


bobbens:
Alternate economy system has been merged.

PR #100, opened 2013-06-16
[Mission] Ship-Shipment

w3stley:
Here is my Ship-Shipment-mission.

### DIFF
--- a/dat/commodity.xml
+++ b/dat/commodity.xml
@@ -51,5 +51,11 @@
  <commodity name="Weapons"/>
  <commodity name="Drugs"/>
  <commodity name="Exotic Animals"/>
- <commodity name="Radioactive Materials"/>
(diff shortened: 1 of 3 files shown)


onpon4:
I'm not sure what I think about the idea, as it's implemented; it seems that it picks an amount of cargo between 25 and 50 tons and calls that the "ship". But the thing is... that's *extremely* light. Even really lightweight ships are usually over 100 tons.

Besides, it's a missed opportunity: wouldn't it be a lot more interesting for the mission to make you fly a randomly generated ship, which might not handle the way you're used to, than to just make it a different type of cargo in the shipment mission?

bobbens:
Yes, I like the idea, but I think the implementation does not exploit it's potential. I believe I mentioned that on IRC.

bobbens:
Given that ships no longer need to be shipped, I'm closing.

PR #136, opened 2014-07-24
FLF Improvements

onpon4:
This improves the FLF in a few ways:

- The FLF's collor is yellow (same as the Frontier), not orange (same as Soromid).
- The FLF has a basic outfitter, shipyard, and mission computer. The old FLF base was basically useless; all it had was a bar. The outfitter can't supply very much, however.
- The pre02 mission has been rewritten. Instead of going with a group to combat three Dvaered patrols, you go on your own to combat one small Dvaered patrol, and are rescued by several FLF ships when it turns out there are more ships than you can handle. This new mission doesn't require any particular ship type, but actually is best done with light, fast ships simply because it's not possible to survive without avoiding enemy gunfire.
- A new generic mission, eliminating a Dvaered patrol, is unlocked after beating pre02. It's basically the same as pre02, but with varying difficulty and no dramatic rescue. Harder patrols include FLF escorts which sometimes arrive late.
- Hidden jump points for the FLF have been added, making travel from FLF space to Frontier space and back much more practical, and also making entry into and exit from Dvaered space much more practical. Every system with FLF presence can now be reached in 3 jumps or less (usually  no more than 2 jumps).
- The FLF and the Empire are now enemies. It made no sense for them to be peaceful with each other; the player, allying with the FLF, can't be peaceful with the Empire, after all.


onpon4:
I haven't gotten any comments on this. Any word on what's good, bad, and ugly about this pull request?

Zireael07:
I like it!

bobbens:
I would like to merge it but see my comment on #137 about dual licensing. Would it be possible to do gpl + apache?

onpon4:
Just in case you miss my comment on #137: whoever told you that the Apache License is incompatible with version *3* of the GNU GPL is incorrect. See: https://gnu.org/licenses/license-list.html#apache2

bobbens:
Changed it so FLF and Empire aren't explicit enemies. They shouldn't care about each other, only when Dvaered is fighting FLF will empire sort of feel an "obligation" to help, but otherwise Empire doesn't want Dvaered to get too cocky and prefers them to be distracted with the FLF. Thanks for the patch.

PR #137, opened 2014-07-24
Genericized Bounty Mission

onpon4:
These commits repurpose the Empire Pirate Bounty mission into a generic Pirate Bounty mission. The new mission doesn't improve standing with any faction, but pays very well. The missions are much harder to do than cargo missions, however, and the bigger ones require either a very good ship or very refined skills.


onpon4:
Can I get some feedback on how this is with the new Mercenary License, please?

bobbens:
OK, looked over it. It looks good. Personally I am more of a fan of not being able to buy it directly, however, since you won't be able to get good equipment unless you mission anyway, it might not be a problem.

One thing, I'm told by people that because naev itself is gplv3 it's a conflict with apache license or something like that. Would it be possible for you to dual license gplv3 / apache? Would be so for the other stuff I believe I pulled already.

onpon4:
You've been misinformed about the Apache License. It's incompatible with version *2* of the GNU GPL. Version 3 gained compatibility with this license.

onpon4:
See: https://gnu.org/licenses/license-list.html#apache2

bobbens:
OK, pulled thanks.

Deiz:
My suspicion is that it's not possible to write missions, etc. for Naev that aren't GPLv3, since both the Lua API and the various utility scripts are all GPLv3 (as is the standard mission boilerplate, but that's probably ineligible for copyright).

I really hate dealing with licensing, as it's all murky at best, which is why I'd prefer contributed code to be GPLv3 or dual-licensed.

onpon4:
> My suspicion is that it's not possible to write missions, etc. for Naev that aren't GPLv3

No, that's not quite correct. The *whole* must be under the GPLv3. Permissive licenses are fine as long as they don't add any incompatible restrictions, because permissive licenses don't prohibit adding additional restrictions (e.g. relicensing under the GNU GPL). Dual-licensing in this case is in fact completely superfluous.

PR #144, opened 2014-08-01
Changed energy regeneration to simple linear.

onpon4:
I don't know if the way that was used before was more realistic than
simple linear regeneration, but regardless, it's unintuitive,
game-breaking, and much more complicated than the system needs to
be. I've removed all references to "energy_tau", and energy_regen
just gets added directly to the battery in a linear fashion now.


onpon4:
This isn't even a part of my master branch anymore, so I'm closing this.

PR #150, opened 2014-08-05
Economy System (based on BariumBlue's)

onpon4:
This is an updated #94. I merged BariumBlue's economy branch, resolved the conflicts, and reverted certain changes (namely, some changes from British English to American English in various places, and the change to commodity_renderMod). It seems to work perfectly; I don't understand why it wasn't merged when #94 was opened.


onpon4:
Say, bobbens... where did you even see medicine being worth 50,000 credits? Everywhere I look, it's around 5,000, which is more than what it currently is in Naev, but nowhere near as ridiculously high as you made it out to be...

ids1024:
@onpon4 The discussion about the economy seems to have been on IRC.

http://colabti.org/irclogger//irclogger_log_search/naev?search=economy&action=search&text=checked

bobbens:
Semper is 51k for medicine, 9k for food. Seeing this all over. It is consistently like 10x over what is expected and over the set values. I have changed some whitespace, fixed some bugs, merged it and also fixed a few memory leaks, but I don't think it's ready. My main concern is we have virtually no control over the prices, would have to look at it into more detail to see what it is going on, and at the moment not sure I'll be able to try to solve it.

If you would make it easier to control the prices so it actually respects the prices set in the xml to some extent (example is tau prime which is no where what it says in the xml) I think it could be merged, but atm it's not ready.

You should pull from my economy branch to get my changes.

onpon4:
After investigating this, I see where the confusion results:

This economy system combines two factors: one, the price of commodities. Two, the price of commodities in any given system. These two factors are *multiplied together*. Medicine's base price is 10, so yes, when a system's medicine price is 5,000, it will result in 50,000. Food's price is set to 100, so it would be 100x the amount indicated for the system! I'm not quite sure why BariumBlue included this feature.

So the solution is simply to reduce the price of commodities (in dat/commodity.xml), or reduce the prices indicated for systems to get the desired result with the commodity prices. Making it less confusing and more direct is simply a matter of changing the base values to 1.

onpon4:
Thinking about it a bit, I suppose the idea was probably to determine what kinds of prices each commodity could have so you could use the same system prices for all commodities. Or if not, it could be used that way. For example, the price could be set such that 100 is always an average amount for that commodity. Then if you decide to make one commodity more or less expensive, you just have to change one number, and not dozens of numbers.

bobbens:
http://colabti.org/irclogger/irclogger_log/naev?date=2015-02-08#l421

I agree with BTAxis that the current implementation leaves a bit desired. Instead of setting it manually having it automatically get set based on asset metadata seems fairly more complex, but it's more of a question of running some tests and finding some heuristics that give reasonable numbers. Ideally each commodity would be tuned with 4-5 parameters related to some higher level concept that would automatically set the prices around the universe. Would need a bit more thought.

BTAxis:
For what it's worth, I had a discussion with BariumBlue back in the day in which I explained my misgivings about his system. Log begins here:
http://colabti.org/irclogger/irclogger_log/naev?date=2013-04-23#l76

onpon4:
> Instead of setting it manually having it automatically get set based on asset metadata seems fairly more complex, but it's more of a question of running some tests and finding some heuristics that give reasonable numbers. Ideally each commodity would be tuned with 4-5 parameters related to some higher level concept that would automatically set the prices around the universe.

Personally, I don't see that as beneficial, and here's why: we still have to add something manually, because there simply isn't currently any data that can be used to infer what kinds of prices we should have. The only difference I can think of is it wouldn't be simple numbers, but instead a more complex set of factors to adjust; and, of course, the A.I. that determines prices from all these factors would have to be developed.

I really don't think it would make adjusting the economy any easier, much less make the game any more fun.

BTAxis:
In the short term, maybe. But it's the difference between digging a ditch, and building a machine that digs a ditch for you. Given proper planet metadata and a system that can effectively reason about it you can generate prices for however many commodities you like, instead of having to edit the *entire galaxy* every time you want to add a new one.

Additionally, planet metadata is going to be needed anyway for other reasons. As an example I've given The Wringer station in Suna, which is supposed to be a seedy place where you can find shady types, but right now the game has no way of knowing it shouldn't spawn ordinary NPCs there. You could dick around with factions to accomplish it, but that would be the wrong way to go about it.

PhoenixRiver:
TLDR - a code light, highly adaptable and extensible economy system
based on BriumBlue's economy and moving towards greater automation of
metadata creation with few actual additions or substractions to the
current file load.

I have been thinking about this ever since I came across BariumBlue's
economy model (one of the first things I looked at modding into the
game) and it is part of the reason that I decided that it was worth
going through the planet metadata and system metadata files and
cleaning them up*. I think that a primitive set of metadata already
exists - maybe not enough to work with an economy system but close if
we make a couple of assumptions (which are at least true for the
systems and planets I've worked through - maybe a third at present).

The assumptions - plus extra data that would need created:
(I should also note that I make implicit assumptions about goods
belonging to different classes, what is interesting is that these need
not be made explicit (though I'd suggest they are so as to make adding
new goods easier) - the classes are: basic, industrial, military and
luxury (basic is also split into biologicals, chemicals and
minerals).)

Planet Class is related to ability to produce basic goods
(biologicals, chemicals and minerals).

Planet colour and background image <GFX> is related to the
atmospheric/crust makeup of the planet and therefore which
biologicals/chemicals/minerals are possible (need a lookup table to
match these up), note: this assumption is only about half true for
planets I've not editted, many appear to have placeholder images which
are unrelated to the description.

Population determines the need for basic materials and the production
rate of the planet modulated by the technical automation that faction
is capable and the populations ability to work under difficult
conditions - though this is less important with high automation
(population we have, a datapoint for each faction as to its level of
automation is probably well know so easy enough to add in, as would
their survivability - the Soromid being the main winners here).

The need for industrial goods would be based on the level of
production of biologicals/chemicals/minerals (based on the data
above).

The need for military goods would be based on the total presence value
and range in a system and on the factions there; greater range means
more fuel (for example), value means better ships, faction moderates
the types of ships and level of militancy (this last would need to be
another new data point in for each faction but it realativly well
know, no? (Pirates high, Dvaered, FLF higher, Soromid lower,
Trader/Merchant lowest)).

The amount of trade in a system would depend on nebula activity and
interference (high -> less trade, low -> more trade), discrepancy
between production and consumption (high discrepancy means profits to
be made!), visibility of jump points and of planets - if you don't
know they're their how can you trade with them - trade ships are not
known for having great sensor suites either ;) and a danger value
based on the prescence of enemies and allies to the Trader faction in
the system moderated by the amount of military goods in the system
(more military probably means it's more dangerous and/or difficult to
get the relevant paperwork to trade here and so less trade) - (all of
this data either exists is created above)

The amount of trade impacts luxury goods - the more trade the
wealthier a system/planet is, the more luxury goods they want - if
they can produce enough then they will have higher industrial and
basic wants (so they can manufacture the luxuries) as well as much
higher luxury prices.


Almost everything necessary for this already exists or the system
itself generates the data necessary. What I like about this system is
that it needs little data, is based on how economies work (supply and
demand moderated by military and factional influence) and could be
extended in future to decide which _default_ services (in addition to
any described by the planet xml) based on the amount of trade and what
classes of goods are mostly traded. Based on the factions, their
automation, their militancy and the level of military goods traded you
could also create a _default_ set of techs in the system too - as well
as having throwing in more trade ships in systems with high amounts of
trade.


As an extension you could split trade into two different types - local
(system) and regional (multi-system) ("global" is probably so small,
i.e. potentially mostly just the player, as to make little
difference). Local trade would know the area and not have to worry
about knowing where things are so much and worry less about the danger
and local paperwork side of things because they are well known but
equally may have to make do with smaller profits (this would cover
things like a system having an agricultural world and a mineral world
exchanging their goods with each other and therefore reducing
potential high prices on both worlds since they supply each other
locally). Regional trade would worry more about danger, militancy and
prescence range but could expect higher profits (this is more like the
sort of trade the player would engage in).
It might even be possible to create a map on startup of which systems
traders would trade between because some systems are too dangerous to
cross for regional traders (e.g. due to wars between factions etc.)
which could be moderated by a metadata file containing "global"
factors - if this map generation could be triggered in-game it it
could also be used to modify the "global" balance based on player
actions (particularly when they get higher up certain factions and
have a larger impact or to make a change alongside changing planetary
metadata).

All of this is relatively code light and highly extensible but it will
take a lot of tweeking of small factors and rooting out discrepancies
in planetary and system metadata (certainly to the point for me to
bother creating tools to automate the collection of this data - I
already have a number of tools for that sort of thing due to trying to
get a handle of how much work is left with cleaning the planet and
system metadata (so many Class A worlds!!). If you think it's a good
idea I am happy to keep working on the metadata files and work on
creating some more planet images in case of gaps in what we have and
to give a bit more variety (particularly moons with atmospheres if I
remember rightly).

I hope this helps and apologies for such a long essay - I didn't know
where else to put it but it's been building up over many, many months
(almost years at this point!)

I would have done this already were this my job but I';m sure you all
know how the world works ;)

Take care and keep up the good work :)

River *<:@)

*(this has stalled due to things happening at this end but I've been
trying to keep up with developments in Naev in the hope that I might
have more time again at some point - which might actually be happening
now - hence this email.)

onpon4:
I've adjusted the prices to hopefully make them better. After a lot of experimentation trying to take advantage of the averaging system, I finally decided to just set the system prices based on faction. I could leave it this way, or I could go through all the planets and set the prices on that level as well. In any case, system and planet prices are now factors (1 for normal, 0.5 for half-price, etc), and the base costs of goods are as they were previously.

onpon4:
I'm abandoning this economy system. At this point, I don't think it's worth it. In fact, I'm starting to question the worth of having an economy system  at all.

However, what I *would* still like, is something that allows Lua scripts to adjust prices. In particular, I'm interested in creating something similar to Dopewars: just randomly increasing or decreasing prices when changing systems, sometimes extremely. I think the randomness of something like this would be much more interesting than a static variation across the galaxy, which would just encourage you to go back and forth between the same two systems over and over again.

For the record, I discovered a serious bug of some kind in this system that was making food and medicine much more expensive than they should have been in Dowue. So if we do adopt this system, that needs to be found and fixed. But again, I think the whole idea of an economy system should just be abandoned and replaced with simple Lua price adjustment capability.

onpon4:
By the way, #94 really should be closed. In case someone wants to salvage this, I'm leaving my "economy" branch open.

PR #152, opened 2014-08-07
Removed initiating autonav with left-click on planets.

onpon4:
If you're close enough to click on a planet, using Autonav to get
closer to it is basically useless. If anything, it can be confusing
for new players. I changed it to just hail or land; the message
informing you that you're too far away to land is more useful.


onpon4:
If you want to maintain the ability to click on the arrows to fly there so much, why not make the distance greater instead, then, so that you can be pretty sure the planet is out of view? That would be decent enough.

PR #155, opened 2014-08-10
Gave each file its own copyright/license notice.

onpon4:
This is the normal way to do things. I don't think trying to contain all of this information in one file is worth it.

I considered going through each file, looking through the commit histories, and seeing what all of the copyright holders are, but doing this for 274 files myself would be too much. I've decided to leave it up to each contributor to handle their own copyright notices, instead, except for a couple of files that didn't point you to naev.h previously, where I indicated the person responsible for starting the respective file.


ids1024:
> I considered going through each file, looking through the commit histories, and seeing what all of the copyright holders are, but doing this for 274 files myself would be too much.

Wouldn't it be possible to write a script to automate this?  It should not be to hard to extract the names of the people who have contributed to each file from the git history in python or some other language.

onpon4:
I suppose, but I don't know enough about Git to write such a script.

ids1024:
It could probably be done with a python script that runs `git blame` on each file and extracts the list of contributors from the output.  Give me a moment and I'll write up one to do this.

ids1024:
Here is my script: https://gist.github.com/ids1024/21085226493180c2edf6

It takes a single file as a parameter, although more could be automated.  Note that bobbens may be listed twice, because his earliest commits appear to not have an email and the crude script sees them as different.

onpon4:
Closing this as it's not in my fork anymore and it's not terribly important anyway.

PR #159, opened 2014-10-31
Minor fixes

BrknRobot:
Fixed some stuff I came across while playing.

### DIFF
--- a/src/player.c
+++ b/src/player.c
@@ -2170,6 +2170,9 @@ void player_destroyed (void)
    /* Stop sounds. */
    player_soundStop();
 
+   /* Stop autonav */
+   player_autonavEnd();
+
    /* Reset time compression when player dies. */
    pause_setSpeed( 1. );
 }


bobbens:
I'm not sure what's best actually. Reactivating is not too hard, plus hailing events are fairly rare and we want to encourage the player to actually try them. On the other hand introducing a delay before the speed up is not a bad idea.

nloewen:
It can be difficult to respond to the hail before going out of range when time compression reactivates immediately, and it is likely more frustrating to miss the hail even when you wanted to respond then to have to reactivate autonav. I'm okay with putting the slowdown back in, but I would consider a delay required.

jeidon:
Hey, if I can throw in my two cents.

What about re-enabling autonav not after a pause but after you are out of range? Is that too long a wait?

pydsigner:
How about not disabling autonav, but introducing time dialation until out of hail range?

bobbens:
I like @pydsigner idea of just disabling autonav until out of range, and having it reactivate when you toggle it. A bit of a pain to implement though.

PR #161, opened 2015-01-18
Fixed equipment_transportPrice

thunkk:
equipment_transportPrice returned huge prices for ships that were in the same system due to map_getJumpPath not generating a path for them.


bobbens:
Looks nice, thanks.

PR #162, opened 2015-02-06
Reduced incidence of autonav starting up when clicking planets (and planet indicators)

onpon4:
Changed the distance required for clicking on a planet. Now 5x the radius. This prevents the annoying behavior of causing you to zip past the planet when you're trying to use the mouse to land, without preventing you from starting autonav by clicking the indicator on the edge.

This is an alternative to #152, which deleted the behavior of engaging autonav this way entirely. This one just reduces it. Incidentally, it seems the reduction amount I chose is perfect; when zoomed out due to targeting a ship, it seems to be exactly the distance when the planet indicators on the edge show up.

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -1237,7 +1237,7 @@ static void input_clickevent( SDL_Event* event )
          if (planet_hasService(pnt, PLANET_SERVICE_LAND)) {
             if ((pnt->faction >= 0) && (areEnemies( player.p->faction, pnt->faction ) && !pnt->bribed))
                player_hailPlanet();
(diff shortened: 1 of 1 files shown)


Deiz:
I'd prefer to unify the two mouse input models. The normal one (used for on-screen, border, and radar) performs actions on double left-click, while the overlay's only action is autonav via right-click.

The overlay is handled separately because it'd be thoroughly annoying to accidentally hail a pilot (or similar) when trying to autonav, or vice versa. I'm inclined to remove left-click autonav, so that mouse autonav can only be initiated via right-click.

bobbens:
Using deiz's rehauled system that splits left and right click. Thanks.

onpon4:
Awesome, works great!

PR #163, opened 2015-02-06
Removed the "pause instead of abort" option.

onpon4:
This was clearly designed to use in conjunction with the old autonav
abortion behavior. player_autonavAbort is still used in some places,
but I don't think it's particularly useful to give the player the
option to pause instead of abort in these places.

Also includes a comment fix that I apparently forgot to merge previously.


bobbens:
Merged, thanks.

PR #165, opened 2015-02-08
Make escorts always visible to parent

ids1024:


bobbens:
Escorts are still half-done, we talked about trying to doing something fancier on IRC like shared visibility, but instead of half-assing it we will postpone it for the future. However, this patch does seem like a small compromise. Merged, thanks.

PR #167, opened 2015-02-08
Generic Dvaered Patrols

onpon4:
(Updated #146, since my indiscriminate actions with Git screwed that one up.)

This makes two changes to the Dvaered patrol mission:

1. It is no longer necessary to complete the Empire campaign to access them; the Dvaereds will allow any licensed mercenary to do the job.

2. The FLF no longer get mad at you just for doing Dvaered patrols. Not all Dvaered patrols involve encounters with the FLF, so this behavior doesn't make sense, and you'll lose points for doing your duty shooting down FLF ships, anyway.


bobbens:
Looks good, thanks.

PR #168, opened 2015-02-10
Make Gui - Slimv2 available for sale

ids1024:
Slimv2 works at least as well as the brushed and legacy GUIs, so there is no reason for it to not be available while the others are.


bobbens:
OK, I don't think it's finished, but I guess it can serve as an encouragement. :)

PR #170, opened 2015-02-13
Made it possible to activate outfits that are cooling down.

onpon4:
It makes no sense for a somewhat hot outfit to have to be completely
cool before you can start it up again, and it degrades the usefulness
of outfits with a cooldown period (such as the Emergency Shield Booster)
in a way that's annoying. With this change, the cooldown period still limits
the use of such outfits the same amount, but gives the player full
control over when this use is.

I wanted to do this with beam weapons as well, but couldn't figure
out how to.

### DIFF
--- a/src/pilot_weapon.c
+++ b/src/pilot_weapon.c
@@ -165,6 +165,7 @@ void pilot_weapSetAIClear( Pilot* p )
 void pilot_weapSetPress( Pilot* p, int id, int type )
 {
    int i, l, on, n;
+   double d;
    PilotWeaponSet *ws;
 
    ws = pilot_weapSet(p,id);
@@ -194,7 +195,7 @@ void pilot_weapSetPress( Pilot* p, int id, int type )
 
       case WEAPSET_TYPE_ACTIVE:
          /* The behaviour here is more complex. What we do is consider a group
-          * to be entirely off if not all outfits are either on or cooling down.
+          * to be entirely off if not all outfits are on.
           * In the case it's deemed to be off, all outfits that are off get turned
           * on, otherwise all outfits that are on are turrned to cooling down. */
          /* Only care about presses. */
@@ -209,7 +210,7 @@ void pilot_weapSetPress( Pilot* p, int id, int type )
          on = 1;
          l  = array_size(ws->slots);
          for (i=0; i<l; i++) {
-            if (ws->slots[i].slot->state == PILOT_OUTFIT_OFF) {
+            if (ws->slots[i].slot->state != PILOT_OUTFIT_ON) {
                on = 0;
                break;
             }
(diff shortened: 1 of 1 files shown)


PhoenixRiver:
> It makes no sense for a somewhat hot outfit to have to be completely
> cool before you can start it up again, and it degrades the usefulness
> of outfits with a cooldown period (such as the Emergency Shield Booster)
> in a way that's annoying. With this change, the cooldown period still
> limits the use of such outfits the same amount, but gives the player full
> control over when this use is.

I'd agree for things that are press button to entirely use it but I'd
have said it reduces the skill level for those where you don't need to
keep using it (e.g. beam weapons) - part of controlling a fighter is
balancing your use against any inherent problems with it - overusing
beams is the price you pay for not having to buy ammunition for
them... I'd prefer a game promoting the importance of skill rather
than heading towards dumming down these aspects - personal preference
though - anyone else?

Johann

PhoenixRiver:
Sorry that wasn't so clear - I think that having a lockout and
complete cooldown for things such as beams is a good thing - it makes
you try to not overuse them, which, as I said, I think is a good
thing.

Johann

onpon4:
I don't think it's necessary and inconsistent.

Now, if using something for maximum duration caused an extra cooldown penalty, like what happens with the laser in Project: Starfighter, that would make sense. That isn't the case in Naev, and even in Starfighter you could use the laser at any time *as long as you didn't max it out*.

In the case of the laser, why should you be able to shoot a semi-constant stream if it's almost completely cool, but not if it's, say, half-way between completely cool and completely hot? After all, it would be quite silly and annoying if, after turning off your computer, it refused to turn on again until all of its components reached room temperature!

bobbens:
This is a gameplay decision and is implemented on purpose. As with other gameplay decisions, it is not necessarily realistic, but adds an additional tactical element the player must take into account.

onpon4:
I'm not concerned about realism (I would still like #144 to be merged, for example), but I don't consider this "tactical element" to be fun, just annoying. With the Emergency Shield Booster (the only outfit currently affected), there's already a perfectly good tactical element: it drains your battery tremendously, which makes it more difficult to use weapons. I don't see why the cooldown time should be anything more than a hard limit on use.

PhoenixRiver:
Hello,

> In the case of the laser, why should you be able to shoot a semi-constant
> stream if it's almost completely cool, but not if it's, say, half-way
> between completely cool and completely hot? After all, it would be quite
> silly and annoying if, after turning off your computer, it refused to turn
> on again until all of its components reached room temperature!

In the case of your computer this would be annoying but if you
overheat the CPU a thermal cutoff will cause a hard hardware shutdown
and you will indeed have to wait approximately 10 to 30 seconds before
the machine will let you turn it back on (at least on machines that
have a thermal inlocks, machines that don't (often older) will let you
power them on but then the thermal cutoff will kick in again and,
again, shut down the machine).

If you overheat a laboratory laser you will more likely melt
components in it and risk blinding... I'd hate to have that happen
with a military grade laser...

As for emergency shield boosts... it's supposed to be for an
emergency, not regular use, some games have them burn out after one
use... if you are still fighting when you are using it you are doing
something wrong unless you're too slow (shouldn't have gotten into the
fight in the first place), or have turrets (pick smalle ships to fight
- it's not like a turreted ship can't take on all manner of ships in
this game!)

I promise to be quiet now though, sorry been extra chatty recently,
probably too much.

Johann

BTAxis:
I'm not concerned about realism (I would still like #144 to be merged, for example), but I
        don't consider this "tactical element" to be fun, just annoying.
        With the Emergency Shield Booster (the only outfit currently
        affected), there's already a perfectly good tactical element: it
        drains your battery tremendously, which makes it more difficult
        to use weapons. I don't see why the cooldown time should be
        anything more than a hard limit on use.
    
    If the issue is only with the ESB (and mind you, I don't agree that
    there is an issue here), then don't try to fix it with a change that
    affects all outfits. Instead you could just have removed the
    cooldown from the ESB outfit.

onpon4:
It's currently only the emergency shield booster, but I don't consider it to be good behavior for anything.

Barring energy depletion, the behavior usually makes it "tactically" best to use the outfit until it completely overheats, or else use it in extremely short bursts, because in battle, there's little difference between having to wait 25 seconds and having to wait 15 seconds.

> if you overheat the CPU a thermal cutoff will cause a hard hardware shutdown and you will indeed have to wait approximately 10 to 30 seconds before the machine will let you turn it back on

Project: Starfighter simulates this type of setback in its laser weapon by making the amount of heat instantly become something like 125% when it reaches 100%. So there's an additional penalty if you let it overheat completely. This is pretty much the *opposite* of how Naev's outfit cooldowns end up working (see above).

PR #171, opened 2015-02-13
Set faction of the teenagers in Teenagers mission to "Dummy".

onpon4:
So police forces don't start going after you for doing the mission.


PhoenixRiver:
> So police forces don't start going after you for doing the mission.
> You can view, comment on, or merge this pull request online at:

Makes sense :)

Johann

bobbens:
Why aggressive?

onpon4:
Because my previous assumption that police defense is defined by faction was incorrect, I'm not sure changing the faction to "Dummy" actually made a difference, now (could have just been coincidence, as far as I know).

PhoenixRiver:
Hello,

> Because my previous assumption that police defense is defined by faction was
> incorrect, I'm not sure changing the faction to "Dummy" actually made a
> difference, now (could have just been coincidence, as far as I know).

Thinking back I didn't have this problem with this mission but did
have a problem with another where one of my shots accidentally hit
another ship...

Johann

bobbens:
Looks good, thanks.

PR #172, opened 2015-02-13
Don't lower the player's reputation unless in the faction's territory.

onpon4:
This was something I suggested on IRC to get rid of the problem where
fighting Empire ships that join in fights against you with the Dvaered
negatively affects reputation. I'm not quite sure how well it works, since I haven't tested it yet.

One particular quirk with this is piracy in uninhabited systems has no
permanent consequences. But it could be that this isn't such a bad
thing.

One problem with this is it doesn't work for the FLF or traders, since
both of these factions are typically encountered nowhere near their
own territory. To solve this, I added a variable indicating whether
the old or new behavior should be used (default new), and set the FLF
and Trader faction to use the old behavior.

### DIFF
--- a/dat/factions/standing/flf.lua
+++ b/dat/factions/standing/flf.lua
@@ -8,6 +8,7 @@ _fdelta_distress = {-0.5, 0} -- Maximum change constraints
 _fdelta_kill     = {-5, 1.5} -- Maximum change constraints
 _fcap_misn     = 50 -- Starting mission cap, gets overwritten
 _fcap_misn_var = "_fcap_flf"
+_fextern_penalty = true
 _fthis         = faction.get("FLF")
 
 
--- a/dat/factions/standing/skel.lua
+++ b/dat/factions/standing/skel.lua
@@ -12,6 +12,10 @@ _fcap_misn       = 30 -- Starting mission cap, gets overwritten
 _fcap_misn_var   = nil -- Mission variable to use for limits
 _fcap_mod_sec    = 0.3 -- Modulation from secondary
 
+-- Whether or not the faction should get mad for attacking them or their
+-- allies outside of their territory.
+_fextern_penalty = false
+
 
 --[[
    @brief Clamps a value x between low and high.
@@ -94,6 +98,21 @@ function default_hit( current, amount, source, secondary )
       end
    end
 
+   if not _fextern_penalty and (source == "distress" or source == "kill") and amount < 0 then
+      -- Only take a hit if in the faction's territory
+      local has_planet
+      for _, planet in ipairs(system.cur():planets()) do
+         if planet:faction() == _fthis then
+            has_planet = true
+            break
+         end
+      end
+      if not has_planet then
+         -- Planet belonging to this faction not found. Don't modify reputation.
+         return f
+      end
+   end
+
    -- Adjust for secondary hit
    if secondary then mod = mod * _fcap_mod_sec end
    amount = mod * amount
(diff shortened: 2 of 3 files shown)


Zireael07:
Great!

PhoenixRiver:
Hello,

> This was something I suggested on IRC to get rid of the problem where
> fighting Empire ships that join in fights against you with the Dvaered
> negatively affects reputation. I'm not quite sure how well it works, since I
> haven't tested it yet.

I thought that the Empire and Dvaered were allies so then it should
negetively affect your reputation with both, no? Or do I misunderstand
you?

> One particular quirk with this is piracy in uninhabited systems has no
> permanent consequences. But it could be that this isn't such a bad
> thing.

I assume that this is with the new system... and I'd have said it is
highly likely that a distress signal would have been sent by any
attacked ship so a perminent consequence should definitely be applied.

> One problem with this is it doesn't work for the FLF or traders, since
> both of these factions are typically encountered nowhere near their
> own territory. To solve this, I added a variable indicating whether
> the old or new behavior should be used (default new), and set the FLF
> and Trader faction to use the old behavior.

Pirates as well? and anywhere where there are border disputes and you
side with the defenders rather than the invaders - but the new system
wouldn't note that since you are in the defenders space you don't
loose points with the invaders which makes no sense - you should now
be known by them as hostile to their cause...

I'm confused why this needed to be changed I suspect because there
doesn't seem to be a need to change this mechanism at all - maybe it
is the faction alliances that need changing if anything?

Johann

onpon4:
> Pirates as well?

No, because attacking pirates doesn't lower your reputation with them anyway (by design).

> and anywhere where there are border disputes and you side with the defenders rather than the invaders - but the new system wouldn't note that since you are in the defenders space you don't loose points with the invaders which makes no sense - you should now be known by them as hostile to their cause...

I'm not really aware of a solution to this potential problem, but do note that *any* presence of the faction you are attacking results in a hit, so in many cases the problem doesn't exist. Most notably, while Haleb is mainly a Frontier world, it does have Dvaered presence, so attacking the Dvaered in that system would cause a hit to your reputation with them.

> I'm confused why this needed to be changed I suspect because there doesn't seem to be a need to change this mechanism at all - maybe it is the faction alliances that need changing if anything?

My original change was to make the FLF and Empire enemies, but bobbens rejected this change and explicitly stated that they are not enemies. On IRC, it was further clarified that the system should accomodate the Empire siding with the Dvaered in conflicts, but not lose reputation with you if you side against the Dvaered. That's what this commit is meant to solve. (There are other possible mechanisms, of course, and they might be better, but this was a simple solution I thought of.)

onpon4:
By the way, to clarify: when I said "presence" above, I actually meant "a planet".

bobbens:
While I do see what you're trying to get at. I do not think this is the best approach, as it gives a pretty high level of impunity. Some more fine grained level of control would be better. To for example not give a faction penalty when being engaged by an Empire ship while skirmishing with the Dvaered in Dvaered territory. However, if you do attack the Empire ship and/or destroy it, you should get a penalty.

onpon4:
Ultimately, the proper solution is probably to rework the whole system so that ships attacking you can *never* lower your standing with anyone. But that there is a drop in reputation from destroying the ships of allies is yet another problem, albeit unique to the FLF/Dvaered (it's otherwise a pretty good policy).

PhoenixRiver:
Hello,

> Ultimately, the proper solution is probably to rework the whole system so
> that ships attacking you can *never* lower your standing with anyone. But
> that there is a drop in reputation from destroying the ships of allies is
> yet another problem, albeit unique to the FLF/Dvaered (it's otherwise a
> pretty good policy).

I'd have said a system that allows attacking ships to lower your
rating (if you kill them) as well would be useful (even if it wasn't
the default) this would allow missions that are, in effect, setups or
traps - you do have the choice not to fight after all - but also for
Police to initiate combat with known criminals and for you to become
more of a criminal in their eyes if you engage and kill more of them
(possibly this might enable the raising of the stakes with stronger
police or the military getting involved).

Just a thought :)

Johann

onpon4:
After some experimenting, I think I can conclude that bobbens is right:

> I do not think this is the best approach, as it gives a pretty high level of impunity.

Not only can you pirate as long as it's in an empty system, but you can also attack, for instance, Dvaered ships without penalty in an empty system. In this way, you can fight the Dvaered without making enemies out of them.

Anyway, while I'm removing this from my master branch, I'm still leaving my "forgiveness" branch open.

PR #173, opened 2015-02-13
Abort, not End, autonav when hailed.

onpon4:
As long as we're doing it this way, we'd best be doing it properly.
player_autonavEnd is not meant to be used to suddenly stop Autonav,
and I don't know if it even works.


bobbens:
Looks good, thanks.

PR #174, opened 2015-02-14
Made the Progeny "aggressive".

onpon4:
Workaround to prevent police forces from defending the kidnappers.


PhoenixRiver:
Hello,

> Workaround to prevent police forces from defending the kidnappers.

Don't remember this being a problem... is there something that has
changed about how agressively the police protect ships? This seems to
be a common theme recently...

Johann

bobbens:
OK, looks good.

PR #176, opened 2015-02-14
Don't offer pirate bounties in the current system.

onpon4:
I always get the same result when I get a pirate bounty in the current
system: I can't find the pirate, and it gets killed by someone else.
I could have solved this by making it spawn close to the planet, but
not doing this was a deliberate measure to prevent the player from
grinding in one system over and over again, so I've just removed the
option for bounties in the current system.

### DIFF
--- a/dat/missions/neutral/pirbounty_dead.lua
+++ b/dat/missions/neutral/pirbounty_dead.lua
@@ -79,7 +79,7 @@ end
 function create ()
    paying_faction = planet.cur():faction()
 
-   local systems = getsysatdistance( system.cur(), 0, 3,
+   local systems = getsysatdistance( system.cur(), 1, 3,
       function(s) return s:presences()["Pirate"] end )
 
    if #systems == 0 then


onpon4:
One thing I should mention, and this was a major factor in deciding to remove current-system bounties entirely:

I found that usually, you simply can't make it to the pirate's location in time to catch him before he is killed by police, *even if you know where he is*. Planetary systems are just too huge for you to realistically be able to get to them most of the time, unless you're lucky and you're really lucky.

The only way current-system bounties could work is if police forces don't go after your target. I've considered adding this, especially for Alive bounties, but I've already experimented with getting rid of all other ships, and found that to be much more boring; and I think having police forces ignore the bounty targets would have a similar effect.

ids1024:
I have found alive bounties almost impossible, at least in well policed systems where the bounty is for a small ship. I disable the target, but even then the police shoot it out of space before I can board it.

Something should be changed to make this work.  If the police (who are presumably the ones offering the bounties) want the pilot alive, they should not destroy his ship.

Perhaps the normal police could be made to ignore the ship, but one or two bounty hunters could be spawned who are also trying to kill/capture the target?

onpon4:
For one thing, police forces shouldn't continue to attack disabled ships.

For the other half, that could probably be solved by adding a capability to mark a ship as "nokill" or something, and causing all ships to only use non-lethal weapons on those ships unless directly threatened. This could also be useful for pirates, who should be trying to board civilian and trader ships, not attack them.

bobbens:
OK, looks good.

PR #177, opened 2015-02-14
Fixed possibility of pirate bounty missions failing after they were won.

onpon4:
It happened when you disabled the pirate ship as it was jumping.
Could happen in other circumstances as well, possibly.


bobbens:
If I understand correctly, the problem is you disable a ship, then it wakes up and jumps away. I have put a :disable() command to make sure the ship never reenables, which should also solve this, but having both doesn't hurt. Thanks.

onpon4:
Actually, if my eyes didn't decieve me, the ship was actually destroyed, rather than waking up. Odd that being destroyed would cause an aborted jumpout to be completed, but I think that's what happened.

PR #181, opened 2015-02-16
Various minor improvements to patrol missions.

onpon4:


bobbens:
When the diff is a bit hard to interpret, could you please do a small overview of the changes? From the commits I'm not exactly clear what you've done and looking at the diff it's bit hard figure out all the changes you did. Thanks,

onpon4:
Sorry, I was doing all kinds of stuff at once and forgot the changes as I was doing them. Differences are:

* Dvaered patrol is hostile from the beginning.
* modPlayer is used to give your reputation bonus, rather than modPlayerSingle.
* Patrol ships are no longer renamed.
* Vigilance hail or FLF arrival time in pre02 is now 30 seconds after the arrival of Dvaered reinforcements, rather than 10 seconds.
* Vigilance keeps hailing you every 8 seconds until you answer.
* FLF are not spawned at their scheduled time in pre02 if the Vigilance is dead. (It's kind of weird for the FLF to come and save you if you're actually curb-stomping the reinforcements, which is usually the case if the Vigilance is dead 30 seconds after arrival.)
* Dvaered reputation modification which is supposed to set your reputation to 0 now uses modPlayerRaw instead of modPlayerSingle.
* Defeating all the patrol ships and reinforcements in pre02 immediately sets your standing with the Dvaered to -1 if it was greater than or equal to 0.

onpon4:
So, are any of these changes problematic?

bobbens:
No, I've just been swamped with work. I expect to look at it and merge it this weekend.

bobbens:
OK, looks good. Thanks,

PR #182, opened 2015-02-18
Check for alternative Lua package names

Partmedia:
While here, rewrite the Lua package checking logic using a m4 foreach loop.


bobbens:
Looks good, thanks.

PR #184, opened 2015-03-10
Added to pirate chatter

EveColonyCommander:
~70 additions to pirate chatter. Includes additions to bribes, bribe acceptance, and taunts.


bobbens:
Pulled, thanks.

PR #185, opened 2015-03-13
Adding a way to handle fleets via a lua script

l0k1:
This is a script to handle fleets. It's kind of a fleet AI. As it stands, it works as intended and it works well IMO.


l0k1:
I should clarify - this isn't intended to be a full-scaled fleet AI. This is intended more for usage in missions and events in the current Naev release, to make large amounts of ships coordinate together with relative ease to the scripter. It is built on top of the current AI system and doesn't change the AI of the ships in the fleet at all. It makes heavy use of pilot.control() and pilot.goto() functions.
It works by passing a table of pilots to the Forma:new() function, and creating a fleet object. The fleet is given a formation, and a distance from which to break formation and engage the enemy (defaults to 3000m from the fleet leader to the nearest enemy). The fleet (and it's AI), is built around the fleet leader, and the scripter would control the leader to control the entire script. Or just allow Naev's AI to run the fleet.
It automatically handles removing ships from the fleet, the fleet jumping, or landing. It is simple and built entirely on top of Naev's AI.

bobbens:
Yes, I would have liked a full proper integration in the engine, but I guess that's my job :P. Thanks.

bobbens:
I pulled it, but to avoid adding merge spam to the git log, I cherry-picked the change. For future reference you should not work in your master when you do a pull request and create a new branch. You'll probably have to reset your head to naev/naev.

PR #429, opened 2015-03-19
Fix compile errors when the flag -Werror=format-security is set

jdieter:
In Fedora, we compile all of our packages with -Werror=format-security.  This patch fixes a few places where types aren't specified, which allows naev to compile with -Werror=format-security.


bobbens:
Looks good, thanks.

PR #432, opened 2015-03-21
Correctly pass boarding ship to pilot board hook

ids1024:


bobbens:
Hello, I believe it is correct as it is, i.e., the pilot being boarded is hooked, not the pilot doing the boarding. If you want the pilot doing the boarding to be hooked, another hook should be added.

BTAxis:
The board hook always assumes the boarding ship is the player. AI->AI boarding is not supported.

bobbens:
Actually AI does board. I implemented sometime, but never moved the hooks over. In particular it is possible for pirates to plunder merchants. They are the onnly ones who actually board iirc

PR #439, opened 2015-03-29
fixing bug in fleet script; allowing scripter to set preferred fleet leader

l0k1:
Just cleaning up some stuff. And, in writing a mission, found it would be very handy to be able to set who the fleet leader is.


l0k1:
After looking at this, it needs some more work.

bobbens:
Looks good, thanks.

PR #441, opened 2015-03-29
make AI speed aware

toams:
Add a lua call so the AI can check a pilots speed. This way the speed of a ship can be controlled by stop accelerating when a certain speed is achieved. Useful for escort missions and probably fleet AI. Maybe even useful for attack functions: no more overshooting of fast ships attacking slow ships.
This pull request only includes the C-code, the actual lua functions are not written yet.


bobbens:
Looks good, thanks.

PR #443, opened 2015-03-30
Made failure on Dead or Alive bounties much less common.

onpon4:
This includes two main changes:

1. If someone else destroys the ship, you still get the bounty if
   you fired more shots on it than it did.

2. If someone else destroys the ship *and* someone else fired more
   shots on it than you did, but that ship was an Empire ship, there
   is a 50% chance of the ship hailing you and offering you a smaller
   amount of credits than the bounty (proportional to the number of
   times you hit the ship).

I've only tested this so far with a Kestrel, which now has no trouble
at all getting a bounty in Empire space. I hope this change makes using
smaller ships more viable, in particular.

To be clear, the only reason only Empire ships currently offer partial bounties is I think there should be different text for each faction; the text I wrote so far would look out-of-place in Sirius, for example. However, some factions, like the Dvaered, seem to me like they legitimately shouldn't offer the smaller reward.

### DIFF
--- a/dat/missions/neutral/pirbounty_dead.lua
+++ b/dat/missions/neutral/pirbounty_dead.lua
@@ -201,6 +201,14 @@ function land ()
 end
 
 
+function update_last_health ()
+   if target_ship ~= nil and target_ship:exists() then
+      last_armour, last_shield = target_ship:health()
+      hook.safe( "update_last_health" )
+   end
+end
+
+
 function pilot_board ()
    player.unboard()
    local t = subdue_text[ rnd.rnd( 1, #subdue_text ) ]:format( name )
@@ -216,10 +224,22 @@ end
 
 function pilot_attacked( p, attacker )
    if attacker ~= nil then
+      local armour
+      local shield
+      local damage = 0
       local found = false
+      
+      armour, shield = target_ship:health()
+      if armour < last_armour then
+         damage = damage + (last_armour - armour)
+      end
+      if shield < last_shield then
+         damage = damage + (last_shield - shield)
+      end
+
       for i, j in ipairs( hunters ) do
          if j == attacker then
-            hunter_hits[i] = hunter_hits[i] + 1
+            hunter_hits[i] = hunter_hits[i] + damage
             found = true
          end
       end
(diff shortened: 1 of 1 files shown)


onpon4:
OK, all ships now have a chance of offering you a share of the bounty now, so you don't have a special advantage by being in Empire space.

In other news, I tested this with a smaller ship, and it does seem helpful. Would be much better if it paid attention to damage, though... is there any way to do that? (For example, there is probably essentially zero benefit to users of the Ripper Cannon, as it is now.)

pydsigner:
@onpon4 "What about damage?" was the first thing that came to my mind when I read "more shots".

onpon4:
Right. I intended to base it on damage from the start, but I can't figure out a reliable way to do it. Using hits is only temporary (I hope).

pydsigner:
This is somewhat of a hack, but you could measure the decrease in health (and possibly shield) from hit to hit and add that instead. For instance:
```
armor, shield = p:health()
change = last_armor - armor
last_armor = armor
```
```
hunter_damage[i] = hunter_damage[i] + change
```

onpon4:
That won't work for shield; in between hits, shield can easily be higher than it was on the previous hit, and that could cause hitting a shielded pirate to inflict a penalty, not at all desirable (and even if you exclude that as a special case, it would still render the hits less meaningful than they should be). The shield regeneration has to be compensated for somehow.

On the other hand, just using armor would render damage to shield meaningless, and I don't want that.

Hm... perhaps the shield_regen stat can help.

pydsigner:
I was thinking about that shield issue. It might be wise to factor armor regeneration speed in as well.

onpon4:
Alright, damage is now factored in!

pydsigner:
How often will that hook run?

onpon4:
The "safe" hook happens every frame, at the end of the frame.

pydsigner:
Ok, makes sense then.

bobbens:
I've edited the attacked function to give the damage done as another parameter. I think it would be best to just use that instead of using the safe hook you have right now as it'll be more exact. Although it won't take into account the regen. What do you think?

onpon4:
Cool, OK.

bobbens:
Looks good, thanks.

PR #444, opened 2015-03-30
Run pilot board hook for all pilots and pass the boarding ship as a parameter

ids1024:


bobbens:
OK, so the problem is you were breaking all the current missions. So I've kept the old "board" behaviour and added a new "boarding" hook which is the one you patched.

PR #445, opened 2015-03-31
Warden's End is now landable and has assets, etc.

l0k1:
Warden's End was a space station in Kretogg that was unlandable. Just adding text in, making it landable, etc.


bobbens:
Looks good, thanks.

PR #450, opened 2015-04-01
Made the "alive" bounties more doable.

onpon4:
Quite simply, I made the pirates in the "alive" bounty a different
faction, which the pirates consider to be an ally (and so they will
assist them), but which no one considers to be an enemy, and their
only enemy is the player. In effect, no one gets in the way by blowing
them up. Since non-lethal weapons are limited in power, and other
pirates continue to be an obstacle, this doesn't seem to cause the
mission to be overtly too easy.

### DIFF
--- a/dat/faction.xml
+++ b/dat/faction.xml
@@ -83,9 +83,6 @@
   <spawn>pirate</spawn>
   <standing>pirate</standing>
   <equip>pirate</equip>
-  <allies>
-   <ally>Wanted_Pirate</ally>
-  </allies>
   <enemies>
    <enemy>Civilian</enemy>
    <enemy>Empire</enemy>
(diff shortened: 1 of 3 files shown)


bobbens:
I see the reasoning behind this, but I feel it might take a bit the immersion out of it as nobody will attack the pirate except the player. Might lead to some weird situations imho.

onpon4:
I changed this so that 70% of the time, all pilots' tasks are cleared.

onpon4:
So, is this good? My playtesting suggests it is.

BTAxis:
I'm inclined to think it's okay. Gameplay is by its nature player-centric anyway, so I don't think there are any immersion issues here.

PR #451, opened 2015-04-01
Added crew stat to the Lua API.

onpon4:
This can allow custom boarding code to take advantage of this number.
I was in particular thinking of using it for the bounty missions, but
it could be used in other places too.


bobbens:
OK, it's likely we'll remove it eventually or leave it as some stat in favor for a new boarding mechanic. Thanks.

PR #453, opened 2015-04-02
More fleet_form.lua stuff.

l0k1:
Just bug-testing and fixing errors. I don't know if I really like the speed limit settings how they are - slow ships take a while to catch up (main algorithm is at line 510); but it works.


ids1024:
Is there any reason this hasn't been merged?

Deiz:
Nope, looks fine to me.

PR #454, opened 2015-04-02
Added more text options to pirate bounty missions.

onpon4:


onpon4:
Um, someone screwed up the text by adding that "pronoun" variable. Some of the instances were *not* supposed to be capitalized.

Deiz:
Ah, whoops. Will fix  that in a minute.

onpon4:
Makes it look like the officer it's talking about is some sort of god or something.

PR #455, opened 2015-04-02
Made capture of a pirate not guaranteed.

onpon4:
Based on the C code that checks whether stealing things succeeds.
Basically, the more crew you have, the more likely it is you'll
succeed.

If you fail to capture a pirate in a dead-or-alive mission, you can still kill the pirate. If you fail to capture them in an alive-only mission, too bad; you just have to abandon the mission.


Deiz:
I think that might be a bit harsh. Incentivizing crew bonuses is a good thing, but I think it may be better if the penalty for failure was something gentler, e.g. spawning a few pirates nearby due to the time it took to capture the target.

onpon4:
Doesn't seem to be a particular problem. But how about if when you fail to capture the pirate, allowing you to kill him for a much lower bounty (say, 1/4 the alive bounty offer)?

Deiz:
That'd work too, although would you script that, or rely on the player to blow up the ship after unboarding?

onpon4:
I don't see any particular reason to script it. I was thinking of either having a random ship from the paying faction hail you, or just giving you a message saying you have permission to kill.

Deiz:
The latter sounds more reliable, because it'll always be shown before the player has a chance to do anything after unboarding.

EveColonyCommander:
I'd go with making the 'permission to kill' available at a chance. Lower bounties have a smaller chance and higher bounties are greater. You can still completely fail but you're more likely to succeed in some way when the risk is higher.

Getting permission to kill every time, especially on low level pirates doesn't make sense unless it's up front at the counter when you sign up to 'capture or kill'

Making it dependent on if a faction ship is in the system could also keep things from breaking walls. So deep space jobs are generally at a risk for failure and border patrol is generally a safe option for cash flow.

All of this can be circumvented by having something akin to 'standing orders' at the start of the mission. The player is told who the bounty is, how much they are wanted for, and how the capture needs to be done(currently only dead/alive). This would include the 'all else fails' clause you are discussing.

onpon4:
OK, here's what I've done:

When you fail to capture a pirate, there's a 25% chance of being contacted by the authorities with authorization to kill the pirate, but with a bounty only 1/5 the original offer. How's that?

onpon4:
So, how is it?

Deiz:
Seems like a reasonable approach.

PR #456, opened 2015-04-02
Fixed some bugs with achack03.

onpon4:
Specifically, police defending Harja, and a case where the OSD was
being inappropriately reset.


onpon4:
I've added another bugfix here: the "buy map" button showing up late (which didn't cause any problems, but looked weird).

BTAxis:
Looking good, thanks.

PR #459, opened 2015-04-05
Changed the text representations of the various standing amounts.

onpon4:
These text representations are more consistent; they now all describe what you are to the faction.


BTAxis:
Actually this should probably be moved into the factions' standing lua files, so we can have personalized titles per faction.

onpon4:
True. :) (Also for localization support.)

onpon4:
This is now obsolete, superseded by #464.

PR #462, opened 2015-04-12
Slimv2 and brushed gui fixes

ids1024:


ids1024:
I noticed that the brushed gui was crashing when the player had fighter bays. I added a fix to this PR.

PR #464, opened 2015-04-16
Made standing text something defined in Lua.

onpon4:
So now, different factions can have different terms for different amounts
of standing, and these text representations can now be localized.


BTAxis:
Nice work, thanks.

PR #466, opened 2015-04-16
Removed dv_patrol, and flf_pre02's dependence on it.

onpon4:
I replaced the patrol mission counting with a simple check of the player's standing with the Dvaereds. Also changed some text to be consistent with this change.


bobbens:
Looks good, thanks.

PR #472, opened 2015-04-29
Cause beams and outfits to always require their full delay.

onpon4:
This reverts a change I made in the past to outfits, and then implements the same constant delay with beams.

My original change which I reverted here was intended for use with an alternative system that functioned more like actual heat, i.e. #170. However, that was rejected. It's obvious that the "cooldown" amount is actually intended to be an analog for the delay between each bullet fired, and I suppose this is a useful enough feature, if currently misused. It can be used for a sort of "pulse" effect.

However, when you can activate an outfit for however long you want, and keeping it activated longer means waiting longer until next time, all that does is encourage activating the outfit in really short bursts, perhaps even doing this with a macro of some sort. I can't think of any possible reason to actually want to do this.

The same goes for the laser, so I also changed the laser delay to be constant as well. Right now this causes a bit of a usability problem due to the rather long delay of all the available lasers, but this would be fixed as soon as lasers take turns firing rather than firing all at once.

### DIFF
--- a/src/nlua_pilot.c
+++ b/src/nlua_pilot.c
@@ -1577,35 +1577,36 @@ static int pilotL_actives( lua_State *L )
       }
 
       /* State and timer. */
-      if (o->cooling)
-      {
-         str = "cooldown";
-         d = outfit_cooldown(o->outfit);
-         if (d==0.)
-            d = 0.;
-         else if (!isinf(o->stimer))
-            d = o->stimer / d;
-         lua_pushstring(L,"cooldown");
-         lua_pushnumber(L, d );
-         lua_rawset(L,-3);
+      switch (o->state) {
+         case PILOT_OUTFIT_OFF:
+            str = "off";
+            break;
+         case PILOT_OUTFIT_WARMUP:
+            str = "warmup";
+            break;
+         case PILOT_OUTFIT_ON:
+            str = "on";
+            d = outfit_duration(o->outfit);
+            if (d==0.)
+               d = 1.;
+            else if (!isinf(o->stimer))
+               d = o->stimer / d;
+            lua_pushstring(L,"duration");
+            lua_pushnumber(L, d );
+            lua_rawset(L,-3);
+            break;
+         case PILOT_OUTFIT_COOLDOWN:
+            str = "cooldown";
+            d = outfit_cooldown(o->outfit);
+            if (d==0.)
+               d = 0.;
+            else if (!isinf(o->stimer))
+               d = o->stimer / d;
+            lua_pushstring(L,"cooldown");
+            lua_pushnumber(L, d );
+            lua_rawset(L,-3);
+            break;
       }
-      else if (o->on)
-      {
-         str = "on";
-         d = outfit_duration(o->outfit);
-         if (d==0.)
-            d = 1.;
-         else if (!isinf(o->stimer))
-            d = o->stimer / d;
-         lua_pushstring(L,"duration");
-         lua_pushnumber(L, d );
-         lua_rawset(L,-3);
-      }
-      else
-      {
-         str = "off";
-      }
-
       lua_pushstring(L,"state");
       lua_pushstring(L,str);
       lua_rawset(L,-3);
--- a/src/nlua_pilot.c
+++ b/src/nlua_pilot.c
@@ -1577,36 +1577,35 @@ static int pilotL_actives( lua_State *L )
       }
 
       /* State and timer. */
-      switch (o->state) {
-         case PILOT_OUTFIT_OFF:
-            str = "off";
-            break;
-         case PILOT_OUTFIT_WARMUP:
-            str = "warmup";
-            break;
-         case PILOT_OUTFIT_ON:
-            str = "on";
-            d = outfit_duration(o->outfit);
-            if (d==0.)
-               d = 1.;
-            else if (!isinf(o->stimer))
-               d = o->stimer / d;
-            lua_pushstring(L,"duration");
-            lua_pushnumber(L, d );
-            lua_rawset(L,-3);
-            break;
-         case PILOT_OUTFIT_COOLDOWN:
-            str = "cooldown";
-            d = outfit_cooldown(o->outfit);
-            if (d==0.)
-               d = 0.;
-            else if (!isinf(o->stimer))
-               d = o->stimer / d;
-            lua_pushstring(L,"cooldown");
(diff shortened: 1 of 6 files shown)


Deiz:
Many beams have a manually-defined minimum duration which they'll always operate for (unless the ship runs out of energy), which is defined as part of the [beam revamp](http://wiki.naev.org/wiki/Beam_Revamp), with the rest being capable of pulsing for short durations.

I assume the other commit would only impact the emergency shield booster currently?

In general I'm fond of being able to use outfits for less than their full duration in exchange for a less-than-maximum delay, but the ability to pulse things without them having a minimum duration is clearly a balance issue.

One possible soft remedy would be to front-load the delay accumulation, so that it accumulates faster (1.5x or something) initially, reaches the XML-defined value at 50% of the duration, and perhaps 0.5x by the end, to incentivize firing for longer periods of time.

onpon4:
I've made another change which complements this: separating outfit activation from cooldown, so that outfits will automatically reactivate once they've finished their cooling cycle.

> One possible soft remedy would be to front-load the delay accumulation, so that it accumulates faster (1.5x or something) initially, reaches the XML-defined value at 50% of the duration, and perhaps 0.5x by the end, to incentivize firing for longer periods of time.

I really don't see the point. The fact that all beams fire at once is a documented shortcoming that's supposed to get fixed, so you'll be able to get a continuous beam just by equipping two at once. Besides, there's already another reason to stop firing prematurely: to avoid draining your battery.

Deiz:
I'm not sure what to do, at present. The thing with beams is that, generally being short-range, it mostly makes sense for them to fire in unison, because they deal spike damage and ideally the attacker wants to get out of the victim's weapon range while the beams recharge.

I think staggering their firing in the same manner that bolt weapons are handled would mostly remove the purpose of the recharge period.

onpon4:
Maybe for the forward-mounted ones ("lances"), though I rather doubt this. But if you're using beam turrets, chances are your ship is pretty slow. It seems to me the ideal use-case is to sacrifice some range for perfect accuracy and high damage. Staggering shots doesn't negate this any more than it negates the usefulness of incredibly high-damage shot weapons. (On that note, I honestly think it would be a potentially useful feature to allow the player to choose whether or not to spread out the firing of weapons like that, regardless of the weapon type.)

As for removing the purpose of the recharge period, even if that's true (I don't think it is; the recharge period is still putting a limit on how much the laser can be used), a version of the effect can be achieved manually anyway (just assign each beam to a different button; either make half of them primary or half of them secondary, or if you want to go all-out, make each of them separate with instant mode).

PhoenixRiver:
> I honestly think it would be a potentially useful feature to allow the
> player to choose whether or not to spread out the firing of weapons like
> that, regardless of the weapon type.)

This made me think that maybe you could either just set it as a
default below a certain speed (either by craft or current flying speed
(the latter is less intuitive though)) or better have it as a very
small module you can buy - either in phase or out of phase - but the
weapons themselves have a default - lasers in unison, bolts
alternating, etc.

Just a thought?

Johann

P.S. I realise that I might just have volunteered to do that!

onpon4:
Why would you need an extra piece of hardware for that? I was just thinking it could be a checkbox in the info/weapons window.

PhoenixRiver:
> Why would you need an extra piece of hardware for that? I was just thinking
> it could be a checkbox in the info/weapons window.

I doesn't need to be but then you could argue, in extremis, that there
should be no modules and that you should have a "build your own ship"
checkbox window - if you want to hang on to the pay thing then just
have options only unlock when you pay for them...

If you have a module that does these things then it can be a bit more
indepth than a checkbox and increases the gameplay aspects of the
game. An example of depth is that the module to have continuous fire
could include a cooling system that lets you fire the weapons for
longer increasing overall damage (because overheats with continuous
fire would normally happen quicker) and the alternating fire module
could instead have a distributed recharging/reloading system that
allows for faster fire from each weapon, giving higher dps... and this
means that although you get a default setup for weapons in how they
would normally be used (e.g. continuous for lasers) you can change
that and get a slight boost in damage and a choice in how your ship is
set up buying going to the trouble to find and buy the correct
module... kinda like you do when choosing between lasers and bolts in
the first place :)

Johann

onpon4:
> I doesn't need to be but then you could argue, in extremis, that there should be no modules and that you should have a "build your own ship" checkbox window

Stop being silly. We're talking about an option to automate a process you can already do by hand. I just got over explaining this. Requiring specialized hardware for this would be like requiring specialized hardware for autonav or auto-braking. No one would use it (there are tons of much more useful things you can equip!), and the fact that you can't get this convenience without weakening your ship would annoy players. Bad design all around.

onpon4:
Anyway, that's got nothing to do with this pull request. All I did here was remove the "partial delay" aspect of lasers and outfits, and make outfits automatically turn themselves back on when their cooldown cycle ends. The only relevence of stacking lasers is that it'll be a way to fire lasers continuously.

onpon4:
I'd still like to get this merged. I have a general sense of Deiz's objections, but I still haven't seen an answer to this:

> As for removing the purpose of the recharge period, even if that's true (I don't think it is; the recharge period is still putting a limit on how much the laser can be used), a version of the effect can be achieved manually anyway (just assign each beam to a different button; either make half of them primary or half of them secondary, or if you want to go all-out, make each of them separate with instant mode).

So, what's the plan? Is the player going to be forbidden from firing one laser at a time in the future? Because if not, I see no conflict; this pull will cause a temporary slight weakening of beam weapons, but that will get fixed in the future when beams become stacked. It's not like beams handle that well right now, anyway.

I'd also like to highlight the fact that this pull request also includes auto-activating outfits when they finish cooling down, so if the rest of this pull request is going to be rejected, I'd at least like that to happen explicitly so the auto-reactivating feature can be merged separately (there shouldn't be any objections to it, I don't think).

onpon4:
I've removed the feature of keeping outfits on while cooling down. The feature turned out to be incredibly annoying and not useful at all. Additionally, it would require a lot more work to get that right; in particular, the HUD is not set up to handle the condition of an outfit being on and cooling down.

I still think it's best to always require the full delay for outfits and lasers, though, since these delays are delays between uses that, not limits on the amount you can use them in total.

onpon4:
Closing this as I have lost interest, but I still think that beams should always require their full delay, possibly in conjunction with a replacement of holding down the button to keep firing with a toggle switch.

PR #473, opened 2015-04-30
Remove some unused lua variables detected by lua-inspect

ids1024:


bobbens:
Looks good, thanks. Didn't know lua-inspect insisted, seems useful.

PR #476, opened 2015-05-05
Simplify copying of pilot flags when creating pilot

ids1024:
If I understand the code correctly, this is a simpler and better method.


Deiz:
Yeah, that seems fine. I think it's acceptable to rely on the caller to provide valid pilot flags, particularly because the `pilot` Lua module doesn't have direct access to set flags on creation.

I'll let @bobbens have the final say though, as I'm unsure if the manual flag copying was done for a particular reason.

bobbens:
It looks like it should be ok. Manual copying is more of a legacy thing. Thanks.

PR #478, opened 2015-05-06
Use luajit by default

ids1024:
Luajit provides noticeably better performance with no real disadvantage.


bobbens:
Last time we benchmarked it, there was no performance difference and a higher memory overhead with luajit. Have you tried benchmarking it?

ids1024:
No, but I have noticed higher framerates during time acceleration. Was the last time you tested it before luajit 2.0? That version is apparently a rewrite and presumably faster than 1.x.

bobbens:
@deiz benchmarked it when I first implemented luajit support. Setting it as default shouldn't be any issue though as luajit should stay compatible with lua 5.1. Thanks,

PR #484, opened 2015-05-11
Make faction.enemies() return the player's enemies for "Player" faction

ids1024:
With this change, `faction.get("Player"):enemies()` returns the factions with whom the player has a reputation of less than 0.  I would have done this for faction.allies() too, but I'm not sure what number would be appropriate, since a reputation of more than 0 does not mean "ally".  This helps with `fleet_form` at the very least, and may be otherwise useful.

### DIFF
--- a/dat/scripts/fleet_form.lua
+++ b/dat/scripts/fleet_form.lua
@@ -128,7 +128,7 @@ function Forma:reorganize()
  
 
    for _,p in ipairs(self.fleet) do
-      if p ~= pilot.player() and p ~= fleader then
+      if p ~= fleader then
          p:setFaction(fleader:faction())
       end
    end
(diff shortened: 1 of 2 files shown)


ids1024:
Ok, I saw `PLAYER_ENEMY` and `PLAYER_ALLY`, so I implemented `faction.allies()` too.  Since `areAllies` and `areEnemies` already handle the special case of the player faction this way, this should be the correct behavior for these functions.

ids1024:
This PR has been open for three weeks without comment. Is there a specific reason it hasn't been merged (my knowledge of c is fairly rudimentary, so it may not be done quite right) or just lack of time? I think the functionality provided is correct, since that is how areAllies() and areEnemies() are implemented for the player faction. Just asking, don't mean to be rude.

PR #485, opened 2015-05-11
Police AI for the frontier

nenau:
I created a lua file in dat/ai based on the dvaered AI and use it for the frontier Lancelots.

### DIFF
--- a/dat/ai/frontier_police.lua
+++ b/dat/ai/frontier_police.lua
@@ -53,7 +53,7 @@ function taunt ( target, offense )
 
    -- Offense is not actually used
    taunts = {
-       "Alea jacta est!",
+       "Alea iacta est!",
        "Morituri te salutant!",
        "Your head will make a great trophy!",
        "Cave canem!",
--- a/dat/ai/frontier_police.lua
+++ b/dat/ai/frontier_police.lua
@@ -1,0 +1,64 @@
+include("dat/ai/tpl/generic.lua")
+include("dat/ai/personality/patrol.lua")
+
+-- Settings
+mem.aggressive = true
+
+
+-- Create function
+function create ()
+
+   -- Credits.
+   ai.setcredits( rnd.int(ai.shipprice()/300, ai.shipprice()/100) )
+
+   -- Handle bribing
+   if rnd.int() > 0.4 then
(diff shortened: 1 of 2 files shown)


dat/ai/frontier_police.lua
```diff
@@ -0,0 +1,64 @@
...
+function taunt ( target, offense )
+
+   -- Only 50% of actually taunting.
+   if rnd.rnd(0,1) == 0 then
+      return
+   end
+
+   -- Offense is not actually used
+   taunts = {
+       "Alea jacta est!",
```
ids1024: I think that should probably be spelled "iacta," since that is how the word is spelled in classical Latin (the letter j did not yet exist).

But anyway, neat idea having taunts in Latin!

nenau:
Actually, the English use "iacta" and the French use "jacta". No clue why...
I'll correct this.

PhoenixRiver:
> Actually, the English use "iacta" and the French use "jacta". No clue
> why...
> I'll correct this.

Phonetic similarity in each case is closer to the most recently spoken
version of Latin (Medieval)? Though it might have simply been a
mutation through graphic similarity...
Coming from the police the "French" varient sounds more menecing when
spoken in English which seems appropriate for a taunt from the Police
:P

Johann

BTAxis:
Latin has no J, but the letter I doubled for that particular phoneme. That's why we spell it "Julius Caesar", and not "Juljus" or "Iulius".

ids1024:
> Coming from the police the "French" varient sounds more menecing when
spoken in English which seems appropriate for a taunt from the Police
:P

Actually, it would be pronounced (though not spelled) "yacta" in classical Latin.

> Latin has no J, but the letter I doubled for that particular phoneme. That's why we spell it "Julius Caesar", and not "Juljus" or "Iulius".

The Romans did spell it "Iulius". "I" was either a vowel or a consonant, in this case a consonant since it is either between two vowels or at the start of a word followed by a vowel. In much later forms of Latin and modern romance languages, the consonant began to be written with a tail (i.e. as a "J") to distinguish it from the vowel. "Juljus" is nonsense, because that "I" is a vowel, not a consonant.

PR #487, opened 2015-05-13
Remove unused base64 code

ids1024:


Deiz:
For posterity's sake, it appears this has been unused for six years and was only ever used for mission data saving/loading circa 2008.

PR #492, opened 2015-05-19
event : bounty hunters

nenau:
This event occures in systems that belong to "normal" factions : a group of bounty hunters spawn and attack this faction's enemies. That makes the player see that he isn't the only independant pilot who makes the patrol and bountyhunting missions.

I added it as an event because it avoids to change every faction's spawn file.

### DIFF
--- a/dat/event.xml
+++ b/dat/event.xml
@@ -191,4 +191,9 @@
   <chance>15</chance>
   <cond>player.misnDone("Kidnapped") == false and var.peek("traffic_00_active") == nil and system.cur():name() == "Arcturus"</cond>
  </event>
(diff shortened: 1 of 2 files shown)


bobbens:
I like the idea, however, connecting it to enter makes it so it doesn't happen as naturally, only when you enter the system (which the player would likely miss). Adding it to spawn files would allow it to happen at "any time". Should be fairly simple by adding a helper file to add bounty hunters and parametrize it. Then each faction would be able to have it's modified version running. What do you think?

nenau:
You're probably right.
I won't be able to work on it for some time, but I could try to do it later.

PR #500, opened 2015-06-19
Remove documentation that is duplicated on the wiki

ids1024:


bobbens:
I always liked having documentation in the git, but it is duplicated and out of date here and there. Let me think about this a bit more (or let the nostalgia wear off).

bobbens:
Old wiki is dead, don't think documentation is duplicated ;)

PR #503, opened 2015-06-20
Merge master into race

ids1024:
I was interested in trying the race branch, but without building an ancient version of naev, so I merged it with master. There were several conflicts to handle, so I thought I would submit a pull request so no one ends up duplicating that work.


bobbens:
I assume it's ok. thanks! ;)

PR #508, opened 2015-07-19
New large fuel pod graphic to differentiate from medium fuel pod

ids1024:
Just a simple graphic with four fuel pods instead of just one or two.


pydsigner:
Just a note: the license for the old `large_fuel_pod.png` needs to be updated to refer to `medium_fuel_pod.png`

ids1024:
@pydsigner Good point. That should be fixed.

How exactly should a line in ARTWORK_LICENSE be formatted for derivative works? My graphic is just based on the ordinary fuel pod (though I rendered from the .blend, so it is not based on fuel_pod.png per se)?

Deiz:
We don't do a good job of crediting multiple-author work currently.

There are two blocks that follow a `First Author, License: Name (with modifications by Second Author (License: Name))` scheme which I'll modify and follow for the time being, though it's a bit clunky.

PR #509, opened 2015-07-19
Add leodudepal's pirate license graphic

ids1024:
https://forum.naev.org/viewtopic.php?f=9&t=867
He did other license graphics, but I am not adding them right now
because they do not show what the purpose of the license is, and I think
there should be some indication of that.


bobbens:
Looks great, thanks!

PR #511, opened 2015-07-31
[WIP, not ready for merge] 3D Ships

ids1024:
At least at the moment, this is just the old `3d-ships-working` branch updated with the changes in naev, and with scaling added (for landing/takeoff). I also added 3d graphics for the Admonisher so anyone can test it without figuring out the right blender export options first.

I thought it might be good to have a pull request here for reference. I may try to improve the code further, but I do not really have any experience with opengl and whatnot, so that may not happen.


ids1024:
I finally managed to script my process for exporting to the .obj/.mtl, instead of manually setting the export options in blender and editing the files manually. This works with the Admonisher, but not yet with all ships. The scale of the ship is still not right. I don't know how to determine the right scale value to use to have the ship sized correctly.

https://gist.github.com/f143aba3d8b8584ff06c

bobbens:
Scale should be best left to a parameter in the ship xml definition. That way it is easy to change and play around with scales. Still have to see if there is a simple collision library or something we can use.

ids1024:
@bobbens Ok, the scale is now configurable by setting `_3d_scale` in the `gfx` tag in a ship xml file. The default value is `.01`.

ids1024:
@AvanWolf Do you have the .blend file (or other 3d model) for the Thurion Ingenuity?

UncombedCoconut:
This is merged via #1895. Follow-up will be tracked in #1894. Yay!

PR #513, opened 2015-08-19
Actually test for possession of pirate landing pass

ids1024:
Previously, the pass did nothing, and a variable was set to implement
the functionality. This seems like a slightly cleaner method.


bobbens:
Looks good, thanks. Probably implemented the old way for legacy reasons...

PR #515, opened 2015-08-23
add bountyhunters patrols in Dvaered space

nenau:
In Dvaered controlled space, some bountyhunters spawn alongside dvaered ships.
I followed Booben's advice to write an helper in the spawn directory.
This was heavier than I thought, I had to change the common spawn file in order to make it possible to use the function addRaw.

### DIFF
--- a/dat/factions/spawn/common.lua
+++ b/dat/factions/spawn/common.lua
@@ -69,7 +69,12 @@ end
 scom.spawn = function( pilots )
    local spawned = {}
    for k,v in ipairs(pilots) do
-      local p = pilot.add( v["pilot"] )
+      local p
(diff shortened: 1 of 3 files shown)


bobbens:
Looks good. For future reference it's "Bounty Hunter" not "Bountyhunter". I also toned down the percent from 10% to 5%. Thinking still about making it even lower.

On a side note, the way we generate uniques (dat/events/neutral/unique.lua) is still like how you originally did the patch. It should be done the current way you're doing it.

bobbens:
BTW did the unique thing. I allowed hooking a function to make it much more flexible. Look at 8cb67c2e791dada031015257a2b55d0ba72be268 .

PR #516, opened 2015-08-24
replace setSpeedLimit in fleet_form

nenau:
I use a PD controller in order to avoid using setSpeedLimit in the formation hook.

### DIFF
--- a/dat/scripts/fleet_form.lua
+++ b/dat/scripts/fleet_form.lua
@@ -503,13 +503,18 @@ function Forma:control()
    lead_stats = self.fleader:stats()
    for i, p in ipairs(self.fleet) do
       if not (p == self.fleader) then
-         if p:pos():dist(posit[i]) > 300 then
-            p:setSpeedLimit(0)
-         else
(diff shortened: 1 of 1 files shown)


ids1024:
Quick testing suggests that this works better (I didn't see the circling behavior that had previously been an issue), so this looks like a good change.

ids1024:
Note that there are other uses of `setSpeedLimit()` in fleet_form. Perhaps they could be changed as well.

nenau:
I don't know how to avoid using setSpeedLimit for the fleader, because he is directly controlled by his AI.

ids1024:
Ah yes, I didn't think of that.

bobbens:
Why PD and not PID? No integral component means it will never actually reach the target orientation.

nenau:
PID are usually used when we need to be very accurate, which is not our case imo. And having an integral corrector introduces two problems : 

* If the ship comes from far away, he will exceed it's goal
* I needed to use a memory to store the integral

Of course, there are solutions to avoid the first problem, but, this would increase the code for a minimum gain.

bobbens:
OK, sounds good.

PR #517, opened 2015-08-24
Ai.c improvements

ids1024:
This seems to me like a good first step for improving the ai module. Needless to say, it is not the only improvement to be made.


ids1024:
Come to think of it, do not merge this yet. I will keep pushing changes to further improve ai.c.

bobbens:
For legacy reasons the AI uses a lot of custom stuff. I always wanted to incorporate the standard Lua modules we have into it. I mean there is a nlua_pilot.c and stuff which should be able to work directly in the AI. This would also allow doing coolre things.

ids1024:
@bobbens Yep, I'm working now on removing the duplicate functions and having the ai code use the standard ones.

ids1024:
I'm not quite sure which functions should be kept and which removed. For example, `ai.dist(target)` could be replaced by something like `ai.getPilot():pos():dist(target.pos())`, but `ai.dist()` is used in quite a few places and that is a bit long and I'm sure at least somewhat less efficient.

It is similar with the `ai.rel*()` functions.

ids1024:
Ok, I think this is in somewhat good shape now, though perhaps some testing is needed to make sure everything is working. I have removed the WIP from the title.

ids1024:
@bobbens @Deiz @BTAxis How do the changes here seem?

bobbens:
Haven't looked into it in too much detail, but it looks good. Some of the nice shortcuts have been removed and that makes you use chained commands, but I think it's perfectly mergeable as is. Cheers,

ids1024:
@bobbens A few shortcuts have been removed, but I think it is worth it for the api consistency. I left in shortcuts like `dist` and `relvel` which could be done in lua, but would be somewhat complicated.

PR #518, opened 2015-08-28
Add bounty hunters to more factions

nenau:
I also introduced a parameter for bounty hunter percent in order to make it easier to change.


bobbens:
OK, not sure every faction should be happy with bounty hunters though. Though it should be easy to disable (for example setting pbm to 0 after the require). Thanks,

PR #520, opened 2015-08-31
fix land/jump hooks in fleet_form

nenau:
I think the use of land and jump hooks was wrong, what caused an error. I also made sure to remove the hooks over fleet's pilots when the fleet is disbanded.

I think it would be better to use the system given by the jump hook instead of looking for the closest jump, but I await your opinion.

### DIFF
--- a/dat/scripts/fleet_form.lua
+++ b/dat/scripts/fleet_form.lua
@@ -199,26 +199,14 @@ end
 
 -- Jump hook. Ensures the entire fleet jumps if the fleader jumps.
 -- Takes an ID that tells this function which formation the pilot belonged to.
-function Forma:jumper(jumper)
+function Forma:jumper(jumper, jumpoint)
    if jumper == self.fleader then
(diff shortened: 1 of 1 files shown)


l0k1:
When fleet_form was originally written, land and jump hooks didn't pass the asset and jump points, which is why it was written the way it was. This update looks good though (haven't tested).

nenau:
I have an idea for controlling : I could add a function that manages the tasks of the fleet and ensures that after a battle, the fleader keep following the directives.
The mission script wouldn't use directly fleader:control but instead call this function.

What do you think about that ?

l0k1:
That actually sounds pretty awesome. If you can do it, go for it.
Something to keep in mind though: the fleet_form script was more designed as a band-aide for mission designers until proper fleet support was implemented through AI redesign. But that's probably a ways out still. =]

bobbens:
Yes, something better integrated in the engine would be better. However, that's a while off I think. Patch looks good, thanks.

PR #521, opened 2015-09-04
Fixed a typo.

onpon4:
Wrong name at one point, was causing one possible outcome of bounty
hunts to fail.


bobbens:
Looks good, thanks.

PR #522, opened 2015-09-04
Reduced the active and cooldown times of emergency shield booster.

onpon4:
This gives the outfit more of a "pulse" effect, which is much more
useful since the whole point of the outfit is to dramatically boost
shield usage at a cost of dramatically boosting energy usage.

### DIFF
--- a/dat/outfits/activated/emergency_shield_booster.xml
+++ b/dat/outfits/activated/emergency_shield_booster.xml
@@ -10,7 +10,7 @@
   <limit>emergency_shield_booster</limit>
   </general>
  <specific type="modification" group="8">
-  <active cooldown="3">1</active>
+  <active cooldown="12">4</active>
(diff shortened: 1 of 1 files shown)


bobbens:
Maybe toning it down would be better to something like 12 second cooldown, but I like the idea of somewhat long cooldowns (>10 seconds) as they force you to make more of a tactical decision with higher repercussions than something that can be used every 3 seconds.

onpon4:
How's that? It's half of what it originally was now.

bobbens:
Ah sorry, misread the patch. I'm not too sure on the implications. Will have to ask @Deiz as he's much more familiar with combat then me.

Deiz:
I haven't played with active outfits much, but halving the original duration and cooldown seems fine.

bobbens:
Ok, pulling then. Thanks!

PR #523, opened 2015-09-04
Made enemy faction mod in faction_modPlayer primary, not secondary.

onpon4:
The purpose of "secondary" modding is to reduce the effect of modding
allies. It makes no sense to also apply this to enemies, as this is
in the opposite direction and there's Lua code to handle it.


onpon4:
I've come up with a better way to do what I wanted to do with this change, so I'm closing this pull request (and it's already reverted in my fork).

PR #524, opened 2015-09-04
Removed secondary faction hits from the Empire.

onpon4:
What this effectively means is when you attack House Goddard or
House Dvaered, your standing with the Empire is unaffected. Only
when you actually attack Empire ships is your standing affected.
In particular, this makes it possible to be friendly with the Empire
even when you side with the FLF.


onpon4:
I've got a better idea for how to implement this. Closing, will open a new pull request shortly.

PR #526, opened 2015-09-05
Refer to time units in a way that is easier to understand.

onpon4:
Basically, this causes all real-world references to the time system
to be the following, instead of the standard "STU"/"STP"/"SCU":

- "Second": Used instead of "STU".
- "Minute": Used instead of "hSTU" (100 seconds).
- "Hour": Used instead of "STP" (10,000 seconds).
- "Galactic Day": Used instead of "daSTP" (100,000 seconds).
- "Galactic Year": Used instead of "SCU" (50,000,000 seconds).

For brevity, "galactic day" is typically shortened to "day", and
"galactic year" is typically shortened to "year", in casual
conversation.

The result of this change is that understanding times is *much* easier.


ids1024:
I personally prefer your option 2 from https://forum.naev.org/viewtopic.php?f=6&t=1039, but I do agree that the current abbreviated forms are confusing.

BTAxis:
I generally agree with the sentiment, though I had already introduced the shorthand "cycle" for the SCU, which I felt gave it a little more presence as a non-Earth measure of time.

PhoenixRiver:
I also agree with the setiment to a degree but also would prefer
option two (which I have already used in writing dialog etc.) - and it
just feels wrong and immersion breaking that our time units would be
used in the Naev universe...
River *<:@)

onpon4:
I don't get the appeal of made-up terms for something so mundane as everyday time measurements. I don't think any other work of science fiction does this; correct me if I'm wrong. I think all this does is steepen the learning curve of the game for no good reason.

There were already several references to real time units. Do you guys want those converted to "periods" and "cycles", too? (And what about seconds?)

ids1024:
> I don't think any other work of science fiction does this; correct me if I'm wrong.

Battlestar Galactica comes to mind immediately as something with fictional time units. I'm sure it is not the only.

BTAxis:
The X series, too, uses fictional time units (loosely based on Earth time units).

PhoenixRiver:
> I don't get the appeal of made-up terms for something so mundane as everyday
> time measurements.

Isaac Asimov, Arthur C. Clarke, Larry Niven and Harry Harrison have
all used alternate terms for time measurements (though Asimov's have,
in some cases, now been adopted by astronomers) in one or more of
their stories - particularly ones that relate to story universes where
our time units make no sense (which is prevelent in Scifi/SF even if
writers decide to go with the old naming conventions that make no
sense any more (e.g. Babyon Five if I remember rightly - how alien
races are supposed to understand our time measurements I don't know!
Though this is not relevant for Naev (so far?))
River *<:@)

onpon4:
Well, I'm not in favor of using "units", "periods", and "cycles", and there are *several* references to real time units that already existed in Naev, so I won't be making that change. But if you really want to do that, I'd suggest you base it off of my work here, because finding all instances of the terms "SCU", "STP", and "SCU" was an incredibly tedious chore.

ids1024:
> ...finding all instances of the terms "SCU", "STP", and "SCU" was an incredibly tedious chore.

Um, can't you just use something like `grep STP $(find -type f)`, or use `sed -i` to change all instances?

onpon4:
What you say is true. I didn't know about the find command.

Corax:
Let me chime in with another voice for keeping units, periods, and cycles, and dropping abbreviations where appropriate. I think they do add quite a bit of atmosphere, and their contribution to the overall learning curve is rather minute (haha).

PhoenixRiver:
Since I've a bit of time to start proofreading planet descriptions again I'll check these as I go - can't guarentee I'll not miss any but it should reduce the number to something miniscule :)

onpon4:
I still disagree with the idea of using these weird time names, but it's clear that my preference is not going to be adopted and this is no longer a part of my master branch. Closing.

PR #527, opened 2015-09-05
add a dvaered warlords battle event

nenau:
In this event, the player is caught in a battle between two dvaered fleets. He can watch it, go away or help one of them, in order to get a reward.
Note : the text needs a grammatical proofreading.


l0k1:
+1 for using the fleet_form.lua. This makes me happy.

PR #529, opened 2015-09-05
Made the "Independent" faction invisible and effectively static.

onpon4:
This is meant to prevent the situation where various "independent"
systems and planets appear to have some sort of allegiance.

It would be better if each independent system and planet had its
own hidden faction, but there are so many independents, this would
be impractical and bloated.

(The faction is "effectively" static instead of actually static because making it actually static would cause your reputation with it to show up as "???", which is undesirable.)


ids1024:
Wouldn't that mean that if the player becomes a pirate and constantly attacks independent ships, they will never become enemies with the player? So, independent systems planets will always be landable, and independent ships never attack the player unless attacked? I'm not sure that is desirable.

onpon4:
Yeah, that's exactly what that part means.

nenau:
IMO there shouldn't be too hard to give each independant system its own faction, based on the independant faction : 

* Add them to faction.xml
* Add them as enemy of pirates
* Write small wrappers in the spawn directory, that call the independant, just changing the faction
* Manage the changes in NPC bar event

I could do it.

PhoenixRiver:
> IMO there shouldn't be too hard to give each independant system its own
> faction, based on the independant faction :
>
> * Add them to faction.xml
> * Add them as enemy of pirates
> * Write small wrappers in the spawn directory, that call the independant,
> just changing the faction
> * Manage the changes in NPC bar event

That would work, though confeptually would independants talk to each
other at least or have some sort of loose confederation(s) so maybe
they could be in a few different factions based on geographic area
maybe and just with an aversion to pirates (the reason they'd talk in
the first place if nothing else)... just a thought.

River *<:@)

onpon4:
I've just converted Ulios from the Independent faction to its own faction:

https://github.com/onpon4/naev/commit/678b4f54cf4ceb336825e4c6e1b837e99704a589

In total, the actual conversion took somewhere between 15 and 20 minutes. (It actually took more time because I was testing to make sure everything worked.) This needs to be done for *every* independent system, planet station, and what have you, so we're talking a few hours of work total. There are around a dozen independent planets, to my recollection, and I'm not aware of any way to reliably find all of them. So this is going to have to be a long-term project done one step at a time, not something done all at once.

onpon4:
Oh right, grep in the assets directory. There are 16 remaining.

BTAxis:
Another approach would be to write code to support this sort of thing. You'd define a "federation" faction or something like that, and each asset belonging to that faction gets assigned its own subfaction so standings are separate but all assets are still considered part of the same faction for purposes of presence and scripting.

The idea of subfactions has been floating around for some time - we could use it for the Frontier, DV Warlords, military vs police forces, etc. There's no detailed proposal though, but it's something we can discuss.

nenau:
@onpon4 : you are right, it's quite heavier than I first thought...
About the standing script : I think, it is not necessary : for example, the frontier doesn't have it's own.
Maybe several assets could have the same faction (for example planets in the same system).

nloewen:
I tried putting together a subfaction system years ago but eventually gave up on it. I think what I was trying to do was a little beyond my skills at the time. Haha! I'd love to see a proper implementation of that go in. I think it would add a lot to the game.

ids1024:
+1 for subfactions. It might make sense for other factions as well; for example, the Dvaered could have some subfactions controlled by individual warlords.

onpon4:
I don't know about that. The thing is, it's not universally true that an independent planet is completely independent, it might be a part of an independent *system*. Or, like the Wringer, it might have no sense of union at all. (The Wringer should probably continue to use the static Independent faction, and it could be that others would be best off sticking with it as well.)

For most of the other possible benefits, it seems like it would be possible to do this entirely in Lua. A special faction_standing definition that heeds to another particular other faction's standing, for example. It depends on what exactly you want to do with it.

onpon4:
I still think this pull request should be merged.

Ideally, every independent planet and system should have its own (hidden) faction. But considering the amount of added work and maintenance, and not to mention having to go back to specific planet at the end of patrol and bounty missions (which makes sense, but could be annoying), I think just using a static "Independent" faction as I have done here is best.

bobbens:
After giving it a bit more thought, I will pull. I do think in the mid/long run that something fancier should be done though. Thanks!

PR #532, opened 2015-09-06
Made the Empire not care about its allies being attacked.

onpon4:
The effect of this is that if the player attacks House Goddard or House
Dvaered, the Empire will not become hostile to the player. Only attacks
on the Empire's own ships will provoke this response.

This change is primarily meant to make it possible to align with the FLF, but
maintain good relations with the Empire.

This depends on #531 to have any effect; it uses new variables defined there.

### DIFF
--- a/dat/factions/standing/empire.lua
+++ b/dat/factions/standing/empire.lua
@@ -18,7 +18,7 @@ function faction_hit( current, amount, source, secondary )
    local start_standing = _fthis:playerStanding()
    local hit = default_hit(current, amount, source, secondary)
    if (source == "distress" or source == "kill") and secondary and amount < 0 then
-      if start_standing >= sec_hit_min and start_standing + hit < sec_hit_min:
+      if start_standing >= sec_hit_min and start_standing + hit < sec_hit_min then
          hit = sec_hit_min - start_standing
       end
    end
--- a/dat/factions/standing/empire.lua
+++ b/dat/factions/standing/empire.lua
@@ -9,12 +9,18 @@ _fdelta_kill     = {-5, 1} -- Maximum change constraints
 _fcap_misn     = 30 -- Starting mission cap, gets overwritten
 _fcap_misn_var = "_fcap_empire"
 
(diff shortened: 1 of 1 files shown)


bobbens:
I'm not too sure about this. Being allies with the FLF should have implications to some extent with the Empire in my opinion.

onpon4:
bobbens, it was you who said, black and white, that the Empire and FLF are *not* enemies. Then you clarified that the Empire only feels obligated to help the Dvaereds in individual battles against the FLF, but doesn't engage the FLF on its own. Now you're saying you do want the player to become enemies with the Empire by joining the FLF? This doesn't make sense. (Note: just doing normal FLF missions *will* cause you to become enemies with the Empire, as it currently stands.)

I'm OK with an FLF-aligned player being enemies with the Empire, but only on the condition that either the player caused it himself (not as a result of FLF duties), or that the FLF ships help him against Empire ships. You can't have the player singled out as the only one in a faction hated by another faction for no special reason. It's unfair and annoying.

bobbens:
Yes. I'm a bit torn on the issue for several reasons. I can see several ways to deal with this:

1.) Empire and FLF are full enemies. FLF is considered terrorism.
2.) Empire doesn't care about FLF and doesn't want them bothering them.
3.) Empire encourages FLF and secretly supports them to keep Dvaered week. However, they don't make it public.

If we go on the line of 3.) then your solution is fine. If we go on 2.) then it can be implemented so that if you cause too much trouble, it will annoy the empire to some extent, however, not to become enemies. So basically something like if you have >30 empire standing it'll bring it down, but not below. Of course 1.) is current implementation. Personally I feel like 2.) is the best option, but of course this is up for discussion. Maybe it would be better to just make them not enemies to keep it simple for now.

onpon4:
I kind of like that idea. I've implemented it: rather than just never getting less reputation from attacking Empire allies, it happens only to a certain point (though I chose 10 instead of 30). How's that?

bobbens:
Looks good, thanks.

PR #534, opened 2015-09-08
Added medium and large cargo pods

PhilmacFLy:
, similar to medium and large fuel pods.

Same price increase, same mass increase and same capazity increase.

Didnt do the gfx work but shouldnt be to hard.


bobbens:
Looks good, thanks.

PR #535, opened 2015-09-09
Modified the aim method

nenau:
The way the ai computes the flying time of the ammo was uncorrect (I think).
I changed this and I think that turrets are now deadlier.

### DIFF
--- a/src/ai.c
+++ b/src/ai.c
@@ -2220,11 +2220,12 @@ static int aiL_aim( lua_State *L )
    double x,y;
    double t;
    Pilot *p;
-   Vector2d tv, approach_vector, relative_location;
+   Vector2d tv, approach_vector, relative_location, orthoradial_vector;
    double dist, diff;
    double mod;
    double speed;
    double radial_speed;
+   double orthoradial_speed;
 
    /* Only acceptable parameter is pilot id */
    id = luaL_checklong(L,1);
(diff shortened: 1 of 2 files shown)


onpon4:
This seems to me like the most likely possible source of the wildly inaccurate shots I've been getting lately, as I reported in #543. Has anyone actually checked this before merging it?

onpon4:
I notice that the new line ``vect_cset(&orthoradial_vector, VY(cur_pilot->solid->pos) - VY(p->solid->pos), VX(p->solid->pos) - VX(cur_pilot->solid->pos) );`` puts Y values in the X field and vice-versa. I'm going to see what reversing that does; seems to be likely as a source of this kind of problem.

onpon4:
Well, that change doesn't help. Must be something else.

nenau:
Hello,

The orthoradial vector is right because you have to change coordinates in order to get a vector that is perpendicular to the first one.

The method I implemented is more complex than the previous one, and that's why I had some issues, but I thought I had totally debugged these issues, the last commit about it was merged a few hours ago.

Maybe it's because the shots were shot before the ship was disabled, or because your ship has a huge thrust (the aim method assumes the velocity is a constant)...

For now, I can't reproduce the issue you described, I will keep trying...

onpon4:
There's definitely something about the ships being disabled that causes shots to sometimes go wildly off the mark, because the pattern I'm seeing is: accurate shots, but then the ship gets disabled, and any further shots made after that go in this completely wrong direction.

That's not the only time, though. For example, I was using four Turbolasers and two turreted Vulcans on my Pirate Kestrel, and there was one time when the Turbolasers were aiming just fine, but the Vulcans were something like 60 degrees off the mark.

onpon4:
Was there any particular reason you made the system more complex? I'm aware of one oversimplification that was causing distances that were in range to be considered out of range, but that was the only problem I'm aware of the old system having...

nenau:
There were 3 reasons : 
* As you said : sometimes inrange ships were considered out of range.
* Sometimes the opposite happened (rockets often used to explode before hitting).
* It was quite easy to run at constant speed around a ship without getting hit.

As the vulcan ammo are very slow, they are sometimes shot in a strange direction when you are close to a fast target (in order to hit it if it continues at constant velocity).

I'll try to reproduce what you described...

onpon4:
> It was quite easy to run at constant speed around a ship without getting hit.

What are you trying to do to prevent this, exactly? A "constant speed around a ship" involves frequent turning adjustments. I don't see how you can possibly predict this without either making a potentially incorrect assumption or mind-reading.

The most accurate I think you could possibly make auto-aiming be is to assume that the ship is going to keep moving at its present speed. That's not always going to work, of course; the pilot could accelerate or change direction, but that's out of your control. Even if you try to predict where the ship will be if it keeps turning at the same rate, the pilot could just stop turning, and the effort would become meaningless. Same goes for thrusting, and there's also the complexity of turning and thrusting at the same time. Plus, Naev's aiming isn't even precise, anyway; weapons that have heated up (which happens often in battle) randomly change direction.

There's another problem with trying to be extra-intelligent: it makes the behavior of auto-aiming more difficult to predict. If all it does is basic shot-leading, I as a pilot using auto-aiming turrets can tell when it is and isn't going to work. It's up to me, then, to decide when to start shooting. But if the auto-aiming is trying to predict other factors, this is much more difficult, and it can be extremely frustrating if the predictions are also flawed. Just as an example, you'd think that you can hit the ship easily if it's coming almost straight toward you, right? But this is not the case anymore. When I shoot at a ship that is coming almost straight toward me with that mentioned Turbolaser/Vulcan combination, it's very common for a few shots to miss. It's so bad that my Pirate Kestrel configuration using these outfits, which was previously quite deadly, is now effectively crippled (it's not so bad when it happens once, but this really adds up, especially considering a ship coming right toward you is a lot more immediately dangerous than one that's circling around you).

I can think of one improvement that could be made *specifically* to the pilot AI, but it's in decision-making, not aiming: the improvement that could be possible is figuring out whether it's possible to accurately aim in the given situation, and deciding whether to fire or not. That would prevent the AI from wasting ammo needlessly as it previously did (and still currently does).

PR #538, opened 2015-09-12
Changed the inrange computation method

nenau:
The goal is to tend to avoid wasting ammo : 
* Added a function in pilot_weapon.c that estimates the flight time of ammo.
* Compare this value to the aoom duration to decide if the weapon shoots.
* In order to avoid re-computing the time in aimTurret, I pass the time as argument

The results : 
* Bolt weapons shoot at longer range when the enemy is approaching.
* Bombers don't waste ammo when they are attacking a fleeing enemy.


bobbens:
Looks good, thanks.

PR #540, opened 2015-09-15
Bugfix

nenau:
As now, the AI uses pilots instead of IDs, this change has consequences in nlua_pilot.


ids1024:
Ah yes, this looks correct. I didn't think to change this when I was adjusting the ai module. Nice catch.

bobbens:
Thanks!

PR #541, opened 2015-09-15
Replaced use of the "static" tag with a "static" standing script.

onpon4:
The reason is using the "static" tag now guarantees standing to always
be represented as "???", as this text is now defined in Lua.

### DIFF
--- a/dat/faction.xml
+++ b/dat/faction.xml
@@ -1,9 +1,9 @@
 ﻿<?xml version="1.0" encoding="UTF-8"?>
 <Factions>
  <faction name="Dummy">
-  <static />
   <invisible />
   <player>0</player>
+  <standing>static</standing>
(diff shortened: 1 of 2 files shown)


bobbens:
Looking at the patch more carefully. There are a few things in the code that change. For example, the <static/> tag makes it so the faction is not saved and the standing is reloaded each time. This means if the default value changes, there is no issue with saves. Other than that, there are a few other minor optimizations since it is a flag on the C side. So I'm a bit broken on reverting it. :/

onpon4:
Perhaps *both* the static tag and the "static" standing script should be used. This would require removing that warning about having both the static tag and a standing script. Would you be interested in that?

bobbens:
Yes, that would work. A bit redundnt though.

PR #542, opened 2015-09-15
Only make rival bounty hunters hail the player once.

onpon4:
It doesn't really make sense for them to be so insistent. They're just giving you an offer, a token gesture of sorts.


bobbens:
I agree, thanks!

PR #544, opened 2015-09-16
Fixed some issues/bugs in the missions (mainly Shark campaign)

fartoverflow:
e.g. changing nonexisting planet Sindbad


nenau:
Actually, the existence of Sindbad depends on an unidiff. An error in the mission was causing to call Sindbad even in the case where it didn't exist.
I think I will do one day a pull request to clean up the Shark campaign I wrote as I was not so familiar with the lua API.

PR #546, opened 2015-09-17
Add a follow function with pd controller

nenau:
This function is callable by the pilot lua API by adding the keyword "true" in the follow function's arguments.
It uses the AI memory to pass some parameters.
In my opinion, this could also be used in the future in a fleet AI.

### DIFF
--- a/dat/ai/include/basic.lua
+++ b/dat/ai/include/basic.lua
@@ -175,18 +175,15 @@ function follow_accurate ()
       return
    end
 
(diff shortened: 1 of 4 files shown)


bobbens:
Looks good, but it's not clear exactly what mem.radius and mem.angle do from the description. It seems like they are variables that should be updated and not hyperparameters. Could you make that a bit more clear please?

bobbens:
Looks good, thanks!

PR #547, opened 2015-09-17
Moved a call after checking the pilot exists in ai/tpl/generic

nenau:
This prevents an annoying error to occur in case the target died/jumped.


bobbens:
Looks good, thanks!

PR #548, opened 2015-09-17
Added back in the "static" tags to static factions.

onpon4:
This also involves removing a (now obsolete, actually) warning that you got when you had both the static tag and a standing Lua script.


bobbens:
Looks good, thanks!

PR #552, opened 2015-09-20
Made the AI handle the afterburner

nenau:
* Added a function in ai.c that triggers a weapon set (could not use weapset because it must manage activation and deactivation)
* 2 functions : hasafterburner and hasjammers that work like hasturrets
* Had to move stuff in pilot.c in order to make the afterburner trigger for npcs
* Changed sound_play into sound_playPos in order to avoid spam
* In pilot_weapon.c, removed the condition that pilot must not be on manual control if is a npc

### DIFF
--- a/dat/ai/include/basic.lua
+++ b/dat/ai/include/basic.lua
@@ -312,12 +312,11 @@ function __run_target ()
    local dir   = ai.face(target, true)
    ai.accel()
 
-   --[[
-   -- Todo afterburner handling.
-   if ai.hasafterburner() then
-      ai.afterburn(true)
+   -- Afterburner handling.         
+   if ai.hasafterburner() and ai.pilot():energy() > 10 then
+      ai.activate( 8, true )
    end
-   ]]--
+
    return false
 end
 function __run_turret ()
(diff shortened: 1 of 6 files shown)


ids1024:
> Added a function in ai.c that triggers a weapon set (could not use weapset because it must manage activation and deactivation)

I don't know, is it really necessary to add a function? Perhaps `weapset` could be modified to take a second boolean parameter, and it could be interpreted as `true` if not given. Or if the api has to be changed, you can do a search and replace through the lua code.

But anyway, it seems good to add this functionality to the ai!

nenau:
The problem is that the AI needs to know if the outfit is active before taking the decision to "push the key", while the idea behind weapset is just to push the kew without thinking in order to trigger a weapon.

To have an unique function, it would be necessary to test if the chosen weaponset is an outfit or weapon type, and thus complicate the function.

But as I am not an experienced C programmer, maybe I'm not aware of an advantage of using only one function...

ids1024:
Hm, well I don't know if it would necessarily be better to have one or two functions. The argument in favor is just that it is simpler.

nenau:
On reflection, what you say makes more sense. Let me some time and I will do it.

bobbens:
Looks good, thanks.

PR #553, opened 2015-09-21
Added a Dvaered championship mission

nenau:
The player takes part to a fighter dogfight challenge on Dvaer which goal is to disable the opponents.
As usual, proofreading is welcome.


bobbens:
Have to  test it, but looks good, thanks!

PR #555, opened 2015-09-25
Fixing jumpout bug.

l0k1:
After I had "completed hyperspace prep", and was speeding away, the Sirius Preach event started, and after the jump was completed (and I was out of the event system), I was no longer able to control my ship. This should fix that. Resetting the camera and the cinematics was probably overkill, but I decided better safe than sorry.


bobbens:
Looks good, thanks!

PR #556, opened 2015-09-27
Made the AI wider use missiles

nenau:
This commit tends to have weak ships with missiles keep off dogfight and use their launchers.

I had also to do some changes in the hardcode : 
* Add a function that gives the speed of a weaponSet
* Change aiL_relvel and aiL_follow_accurate to increase their abilities ( add an optional parameter, lua api doesn't change )

And changes to avoid having out of ammo pilots trying to shoot : 
* Have empty launchers being ignored in the weapset range calculation
* Re-compute the range when the last ammo has been shot
* Initialize all the weaponset's range at 0 in weaponAddauto (If a script removes a weapon to a pilot, the range of this weapon must be forgotten).

### DIFF
--- a/dat/ai/include/atk_generic.lua
+++ b/dat/ai/include/atk_generic.lua
@@ -99,20 +99,95 @@ end
 -- Enters ranged combat with the target
 --]]
 function _atk_g_ranged( target, dist )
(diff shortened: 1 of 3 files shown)


bobbens:
Looks good, thanks.

PR #557, opened 2015-09-27
Bugfixes and such, again

onpon4:
This one fixes the types used for a couple of Lua functions (was using a double when it was just boolean; changed this to int), and gives the Frontier and House Goddard their own standing Lua files rather than sharing one with the Independent faction.


bobbens:
Looks good, thanks.

PR #560, opened 2015-10-09
Fix LUA_CFLAGS when using luajit

ids1024:
This was creating a value of "-I/usr/include/luajit-2.0-DHAVE_LUAJIT"
for LUA_CFLAGS, which did not work correctly. I noticed this after
starting again from a clean build tree. I don't know how it managed to
build before.


bobbens:
Looks good, thanks!

PR #562, opened 2015-10-12
Add a careful goto function

nenau:
This commit makes not-army ships try to avoid their enemies before being attacked. 
This impacts right now the civilians, traders, pirate and flf, in the following tasks : attack (at long range), land and jump.
All the light ships also use this new function in runaway.

### DIFF
--- a/src/ai.c
+++ b/src/ai.c
@@ -1897,16 +1897,15 @@ static int aiL_face( lua_State *L )
 static int aiL_careful_face( lua_State *L )
 {
    LuaVector *lv;
-   Vector2d *tv, F, F1, pilot_relpos, target_relpos;
+   Vector2d *tv, F, F1;
    Pilot* p;
    Pilot *p_i;
    double k_diff, k_goal, k_enemy, k_mult,
-          k_dot, d, diff, dist, factor, cur_k;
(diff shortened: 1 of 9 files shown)


src/ai.c
```diff
@@ -1874,6 +1876,111 @@ static int aiL_face( lua_State *L )
...
+      /* Valid pilot isn't self, is in range, isn't the target and isn't disabled */
+      if(p_i->id != cur_pilot->id && pilot_inRangePilot( cur_pilot, p_i) == 1  
+            && p_i->id != p->id && !pilot_isDisabled(p_i) )
+
+      {
+           dist = vect_dist(&p_i->solid->pos, &cur_pilot->solid->pos) + 0.1; /* Avoid / 0*/
+           k_mult = pilot_relhp( p_i, cur_pilot );
+
+           /* Check if friendly or not */
+           if (p_i->faction == FACTION_PLAYER)
```
bobbens: Is there any specific reason you check for player instead of letting areEnemies handle it? (besides small speed boost it may give)

nenau:
No, I simply didn't knew areEnemies could handle it.
I will change this.

bobbens:
Looks good, thanks!

PR #567, opened 2015-11-06
fixed the escort attack function

nenau:
The escort functions were still sending IDs to the lua scripts instead of pilot objects.


bobbens:
Nice catch, thanks!

PR #568, opened 2015-11-07
Trade imagearray

Mutos:
- A gfx_store tag in commodities, referencing an image in /dat/gfx/commodity,
- Images folder taken from Vega Strike,
- Change in commodity structure, parsing and freeing code to take this tag into account,
- Display Commodities Exchange Tab as an array of images instead of text-only list.


bobbens:
OK, looks great. I think for the outfits never actually sold, just setting the garphic to _default would work. Since this is a fairly big change, I'm going to merge after 0.6.1 gets released, which should be soon hopefully.

Mutos:
I close this one and I'll create a new Pull Request, far cleaner, from current project state including #570, and with less commits.

Mutos:
Oh, I just found I can close it myself ^-^

PR #570, opened 2015-11-08
Fixed 2 bugs in Image Widget : see comment for details

Mutos:
Widget border takes image size instead of widget size ; Entire widget not displayed when image is NULL


bobbens:
Looks good, thanks!

PR #581, opened 2015-11-20
Removed military ships from Dolmen in Pastor

nenau:
This issue is frequentiy reported by players. I hesitated between that and making Dolmen unlandable in early game.


bobbens:
Looks good, thanks.

PR #585, opened 2015-11-29
Stderr & Stdout timestamped duplicates

Mutos:
First write to hard-named stderr and stdout, then copy them to timestamped files.

Allows for auto-reload of err/out files in text editor while keping track of runs.

Could later be made customizable in conf.lua, then in the Options panel :
  - A checkbox for writing stderr/out at all (redirect_file, already exists)
  - If checked, 2 sub-checkboxes :
    - Keep if stderr is empty (keep_if_stderr_empty)
    - Create timestamped duplicates (timestamped_stderrout)

I will make another branch for including the new conf.lua parameters, then still another to include them in the Options panel.

### DIFF
--- a/src/nfile.c
+++ b/src/nfile.c
@@ -429,11 +429,8 @@ int nfile_fileExists( const char* path, ... )
 int nfile_backupIfExists( const char* path, ... )
 {
    char file[PATH_MAX];
-   va_list ap;
    char backup[PATH_MAX];
-   FILE *f_in, *f_out;
-   char buf[ 8*1024 ];
-   size_t lr, lw;
+   va_list ap;
 
    if (path == NULL)
       return -1;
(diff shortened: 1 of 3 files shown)


src/nfile.c
```diff
@@ -447,11 +444,40 @@ int nfile_backupIfExists( const char* path, ... )
...
+   char file1[PATH_MAX];
+   char file2[PATH_MAX];
+   FILE *f_in, *f_out;
+   char buf[ 8*1024 ];
+   size_t lr, lw;
+
+   if (path1 == NULL)
+      return -1;
+
+   nsnprintf(file1, PATH_MAX, "%s", path1);
```
bobbens: No reason for a sprintf when you can just use path1 and path2 directly no?

bobbens:
Fixed it up and it looks good, thanks.

Mutos:
Must have been tired when I did this one ^-^ But also my own ignorance, I vaguely remember having wondered what was the difference between both forms of character strings in C.

What is probable : didn't remember the difference from years ago, and didn't have internet available to search at the moment, so I took the sure path and transformed the "char *" form passed into the "char []" form used in the original function.

PR #586, opened 2015-11-30
Bugfix : AI trying to use Reverse thrust resulted in ship never braking

Mutos:
At least I found the issue with the Reverse thrust not working on AI ships. In ai.c, before applying acceleration, it was clamped in [0,1] instead of [-1,1]. This meant an AI ship could never accelerate backwards. I just changed the limits to [-1,1]. One second to change, but hours to find it ^-^


bobbens:
NIce catch!

PR #587, opened 2015-11-30
FLF Empire Negotiations

onpon4:
This is a mission which is going to start the second possible campaign with the FLF, where rather than fighting against the Dvaereds, you make peace with them and establish the FLF as an official military of the Frontier. It cannot be done unless you have wiped out the Collective and have very good standing with the Empire. This first mission is a simple one: you just talk to someone on Marius Station.

The portraits currently used are placeholders, of course. Ideally, someone should draw proper unique portraits for these characters.

I've also added new conditions to the Empire Shipping missions to prevent weirdness, and this has a nice side effect of not being able to start befriending the Empire after you've already sided with the FLF. (Well, not easily.)


dat/missions/flf/empa/flf_empa01.lua
```diff
@@ -0,0 +1,136 @@
...
+   else
+   end
+end
+
+
+function land ()
+   if job_done and planet.cur():faction():name() == "FLF" then
+      tk.msg( title[9], text[9]:format( player.name() ) )
+      player.pay( credits )
+      var.push( "_fcap_flf", math.max( 60, var.peek( "_fcap_flf" ) ) )
```
bobbens: Instead of directly manipulating these variables in missions, I think it would be better to try to wrap it as done for empire in 'dat/missions/empire/common.lua'. If we ever change the mechanics or whatever it'll be cleaner and easier to handle, instead of directly manipulating obscure variables.

dat/missions/flf/empa/flf_empa01.lua
```diff
@@ -0,0 +1,136 @@
...
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+--]]
+
+-- localization stuff
+lang = naev.lang()
+if lang == "notreal" then
```
bobbens: My intentions were good :P

onpon4:
Alright, bobbens, I've added a common function for that.

bobbens:
Looks good, thanks.

PR #595, opened 2015-12-27
Proofread of assets starting with "f" and pirate faction assets

PhoenixRiver:
Continued proofreading the assets directory.
This is a pull request for assets starting with "f" and pirate assets:
 - primary changes are to asset class
 - then exterior image (in case we ever decide that players can land on planets without spaceports and want to show an image; the exterior images are now at least consistant with the planet class)
 - then typos and improvements of flow in planet and bar descriptions
 - and also did all the pirate faction assets because it seemed to make sense to change their stations to being "interfactional" on the basis that most pirates would take business from anyone (in both senses) and anyone from any faction could join them as pirates... assuming they didn't get robbed before landing ;)

NB: At a guess the Sirius Military stations are classed as civilian stations (0) - I am changing this to class 1 (military stations) as per the wiki, I'm basically doing this for all stations that list themselves as having military restrictions and where the description backs this up (which is always my goto - if the description doesn't say and then nothing else does either then I just leave it).


PhoenixRiver:
Now fixed so can be pulled before or after typos_E :)

bobbens:
Yes, the planet classes are a bit wonky.The military restrictions is a bit after the class thing, so I guess it never got updated properly.

Thanks!

PhoenixRiver:
> Yes, the planet classes are a bit wonky.The military restrictions is a bit
> after the class thing, so I guess it never got updated properly.

I guess the class stuff is not used for much so it's not much of an
issue except that if they are wonky I guess no one will want to use
them! I figured I might as well see if I can break the chicken/egg
cycle :)

PR #596, opened 2015-12-27
New exterior planet images

PhoenixRiver:
While working through the checking of planet classes I became frustrated at the lack of exterior shots for uninhabited planets and particularly for those that are toxic, gaseous or asteroidal (mainly because I knew they'd be easy enough to add), so ... here are some to fill in some of the gaps (and reduce the huge number of duplicates used) - I realise that these might only be used if we can land on uninhabited planets but at least some of the work would be done in the case that happens plus they are some sort of base for inhabiter versions of the same.

Included are five different images in seven versions, one for airless planetoids and one of each of red, yellow, green, cyan, blue and magenta for gaseous planetoids.

There is also a fix to make moon-D00.png (a pink atmosphered moon with what is everywhere else considered an airless class of planetoid) an airless moon with the previous image being renamed to a Y class moon because we didn't have one yet :)

Lastly, there is also an updated artwork license.

You're welcome to pick and choose what to pull if you prefer not to add everything and I'd welcome a discussion on the naming scheme I used - I picked it to make it easy enough to quickly pick something appropriate for your planet without having to scope out what they looked like exactly - how many of the exterior shots seem to have been picked in the past.

Might find time to knock together a new traderoom exterior too since that's been massively overused so far (116 times) if there is interest?


bobbens:
Yes, they are a bit painful to make. I remember playing with blender and procedural landscape scripts. Was able to pull it off for the uninhabited graphics used in the collective compaign.I'll pull them all, as they should add flavour. Thanks,

Issue #598, opened 2015-12-29
Shark07 is non-functional


aHick:
The final mission in the Shark's Teeth campaign doesn't appear to do anything when accepted, except for make a blank entry in your mission info page.

At first I thought it might be to an issue in group rankings, (I had +18 for pirates) but no, that didn't fix anything when I hacked the save file to change that. I had a Trader Rehabilitation mission going, (Same reason for both, I had just done the Pirate Hitman missions) thought that might be screwing with something, so I finished it and went back.. Nothing changed. 

I've looked in the script, and I can't seem to tell anything differently done in shark07 that is different than any functional mission like it. (My knowledge of the naev lua api is a lot less than flawless) I don't know why this happens, in other words. 

Naev .6.1 Running on Xubuntu 15.

PhoenixRiver:
I have cleaned up the shark missions and submitted a pull request so you can see if those fixes help... the fixes were in cleaning up hooks once you've completed the previous missions, I'm not sure that this is an issue but it needed done anyway so I did it. I also changed a couple of features that the original author wanted (according to his/her comments) but hadn't yet done - the pirates now spawn in random systems :) - well within a certain distance anyway! ;)
Hope this helps :)

aHick:
Righty, thanks

PhoenixRiver:
Shark missions now cleaned up in the repo - do you mind trying it again to see if it works for you?

aHick:
Works perfectly. I mean, only took me a year to get back around to testing that, sorry (where does the time go!!?!!), but the fix most certainly does work. Closed.

PR #599, opened 2015-12-30
Shark fixes

PhoenixRiver:
Includes tear downs for hooks and OSD in shark missions, not sure if they created the bug reported about them (#598) but maybe? (I don't know enough about state maintenance in Lua).

Also cleaned up mission eight (sh07), although, on second thoughts, the error might be being created by mission seven (sh06), but it's done now - proofread, odd typo etc.

Added random systems for where to find the pirates you need to kill - initially they are three to seven jumps away; change the variables min and max to alter this - as well as changing the way the pirates' names are made not to clash (in that unlikely scenario) to be cleaner.

One regression is that the first pirate no longer idles around a planet - I didn't have time to make sure the systems had to have planets as well as picking a planet for him to idle around... didn't seem like an issue and idle() was left in if someone wants to add it back in or I find more time later :)

### DIFF
--- a/dat/missions/shark/sh00_ancestorkill.lua
+++ b/dat/missions/shark/sh00_ancestorkill.lua
@@ -116,7 +116,9 @@ function accept()
       misn.setTitle(misn_title)
       misn.setReward(misn_reward:format(numstring(reward)))
       misn.setDesc(misn_desc)
-      misn.osdCreate(misn_title, osd_msg)
+      osd = misn.osdCreate(osd_title, osd_msg)
+      misn.osdActive(1)
+
       markeri = misn.markerAdd(missys, "low")
       
       jumpouthook = hook.jumpout("jumpout")
@@ -140,6 +142,10 @@ function land()
    if planet.cur() == mispla and stage == 4 then
       tk.msg(title[5], text[6]:format(battlesys:name(), missys:name()))
       player.pay(reward)
+      misn.osdDestroy(osd)
+      hook.rm(enterhook)
+      hook.rm(landhook)
+      hook.rm(jumpouthook)
       misn.finish(true)
    end
    if stage == 2 then   --You were supposed to kill him, not to go away !
--- a/dat/missions/shark/sh01_corvette.lua
+++ b/dat/missions/shark/sh01_corvette.lua
@@ -83,7 +83,9 @@ function accept()
       misn.setTitle(misn_title)
       misn.setReward(misn_reward:format(numstring(reward/2)))
       misn.setDesc(misn_desc)
-      misn.osdCreate(misn_title, osd_msg)
+      osd = misn.osdCreate(osd_title, osd_msg)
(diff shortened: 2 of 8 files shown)


bobbens:
Shouldn't the idle hook be commented out then? I mean it will be throwing errors as it is now right? Without an idle hook they'll also fly away won't they?

PhoenixRiver:
> Shouldn't the idle hook be commented out then? I mean it will be throwing
> errors as it is now right?

Oops! :/ yes, I'll fix that...

> Without an idle hook they'll also fly away won't they?

That confused me a little because as I read it only the first pirate
uses the idle hook whereas the second and third don't, though at
present the pirates only appear when you appear in the system so they
should be huntable... though maybe they could fly away and not be
interested in fighting you (e.g. you are pirate friendly (though doing
this series of missions and staying pirate friendly seems unlikely
(though not impossible I guess))).
I guess if they might fly away then it'll need the work put in to
choose systems that have planets, find the name of a planet there and
idle the pirate around it... can't guarentee the little one will let
me have the time atm.
Will have a look at this again and see if I can't get it to work as
intended - would be nice to know it would fix the bug report of
course! ;)

PhoenixRiver:
> Without an idle hook they'll also fly away won't they?

A reread suggests to me that they are all using the idle funcion but
the second two sets of pirates are just flying around the point in
space where the planet was that the first pirate originally was flying
around? I think? Dunno how easy that would make them to find so it
does look like the better idle routine is the way to go?

nenau:
Hello,
For the pirate Cruisers, they won't go away because they have to patrol 3 points before leaving the system (because of their AI). Maybe you could use pilot.setHostile() and set the player visible for NPCs in order to make them attack the player.
The first pirate (the one with the Gawain) will try to leave ASAP because he hasn't a patrol AI. That's why you need to find a way to make him wait for the player to engage him (except if you want to make this mission very difficult)

PhoenixRiver:
Hello :)

> For the pirate Cruisers, they won't go away because they have to patrol 3
> points before leaving the system (because of their AI). Maybe you could use
> pilot.setHostile() and set the player visible for NPCs in order to make them
> attack the player.
> The first pirate (the one with the Gawain) will try to leave ASAP because he
> hasn't a patrol AI. That's why you need to find a way to make him wait for
> the player to engage him (except if you want to make this mission very
> difficult)

Thank you for the info. I've never dealt with pilot AI before hence my
total lack of understanding, apologies... I had hoped there would be a
quick fix but apparently not so much... I'll delve into the mechanics
of it all and see if we can't come up with something more
appropriate... maybe have him near the sun so he can sun bath with his
escorts patrolling around him (from memory he is supposed to be on
holiday after all :P

PhoenixRiver:
Received a comprehensive fix from Nenau for the idle-hook problem - we think this is now good to go... other thoughts?

bobbens:
Looks good, thanks!

PR #601, opened 2016-01-24
2 New Missions

iwaschosen:
Add two new missions


bobbens:
Looks good, thanks. The part of the "ton" of letters is a nice touch ;)

PR #605, opened 2016-02-14
Add/fix ai.c luadocs and a couple small improvements to ai.c

ids1024:


bobbens:
incombat doesn't even appear to be even currently used. Thanks!

PR #606, opened 2016-02-15
Make LuaPilot an int instead of a struct, simplifying things

ids1024:
Creating a `LuaPilot` struct and setting the `.pilot` property repeatedly seems verbose to me (the same applies for the other lua types).

If there is no problem with this, I will go on to make the same changes for the other lua types.


bobbens:
I wonder why I wrote it that way in the first place. Looks good, thanks!

PR #608, opened 2016-02-16
Use lua registry to make push(sub)task support data of any type

ids1024:


bobbens:
Looks good, thanks. Added a bit of explanation to why TASKDATAs still exist.

PR #610, opened 2016-02-21
Allow storing system, planet, or faction in mission vars

ids1024:
There are various other types in naev which may also be useful to store in mission vars, but these are the most useful.

Tables would be useful too (as has been mentioned before), but they are more complicated.

### DIFF
--- a/src/faction.c
+++ b/src/faction.c
@@ -110,6 +110,11 @@ int pfaction_load( xmlNodePtr parent );
 int faction_get( const char* name )
 {
    int i;
+
+   /* Escorts are part of the "player" faction. */
+   if (strcmp(name, "Escort") == 0)
+      return FACTION_PLAYER;
+
    if (name != NULL) {
       for (i=0; i<faction_nstack; i++)
          if (strcmp(faction_stack[i].name, name)==0)
--- a/src/nlua_var.c
+++ b/src/nlua_var.c
@@ -373,7 +373,7 @@ static int var_pop( lua_State *L )
  *
  *    @luaparam name Name to use for the new mission variable.
  *    @luaparam value Value of the new mission variable.  Accepted types are:
- *                  nil, bool, string, number, or faction.
+ *                  nil, bool, string, number, system, planet, or faction.
  * @luafunc push( name, value )
  */
 static int var_push( lua_State *L )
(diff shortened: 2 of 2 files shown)


BTAxis:
Is this really that useful? These things can trivially be handled by storing a string and encapsulating the var.peek() in the relevant get() function.

ids1024:
@BTAxis I think it is easier to have it handled in the API, rather than repeatedly calling those functions. But true, it can already be done that way (and is).

BTAxis:
My reservation mainly comes from the fact that you've added a few specific Naev data types, which makes the whole var business a bit "messy" in that some select types are supported and others not. Makes the module that much more esoteric. At least ints and strings have the advantage of being basic types.

I have previously advocated storing arbitrary data in the player save (such as tables or any Lua objects), but as you said, that's tricky. Not just on the implementation side, but garbage collection becomes more of an issue.

ids1024:
Fair enough; though the first problem would be solved by supporting the other naev types (other than pilot, which wouldn't make any sense).

Support for tables should definitely be added; the complexity depends on how much needs to be supported. Supporting tables containing the existing supported types, and not tables would probably not be too complicated. Supporting tables in tables becomes much more complicated, as a table can even include itself, and cases like that would have to be dealt with in some way. I don't know if any of the existing implementations of table serialization would be useful for naev: http://lua-users.org/wiki/TableSerialization

I can implement this for the other naev data types if desired.

BTAxis:
Okay, I'm not going to make the call on this pr either way, I think bobbens is better qualified to do that.

ids1024:
Anyway, I think https://github.com/ids1024/naev/commit/23141f13322a3474fd46435cf055e8acacded16e should be merged either way, as otherwise `faction.get(player.pilot():faction():name())` fails.

bobbens:
I cherry-picked the faction change. On the other changes, I think that as it currently is, just using a string is enough. However, I do agree that if support for tables is implemented. It would be much more comfortable to directly store factions or whatever in the table, and save those naturally instead of having to fiddle around with strings and faction.get() or whatever. You could use an implementation like the one we have in nxml_lua.c to persist the lua (which handles tables with __save fields). With tables and support for all naev datatypes I think it would be a good contribution to help mission development.

bobbens:
Support for tables has been implemented.

PR #618, opened 2016-03-19
small improvement on zlk test mission

nenau:
* The player can't anymore run 2 zlk_test missions in the same time
* I corrected a wrong line (188) that was causing an error


bobbens:
Looks good, thanks!

PR #621, opened 2016-04-11
Master

micahmumper:
New long-distance shipping campaign and repeatable mission as part of the Empire Shipping.


BTAxis:
Thanks for the missions. I squashed your two commits into one, since there's no point having the XML file without the missions.

micahmumper:
Thanks. It was a silly oversight in my first commit. Still getting used to
github.

PR #622, opened 2016-04-15
Master

micahmumper:
Repeatable mission. Ferry tourists to various locations and then return
them to their original planet. Based on the patrol missions, but much
less dangerous because the missions only go to systems with a high
friendly presence.

Give 25,000 credits and no reputation gain.

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -290,7 +290,7 @@
   </avail>
  </mission> 
 <mission name="Long Distance Empire Shipping">
-  <lua>empire/es_longdistancecargo</lua>
+  <lua>empire/longdistanceshipping/es_longdistancecargo</lua>
   <avail>
    <priority>3</priority>
    <cond>faction.playerStanding("Empire") &gt;= 0 and var.peek("es_cargo") == true</cond>
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -1168,4 +1168,3 @@
    <faction>Za'lek</faction>
   </avail>
  </mission>
-</Missions>
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -205,7 +205,102 @@
    <planet>Halir</planet>
   </avail>
  </mission>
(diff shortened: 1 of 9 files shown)


BTAxis:
Could you resolve the conflict first? I'd do it, but it seems like you need to fix it on your end regardless, or this will keep coming up.

micahmumper:
I'm sorry, I've never used git before. Where is the conflict and how can I
fix it/avoid it in the future?

-Micah

On Fri, Apr 15, 2016 at 4:16 AM, BTAxis <notifications@github.com> wrote:

> Could you resolve the conflict first? I'd do it, but it seems like you
> need to fix it on your end regardless, or this will keep coming up.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/naev/naev/pull/622#issuecomment-210352205>
>



-- 
Micah Mumper, B.A.
Cognitive Psychology
Stony Brook University
808-389-5034

BTAxis:
You're going to have to git pull the naev repo to bring your local master branch up to date with the naev master branch. So you're going to want to do a git pull.

This will give you a merge conflict, which you need to fix. Edit dat/mission.xml and change it so it makes sense again (the conflict is on lines 309~313). Then commit your change and do another PR from the result.

micahmumper:
Okay, I think I fixed it. So, if i understand correctly, this issue was caused because there was a change to the naev master branch that wasn't reflected in my local master branch, causing an issue when I attempted to push my new changes?

BTAxis:
Yes, pretty much. In the future, you should always make sure your local repo is up to date with the official naev repo before you push, so always do a pull and fix any conflicts that pop up beforehand.

Right now, I've added your commits to the Naev master branch, but I omitted those two reverts. They're in your history though, so you might want to hard reset to the Naev master branch using 'git reset --hard naev/master' (assuming your remote for Naev is called "naev"). Note that this will nuke any commits that aren't in the Naev master branch.

PR #626, opened 2016-04-17
Add Piracy Risk to cargo missions

micahmumper:
Bulk cargo trading is back baby!

Essentially, players can make much more credits by getting a small fast ship and doing cargo rush missions. There is little incentive to do regular cargo missions because the pay is so low since “difficulty” is low. Well, when you’re trying to haul a ton of cargo into pirate infested space, it’s not so easy. Given the small rewards, it’s usually not worth it do bulk cargo missions.

So to fix this, I’ve upped the reward for delivering cargo to pirate infested space. To do this, I calculate the pirates presence along the systems to be travelled and take an average. Depending on the average, the risk is calculated as “none” (no pirates), “low” (avg presence less
than 25), “medium” (avg presence between 25 - 75), or “high” (greater than 75). Players then receive a bonus for doing routes with more risk.

What this does, in effect, is offer much more lucrative rewards for doing bulk cargo missions into pirate infested space. This gives players an incentive to use the large trader ships because the rewards are now pretty comparable with the cargo rush missions when cargo
amount and piracy is high.

Ofc, this doesn’t solve the fact that it’s easier to make credits quick using cargo rush missions in safe spaces rather than slower bulk cargo missions in dangerous territory. But, this still is an improvement for players who want more excitement while trading.

### DIFF
--- a/dat/missions/neutral/cargo.lua
+++ b/dat/missions/neutral/cargo.lua
@@ -29,7 +29,8 @@ else -- default english
     title_p2 = [[ 
 Cargo: %s (%d tons)
 Jumps: %d
-Travel distance: %d]]
+Travel distance: %d
+Piracy Risk: %s]]
 
     full = {}
     full[1] = "No room in ship"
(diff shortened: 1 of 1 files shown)


PhoenixRiver:
> Ofc, this doesn’t solve the fact that it’s easier to make credits quick
> using cargo rush missions in safe spaces rather than slower bulk cargo
> missions in dangerous territory. But, this still is an improvement for
> players who want more excitement while trading.

But you could use the same system for reducing the pay of rush cargo
missions... ;)

Phoenix *<:@)

micahmumper:
Yes, I agree. I wanted to get feedback on the initial implementation. If its well-received, I could easily import the same code into the rush cargo missions.

PR #627, opened 2016-04-22
Cargo overhaul

micahmumper:
Brief Summary:
Cargo delivery missions now only deliver commodities offered at the origin planet to planets that do not sell that commodity. Jump and distance rewards depend on the base price of the commodity being delivered. Rewards are also higher for traversing pirated space.

A new mission was created that asks players to deliver a commodity not sold on that planet. Players must buy the commodity and return. Pay is dependent on base price of the commodity and how much is delivered. 

A more detailed description of the change is available on the forum at: http://bit.ly/1XNq7Xi

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -91,6 +91,23 @@
    <faction>Trader</faction>
   </avail>
  </mission>
+ <mission name="Commodity Run">
+  <lua>neutral/commodityRun</lua>
+  <avail>
+   <priority>5</priority>
+   <chance>60</chance>
+   <location>Computer</location>
+   <faction>Dvaered</faction>
+   <faction>Empire</faction>
+   <faction>Frontier</faction>
+   <faction>Goddard</faction>
+   <faction>Independent</faction>
+   <faction>Sirius</faction>
+   <faction>Soromid</faction>
+   <faction>Za'lek</faction>
+   <faction>Trader</faction>
+  </avail>
+ </mission>
  <mission name="Dead Or Alive Bounty">
   <lua>neutral/pirbounty_dead</lua>
   <avail>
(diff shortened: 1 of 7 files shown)


BTAxis:
So one minor nitpick I have (and it is really pretty minor) is that you're assigning a text string in the middle of your code (the piracy difficulty indicators). With potential translations into other languages in mind, the standard practice is to place all user-viewable text at the top of the script, in an if-then clause based on language selection. Of course we don't actually have any translations of the sort, but it's still the idea.

Also, the code for simulating the trip is duplicated across three files. It might be a good idea to split this off into a separate helper script and include it in the files that need it.

micahmumper:
Okay, I have committed changes to address these issues. I added the piracy
risk calculation as part of the cargo_common script. If you think it
belongs somewhere else, I can make it a separate script.

Also, the piracy difficulty have been moved to a list at the top of the
script.

On Sat, Apr 23, 2016 at 4:46 AM, BTAxis <notifications@github.com> wrote:

> So one minor nitpick I have (and it is really pretty minor) is that you're
> assigning a text string in the middle of your code (the piracy difficulty
> indicators). With potential translations into other languages in mind, the
> standard practice is to place all user-viewable text at the top of the
> script, in an if-then clause based on language selection. Of course we
> don't actually have any translations of the sort, but it's still the idea.
>
> Also, the code for simulating the trip is duplicated across three files.
> It might be a good idea to split this off into a separate helper script and
> include it in the files that need it.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/naev/naev/pull/627#issuecomment-213694752>
>



-- 
Micah Mumper, B.A.
Cognitive Psychology
Stony Brook University
808-389-5034

ids1024:
It looks like this broke the pirate cargo missions (#633).

PR #632, opened 2016-04-28
corrected a small bug in ai

nenau:
The function ai.getenemy_size wasn't doing what was on the description.
This seems to correct some oddities of the ai, like the fact that fighters seemed to hesitate between targets and the fact that they attacked strongest ships first.


bobbens:
Nice catch, looks good. Thanks! (merged and rebased to master)

PR #640, opened 2016-06-10
Insane Businessman Mission

a-posadas:
Well, I'm not sure if I'm doing this right but here I go. 

These are three missions I coded based on the mission ideas in the wiki. It's probably a little rough. I welcome feedback and assistance for getting this merged eventually. 

The coding probably needs work. 
The text needs to be cleaned up, improved on. 
I have some other questions I will raise when appropriate. 
It probably needs one more concluding.

Thank you

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -1318,8 +1318,8 @@
  <mission name="Insane Businessman">
   <lua>neutral/insanebusinessman/insanebusinessman</lua>
   <flags>
-  <unique />
- </flags>
+   <unique />
+  </flags>
   <avail>
    <chance>100</chance>
    <location>Bar</location>
@@ -1328,6 +1328,9 @@
  </mission>
  <mission name="Insane Businessman Part 2">
   <lua>neutral/insanebusinessman/insanebusinessman2</lua>
+  <flags>
+    <unique />
+  </flags>
   <avail>
    <chance>100</chance>
    <location>Bar</location>
@@ -1335,5 +1338,13 @@
    <done>Insane Businessman</done>
   </avail>
  </mission>
-
+ <mission name="Insane Businessman Part 3">
+  <lua>neutral/insanebusinessman/insanebusinessman3</lua>
+  <avail>
+   <chance>100</chance>
+   <location>Bar</location>
+   <planet>Vault</planet>
+   <done>Insane Businessman Part 2</done>
+  </avail>
+ </mission>
 </Missions>
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -1315,7 +1315,7 @@
   </avail>
  </mission>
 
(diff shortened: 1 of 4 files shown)


BTAxis:
Okay, some critique:
- You might want to remove the comments you copied over from the tutorial. People reading this script can be assumed to be familiar with Lua and the Naev API, so the tutorial comments only get in the way here.
- You marked your missions as unique but use misn.finish(false) on failure or abort. This will prevent the mission from ever appearing again. It might be better if it did reappear later, many of our existing unique missions work that way. The idea is that we don't want to punish the player for aborting a mission unless there's a good reason.

I think these two things should probably be addressed prior to merging, the rest is just generic coding stuff that isn't critical. For example, you can eliminate the table reference for single pilots you spawn by assigning the pilot directly rather than the table: 
pilot1 = pilot.addRaw( "Admonisher", "mercenary" , from_system , "Dummy")[1]
pilot1:rename...

a-posadas:
Alright, I removed the comments and made some code neater. I'll have to test it to make sure there are no errros. 

In any case, I need some clarification on your second point. The wiki says that with the unique tag and mission.finish(false) the mission will still appear. But really, I need a little help deciding when and how the mission shows up. I could use suggestions. 

http://wiki.naev.org/wiki/Mission.xml

> If a mission exits with mission.finish( false) that mission may appear again, even if the <unique> flag is present in the mission XML.

BTAxis:
It appears you're correct, I was going by the comment from the tutorial script. This comment is in error, I shall fix it. So ignore the second point of my earlier post (though you can omit the abort function entirely and still get the same behavior regardless).

a-posadas:
ok, I removed the abort functions. I also did some minor edits. 

I have a question about the ships in part three. I equipped them with lasers and whatnot but not missiles. However, it seems in the mission they have lasers and missiles. I used a dummy faction but a mercenary ai so neutral or friendly players won't attack you when you attack them, but they aren't associated with any factions. Still, it seems by default they are equipped with missles. 

How do I find out what they are equipped with by default or are the equip lines I'm using being ignored for some reason?

a-posadas:
I added a concluding event. This occurs with a chance of 50. It can happen only after you complete part 3. In the event, Crumb shows up flying around Mason and if you disable the ship you find one million credits (because Crumb )

BTAxis:
The default equipment for ships is determined by the equip scripts, located in dat/factions/equip. I'm unsure which script (if any) the dummy faction uses. There may be a hardcoded default but I'm not certain. In any case you can override this equipment loadout by un-equipping everything from a pilot before anything else using pilot.rmOutfit("all"). Note that this will not strip core outfits.

TheGiraffe3:
@a-posadas can you get around to this any time soon?

PR #642, opened 2016-06-25
LDoc

ids1024:
Naev currently uses `luadoc` to generate the documentation at `api.naev.org`, but that has been abandoned in favor of `LDoc`. In the branch I have started to get the documentation to build with `LDoc`.

There are still some issues, so this is not ready for merge yet.


ids1024:
I think this is working correctly, so it should be ready for merge. There may be issues that i've missed, however.

ids1024:
I have pushed a commit removing `docs/html/`, which can be build with `luadoc.sh` and is avaiable at http://api.naev.org anyway. If this is not desired, it should instead be regenerated with `luadoc.sh` so it is up to date.

PR #643, opened 2016-07-04
Add missing return

ids1024:
I presume there should be a return statement here.


bobbens:
Ah thanks! That could explain some issues we've had with intel boards in the past.

PR #644, opened 2016-07-04
Layeredimage

Kinniken:
LayedImage and LayedImageArray, plus sample use in outfit pictures.

Originally developped for use in my Long Night fork, details on the forum: https://forum.naev.org/viewtopic.php?f=3&t=1196


Kinniken:
Can you send me your texture fix patch? I'd like to reimport it in Nox.

For the missing images, what were they? That's odd, never had that problem. Must be an issue with my commit.

bobbens:
You probably forgot to commit the three files in question:
1. 'dat/gfx/outfit/store/layers/slot_special_Core'
2. 'dat/gfx/outfit/store/layers/slot_special_Engines.png'
3. 'dat/gfx/outfit/store/layers/slot_special_Hull Modifications.png'
The fixes can be found at https://github.com/bobbens/naev/commit/ccf68ad782f065ebc1af176970397f765cdbacc2

Kinniken:
Actually no... it's a bug I only noticed and fixed recently in Nox: https://github.com/Kinniken/NoxImperii/commit/c17a984b24ae1e76631a0dd38d93aecb3490a7ce

Basically I was using the display name of the slot in my picture instead of the id, and doing so in a case-insensitive manner which only worked on Windows. The solution is to import that commit above and to rename the picture files accordingly (slot_special_core systems => slot_special_core and slot_special_hull modifications => slot_special_hull).

Can I amend a pull?

And thanks for the fix.

bobbens:
Yes, go for it. I'm still not sure how we will use it in naev. The stars indicating strength does seem like something I like. However, I also liked the colour backgrounds we have for slot size. I guess before merging I'll probably mess with it more.

Kinniken:
Committed.

Kinniken:
Btw, regarding usage, I'm planning an overhaul of the trade screen using a similar system. That one will have to be Nox only though as my trade system is custom.

bobbens:
Merged with current master in the layeredimage branch. Still needs some work to get it looking good, and thinking it just might be easier to merge the new layeredimagearray widget with the imagearray widget.

bobbens:
Ended up using a more streamlined alternative implementation that I have pushed to master. Although it is implemented, it is still not used in game.

PR #651, opened 2016-08-07
Removed combat elements from the sightseeing mission.

onpon4:
These elements were left over from the patrol mission where the
whole point was to engage enemies of the paying faction. I don't
think it makes sense for the sightseeing mission to do anything
special with regard to this; hostiles (e.g. pirates) will attack
you of their own volition, and that will of course mean you have
to fight back so you don't die (or escape, which the patrol mission
code doesn't necessarily allow).

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -465,7 +465,7 @@
  <mission name="Assault on Unicorn">
   <lua>dvaered/assault_on_unicorn</lua>
   <avail>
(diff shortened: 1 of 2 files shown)


onpon4:
I've pushed some other stuff related to the sightseeing mission here:

* Improved the ship class check (it's now done when you accept the mission, and it double-checks to make sure you didn't switch out).

* Removed a faction check (which only allowed tours in systems with the origin planet's faction's presence), since while it makes sense for patrols to only be in systems with the faction's presence, tourists aren't going to insist on that.

* Removed a cap on the number of attractions inherited from the patrol mission; it is now possible for more than 5 attractions to be visited.

bobbens:
Looks good, thanks!

PR #652, opened 2016-08-07
A couple more minor things.

onpon4:
Should be uncontroversial.


onpon4:
A bunch more bugfixes have been pushed into here since this was opened. So here is the full list of what this pull request has:

* Added use of numstring to the commodity run mission.

* Renamed a couple oddly-named mission files.

* Fixed the target ship not showing up in traffic_00 if you took off from a planet.

* Fixed a flawed check in the warlords battle event (caused it to continue where it should abort, e.g. in the Goddard system, and then have an error later).

* Fixed mixing up "she" with "him" and "his" in the bounty missions.

* Added a mercenary license check for Assault on Unicorn.

bobbens:
Looks good, thanks!

PR #653, opened 2016-08-07
Removed an alarming number of empty Commodity services.

onpon4:
These empty Commodity services are a problem if you're actually
looking for a place to buy/sell commodities, because the indicator
that a planet has commodities becomes almost useless. This took some
fiddling about with grep, but I think I've removed all instances
of planets with empty Commodity services (i.e. offering the service
but no commodities).

### DIFF
--- a/dat/assets/agino.xml
+++ b/dat/assets/agino.xml
@@ -22,7 +22,6 @@
    <refuel/>
    <bar/>
    <missions/>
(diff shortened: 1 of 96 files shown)


bobbens:
I guess this is due to copy-pasting. As it doesn't make sense to have them since you can't do anything, I'm merging. Thanks!

PR #654, opened 2016-08-09
Removed the faction requirement for the commodity run mission.

onpon4:
This mission's important conditions are related to the commodities
available on the specific planet. It makes no sense to restrict it
only to specific factions. Pirates and the FLF need basic commodities,
too.

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -97,15 +97,6 @@
    <priority>5</priority>
    <chance>90</chance>
    <location>Computer</location>
-   <faction>Dvaered</faction>
-   <faction>Empire</faction>
-   <faction>Frontier</faction>
-   <faction>Goddard</faction>
-   <faction>Independent</faction>
-   <faction>Sirius</faction>
-   <faction>Soromid</faction>
-   <faction>Za'lek</faction>
-   <faction>Trader</faction>
   </avail>
  </mission>
  <mission name="Trader Escort">


PhoenixRiver:
Hello :)

> This mission's important conditions are related to the commodities
> available on the specific planet. It makes no sense to restrict it
> only to specific factions. Pirates and the FLF need basic commodities,
> too.

True and I hate to disagree but ... associating with these groups (and
soliciting for commodity runs to their planets) is, at minimum, a
minor crime in a lot of the known galaxy and probably a felony
equivalent in some of those - more so the pirates than the FLF I'll
grant but never-the-less.

If you are going to remove those restrictions they should be replaced
with faction based restrictions - bonus points for using the faction
file to generate a reasonable set ;) - if you feel strongly but don't
have the time I might in a week or so... I guess code to do that might
also be useful for other general types of mission too e.g. cargo runs
etc.

Hope this helps,

Johann *<:@)

onpon4:
I don't understand your objection. This mission is where *you* go out and buy commodities from some other planet (of your choosing) to bring back to the current planet. The process is exactly the same regardless of whether or not the current planet is criminal. The only thing that might be different is pirates might want to get illegal commodities supplied, but they're going to want normal commodities too, so that's not an issue here. (Any contraband commodity run for pirate worlds can be offered alongside the normal commodity run missions.)

ids1024:
It makes sense to allow the missions, but with a risk of a reputation hit to pilots supplying pirates or the FLF. Not something that is guaranteed to happen, but the player could get "caught" triggering some message and reputation drop.

onpon4:
No, that doesn't make sense. Who is there to be a witness? If you got caught selling things to New Haven, then New Haven's location gets discovered. And it's not like you can get "caught" buying food, a perfectly normal, legal activity.

ids1024:
It's just an idea. It certainly wouldn't be extremely likely that the player would get caught.

In a well ordered world such things may be noticed, but perhaps you're quite right in the case of a more chaotic universe with weak government, as in naev.

onpon4:
> In a well ordered world such things may be noticed

I don't know about that. Consider the case of UPS delivering packages to a drug dealer, the packages containing perfectly legal things like clothing. Would anyone notice? Of course not. If someone did notice, would UPS get in trouble? Of course not.

If it turned out that UPS was helping the drug dealer do something illegal (say, supplying him with drugs), that would be different. But even if that happened, perfectly normal, legal deliveries to the drug dealer's house wouldn't be penalized.

PhoenixRiver:
>> In a well ordered world such things may be noticed
>
> I don't know about that. Consider the case of UPS delivering packages to a
> drug dealer, the packages containing perfectly legal things like clothing.
> Would anyone notice? Of course not. If someone did notice, would UPS get in
> trouble? Of course not.

The UPS/drug dealer comparison is a little off in my opinion - this is
more like (as far as some factions are concerned) an individual (not a
global corporation who are harder to police) supplying goods to
terrorists (FLF) or rogue nations (North Korea, Iran, Cuba in the 60s
etc.)

News that a ship matching the description of yours repeatedly heads
towards (or uses the jump point to) a pirate haven is going to be
noticed eventually by locals and they might have an interest (given
they often won't like the pirates) in telling someone who can do
something about it. Governments know that starving a rogue nation of
standard commodities is a good way to weaken them (and do it regularly
in RL c.f. large number of trade embargoes countries have against each
other). Stopping you is in their interests if they want to stop their
enemy.

> If it turned out that UPS was helping the drug dealer do something illegal
> (say, supplying him with drugs), that would be different. But even if that
> happened, perfectly normal, legal deliveries to the drug dealer's house
> wouldn't be penalized.

Of course a major galaxy corp would not suffer (badly anyway) but a
individual that's causing trouble can easily have an "accident" if a
large galaxy government doesn't feel like publicly reprimanding
terrorist/rogue nation sympathisers... never mind supports... and how
do they know what you are carrying is only commodities... maybe you
are taking things that are not what they seem? Easier to arrest you
and _then_ make charges stick.

And what we're talking here is simple reputation, not bounties on you,
police action or underhand assassin ships  - though those would all be
cool additions :P

As long as it is pointed out that working for pirate/freedom fighter
organisations has it's drawbacks (and possible benefits) then the
player has the choice... though in the case of pirates they already
pretty much made that choice just getting to the pirate haven!

Phoenix

BTAxis:
Putting aside the morality/legality debate, what is curious to me that an organization like the FLF would post a delivery mission to their, presumably secret, bases on a public mission board. It seems to me their chief of security might have something to say about that.

onpon4:
I've always considered the mission computer on Sindbad to be private. After all, you can't access missions for other planets from there, and why would they enable any sort of wireless communication with other planets to begin with?

onpon4:
> News that a ship matching the description of yours repeatedly heads towards (or uses the jump point to) a pirate haven is going to be noticed eventually by locals and they might have an interest (given they often won't like the pirates) in telling someone who can do something about it.

I guess, maybe, except for one problem: these pirate bases, and Sindbad, are not even known by the governments to exist. If their existence and locations were known, they would be attacked and almost immediately destroyed. So all anyone knows, barring mass surveillance which cannot be occurring if these secret bases can exist, is that you're buying commodities, a normal activity. No one can tell where you're taking them to or when you're selling them. And since it's a perfectly legal activity to buy commodities, there's no reason for you to be investigated for it.

onpon4:
Let me put it another way: if there was an event where New Haven was discovered and attacked by the Empire, and you have done business with New Haven, it might make sense for *this event* to possibly cause a drop in reputation for all of the Pirates' enemies. But that would be contingent on merely landing on New Haven; surely buying outfits and ships from them would apply just as much as selling ore or food to them.

PhoenixRiver:
> Let me put it another way: if there was an event where New Haven was
> discovered and attacked by the Empire, and you have done business with New
> Haven, it might make sense for *this event* to possibly cause a drop in
> reputation for all of the Pirates' enemies. But that would be contingent on
> merely landing on New Haven; surely buying outfits and ships from them would
> apply just as much as selling ore or food to them.

Well put and I agree with all of it... as long as the governments do
actually want to destroy the pirates entirely and if they do, are they
even capable? ... North Korea, Iran and Cuba all exist and were never
destroyed outright despite the United States clearly having the
ability and apparent will to do so - c.f. Iraq and Afghanistan - oh,
wait those would be examples of a failure to deal with the problem...
oops. So... I agree with the above but only with the provisos
mentioned above - the will and ability to do the deed but I'm yet to
be convinced that either is true.

But I don't get the call so I'll leave it to them that do to decide :)

Phoenix *<:@)

PS BTAxis: I also agree that I find it odd that a small and clearly
well organised and structured organisation (else how would it manage
to run as a freedom fighter / terrorist organisation for any real
length of time and stay hidden) like the FLF would trust runs like
this to anyone with access to the board - especially such "ordinary"
things that the lack of could cause serious resource problems for the
group. On the other hand, maybe they are that disorganised and the
Daevred are right to want them gone ;)

onpon4:
> as long as the governments do actually want to destroy the pirates entirely and if they do, are they even capable?

Yes. We have a previous event as evidence of this: the destruction of Haven by the Empire. There isn't a hidden jump point from Haven to New Haven, which further suggests that the pirates are afraid of this happening again. :)

> FLF would trust runs like this to anyone with access to the board

It's strongly implied to be incredibly unusual that you get to access Sindbad before joining the FLF, and explicitly stated that if you didn't have someone from the FLF to vouch for you, you would be imprisoned if not outright killed (as the guards try to do before leading you to the base).

So no, it's not odd; the only ones with access to the FLF's mission computer are normally FLF members. The player is a special exception, but they're already trusting you not to just straight-up tell the Dvaereds where Sindbad is, ensuring their destruction. Also trusting you on a non-essential mission to go on a supply run for them is not all that strange.

bobbens:
I'm a bit torn on this. On one side, I think it makes total sense for FLF/Pirates to ask for cargo. However, on the other side, it sort of breaks the entire "super secret base" feeling. It's nice to be at a FLF/Pirate base and see that all the missions are fancy and not just your standard boring cargo missions.

onpon4:
Most of them are still fancy. ;) This just means that one particular standard mission is available on these special worlds, in addition to the fancy missions.

But if you'd be more comfortable with it this way, I could instead create special versions of the commodity run mission for pirates and the FLF that use different wording. Would that be better?

bobbens:
Yes, trying to make them a bit more flavourful would help. Such as the news and stuff is different in the pirate places (and rather silly at times).

onpon4:
Alright, done in #671.

PR #658, opened 2016-08-11
Adjusted the fleets in Operation Cold Metal to make it more fun/climactic.

onpon4:
There were two problems with Cold Metal: one, you don't even have to
participate; you can just sit back and let the Empire ships do all the
work and the mission will succeed. Two, if you go in and try to help
anyway, the Collective drones were so concentrated in one spot that
this was practically a death sentence if you didn't have a super-heavy
ship.

I can't think of a way to address the first problem, and it's not that
big of a deal, anyway (and might even be considered a good thing). But
I've addressed the second problem here by spreading out the Collective
drones. Being more spread out means you can join in the fight without
getting your butt blasted every time, so that at the very least makes
your participation more fun, if still entirely pointless.

I've also added some heavy Collective drones to the Collective fleet,
and reduced the size of the Empire fleet (now one Peacemaker instead
of three and six Pacifiers instead of seven), to at least make your
participation *feel* a little less pointless.

Sorry about the duplicate; I accidentally deleted the branch when I meant to delete a different one.

### DIFF
--- a/dat/missions/empire/collective/ec06.lua
+++ b/dat/missions/empire/collective/ec06.lua
@@ -123,12 +123,10 @@ function jumpin ()
             local fleetCpos = vec2.new(0, 0)
             deathsC = 0
 
-            fleetE[#fleetE + 1] = pilot.add("Empire Hawking", nil, fleetEpos)[1]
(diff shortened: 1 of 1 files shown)


bobbens:
I guess this is because this mission was originally tuned for before core slots and all the major engine changes. So balance has completely changed. Thanks!

PR #659, opened 2016-08-11
Changed Cold Metal to warp the Empire fleet in instead of them waiting for you.

### LINKED ISSUE #597: [Final Mission of] Collective Campaign Sequence Breaking
Once you get to C-00 in the final mission of the Collective campaign (ec06 as far as I know), the script tells you to "Meet with the Imperial Fleet." However, if you don't do that, you can just go around them and take out every single one of the opposing fleet's members, (they don't wake up when you shoot at them) then return to base.

If you go back to where the fleet is after destroying the opposing fleet, of course it will go to the second part of the mission, which will render the mission unfinishable due to the Starfire and Trinity already being destroyed.

A good solution to this would either be to move the Imperial fleet back into C-28 near the jump and have them all jump through to C-00 to fight the fight, or have them really close to the jump point at C-00 so that you can't go around them, so that the battle starts as soon as you enter.

If a fix is not created a really good while after I've posted this (which one will :)), I will attempt to change the file to a better working thing.

This was running Naev 6.1 on Xubuntu 15.

onpon4:
This is mainly to fix #597. I could have added code to punish you for trying to sequence-break, but I felt the whole "go to the Empire fleet first" thing was pointless anyway. I even went further and made it not matter which system you enter from.

Because of the extent of conflicts, this pull request implies #658 as well.

### DIFF
--- a/dat/missions/empire/collective/ec06.lua
+++ b/dat/missions/empire/collective/ec06.lua
@@ -32,7 +32,7 @@ else -- default english
    text = {}
    text[1] = [[You join Commodore Keer at her table.
     She begins, "We're going to finally attack the Collective. We've gotten the Emperor himself to bless the mission and send some of his better pilots. Would you be interested in joining the destruction of the Collective?"]]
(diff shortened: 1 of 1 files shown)


bobbens:
Warping in should give the mission a more epic feeling. Thanks!

PR #660, opened 2016-08-11
Added a soft limit of 3 commodity runs at a time.

onpon4:
This should ensure that the player isn't just going around the galaxy
accepting a million commodity runs at a time and then only delivering
on whichever ones happen to be easiest.


bobbens:
Bit worried the counter might get screwed, but it shouldn't unless the Lua crashes. Thanks!

PR #661, opened 2016-08-11
Made it so that the refueling ship in Cold Metal never leaves.

onpon4:
I haven't tested this since I don't have a save where I'm at or
near that point, but it should do the trick. It's really annoying
if you fail to respond to the refueling ship and it runs away,
effectively forcing you to start the mission over again since there's
no other way to refuel for three systems.


bobbens:
Should work, thanks!

PR #662, opened 2016-08-11
Made bribed ships show up as such, not "Neutral".

onpon4:
Showing up as "Neutral" is a relic of a previous engine limitation
which no longer exists. It makes no sense to keep this ambiguity.


bobbens:
Looks good, thanks!

PR #663, opened 2016-08-12
added a min dist for the careful_face function

nenau:
* Fleeing and trader ships don't take into account ships that are too close (to fix the behaviour when a heavy ship is attacked by a fighter that revolves around it)
* reldps is also taken into account
* the condition to take a ship into account is now cleaner imo.


bobbens:
Looks good and should be more stable than adding a fixed value to avoid small divisions. Thanks!

PR #664, opened 2016-08-12
Made the Empire's limitation on secondary faction hits apply to missions.

onpon4:
The secondary faction hit limitation is a way to prevent FLF-aligning
players from becoming flat-out enemies with the Empire from attacking
the Dvaereds. This change makes that limitation apply to missions,
rather than just aggression against Dvaered ships.

(Honestly, I don't remember why I limited it to distress and death in the first place. But it's clearly a bad decision.)

### DIFF
--- a/dat/factions/standing/empire.lua
+++ b/dat/factions/standing/empire.lua
@@ -17,8 +17,7 @@ sec_hit_min = 10
 function faction_hit( current, amount, source, secondary )
    local start_standing = _fthis:playerStanding()
    local f = default_hit( current, amount, source, secondary )
-   if ( ( source == "distress" or source == "kill" ) and secondary and
-         amount < 0 and f < sec_hit_min ) then
+   if ( secondary and amount < 0 and f < sec_hit_min ) then
       f = math.min( start_standing, sec_hit_min )
    end
    return f


PhoenixRiver:
> The secondary faction hit limitation is a way to prevent FLF-aligning
> players from becoming flat-out enemies with the Empire from attacking
> the Dvaereds. This change makes that limitation apply to missions,
> rather than just aggression against Dvaered ships.
>
> (Honestly, I don't remember why I limited it to distress and death in the
> first place. But it's clearly a bad decision.)

Does this mean being attacked by the Dvaered when not on a mission
means you get worse Empire standings? if so then that's why it has to
always apply since if it doesn't then anyone aligning with the FLF and
spending any time on non-mission stuff is going to end up in the same
position as without this at all, no? Because the Dvaered will
auto-attack an FLF aligned player... or did I miss something?

onpon4:
> Does this mean being attacked by the Dvaered when not on a mission means you get worse Empire standings?

No.

This is how the reputation system works (for *all* factions, not just the Empire): if you attack a ship, it "distresses", and this causes a faction hit. If you destroy a ship, this causes a faction hit. Missions can cause faction changes in either direction, but are usually gains.

Whenever your standing with a faction changes (with some exceptions in missions), that faction's allies and enemies also experience changes, but on a smaller scale. Allies have their standing changed in the same direction, while enemies have their standing changed in the opposite direction. This is called a "secondary" standing change.

Some time ago, I made a tweak to the Empire's standing script (which is already merged) that made it so that secondary faction hits resulting from pilot death or distress can never bring your standing with the Empire below 10%. This was designed to prevent FLF-aligned players from becoming enemies with the Empire automatically; you can avoid becoming enemies with the Empire by simply not fighting back against any Empire ships that join Dvaered ships in aggression against you. Your reputation will still drop, but not below 10%.

The only thing this pull request does is make that 10% minimum for secondary faction hits also apply to faction changes caused by missions. In practice, this means that completing FLF missions does not drain your Empire standing lower than 10% (unless you happen to attack or destroy Empire ships along the way).

bobbens:
Sounds good, thanks!

PhoenixRiver:
> This is how the reputation system works (for *all* factions, not just the
> Empire): if you attack a ship, it "distresses", and this causes a faction
> hit. If you destroy a ship, this causes a faction hit. Missions can cause
> faction changes in either direction, but are usually gains.
>
> Whenever your standing with a faction changes (with some exceptions in
> missions), that faction's allies and enemies also experience changes, but on
> a smaller scale. Allies have their standing changed in the same direction,
> while enemies have their standing changed in the opposite direction. This is
> called a "secondary" standing change.
>
> Some time ago, I made a tweak to the Empire's standing script (which is
> already merged) that made it so that secondary faction hits resulting from
> pilot death or distress can never bring your standing with the Empire below
> 10%. This was designed to prevent FLF-aligned players from becoming enemies
> with the Empire automatically; you can avoid becoming enemies with the
> Empire by simply not fighting back against any Empire ships that join
> Dvaered ships in aggression against you. Your reputation will still drop,
> but not below 10%.
>
> The only thing this pull request does is make that 10% minimum for secondary
> faction hits also apply to faction changes caused by missions. In practice,
> this means that completing FLF missions does not drain your Empire standing
> lower than 10% (unless you happen to attack or destroy Empire ships along
> the way).

Thank you for the detailed explanation, it seems that I was confused
about how it worked and even the change you had made - apologies.

PR #665, opened 2016-08-12
Made the cargo hulls give less protection.

onpon4:
There really wasn't much difference in terms of protection between
the cargo hulls and the high-end combat platings. I think this is
a bad design, so I have changed the cargo hulls to all have low-end
protection (equivalent to the standard Unicorp components of the
same size) to make it more of a trade-off.


bobbens:
Yes, it makes sense. Thanks!

PR #666, opened 2016-08-12
Removed positive characteristics from "beat up" engines.

onpon4:
These engines previously had particularly high engine mass limits
and particularly low energy usage. I have changed it so that they
have standard engine mass limits and particularly high energy usage.
These engines are supposed to be a last resort, after all.


bobbens:
Makes sense, thanks!

PR #667, opened 2016-08-12
Removed the map for Sol's location.

onpon4:
The existence of a map suggests that the area is meant to be
explored or has something important in it. Neither is the case for
Sol; the inner nebula is completely impossible to pass through, and
there are no plans to change this in the future (instead, a way
around it is supposed to be added). Considering this, the Sol map
just creates confusion for players and is an expensive, worthless
item.


PhoenixRiver:
> The existence of a map suggests that the area is meant to be
> explored or has something important in it. Neither is the case for
> Sol; the inner nebula is completely impossible to pass through, and
> there are no plans to change this in the future (instead, a way
> around it is supposed to be added). Considering this, the Sol map
> just creates confusion for players and is an expensive, worthless
> item.

But the player doesn't know that and they are supposed to find out the
hard way if they don't pay heed to the warnings of what happened to
Sol and the price of the map is what you'd expect given the danger of
getting the information. So in essence the reasons you describe for
wanting to remove the map are the exact same reasons the story says
that it should exist - in fact it would be strange if a map for such a
dangerous area wasn't being sold... just look at the proliferation of
maps (good and bad) that occurred during the era of exploration that
led to the discovery of the "New" World...

onpon4:
> supposed to find out the hard way if they don't pay heed to the warnings of what happened to Sol

Are you suggesting that players should be sent on a fool's errand on purpose just so they can find out that the inner nebula is impassable? I can't say I agree or even understand why you think this, if I'm reading what you wrote correctly.

> the map is what you'd expect given the danger of getting the information

No, it's pretty clear that the map is a leftover from pre-Incident records; anyone who goes into the nebula never comes back, as the NPC dialogs make clear. Besides, if the route to Sol was known, that would be a part of the map.

But I digress. The price of the map is what makes it wasteful. The main reason I want it removed is it's confusing.

> So in essence the reasons you describe for wanting to remove the map are the exact same reasons the story says that it should exist - in fact it would be strange if a map for such a dangerous area wasn't being sold

I don't follow. The danger of the inner nebula is very clearly stated by NPCs in bars. A map painting Sol as a mysterious system that no one knows anything about is exactly the opposite of warning about the dangers of the inner nebula.

> just look at the proliferation of maps (good and bad) that occurred during the era of exploration that led to the discovery of the "New" World...

Sol isn't a discovery, though. It's where people came from and known to be a dead system (rendered dead by the Incident). The systems beyond Sol are a discovery, but you're going to get them not by going to Sol, but by going around the entire inner nebula somehow, so knowing where Sol is will not help in any way.

bobbens:
On one side, I do think it can be a bit misleading to players. However, I think it both adds flavour and creates a bit of game atmosphere. Furthermore, Sol should be reachable in some of the main campaigns (assuming they ever get finished). We have nebula damage decreasing armour and stuff for that. It is also likely it can get terraformed or whatever. So I'm inclined to leave it as is, maybe make the description a bit more explicit, such as: "This is a map that leads to birthplace of mankind, rendered dead by the poisonous nebulae." or something like that.

onpon4:
> So I'm inclined to leave it as is, maybe make the description a bit more explicit, such as: "This is a map that leads to birthplace of mankind, rendered dead by the poisonous nebulae." or something like that.

Alright then. I'll do just that.

PR #668, opened 2016-08-12
Changed flf_empa01 to use modPlayerSingle.

onpon4:
This prevents a reputation hit with the Empire when you complete it.


bobbens:
Nice catch, thanks!

PR #669, opened 2016-08-13
Changed the Sol map's description to better fit Sol's purpose.

onpon4:
The description now explains that Sol is "rendered dead" and covered
in an unstable nebula, with a teaser about what lies beyond.


bobbens:
Looks good, thanks!

PR #670, opened 2016-08-13
FLF-Empire alliance campaign progress

onpon4:
Added the second mission in this campaign, where you have to get the agreement from Petridis and give it to Cheryl.

The next mission is going to be much more exciting, I promise. :)


bobbens:
Looking forward to the next one, thanks!

PR #671, opened 2016-08-13
Added versions of the commodity run mission for the FLF and Pirates.

onpon4:
They are functionally identical, but with different texts.


bobbens:
Looks good, thanks.

PR #672, opened 2016-08-14
Added support for some faction unidiffs.

onpon4:
The diffs are as follows:

* Making a faction visible or invisible (for factions that have
  been destroyed or created)

* Changing faction alignments (e.g. making the Empire and the FLF
  enemies mid-game, or making the FLF and Dvaered neutral mid-game)

Use of this to make the Collective invisible once it is defeated has
been added.


onpon4:
It seems this is causing a segfault any time Naev quits (while a relevant unidiff is applied, i.e. the collective_dead diff). I must have done something wrong with memory management.

zid:
Got a backtrace?

zid:
faction_rmAlly() looks heavily broken, regardless.

The indentation betrays which code is inside or outside the if, and only supports removing an ally from the end of the list, not from the middle.

onpon4:
Oh, of course.

The indentation represents what I intended it to do. Stupid me. I guess I'm just a little too used to Python...

onpon4:
Alright, that's been fixed. It's not the cause of the segfaults, though, so there must be something else wrong.

zid:
A backtrace would help, I'm not set up to compile it atm.

Realloc is always a scary function to see in the context of a segfault, it invalidates every pointer pointing into the object, so it won't show up until something /else/ tries to access the object through the stale pointer. Sounds like something has a stale reference to an ally or faction or something and the cleanup code is trying to free it. 

If possible I'd recommend splitting this patch into two. One patch adding the functionality, the 2nd patch using it. It makes for better bisection points.


No love for the lack of removing from the middle of the ally list?

onpon4:
> No love for the lack of removing from the middle of the ally list?

No, it removes things from the middle. What it's doing is copying the last value into the spot on the list where the item that should be deleted is found.

TBH, I doubt rmAlly and rmEnemy are the cause, because I'm pretty sure they're not actually being used anywhere. The only diff actually being used is the one to turn a faction (the Collective) invisible. So I think the code that is most probably the problem is the code I added to unidiff.c. It doesn't help that I only have a vague understanding of how that code works, so I relied on a lot of almost straight-up copying/pasting.

zid:
> What it's doing is copying the last value into the spot on the list where the item that should be deleted is found.

Oh of course, I should have read it harder, I do that all the time.

I'll give the unidiff part a look over and see if I can see anything obvious.

onpon4:
OK, I think I got a backtrace... did I do this right? Sorry, I'm not familiar with using the debugger:

http://pastebin.com/AAdyG3nA

onpon4:
Here's the save I've got:

[save.zip](https://github.com/naev/naev/files/418079/save.zip)

onpon4:
Alright, that segfault is fixed, thanks to zid helping look over my code. Simple solution: I needed to assign something to hunk.u.name, which wasn't used by the visible/invisible diffs, but because nothing was assigned to them, some garbage values were there and getting cleaned up later. Fixed by assigning this as NULL.

onpon4:
This is good to go. Anyone want to review and pull? @bobbens? @BTAxis? @Deiz?

Same goes for #673.

PR #673, opened 2016-08-15
Next mission in the FLF-Empire campaign.

onpon4:
This is where it starts getting exciting.

To earn their trust, the Empire's conditions include disarmament of several explosives that were placed in Raelid Outpost (which would be used to destroy Raelid Outpost in the flf_dvk campaign).

However, many FLF members are not happy with the terms set forth by the Empire. So you have a choice: you can follow through with the plan, and you have hope for the future, but gain the ire of many aboard Sindbad. Alternatively, you can betray the Empire and blame the Dvaereds, pleasing those FLF members, but leading the FLF on a path that looks very grim and includes making the Empire an official enemy of the FLF.

Allying with the Empire is going to cause a massive faction split in the FLF later, while betraying the Empire will result in changes in leadership followed by turmoil as the FLF struggles against the combined power of the Empire and House Dvaered. The FLF in the betrayal path, and the defectors in the alliance path, will likely join together with the pirates and possibly even the Protorons in a last-ditch effort to maintain a foothold.

Note: to work entirely, this requires #672.


BTAxis:
You might want to consider putting off opening a pull request until after you've made all your commits. Each commit generates an email, so it gets a little spammy if you have to make minor corrections repeatedly.

onpon4:
I typically do, and I though I was done with 47733f (when I made the pull request); all those added commits were mistakes I discovered later. This was just a case where I made a huge number of short-sighted mistakes, probably because I was doing this in the middle of the night.

It's done now.

onpon4:
Anyone want to merge this?

PR #675, opened 2016-08-16
use a table in enumerate function, for lua 5.1 compatibility

Pilcrow182:
Using the argument "..." in a function is a lua 5.2 feature for handling a variable number of arguments. However, this feature does not exist at all in lua 5.1. Instead, providing the function a single table of strings ensures lua 5.1 compatibility without breaking compatibility with 5.2...

This is also the only function that uses this feature, so replacing it is a trivial matter.

### DIFF
--- a/dat/events/tutorial/tutorial-heat.lua
+++ b/dat/events/tutorial/tutorial-heat.lua
@@ -80,7 +80,7 @@ function create()
     tk.msg(title1, message1)
     tk.msg(title1, message2:format(tutGetKey("primary")))
     
-    stages = enumerate("weapons", "wait", "coolwait", "cool")
+    stages = enumerate({"weapons", "wait", "coolwait", "cool"})
     stage = stages.weapons
 
     flytime = 10 -- seconds of fly time
--- a/dat/missions/sirius/achack/achack04.lua
+++ b/dat/missions/sirius/achack/achack04.lua
@@ -95,7 +95,7 @@ function create()
    startplanet, startsys = planet.get("Eenerim")
    tk.msg(title1, text1:format(player.name(), startplanet:name(), startsys:name(), startplanet:name()))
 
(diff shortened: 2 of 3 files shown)


ids1024:
> without breaking compatibility with 5.2...

It doesn't really have any impact on this PR, but for reference, support for lua 5.2 does not matter. Naev supports lua 5.1 or luajit (which is backwards compatible to 5.1).

Since luajit is the default now, it may be worth dropping support for lua 5.1. The only real disadvantage I can think of is the fact that luajit supports a limited number of architectures, but that seems rather unimportant  since it supports anything I could really see a need to support (x86, arm, mips, and ppc).

Pilcrow182:
The "..." parameter as a function argument doesn't seem to work in luajit either (at least, not in luajit 2.0; 2.1 may be a different story).

I've compiled and run naev in a slightly outdated Linux distro (using luajit 2.0.0b9) and selecting the "Tutorial: Heat" mission results in the following:

    Warning: [event_runLuaFunc] Event 'Tutorial: Heat' -> 'create': [string "enum.lua"]:7: bad argument #1 to 'ipairs' (table expected, got nil)
    stack traceback:
            [C]: in function 'ipairs'
            [string "enum.lua"]:7: in function 'enumerate'
            [string "dat/events/tutorial/tutorial-heat.lua"]:83: in function <[string "dat/events/tutorial/tutorial-heat.lua"]:57>

This simple PR fixes that.

ids1024:
> The "..." parameter as a function argument doesn't seem to work in luajit either

Ok, fair enough. Dropping lua 5.1 support may be a good idea regardless, but it has no bearing on this PR.

bobbens:
OK. Yeah, Lua is getting a bit messy with the Lua 5.1, Lua 5.2 and Luajit which are all different subsets. I wonder why we weren't getting any issues with this. Thanks,

PR #676, opened 2016-08-18
Correct calls to glTexImage2D

ids1024:
I think this is correct. I noticed the issue while trying to port naev
to emscripten. Under emscripten, this made texture rendering work.


bobbens:
Nice catch!

PR #678, opened 2016-09-01
Tweaking robots and their assorted weapons.

Lineth:
Changed my mind of the Collective Heavies being Corvettes. It's probably better for them to be heavy fighters. Also, gave a little love to Za'lek drones. Nerfed the light drones as their shields were about as strong as the heavies (wot). Gave the EBCs to Za'lek (not ideal, tried to make a clone of it but decided eh too messy. Some lore expansion on Za'lek and Collective could justify it). Hopefully it's all in order. If anything's wrong, please let me know what!


bobbens:
Looks good, thanks!

PR #681, opened 2016-09-10
Got the traders to do convoys

nenau:
Every trader ship that jumps in looks for a suitable boss, and follows him. As a result, small convoys are sometimes put together. For now, it's mostly cosmetic because when the pilots feel danger, the convoy disbands itself.
In my opinion, that's a cheap way to get pseudo-fleets that can maybe be transformed into a real fleet system.
If you like it, I can do the same for patrol ships.


BTAxis:
Sounds like an improvement over what we have now, so I'll merge it.

PR #682, opened 2016-09-10
Patrol ship also make convoys

nenau:
I extended the convoy stuff to patrol ships, and I corrected an indentation error on previous commit.

### DIFF
--- a/dat/ai/personality/patrol.lua
+++ b/dat/ai/personality/patrol.lua
@@ -14,10 +14,20 @@ function idle ()
              ai.pushtask("land")
           end
        end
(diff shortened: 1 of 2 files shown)


BTAxis:
This sounds less of a desirable change because patrol ships are supposed to be patrolling, which means spreading out to cover as much ground as possible. The military should probably fly in a convoy, the police force not so much.

nenau:
Right, I had forgotten that point.
I guess, patrol ship could do convoys if there is a strong enemy presence in the system...
Anyway, I'll close this and propose a PR for the indentation fix.

PR #684, opened 2016-09-28
added a landingport image for rocky planets

Bisc88:
An image of a small outpost built arround a crashed ship on a moonlike planet.
I am not a good modeler, then I added a lot of post effects to make the image looks good. Lets hope these effects dont spoil it.
rock scene color balanced


bobbens:
Looks good, thanks!

PR #685, opened 2016-10-09
Map decorators for Naev, with one as example

Kinniken:


Lineth:
I like the sound of this. Would add a little something to the map Maybe another use of the faction logos :3?

Kinniken:
If it gives people ideas, here is what it looks like in Nox:
image
image

bobbens:
Sorry for the late merge. Cleaned up the code a bit and added a free function before merging. Thanks!

PR #686, opened 2016-10-21
Fix ldoc build

ids1024:
The sed/awk conversion is a hack and ideally would be eliminated, since ldoc (unlike luadoc) supports [documenting c code](http://stevedonovan.github.io/ldoc/manual/doc.md.html#Extension_modules_written_in_C). Unfortunately naev also uses doxygen comments to document the code, and those would confuse ldoc. I'm not sure how that could be dealt with...

As I've mentioned before, I'd also like to set up automatic building of the API documentation. Travis-CI might be the easiest way to do that.

### DIFF
--- a/docs/lua/config.ld
+++ b/docs/lua/config.ld
@@ -1,6 +1,7 @@
 dir = "../html"
 style = "./"
 template = "./"
+not_luadoc = true
 title = "Naev Lua API"
 project = "Naev Lua API"
 description = "Naev Lua API"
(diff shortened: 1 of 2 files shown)


bobbens:
luadoc is apparently broken in arch linux. Will have to try on another machine when I get back home (on a business trip) before merging.

ids1024:
> luadoc is apparently broken in arch linux.

I'm running Arch and it's working fine here. Just install `ldoc` from the community repository.

Luadoc may be broken, but as of #642 it uses the more modern ldoc, not luadoc.

bobbens:
I see. Had to rerun ./configure. It was also horribly broken due to my multiple lua installations, but installing over it with luarocks seems to let it work. So, looks good, thanks!

Issue #687, opened 2016-10-22
Use Travis-CI to build lua API documentation


ids1024:
The http://api.naev.org site has a tendency of getting out of date since it has to be manually updated. It would be better to automate that. Travis-CI seems like a good way to achieve that.

I've gotten this to work with my own server at https://github.com/ids1024/naev/tree/travis.  It will need to be updated to work with the official naev repository and the server naev.org is hosted on, and there are some small improvements to the .travis.yml that should be made (uploading docs only when building the master branch, etc.), so I haven't created a PR yet.

Does this sound like a good idea?

bobbens:
I think it would be cool, although I am not familiar at all with travis-cl and not entirely sure about the downsides there might be.

ids1024:
Closing. `api.naev.org` has been built by Travis for a few months now.

PR #688, opened 2016-10-22
Adjust .travis.yml to build against sdl2, and to build lua api documentation

ids1024:
This does not implement deployment (#687) of the api docs to http://api.naev.org, but is a step toward that. Travis also has the usual advantages of continuous integration; verifying if PR's build properly, verifying that the lua documentation builds properly (perhaps more important, since that is not usually tested each commit), etc.

I have successfully implemented and tested a Travis setup for deploying via rsync to my own server, but that is not included in this PR because it will need additional configuration for naev's server.

Travis will also need to be activated for the naev/naev repository by an admin at https://travis-ci.org.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,8 @@ compiler:
    - clang
 before_install:
    - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


bobbens:
Not really familiar with travis, but can try. Merged.

ids1024:
OK, the build is passing at https://travis-ci.org/naev/naev/, so this is working so far.

@bobbens If you want to set Travis up to automatically update http://api.naev.org, you'll need an ssh key for the server, which is then encrypted with a key provided by Travis (https://docs.travis-ci.com/user/encrypting-files/) and then committed to the repository. Then it can use rsync to upload (rsync wouldn't be the only option, but it seems like the best if possible).

As I've mentioned, I already have the deployment working on a branch of my fork and my own vps, so the (fairly simple) scripting is already figured out; it just requires configuration for the server.

PR #689, opened 2016-11-01
Fix memoryCheck() and memory() when called from same Lua state as AI

ids1024:
See https://forum.naev.org/viewtopic.php?f=15&t=1220 for discussion of the problem.


bobbens:
Looks good, thanks. Didn't think about that when I originalyl implemented.

PR #694, opened 2017-01-17
Use one Lua state for everything

ids1024:
This shouldn't need too much explanation, since I've discussed it [on the forum](https://forum.naev.org/viewtopic.php?f=15&t=1220) and on IRC.

The general point is to allow different parts of Naev to share arbitrary Lua objects. This could be useful for providing APIs for communication between pilots (such as for fleet AI), and could also help in debugging (by allowing arbitrary access to a mission or event's variables from the Lua console).

Those things are not implemented yet as of this PR. I also have not yet changed how `pilot.memory()` and `pilot.memoryCheck()` work for reasons mentioned on IRC.

As part of this, the way read-only Lua code is handled has also changed. Now the functions will still be visible in the module, but will produce an error message. In addition to being easier (with one  Lua state), this will result in a clear error message if someone tries to use a RW API in read-only Lua, so this provides an additional benefit.

In the process, I ended up refactoring some duplicated code into methods, in particular the `nlua_pcall()` function handles the previously redundant error handling that went with each call to `lua_pcall()`, and `nlua_register()` handles some duplicated code in registering modules.

Thus, this PR results in a net decrease in lines of code, which is always a plus.


bobbens:
I haven't gone over it all in detail, but I have tested it and checked some things. For example changed __RW to __NLUA_RW. I think it does look more than usable, so I'll merge it and we can see if people file bug reports ;). Thanks for the big patch ;)

PR #696, opened 2017-01-23
Remove some functions from Lua blacklist

ids1024:
I don't think these are an issue except with intentionally malicious code, which Naev probably doesn't protect well against anyway.

I am trying to use some Lua code (specifically https://github.com/kikito/inspect.lua) that uses getmetatable() and rawget(). If these functions are really an issue, I can probably work around it, but I don't think it's necessary to blacklist them.

### DIFF
--- a/src/nlua.c
+++ b/src/nlua.c
@@ -265,23 +265,18 @@ int nlua_loadBasic( lua_State* L )
          "collectgarbage",
          "dofile",
          "getfenv",
-         "getmetatable",
          "load",
          "loadfile",
          "loadstring",
-         "rawequal",
-         "rawget",
-         "rawset",
          "setfenv",
-         /*"setmetatable",*/
-         "END"
+	 NULL
    };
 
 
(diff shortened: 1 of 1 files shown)


bobbens:
Although I do think sandboxing is beneficial in avoiding bad practices and behaviours when possible, you're right in that we are not running arbitrary code from the internet, so it daesn't make too much sense to be paranoid about sandboxing. Although there is always the extremely far possibility of creating a sort of plugin system that lets you download stuff from the internet and use that, which would require a bit more sandboxing.
I'm tempted to only allow these functions when debug mode is enabled at compile-time, but am worried about that leading to code not working in release version issues, so for now I'll just pull ;)

ids1024:
Yeah, I know a plugin system is a vague post-1.0 goal. I don't think these functions I've enabled do any harm to the system (I think they could only interfere with other Lua code in Naev) but sandboxing against malicious code is... hard.

PR #707, opened 2017-02-26
The 2 first shark missions don't need operation cold metal to me done anymore

nenau:
The aim is to increase the number of missions available before the player gets heavy ships (needed for the collective campaign).


bobbens:
Sounds good, thanks!

PR #709, opened 2017-03-03
Asteroids

nenau:
I added asteroids as decoration for now. Read the forum subject for more info.
I had to move all the asteroids gfx into a sub-directory.
There was a mistake (or maybe something I definitely didn't understand) in the Euler function.

For now, asteroids don't add any gameplay feature at all, they are only "nice to look at".


PhoenixRiver:
Hello :)

Sudden thought as relates to the AI in asteroid fields as well as the
player... an outfit could be added that reduces or removes the effect
of damage from asteroids - call it a tactical navigation computer or
tactical navigation thrusters (or both) {it/these might also give a
benefit in combat; but probably not when also in an asteroid field
e.g. 5% of shots at you miss even if they look like they should have
hit - I guess this would be a change to collision detection} by saying
that the outfits move you in the 3rd dimension to avoid the asteroids
{shots} - as is done with ships already - but reduce your max speed
when turned on.
AI ships then come with them installed by default in areas that have
asteroids so they don't hit them by default.
The outfits also operate automatically so if you want to mine
asteroids you have to turn them off (just to make mining a little more
dangerous - as it would be!) and just get near enough an asteroid to
"land" (mine) it but not hit it...
And to extend even more - some asteroids might be "explosive" and so
shooting them might cause shrapnel to blow off and damage nearby
ships... possibly ice ones (they would also be easier to spot due to
being a different colour), I don't know if the AI could handle this
but I'm sure players would appreciate the extra potential tactics
involved in asteroid dogfighting! :)

Thoughts?

nenau:
My next goal with asteroids is to add mining functionnalities. I think that the damage-dealt-by-asteroids part will cause more difficulties to implement right.
Using asteroid fields as defensive field and as a weapon can really add things to the gameplay if it's done well, but I'm sure we'll have plenty of time discussing this on the forum.

bobbens:
Some comments:
For effects from asteroid field, we always mentioned that for dense asteroid fields and large ships, damage could be considered, although small ships should be fine. For nowe the easiest thing to do would be to have asteroids make it harder to hit your ships by increasing ew_hide (electronic warfare). This would have the effect of making turrets and swivel bolts have a much harder time tracking. As a bonus, lockon times for missiles would also be increased.

Implementation-wise, I think it is nice, but it might be interesting to split asteroids into two layers, have one "on top" of the player and hove one "below". I would also consider making the asteroids in general move much slower, and be perhaps much larger.

Keeping track of all the asteroids might be process intensive if we go with very large asteroid fields. I aways thought of making them region based. For example, define the asteroid field in the editor using a polygon, and have the effects work on all ships in them. Since the effects would be location based (and not for particular asteroids), they could only be rendered as needed and only shown in the player's viewport. Think somewhat like the nebulae puffs, which only give a nice visual effect. Of course, when interacting with the asteroids, it would be necessary to allow the player to shoot them and such. One possible way would be to, say, seperate "visual" asteroids from "real" asteroids. 

Another thing would be to have asteroids appear in the overlay map, not as individual asteroids, but as an asteroid region. If defined by polygons, it would be as simple as to display the polygon.

What do you think of doing this changes?
1. Radius -> Polygon
2. Large areas
3. Player-based rendering
4. EW effects

nenau:
EW effects seem quite easy to implement.
About larger areas, it's true that my implementation won't support hunreds of asteroids, that's why player-based rendering is necessary, but I have no idea abour how to do it right now (because the fields have borders, unlike nebulae)
About the polygon stuff, I don't know how to keep the asteroids inside a polygon with a smooth cinematic.

One way could be to use stationary asteroids, but it's somewhat lame.
I'll think about it...

bobbens:
There are two options I think. One would be to just have them sort of fade out and disappear (or shrink into the background like landing on the planets). That would also allow them to fade in around the border to keep it nice looking. The goal would be to make the density seem constant. This would probably be my recommended approach.

Another would be to do somewhat like now. That is, apply a negative force as they leave to push them back into the polygon. However, that might look weird as asteroids would accelerate and such.

nenau:
Ok, I'm on it !
Here are the limits of what I am about to implement :
* The polygon will have to be convex (to simplify inside/outside detection)
* I won't implement player based rendering because it's kind of complicated with the moving screen and the fixed boundaries of the field, but as I plan to make the computation lighter, I hope it will be still possible to have hundreds of asteroids without any lag.

Of course, these limitations can be overcome (and will be eventually) but for now, I'd prefer not to do it.
Do you agree with my plan ?

bobbens:
Yes, it should be fine. It should be possible to convert arbitrary polygons into a set of convex polygons anyway for the collision code (worst case just turn everything into triangles).

PR #711, opened 2017-03-25
Update on asteroids

nenau:
* Added support for **some kinds** of non-convex fields. There are still limitations that will lead to strange distributions of asteroids if a fied that is too complex is added.
* Added player-rendered "debris" giving the most of a field's consistency at a small cost (specially for large fields).
* Added a file "asteroids.xml", to choose the gfxs of the big asteroids.

My plan is to have in the future several asteroids types defined in "asteroids.xml", that give different resources during mining missions.


bobbens:
If possible, the limitations of the asteroid fields should be discussed, so people don't screw up. I also set the asteroid field type name to a string, as I think it's more "human understandable". However, as the code is currently, it only uses the first one. However, it should be easy to add code to choose which one to use and have it use the "default" otherwise. Other than that, looks good, thanks.

src/space.c
```diff
@@ -2792,11 +2872,127 @@ static int system_parseAsteroidField( const xmlNodePtr node, StarSystem *sys )
...
+
+               /* Add some nodes to the new subset */
+               k = 0;
+               for (m=n; m<j+1; m++) {
+                  newsub->corners[k] = sub->corners[m];
+                  k++;
+               }
+
+               /* Remove some nodes to the old subset */
+               memmove(&sub->corners[n+1], &sub->corners[j], sizeof(AsteroidSubset)*(j-n-1));
```
bobbens: Trying the code, it seems like there is an invalid read here. 8 bytes passed the allocated area, so maybe an off-by-one miss?

nenau: Yes, I've messed up there.
I guess there should be as last argument `sizeof(AsteroidSubset)*(sub->ncorners - j)` as it's the number of bytes to move.

bobbens: ok!

PR #712, opened 2017-04-07
Enables asteroid mining

nenau:
This PR contains the following : 
+ Asteroids can be destroyed by a special ammo (buyable only on Bon Sebb)
+ They spawn what I called "gatherables objects" that the player (or any npc) can scoop.
+ For now, no outfit is necessary to scoop stuff.
+ A new kind of asteroid, containing some gold (only in Zied)
+ A new commodity, gold. I decided to add this one because I wanted to test my "multi-commodities" asteroids.
+ I hope there is as few memory leaks as possible.


ids1024:
It doesn't work with "Only shoot weapons that are in range" enabled, since the asteroids cannot be targeted.

Perhaps that option should be ignored when firing this kind of missile (or asteroids should be forgettable).

nenau:
In my opinion, this is not very important because the player should have one weaponset with his regular weapons and "only inrange" enabled, and one weaponset with only the anti-asteroid stuff and "only inrange" disabled.

ids1024:
Well, it took me a while before I figured out how to shoot at asteroids (since I tend to keep that option on, and didn't think of it). But yes, that probably is how the player should configure things.

nenau:
If you got problems with that, it will be awful for a beginneer (I just rembered the first time I equipped new weapons and they were not in the right weaponset) .
I guess I'll have to add an entry about mining in the tutorial once stuff is stabilized. Maybe, it could be good to have also npcs at bar talking about that (and about the weaponset system in general).

BTAxis:
A problem with the tutorials is that we cannot effectively guide players through the weapon set system, because there's no way to do so interactively while the menu is up. So we're basically limited to walls of text explaining how it works, which is a crap way of explaining anything.

PhoenixRiver:
Since the weapons used are currently only used for asteroids and they
don't work with only_weapons_in_range enabled but can only be used for
asteroids then it makes no sense that it would pay any attention to
only_weapons_in_range. That option is completely useless for asteroids
(at the moment) so if it's a small change it would make sense to
ignore only_weapons_in_range for the moment. If it's a major change
then maybe that's no viable for a possibly temporary problem (since
further changes may make this a moot point).

Just my two cents,

River

nenau:
It's probably a small change to ignore `only_weapon_in_range` for asteroid wapons, but those weapons are also usable against regular targets. Maybe players and Ais will want to use it in desperate situations...
Personnaly, I would prefer to be able to target asteroids (that requires quite a lot of coding) than to ignore `only_weapon_in_range` for asteroid wapons.

bobbens:
Sorry for the late reply. I tried the branch and have some comments:
* Asteroids that can be destroyed should be easily identifiable.
* Instead of low-ammo missiles, probably mining lasers would be more enjoyable to use. At the moment you run out of ammo fast.
* I wasn't able to scoop stuff, should be made easier (or have an active outfit that drags stuff towards you).
Overall I like the idea, but it should be made more user friendly. Another option would be instead of making the player do stuff himself, have an outfit that sends little drones out and mines for the player. This would change the mining task from actively targeting and destroying asteroids to having to stay around and defend the little drones, as with safety lanes (hopefully implemented some day), asteroid fields would be prime areas for pirate activity.

nenau:
Hi,

* I guess, if I implement a targeting system for asteroids, it could be reliable a way to distinguish destroyable and not destroyable asteroids, beside solving the only_weapon_in_range issue.
* About ammo, the aim was to make the player have to be kind of careful : if you have unlimited ammo, you can blow up every asteroid and then try to scoop stuff, but in the other case, you have to be parsimonious in order that this activity gives benefit. Of course, the amount of ammo per launcher is adjustable. I put this value because I saw that it was enough for my shark's commodity capacity. And of course, if you still think that unlimited ammo is better, I can change that point.
* Quantity, price of ammo, and distance and relative velocity necessary to scoop are balance parameters that can be adjusted at will. My first idea was to make only small ships able to do mining, because a player that has a Kestrel has more interesting and dangerous things to do.
* About drones : I agree, it could be a good idea. For now, my idea was to first implement the mining mechanism, than make a dedicaced AI, and after that, have drones that use this AI.

By the way, well done with the release. I'm gonna play it, and afterwards, I will start working again on this mining stuff.

nenau:
Hello, I updated the stuff in order to correspond to what we agreed:

* A targeting system for asteroids was implemented (it solves the 'only weapons in range' issue)
* A bolt mining weapon was added (asterokiller gauss gun). Right now, it is possible to purchase both the launcher and the gun on Bon Sebb
* The ammo of the launcher was increased
* The scooping was made easier

src/economy.c
```diff
@@ -299,6 +316,120 @@ void commodity_Jettison( int pilot, Commodity* com, int quantity )
...
+      gat = &gatherable_stack[i];
+
+      if ( .03*vect_dist( &p->solid->pos, &gat->pos ) +
+           .03*vect_dist( &p->solid->vel, &gat->vel )  < 1. ) {
+         /* Add cargo to pilot. */
+         q = pilot_cargoAdd( p, gat->type, RNG(1,5), 0 );
+
+         if ( q>0 ) {
+            if (pilot_isPlayer(p))
+               player_message( _("%i tons gathered"), q );
```
ids1024: This should probably say "tons of ore/gold/etc. gathered".

src/player.c
```diff
@@ -1304,6 +1315,44 @@ void player_targetPlanetSet( int id )
...
+   if (old != id) {
+      if (id >= 0)
+         player_soundPlayGUI(snd_nav, 1);
+
+         /* Recover and display some info about the asteroid */
+         anchor = &cur_system->asteroids[field];
+         ast = &anchor->asteroids[id];
+         at = space_getType( ast->type );
+
+         player_message( _("Asteroid targeted, composition: ") );
```
ids1024: Adding this was a nice touch. But perhaps the player should have to buy a special scanner to get this information?

src/economy.c
```diff
@@ -299,6 +316,120 @@ void commodity_Jettison( int pilot, Commodity* com, int quantity )
...
+
+   p = pilot_get( pilot );
+
+   for (i=0; i < gatherable_nstack; i++) {
+      gat = &gatherable_stack[i];
+
+      if ( .03*vect_dist( &p->solid->pos, &gat->pos ) +
+           .03*vect_dist( &p->solid->vel, &gat->vel )  < 1. ) {
+         /* Add cargo to pilot. */
+         q = pilot_cargoAdd( p, gat->type, RNG(1,5), 0 );
```
ids1024: This can fail if the ship does not have cargo space left. It may be worth displaying a message in this case; otherwise it is confusing when nothing is gathered.

nenau:
I did the modfications (and I cleaned up a bit)

bobbens:
Hello, tried the branch, looks much better. Fixed a small issue with plurals (1 tons) and pushed to mining branch. One thing that isn't entire convincing me is the way the ateroid scanner works. Instead of having it just print a message to the console, it should automatically display targetting reticles on all "shootable" asteroids or something like that and add what ores as a small icon. I was also thinking that non-shootables could be darkened a bit and give more of a sense of background. Furthermore, adding "foreground" asteroids would also give a more sense of depth to the field.

Another thing, you use malloc, realloc and those commands often, but I would encourage you to look at src/array.h that is a wrapper for vectors and can simplify the code quite a bit.

Finally, given that the lasers can blast asteroids really fast, it might be better to encourage the use of missiles by giving them some sort of advantage. Although as a side note, I think moving to a system where there is no explicit need to buy "ammo" for missiles is a better route.

What do you think?

nenau:
Hi,

* About the scanner, I'm not sure how to do it properly : if all the asteroids get a recticle and icons, it could become confusing...
* Darker not shootable and foreground asteroids seems good to me, I'll do it if I don't have any technical issue.
* I'll look at array.h for further work with vectors, thanks for the hint.
* I've kept both systems (missile and laser) because I didn't want to choose between them. The problem with endless and priceless ammo imho is that the player will be tempted to shoot as many asteroids as possible in order to recover the ore easily. On the other hand, it can be boring to always have to take care of the amount of missiles you have to buy before taking off. Anyway, the choice between both can be done easily at the last moment, and reverted at will after other players have tested it.

bobbens:
OK. I merged it. Some comments:

1. I believe saw an asteroid disappear near the edge of the field. It would be better to have them fade out gradually (fade to black then fade to transparent) to be less jarring to the player. Or have them change speed to move back to the center of the asteroid field.
2. Opened ticket #761 on better graphics for debris.
3. Opened ticket #762 for an outfit to make it easier to collect debris (it's a PITA in a capship ;) ). If you think it's OP it could be easily made a mission reward.
4. Opened ticket #763 related to incorporating asteroids and mining into the game so the player feels more involved.

Thanks for the great work!

bobbens:
Didn't get closed automatically.

PR #713, opened 2017-04-09
fleeing ships use planets if they are closer

nenau:
I had to modify `aiL_getnearestplanet`. I use `1./0.` please tell me if there is a problem with that...


bobbens:
Looks good, thanks!

PR #715, opened 2017-04-27
Alternate system for allying with the pirates

onpon4:
Here's a rework of the way you align yourself with the pirates. I haven't tested it much, so it could be that exact numbers should be adjusted, but I think the concept works well and it seems to me to be better than the current system, at least.

Here's how it works:

1. You can gain reputation with the pirates by killing, attacking, and all-around causing mayhem. You lose reputation by killing pirates.
2. Every time you change systems, there's a chance of losing reputation. This was already in place, but I modified it so that the floor starts at -20 and is adjusted up by missions, rather than being a static 20. It cannot rise above -1. As a result, you have to continue committing piracy to stay friendly with the pirates, no matter how many campaigns you've completed.
3. Campaigns are used to boost the mission cap (as with other factions) and, additionally, to boost the reputation floor. Currently, the Hitman campaign just adjusts the reputation floor.
4. The "Pirate Landing Pass" is no longer given to the player, though I haven't removed it outright at this point.
5. Missions (which I haven't worked on yet) should give both reputation and money.
6. It's essentially impossible to both be a pirate and be friendly with the other factions.

So in a nutshell, this system is simpler, yet more difficult. I think this is a far better system for someone who wants to commit galactic piracy.

Of course, I realize this is a pretty substantial change, so if there is disagreement on this, I am open to suggestions.

### DIFF
--- a/dat/events/pirate_fame.lua
+++ b/dat/events/pirate_fame.lua
@@ -66,27 +66,22 @@ end
 function create()
    local fame = faction.playerStanding("Pirate")
 
-   -- If the player already has a low reputation, just stop there. 20 is the
-   -- minimum required to land on a pirate planet without having to bribe the
-   -- guy in charge of landing ships.
    local floor = var.peek("_ffloor_decay_pirate")
    if floor == nil then floor = -20 end
    if fame <= floor then
       evt.finish()
    end
 
-   local maybe
+   local amt
    if using_pirate_ship() then
-      maybe = 0.05
+      amt = 0.15 + rnd.sigma() * 0.05
    elseif using_impressive_ship() then
-      maybe = 0.10
+      amt = 0.5 + rnd.sigma() * 0.10
    else
-      maybe = 0.15
+      amt = 1 + rnd.sigma() * 0.25
    end
 
-   if rnd.rnd() < maybe then
-      faction.modPlayerSingle("Pirate",-1)
-   end
+   faction.modPlayerSingle( "Pirate", -amt )
 
    evt.finish()
 end
--- a/dat/events/pirate_fame.lua
+++ b/dat/events/pirate_fame.lua
@@ -69,7 +69,9 @@ function create()
    -- If the player already has a low reputation, just stop there. 20 is the
    -- minimum required to land on a pirate planet without having to bribe the
    -- guy in charge of landing ships.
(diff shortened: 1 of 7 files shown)


onpon4:
Update: I've done some testing on this, and based on that, I've made some tweaks to the numbers. I can say now with confidence that the system works well. It just feels like what a pirate-aligning player should do: be extremely hostile to anyone who isn't a pirate, leave no survivors, etc. It fits really well.

ids1024:
This sounds to me like it's generally a good idea. How should the player find out that they can gain status with pirates this way? (Although that is not a new problem, since the previous was fairly opaque; probably more so.)

It also is generally quite hard to play as a pirate; but that's a somewhat separate issue. (Although it does potentially impact this, if this change makes it harder to be able to land on pirate planets and also have a non-negative reputation with other factions).

onpon4:
I figured a couple of things:

1. Some text could be added to the random citizen dialog discussing this. They could also discuss the existence of the pirate strongholds a bit more transparently, since that too is not exactly obvious. This would be enough to get the general idea of how to become a pirate.

2. Pirate-allying missions can still be offered in Alteris, or any pirate-infested independent system really, and they could give you hints about how the whole system works (as well as, again, the existence of pirate strongholds). This would give enough information for the player to know how to *reliably* become a pirate, e.g. by pointing out that it's a lot easier if you pay off (bribe) the nearby pirates rather than killing them.

3. Pirates at pirate strongholds could explain the nuances, like how using a pirate ship keeps your reputation up longer than using a civilian ship.

onpon4:
Are there any other thoughts on this? Anything that makes it unfit to merge?

bobbens:
Sorry about the delay. I think it works fine, but with pirates there's always been a sense that it's incomplete. The way original EV did it was lackluster and we sort of followed that, in the sense that being a pirate is high-maintenance and annoying. Being a pirate shouldn't be about just flying around killing everything, but being much more stealthy and avoiding patrols and stuff. For landing, bribing when necessary, but there should be a bunch of pirate "back-ports" that once you get access to would let you land normally, just having to avoid patrols and being scanned. However, that is fairly another topic and the proposed pull request does improve the current stuff.
Thanks for the pull request, merged!

PR #716, opened 2017-04-27
Fixed failure to account for positive distress.

onpon4:
I found this while I was working on #715. An implicit assumption causes the Lua code to fail if any faction allows a positive distress response, as I made the case for the Pirate faction there. This is just the fix to that problem without the other stuff in #715.


onpon4:
I've just pushed another fix into this branch: Wardens End incorrectly didn't have the "blackmarket" tag, so I added it in.

onpon4:
One more fix: as an easy way to make this property configurable in the editor, and to make the XML a little cleaner, a planet is made a black market by having the "blackmarket" service instead of the dedicated "blackmarket" tag. The old tag is still supported to maintain compatibility.

bobbens:
Blackmarket as a service instead of a tag sounds good. Thanks!

PR #723, opened 2017-05-17
Removed a duplicate entry.

onpon4:
I'm not sure how it happened, but a perfect duplicate of the "FLF Instability" mission entry was added somewhere down the line. Fixed.


bobbens:
Looks good, thanks!

PR #725, opened 2017-05-22
[Docs] Replace '=' with '=='

Fayti1703:
Comparison is done with the latter, not the former.


bobbens:
Nice catch, thanks!

Fayti1703:
Just remember to rebuild the LuaDoc sometime.

ids1024:
> Just remember to rebuild the LuaDoc sometime.

https://naevdocs.ids1024.com updates automatically on push. And so will `api.naev.org`, if someone with access to that sets it up.

PR #729, opened 2017-07-11
Improve macOS support, automated builds

stephank:
In actual code, I've made the following changes:

 - `GIT_COMMIT` is now short hash only. The previous format depended on how the repo was acquired. E.g., Travis does a shallow clone, has no tags, so `GIT_COMMIT` was already just the short hash.
 - Added a glue file that can talk to ObjC land.
 - Changed paths on macOS from XDG to macOS-specific paths:
   - Config: `~/Library/Preferences/org.naev.Naev`
   - Data: `~/Library/Application Support/org.naev.Naev`
   - Cache: `~/Library/Caches/org.naev.Naev`
 - Added check for ndata in app bundle `Resources` directory.
 - Removed the Xcode project, builds are now produced using regular build infra.
 - Added a shell script to create the app bundle, including dependencies.

This PR also contains changes to Travis to create automated builds, and upload them to a free account on bintray: https://bintray.com/naev/naev/naev

Currently, this builds release versions for 64-bit Linux and macOS. The macOS builds should support at least 10.11 (from Sep 2015), and may go back a bit further if we're lucky. The Linux builds, I really don't know. 😇 (32-bit Linux is also a bit more work on Travis.)

To properly merge this, we need to setup travis/bintray for `naev/naev`, because it's currently setup for `stephank/naev`. The steps are:

 - Someone needs to create a bintray account
 - I add them to the bintray org (and probably transfer ownership as well)
 - They encrypt their bintray API key for Travis: `travis encrypt -r naev/naev $API_KEY`
 - We update `.travis.yml` with the new API key

The build failure here is merely because I'm also executing deploy steps for `stephank/naev`, and the luadoc deploy fails (because travis won't decrypt the secret for my fork).


ids1024:
Good work. I was meaning to try to get Travis to build on Mac and possibly upload the result, but I'm not really familar with macOS.

We could probably set up AppVeyor to have a Windows build too.

> Changed paths on macOS from XDG to macOS-specific paths:

Does that mean that someone who upgrades won't see their old savegames, and their configuration will be reset?

stephank:
@ids1024 That's correct. Not sure what we want to upgrade path to look like (just docs or code to handle it?), but putting the files there is the way forward, IMHO. :)

bobbens:
@stephank what we did when moving to xdg was add some script to move files to the right place. See naev-confupdate.sh and src/naev.c:404.

stephank:
Seems simple enough. There's a `TODO` to remove that code. If I'm going to meddle with this, should I remove the old code, or do we want to keep it? It appears that migration was 6 years ago. :)

bobbens:
You can remove the old code. It's bitrot ;)

stephank:
I rebased, added the migration stuff, and tested the last build on a fresh VM. The secret in `.travis.yml` is now for `naev/naev`. :)

bobbens:
Looks good, thanks!

PR #730, opened 2017-07-13
macOS fixes

stephank:
Some macOS build fixes, and also app sandbox preparation.

While Naev now works in the app sandbox, it means ndata must be inside the app bundle. Outside of the bundle, you can drag ndata to Naev and it'll work, but this creates a temporary sandbox exception that is forgotten when Naev closes. (So you'll get `insert ndata` on every launch.)

So we'd have to either instruct users to place ndata inside the bundle, or (better) distribute Naev with ndata already in place. (Better, because I believe we can then also sign the entire bundle including ndata.)

The other thing is that app sandbox involves a manual migration of config and save games. Apps running in the sandbox get a private home directory. We can't really help the user, because the old locations are all outside the grasp of the sandboxed app.

So yeah, I've left that off for now, but it's there if we want to. Biggest benefit is that we could eventually get a $99/year developer cert from apple, and launch on the App Store. :)

### DIFF
--- a/extras/macos/Info.plist.in
+++ b/extras/macos/Info.plist.in
@@ -7,7 +7,7 @@
 	<key>CFBundleIconFile</key>
 	<string>naev.icns</string>
 	<key>CFBundleIdentifier</key>
-  <string>org.naev.naev</string>
+	<string>org.naev.Naev</string>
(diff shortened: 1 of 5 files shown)


bobbens:
Looks good, thanks!

PR #741, opened 2017-08-08
Adding Outfits to Thurions

ValerioCietto:
It would be cool to have Thurions to sell some of their specialities.


bobbens:
Looks good (for now), however, they shouldn't be acquirable without doing missions and working for the Thurion. I guess making landing permits necessary would be a good way to solve it.

PR #752, opened 2017-12-30
Added a landingport image for residential planets

Bisc88:
An image of a neighborhood with futuristic cars, futuristic houses and futuristic people.

residential

I tried to texturize the houses but white looks more futuristic (I don't know why, I think it is because it looks clean).

Happy new year.


Bisc88:
And sorry for the excess of commits, I messed up with the gitcode, but I think it is now fixed.

Bisc88:
Okay, it is not fixed, I will delete this branch and open a new one.

Issue #755, opened 2018-01-02
The "Sharkman is back" mission is broken


abenkovskii:
Naev v0.7.0
How to reproduce:
1. Accept the "Sharkman is back" mission
1. Buy a destroyer class ship
1. Fly to Toaxis
1. Let the shark disable your ship

Result:
Somewhere after your ship is disabled and before you can move again the mission mails without a message pop-up explaining why it failed.

abenkovskii:
My observations so far:
1. After the shark disables my ship it flies to the Raelid warp gate and just sits there randomly spinning
1. This is apparently a very fast and reliable way to die
1. When the shark dies to pirates the mission fails

PR #756, opened 2018-01-02
Fixed the "Sharkman is back" mission

abenkovskii:
Fixes naev/naev#755
I experimented a bit. It seams that the reason the shark was behaving so strange is that it was out of fuel.
This pull:
1. Ensures that the shark has fuel to jump out of the system once the battle is over
1. Ensures that the shark jumps out through the Ingot warp gate that doesn't spawn any pirates
1. Adds the missing pop-ups that explain to the player why the mission failed


bobbens:
Nice catch, thanks for the pull request.

PR #760, opened 2018-01-13
Changed the way spawn position is chosen

nenau:
I propose a new implementation of the way to ensure that all the ships in a fleet spawn from the same jump/planet. I belive, this implementation is more handy as the previous one.
The only visible effect of this modification is that the escort of a capship now jumps-in in the same time as the capship.


bobbens:
Looks good, thanks!

PR #764, opened 2018-01-16
Corrected a funny bug in the Sirius preach event

nenau:
This one was causing the event to crash in a very annoying way. _ was an integer !

### DIFF
--- a/dat/events/sirius/preach.lua
+++ b/dat/events/sirius/preach.lua
@@ -162,7 +162,7 @@ function theFunBegins()
    end
 
    followers = addRawShips(followers, "sirius_norun", curr, "Sirius") -- The table now contains pilots, not ship names.
-   for _,j in ipairs(followers) do
+   for k,j in ipairs(followers) do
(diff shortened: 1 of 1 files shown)


BTAxis:
That is valid Lua though. Why would it cause a crash?

BTAxis:
Ah, looks like _ is not a special variable but just a convention. Hokay.

ids1024:
Grepping for `for _` shows a lot of code doing this (and probably there are other uses of `_` that aren't caught by that search). Although most of those cases don't then call `_()`.

This is sort of awkward. Using `_` for gettext in some places and in assignment where the value isn't needed is likely to cause issues. So one of those uses should probably be avoided...

bobbens:
Mmmm. Using _,k in enumerate(...) was (is?) encouraged in Lua, but using _(...) for gettext is standard practice which is why this has happened. Things like this really make me appreciate static typed languages ;)

PhoenixRiver:
Yeah, that does suck a bit and static typing definitely does have its
benefits! :D

PR #765, opened 2018-01-20
Correction of a warning in unique

nenau:
The variable weapon did grow at each call of the unique function, causing warnings of the form `Attempting to equip more that 8 weapons on Empire Hawking`. I guess, this means that the ships didn't get exactly the equipment we expected them to have.


ids1024:
I suppose this code should probably be refactored so that `weapons` isn't a global.

PR #777, opened 2018-04-30
Add mechanism for persisting pilots on jump; use for escorts

ids1024:
This has the advantage that damage is persisted, and the same mechanism
can potentially be used for other purposes (enemy ships in a mission).


BTAxis:
Nice one.

PR #782, opened 2018-08-18
Update ARTWORK_LICENSE

FedericoCeratto:
Reflow file to allow machine parsing. Correct typos and missing paths.
Remove nonexistent file uninhabited-jungle.png


dat/gfx/ARTWORK_LICENSE
```diff
@@ -1142,9 +1124,8 @@
        * uninhabited-rocky-magenta-03.png
        * uninhabited-rocky-magenta-04.png
 
- * Author: Fernando L. Copercini "Bisc88" <fernandoluis1732@yahoo.com.br>, License: GPLv3+
+ * Author: Fernando L. Copercini "Bisc88" <fernandoluis1732@yahoo.com.br>, License: GPLv3
```
ids1024: Why remove the + from GPLv3+ here? These are distinct licensing terms. Software licensed GPLv3+ can also be used under the terms of version 4 or later of the GPL, if GNU ever releases another version.

dat/gfx/ARTWORK_LICENSE
```diff
@@ -1142,9 +1124,8 @@
...
    * planet
      * exterior
        * snowy_cliffs.png
        * rocky_small.png
-       * uninhabited-jungle.png
```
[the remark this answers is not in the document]
FedericoCeratto: The file is missing in the 0.7.0 tarball. I'll update.

dat/gfx/ARTWORK_LICENSE
FedericoCeratto: It looked like a mistake as all other files are under GPLv3 but I found the commit from the author.

dat/gfx/ARTWORK_LICENSE
```diff
@@ -918,20 +918,12 @@
 
  * Author: Unnamed, License: CC-by-sa 3.0
    * outfit
-     * space
```
ids1024: Likewise, these shouldn't be removed. I think all the files here are in git.

onpon4:
I've implemented most of the proposed changes in this PR, excluding removals (which as @ids1024 mentioned isn't something we want), via d855c9f.

PR #787, opened 2018-10-05
Increased simulation time on warp-in from 15 seconds to 5 minutes.

onpon4:
This makes it so that you're not effectively guaranteed to warp into
the middle of a battle zone, which can be a problem if you're
in hostile territory.

### DIFF
--- a/src/space.h
+++ b/src/space.h
@@ -18,7 +18,7 @@
 #include "explosion.h"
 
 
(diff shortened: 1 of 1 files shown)


onpon4:
Cut it down to 2 minutes since bobbens thought 5 minutes might be too much for older hardware.

PR #788, opened 2018-10-07
Bugfix

onpon4:
The code was treating every ship on the map as an escort, from the looks of it. An additional check stops this problem from occurring. Yay!

Please excuse the long list of commits; I pulled my entire master commit history into this so I could get my master in-line with naev/master. There was one difference I wasn't able to make go away (the ending of some files), but all other differences besides the bugfix have been neutralized.


ids1024:
You should probably use `git rebase` rather than merging all these commits into naev/naev.

onpon4:
Sorry... doing a git rebase gives me so many conflicts from years ago, I don't even know where to begin with it. I probably should have rebased all the way back in 2015, but I didn't know about that command at the time.

Of course, if merging all of this junk won't do, cherry-picking 91a801b will pull just the bugfix.

BTAxis:
Make a new branch based off naev/master, then cherry-pick your patch and make a new pull request with that.

PhoenixRiver:
> The code was treating every ship on the map as an escort, from the looks of
> it. An additional check stops this problem from occurring. Yay!

Sounds like a good fix...

> Please excuse the long list of commits; I pulled my entire master commit
> history into this so I could get my master in-line with naev/master. There
> was one difference I wasn't able to make go away (the ending of some files),
> but all other differences besides the bugfix have been neutralized.
> You can view, comment on, or merge this pull request online at:

but I'm having trouble working out which thing it is amongst the mess
:( - that's why this sort of thing is unhelpful - particularly if
someone is bug hunting through the update history later...

River

PhoenixRiver:
> Of course, if merging all of this junk won't do, cherry-picking 91a801b will
> pull just the bugfix.

Nice :)

PR #789, opened 2018-10-07
Fixed the every-ship-warps-in-with-you bug.

onpon4:
The code was treating every ship on the map as an escort, from the
looks of it. An additional check stops this problem from occurring.
Yay!


onpon4:
Sorry, this one has just the bugfix.

Issue #779, opened 2018-05-04
Fuel remains at 99.9999999 for the last jump in autonav


amusedberry:
When using autonav sometimes fuel drops to 99.999999 during the last jump, producing the cancellation of the autonav feature and forcing to refuel in a planet.

If you are running against the clock, like in several urgent missions, then you will miss the deadline, as stoping in a planet spends 1 cycle.

![image](https://user-images.githubusercontent.com/32961675/39610555-db17bd14-4f16-11e8-9d1d-9618cb49b6eb.png)
![image](https://user-images.githubusercontent.com/32961675/39610595-1e596f50-4f17-11e8-80cf-443441274437.png)

spectrevil:
Hi. I'm still not completely sure, but this issue only seems to happen when there is an Afterburner equipped. I have been testing with mules filled with fuel pods and it does not happen even if autonav is set to 8-10 jumps.

bobbens:
Ok, it's likely a rounding issue somewhere or other. I think it might make more sense to just stop using floating point for fuel.

PR #791, opened 2018-10-17
Change fuel and fuel consumption from double to int

thunkk:
Might fix https://github.com/naev/naev/issues/779


ids1024:
I was going to comment that the `player.fuel()` and `player.refuel()` Lua functions should be changed to handle integers, but apparently before 5.3 Lua only has doubles, with no separate integer type. I don't know if Luajit handles that differently... or if that has any implications we should care about.

PR #793, opened 2018-10-18
Some stuff that was missed while converting fuel to int

thunkk:


bobbens:
Looks good, thanks! (rebased and pushed)

PR #801, opened 2018-11-20
Modified player_land() so that you only need to push the land key once.

onpon4:
Simple modification: in player_land(), if permission to land is
granted, it calls itself a second time so that the actual landing
procedure begins. Simple rationale: nothing is gained by making landing
a two-step process, it just makes doing so take more effort needlessly.

The tutorial has also been updated to reflect the change, of course.


bobbens:
Looks good, thanks!

PR #802, opened 2018-11-20
Fix sillyness in sensor array description

Fayti1703:
More powerful sensors should be able to detect ships further way, thus increasing the distance at which ships can be detected.

Discovered while looking through #800


ids1024:
Merged, since this seems uncontroversially a silly mistake. Thanks!

PR #804, opened 2018-11-24
Fixed a mistake that causes a crash

onpon4:
Sorry about that.


onpon4:
I snuck in a change to where the FPS is displayed as well. Again, sorry for this sloppiness. on my part.

PR #814, opened 2019-01-03
Added variation to passage of time between ships.

onpon4:
I briefly mentioned this idea I had on IRC yesterday, and thinking it over, I decided to implement it. I'm very happy with the results.

This is a pretty simple idea: heavier classes of ships have time pass at a faster rate than lighter classes. In effect, this simulates the phenomenon where larger animals "feel" time passing faster than smaller animals, and it has two effects:

1. Lighter ships now have an inherent advantage over larger ships in that they enable better reaction time.
2. Heavier ships feel much less slow, making piloting them feel a lot nicer.

I decided to make Corvettes go at 1.5x, Destroyers and medium transports go at 2x, Cruisers go at 2.5x, and carriers go at 3x. Fighters, couriers, and yachts go at 1x, as before.

Additionally, I made the Hyena a special case that goes at 0.75x, along with the regular Collective Drone. I also buffed up the Hyena's maneuverability so that players can better take advantage of the slower time passage. It works very well, and it makes the Hyena much more competitive. This also has the nice side-effect of making pirates a bit more threatening. (Note: the Collective Drone did *not* have such a buff applied to it; it's the same as before, just with slower time passage.)

While I was doing this, I also noticed that the feature that speeds sound up would completely silence sound at above 2x or so, so I capped it to that for now, until someone can come up with a more complete solution.


bobbens:
Thanks! Changed to time dilation and pushed (rebased to master).

PR #824, opened 2019-05-07
Display build in stdout, as for instance : Naev v0.7.0 build for Win64

Mutos:
Display build in stdout, as for instance : Naev v0.7.0 build for Win64

### DIFF
--- a/src/naev.c
+++ b/src/naev.c
@@ -1219,7 +1219,7 @@ char *naev_version( int long_version )
    if (long_version) {
       if (human_version[0] == '\0')
          nsnprintf( human_version, sizeof(human_version),
-               " "APPNAME" v%s%s - %s", short_version,
+               " "APPNAME" v%s (%s)%s - %s", short_version, naev_build(),
 #ifdef DEBUGGING
                _(" debug"),
 #else /* DEBUGGING */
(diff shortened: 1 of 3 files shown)


Mutos:
Hi all,

Tried to understand what happenned, understood at last what Travis is, but no clue to what happenned to the Mac build... Someone with Travis XP please could look ? Thanks in advance ^-^

Mutos

ids1024:
The failure seems to be spurious and unrelated to this change; I retried the build and it passed.

What is the motivation for this? Is it mainly to see if Naev is compiled for 32 or 64 bit?

I think there should be a good way to use autotools to do this with less code while supporting far more OSs...

Mutos:
OK for the failure ^-^

For Windows, I got both compilers setup and I compile OOT, so I like to have a clear display of 32/64. Maybe less useful under other OSs... But it may be useful to have the win32/64/others distinction readily available in code.

Also, I display the build in 2 places : stdout and menu. It's OK for me, but others could want it in more or less places...

And of course, if you got a better way to do it, I've no problem with that ^-^

ids1024:
> Maybe less useful under other OSs...

Personally on Linux I'd just use the "file" command to check what type of file it is. But sure, this seems reasonable. It's not uncommon for programs to display this, and it could be moderately useful to have the information if someone posts a log.

Closing in favor of commit https://github.com/naev/naev/commit/d0472cd0464df24bda4968d6401afa06e34dc2ed, which uses a little autoconf magic to provide a more general solution with basically no code. Hopefully this solution fits your requirements well.

 For me this prints "Naev v0.7.0 (x86_64-pc-linux-gnu)".

Issue #826, opened 2019-05-10
Non-convex asteroids field ends up messed up by convex subsets generation


Mutos:
I'm currently adding an Asteroids Fields display to the System Editor Map. It now displays OK for convex and non-convex fields by using the convex subsets.

However, I found out that a non-convex asteroids field is loaded OK, displayed OK, but is saved incorrectly after opening and closing the System Editor. The # of corners is still OK, but the corners coordinates are messed-up with seemingly random numbers.

The issue was already present in master branch, commit b75cf4e91, to find it out, I did the following :
- Checkout master branch,
- Modify Hakoi to have a non-convex field, like in my branch 'error-convex-asteroids',
- Edit system Hakoi, only check automatic save, and close System Editor,
- Inspect the hakoi.xml file : it is now messed-up.

I created a branch 'hakoi-test' in my repo Mutos/naev, with only the hakoi.xml file changed, for you to inspect.

Mutos:
I tracked the error out of the Editors...

I logged asteroids fields properties upon entering Universe Editor, before any change can be made.

Here is the result for my test field on Hakoi :
  [stderr.txt](https://github.com/naev/naev/files/3168983/stderr.txt)

It shows the convex are generated OK, but the corners coordinates are different from what is in the file. This occurs BEFORE entering the Universe Editor.

So my next target is convex generation...

Mutos:
By logging at the system loading stage, I found out the issue lies in convex subsets generation.

Now to fix it, I'll have to dive into it and it can take some time...

[stderr.txt](https://github.com/naev/naev/files/3169014/stderr.txt)

Mutos:
Issue fixed, you just have to merge the PR to get it.

bobbens:
Closing as it seems the fix was merged.

PR #827, opened 2019-05-11
Fixed Issue #826 : Non-convex asteroids field ends up messed up by convex subsets generation

Mutos:
When loading a non-convex Asteroids Fields, the field's corners ended up with incorrect values, seemingly random, hinting at a memory conflict issue.

First thought it was in my new System Editor with Asteroids Fields display, but tracked to issue down to the Universe Editor, then to the system loader itself.

Finally, the issue was caused by making the 1st convex subset's corners array a pointer on the field's same array. When further operations caused corners to get removed from the subset, the field's corner array got downsized as well. Then memory could be reallocated to other purposes, causing random values to appear.

Fix : allocated a separate memory zone for the subset's corners array and copied coordinates from the field's corners into the subset's corners.

Note : maybe this could be done by a memory copy, my inexperience in this matter led me to loop and copy each corner's x and y fields. Feel free to optimize ^-^

### DIFF
--- a/dat/ssys/hakoi.xml
+++ b/dat/ssys/hakoi.xml
@@ -45,20 +45,20 @@
     <y>2500.000000</y>
    </corner>
    <corner>
+    <x>-5000.000000</x>
+    <y>5000.000000</y>
+   </corner>
+   <corner>
+    <x>-5000.000000</x>
+    <y>-5000.000000</y>
+   </corner>
+   <corner>
     <x>0.000000</x>
     <y>-7500.000000</y>
    </corner>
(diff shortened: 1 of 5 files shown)


Mutos:
The right data in dat/ssys/hakoi.xml is the one on my fix-nonconvex-asteroids branch. The display-asteroids branch was polluted by the test file which was rewritten by issue #826 and never fixed. Merge of display-asteroids into the master branch polluted this file on the master branch. Merging the fix-nonconvex-asteroids branch will fix this.

PR #831, opened 2019-07-28
Added several additional factions to the pirate "empire bounty" mission.

onpon4:
The mission, seen by players as "Assassination Job", now applies to
the following factions in addition to the Empire: Dvaered, Soromid,
Sirius, Frontier, Za'lek, Trader, Civilian, Independent.

Because ships available vary, not every faction is available at every
difficulty level. For example, Civilians only generate very low-level
bounties, while on the other hand Za'lek bounties are always high
since it would make no sense for anyone to put a bounty on a drone
(and manned Za'lek ships are all Corvette class and higher).

This also removes the attempt to give names to pilots, since there is
currently no code to do so with any faction other than the Empire.

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -960,7 +960,7 @@
   <avail>
    <priority>2</priority>
    <cond>player.numOutfit("Mercenary License") &gt; 0 or planet.cur():blackmarket()</cond>
-   <chance>660</chance>
+   <chance>860</chance>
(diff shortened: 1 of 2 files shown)


onpon4:
Just want to ping this. :)

onpon4:
Sorry to be pushy, could I get a merge or alternatively an explanation if it can't be merged yet? @BTAxis

PR #834, opened 2019-08-02
Extended the "Assassination Job" mission to work on more factions

onpon4:
The mission, seen by players as "Assassination Job", now applies to
the following factions in addition to the Empire: Dvaered, Soromid,
Sirius, Frontier, Za'lek, Trader, Civilian, Independent.

Because ships available vary, not every faction is available at every
difficulty level. For example, Civilians only generate very low-level
bounties, while on the other hand Za'lek bounties are always high
since it would make no sense for anyone to put a bounty on a drone
(and manned Za'lek ships are all Corvette class and higher).

This also removes the attempt to give names to pilots, since there is
currently no code to do so with any faction other than the Empire.

(This is the same as #831; the reason for the new pull request is because I originally accidentally had it on my master branch.)


onpon4:
I tweaked the mission so that it searches through hidden jumps for target systems. Makes it possible for missions further out to be given (since pirates use a lot of hidden jumps), and opens up the mission to some worlds where it was effectively closed off like Foundation Station.

BTAxis:
This currently has a conflict. Just one include line though.

onpon4:
Yeah, that's because of #839. One sec...

onpon4:
There you go, it's good to merge again now.

PR #837, opened 2019-08-02
Added a pirate version of the patrol mission.

onpon4:
It doesn't actually work right now, not because there's anything wrong
with the mission, but because there just happens to be no system near
any pirate stronghold with both sufficient enemy presence and a
sufficient number of planets. But should that change in the future,
this mission will become available.


onpon4:
One small change: I made it so that the pirate patrol mission considers hidden jumps. This makes it actually show up on New Haven.

PR #842, opened 2019-08-03
Switched some dialog in dv_diversion to gender-neutral.

onpon4:
I noticed this just  after I submitted #841. Dialog in this mission has a lot of references to "he" and "him" directed at the player, and... well, the player isn't necessarily a "he" or a "him". :) Switched to singular "they" in those messages to correct this.

### DIFF
--- a/dat/missions/dvaered/dv_diversion.lua
+++ b/dat/missions/dvaered/dv_diversion.lua
@@ -59,9 +59,9 @@ misn_desc = _("You have been recruited to distract the Dvaered fighter escorts a
 misn_reward = _("Some good money, hopefully")
 
 chatter[0] = _("All right men, this will be Hawk's maiden jump. Continue on course to the %s jump gate.")
-chatter[1] = _("How dare he attack me! Get them!")
-chatter[2] = _("You heard Warlord Khan, blow him to pieces!")
-chatter[3] = _("He's attacking us, blow him to pieces!")
+chatter[1] = _("How dare they attack me! Get them!")
+chatter[2] = _("You heard Warlord Khan, blow them to pieces!")
+chatter[3] = _("They're attacking us, blow them to pieces!")
 chatter[4] = _("Arrgh!")
 chatter[5] = _("Khan is dead! Who will be our warlord now?")
 chatter[6] = _("Obviously the one who killed him, idiot!")


BTAxis:
No - the player is assumed to be male in other missions as well. That was mentioned on the wiki writing article too, but, well. Anyway, no need to make the pronouns gender neutral.

onpon4:
Ah... why? I mean, you put your name in at the start of the game and the game uses that name. Everything about Naev screams to me that it's trying to put the player into the game, but referring to the player as male doesn't jive with that.

If the intention is for the player to be controlling a character who is a man, shouldn't the references to the player's name be replaced with some stock name, and some sort of avatar or portrait be introduced? In fact if that's the intention then it seems to me there needs to be some serious changes to the game's dialog, because it *always* refers to the player's character as "you". But if "you" is actually a character the player is controlling, this needs to be changed to first-person or third-person.

But personally, I think trying to immerse the player in the game is better, and that can only be achieved for everyone (rather than only 50% of the population) by making references to the player gender-neutral. I don't see how there's anything wrong with that.

BTAxis:
Well I basically agree, what I think we should really have is a gender selection of some sort at character creation, and then have Lua formatting in place for all the things that matter to that, like pronouns and honorifics. However lacking that I opted for male pronouns because it was the least fuss and I couldn't live with having to write awkward gender-neutral phrases.

onpon4:
But is it not awkward to be called "he" or "him" when you're a woman? How would you feel playing the game if it referred to you as "she" and "her" because it's "the least fuss" and someone "couldn't live with having to write awkward gender-neutral phrases"?

Besides, most gender-neutral phrases aren't awkward. Even if you don't accept  the recently introduced usage of singular "they", the player's ship also has crew, so who's to say they aren't referring to your ship as a team of crew members?

BTAxis:
Yeah okay, that doesn't make sense to me. The writing is player-centric, as per Escape Velocity and friends, so no, you wouldn't be talking about a team of crew members. And I think using "they" to refer to what is clearly meant to be a single individual is worse than assuming the player character is necessarily male. That's not to say it's ideal to do so, but if we're going to fix it I'd rather just do it right.

onpon4:
Singular "they" goes back hundreds of years. The only recent thing is using it for a *known* individual. But these pilots don't know who you are. So even by ultra-conservative traditionalist logic, it's valid.

It's not just "not ideal" to refer to the player as "he" or "him" no matter what. It makes it really awkward and takes away immersion.

onpon4:
Also, a gender selection system wouldn't be a complete solution anyway because non-binary people are a thing.

BTAxis:
It's not true that other characters don't know who the player is. In many situations - during missions at least - they meet the player face to face. Persisting in a gender-neutral way of address would be highly detrimental to the immersion. It telegraphs the author's intent to be inclusive, at the expense of credibility.

As for non-binary people, I envision a selection menu where you can just select the most common presets, plus a "custom" setting where you, as a player, have to enter all the pronouns and such yourself. The Lua API can just fetch them via player.getPronounPossessive() or something like that. It should cover everything, hopefully.

onpon4:
> It's not true that other characters don't know who the player is. In many situations - during missions at least - they meet the player face to face.

I'm talking about this specific pull request. The dialog lines changed are broadcast comms from other ships you're attacking in space.

> Persisting in a gender-neutral way of address would be highly detrimental to the immersion. It telegraphs the author's intent to be inclusive, at the expense of credibility.

As I said, I already made several things gender-neutral in the past, and I've been doing so for years. Has your immersion been thrown off so far?

It's really not a hard thing to do. One of the easiest ways is to use the passive voice. Can you spot where I've done that? Because I can't, and I actually wrote the changes.

> As for non-binary people, I envision a selection menu where you can just select the most common presets, plus a "custom" setting where you, as a player, have to enter all the pronouns and such yourself.

That's not going to jive well with translation support. Pronouns don't work the same way in all languages, and if you implement a pronoun system like this you'll be forced to basically program how languages work. Just keeping it gender-neutral is much, much simpler for that.

onpon4:
On a side note, @bobbens @Deiz and @ids1024 I'd be interested to hear your opinions about this as well.

BTAxis:
> I'm talking about this specific pull request. The dialog lines changed are broadcast comms from other ships you're attacking in space.

Okay, but I feel this is a thing that applies to the whole of Naev's writing. We HAD a policy for that, as I mentioned before, and while it's fine to change that policy it needs to be done across the board. We can't treat the player's gender one way in one place and another way elsewhere.

> As I said, I already made several things gender-neutral in the past, and I've been doing so for years. Has your immersion been thrown off so far?

I don't think I've read any of it, but if I had, yeah, probably. I absolutely don't mean to imply that you're wrong to want to do it or anything, I just really don't like it. Sorry.

> That's not going to jive well with translation support. Pronouns don't work the same way in all languages, and if you implement a pronoun system like this you'll be forced to basically program how languages work. Just keeping it gender-neutral is much, much simpler for that.

Well... I've never believed in translation support for Naev. Yes there is a rudimentary localization API for the Lua strings, but it's a joke. It might as well not exist. There's so much other stuff that doesn't support translation that I feel comfortable in saying Naev will always be in English only, and I think that is fine.

onpon4:
> Okay, but I feel this is a thing that applies to the whole of Naev's writing. We HAD a policy for that, as I mentioned before, and while it's fine to change that policy it needs to be done across the board. We can't treat the player's gender one way in one place and another way elsewhere.

So far as I'm aware this is the only case; as I said, I already fixed the rest years ago; this is just the latest correction I've made, in fact the pull request you merged just before I opened this one had another.

I think the only problem that comes from having both is that the gendering problem isn't solved yet; gender-neutral references and gendered references don't make each other sound weird.

> Well... I've never believed in translation support for Naev. Yes there is a rudimentary localization API for the Lua strings, but it's a joke.

That was ditched years ago in favor of gettext, and there's already some strings translated into Japanese in this way (albeit not very many).

onpon4:
By the way:

> You and Harja finish the post-landing protocol and meet up at the terminal. Harja seems a little apprehensive - he clearly doesn't like the idea of meeting Joanne face to face much. But he doesn't complain. In this he really does appear to be a man of his word. Together, you make your way to a small conference room Joanne booked for the occasion.
>
> Joanne greets you, and Harja somewhat more stiffly. You notice she looks a bit tired. "My apologies," she says when she notices your glance. "I just came off my shift, and my work can be a bit taxing at times. But never mind that, we're not here to talk about my job today." She turns to Harja. "There's something I want to ask you, Harja. Last time we both had dealings with John here, I was told that you swore your innocence, by Sirichana's name." Harja doesn't respond. He doesn't even meet Joanne's gaze. She continues regardless. "If this is true, then I want you to repeat that oath, here and now, at me directly."
>
> There is silence for a few moments, but then Harja makes up his mind. He looks at Joanne and speaks. "Very well. I did not do the things I have been accused of. I did not tamper in any way with the central computer of the High Academy. By the grace of the Touched and the Word of Sirichana, I so swear."

That's the corrected text that has actually already been pulled; it was a part of the previous pull request. Can you see where gendered language used to be?

onpon4:
Also relevant: https://brians.wsu.edu/2016/05/25/they-their-singular/

MageKing17:
Firstly, as a native English speaker, singular-they is not even remotely awkward to hear; if nothing else, the fact that people use it means it should be downright *expected* to be encountered, at some point, in dialogue (so, IMO, the real "policy decision" should be on whether or not singular-they should be encountered in narration... to which my vote would personally be "yes", but I'm aware others would disagree).

Secondly, I'm not sure "different languages have different numbers of pronouns" is actually all that much of a problem for localization; I've never implemented it myself, but if a language could supply a list of which pronouns it uses, which then just need to be filled out (so instead of a hardcoded `player.getPronounPossessive()`, you'd do something like `player.getPronoun("Possessive")`), with the obvious downside that changing languages with custom pronouns would require re-supplying custom pronouns.

onpon4:
I suppose one way you could implement the idea is to have a list of "pronoun templates" that the engine says nothing about and insert them into text with some special coding, e.g. you could have:

$p01 = "she"
$p02 = "her"
$p03 = "her"
$p04 = "hers"
$p05 = "herself"

Suppose capitalizing the "P" makes the it so that the pronoun is capitalized. And then you'd have stuff that looks like this (note, %s is the player's name):

"$P01's %s. I met $p02 back in February when $p01 was buying $p03 first frisbee by $p05."

So first of all, you can see that this is incredibly unwieldy to write; this is necessary to avoid specifically having to anticipate how language works, you end up instead throwing that burden onto writers. Anyway, you'd get the following with the settings I suggested above, plus making the name of the player "Julie":

"She's Julie. I met her back in February when she was buying her first frisbee by herself.

But, what if we set them to these instead?

$p01 = "they"
$p02 = "them"
$p03 = "their"
$p04 = "theirs"
$p05 = "themselves"

Now we get:

"They's Julie. I met them back in February when they was buying their first frisbee by themselves."

So you see, not only is the method incredibly clumsy, it still doesn't work. You need to somehow also account for the possibility of tenses changing, and I don't even have a theoretical solution for that one. Meanwhile, you're making it very difficult to write dialog because you have to do special weird things rather than just standard strings.

Or you can just write all dialog to not assume the player's gender, as I've been doing for years now. Much easier.

bobbens:
OK, I have no problem making it gender-neutral but it has to be done to avoid difficulty and maintenance. The way I see it:

Writing gender-neutral
+ Simple way to "solve" the gender issue
- Can look funny at times
- Might be harder for non-natives to write

Using Lua API
+ Allows full control of pronouns
- Uglier and trickier to maintain
- Have to handle many tricky cases

This is no clear better solution in this case. What I would be interested to know is how other projects are handling this (although likely not handling it). Is there some sort of guideline on the topic?

I am partial to using Lua API (although it probably sounds better in my head than in practice), but realistically writing gender-neutral would probably be the easiest way to solve this.

onpon4:
I can't really think of any dialog-heavy game that has this problem. The closest I can think of is The Ur-Quan Masters (Star Control II); it has you controlling a character that's clearly a man, is shown to be a man at the end, and defaults to a masculine name (although the name can be changed). Even so, most of the game there's no reference whatsoever to the character's gender, and if you ignore the ending cutscene you could easily interpret your character as a woman.

A confounding factor in that case, though, is that all references to the character are second-person (all conversations are one-on-one in that game, so you never see anyone talking *about* your character), meaning the issue never comes up. So I suspect this was entirely accidental.

Even so, that game did face a similar problem when it was ported from the PC to the 3DO, because text was replaced with voices and they couldn't make them say whatever name you wanted. Their solution, quite simply, was to take out the characters saying the player's name and instead make them say "Captain". It worked rather well, and I think this deliberate lack of reference to the character's name is very similar to my approach of deliberately not referencing the player's gender.

ids1024:
Gender neutral language seems desirable when it doesn't look convoluted. I think that's generally possible in English, though sometimes  a bit more difficult. If we're taking translation seriously (which doesn't seem particularly realistic currently) it may simply be impossible in some languages, where adjectives and such are all gendered.

I could probably rant for a while about singular they (remember 'you' also is grammatically plural; the singular is 'thou'). Personally I mostly use it for a non-specific person or where I don't know someone's gender (such as online). But it looks natural enough to me in this context.

I don't really have strong opinions on gender neutral language vs. allowing the player to select pronouns. I guess the latter improves immersion, but it seems like it might get a bit awkward to implement. I suppose it's easier if the dialog already mostly uses language like 'you', but then it also serves less purpose.

BTAxis:
All right, since nobody seems to have an issue with the gender neutral pronouns other than myself, I will go ahead and merge this. However, I still think it should be made consistent across the board (e.g. "Mr. %s" in flf_pre01), and I still favor a more technical solution that allows interchangeable pronouns.

toams:
In a mmorpg i contributed to a few times they solved it by making the dialog functions gender aware. When writing dialogs you have to specify if the sentence mentions use of the gender of the player and if it does you have to write 2 sentences, one for male and one female. The dialog functions then looks at the gender of the player and displays the correct sentence.

Syntax is something like: dialog_gender("Hello Mr. %s", "Hello Mrs. %s")
a regular non gender specific sentence is displayed by a regular dialog() function

Once implemented i think this is a good solution, but likely it needs many changes to the code base before it can be used.

PhoenixRiver:
I should note that all dialog I have written for Naev was gender
neutral and no-one has batted an eyelid - maybe because I also
included varient resutls for cargo missions that included dialog with
both male and female characters... something I'm not sure is still in
because I haven't checked them recently but I felt was important to
include.

I considered adding gender aware text but I also looked at possibly
doing translations into French and Spanish of Naev however neither of
these languages make a gender setting (male / female / whatever) even
remotely possible - it's as bad as trying to do real world time
accurately - it's not unless you put insane work into it because these
languages alone change verb endings depending on gender, never mind
changing adjectives as well - so pronouns are the least of your
worries. Japanese also has issues since some language is gendered, in
that some language has varient grammar, words and phrases depending on
the gender of the person speaking as well as the gender of the person
being spoken to... this is can be covered by using slightly more
formal language which has more commonality (though it will make
females sound less formal and males sound more formal but at least you
don't have to deal with multiple versions of the Japanese
translation... okay you may say two versions is no problem but then
the reality is that there are also at least three varients depending
on the age of the speaker and the spoken to (for adults) - young,
middle aged and old... though again more formal versions are more
gender and age neutral. All told in these three languages alone gender
neutral language is possible and much, much easier than trying to code
something to deal with it; if other ways of changing language based on
gender of speaker or spoken to exist in any other languages then it
gets even more complicated again... I suggest that the gneder neutral
is the easiest and also the most satisfactory solution until a
generalised solution to this problem is achieved by Google or some
other large entity that can afford the time and effort it will
require.


Just my tu'penny worth,

River *<:@)

Issue #832, opened 2019-07-30
Proposal: Player Escort Hiring System


onpon4:
This is something I won't have time to work on for the moment, but think might be a good idea and something I might be interested in working on at some point in the future.

So, back when I first played Naev, it had this problem where there was this very clear hierarchy: small ships < medium ships < big ships. This was because smaller ships were weaker and more vulnerable, but didn't have any particular advantages over larger ships in terms of things like speed. This was fixed several years ago with the introduction of many mechanics, but most importantly the "mass limit" system, which has done a very good job to make light ships more useful by allowing them to be much more agile than their larger counterparts.

That improvement applied to medium (Corvette and Destroyer class) ships too. However, what I have consistently found is that medium ships find themselves sandwiched between the faster and more numerous light ships, and the slower but heavier armored and deadlier heavy ships. Medium ships are too weak to take the punishment of a capital ship, yet too slow to run away from light ships or evade fire of any ship. This has left medium ships in a peculiar situation where it never makes any sense for the player to use one; the second you enter hostile territory, a medium ship is essentially an open target that is easily taken out.

That is, and this is where my proposal comes in, unless you happened to jump into a system that *also* has very high friendly presence. If you're lucky enough to find yourself in that situation, controlling a medium vessel can be a massive boon; Corvettes and Destroyers are fast enough to mostly keep up with a friendly fleet, but powerful enough to inflict crippling damage to opposing fleets. Additionally, medium cargo ships are some of the best ships for carrying very large amounts of cargo in the game, and the help of friendly ships helps to draw the attention of hostile ships away from you.

Escorts have become an important part of how the A.I. ships fight. But as it is right now, the player is, for the most part, unable to create fleets. The only way for a player to have any kind of escort is through fighter bays, which can only be used on heavy ships or at least ships with heavy weapon slots. So here I would like to propose an expansion of the idea of escorts to match what is seen in A.I. ship fleets: the ability to hire your own escorts to form your own fleet.

There are a few possible ways to do this, but the following options are the ones I'm considering in particular:

1. Allow the player to hire random non-hostile ships by hailing them in space; or
2. Add a hub of sorts, probably presented as a menu (similar to the mission computer), where you can hire new pilots.

[...]

PR #844, opened 2019-08-04
Removed player_getRating and all uses of it.

onpon4:
Inspired by something @bobbens said in #832 that I agree with
("I don't think anybody would miss it if removed").  I didn't go
all-out and remove the combat rating mechanic in C code because I don't think
its existence alone is a problem as long as it's just cosmetic (and
it might even be a nice little nod to EV and a way to sort of get
a feel for how much combat you've done this game). Just actual use
by Lua code, which was rare, has been removed, as well as the
Lua function "player_getRating" itself.

Missions that previously modified their difficulty based on this no
longer do so. In particular:

* emp_hcvl now always gives you an "easy" ship.
* trader_escort now always gives you the full range of options for convoy size.

Also worth noting that dat/missions/common.lua has been removed both
because no Lua code was actually using it, and because everything
there seems to be centered around the combat rating mechanic.
hitman3's Lua file has also been removed because it is no longer in
use (due to being obsolete) and doesn't really have anything that
needs to be saved. (Both of these decisions were a response to the
use of player_getRating() in these files.)

Of course, all events and missions that had <cond> attributes with
getRating() have had those getRating() calls removed.


dat/mission.xml
```diff
@@ -503,8 +502,7 @@
    <unique />
   </flags> 
   <avail>
-   <chance>3</chance>
-   <cond>player.getRating() &gt; 50</cond>
+   <chance>3</chance>\
```
Fayti1703: Shouldn't the backslash here cause the XML parsing to fail?

PR #846, opened 2019-08-04
Revert "Add a 'local' keyword"

onpon4:
This reverts commit 99e96d38aae677ec64116ed4b8ce45ba8a013801.

I left a comment on the commit itself about why; the change breaks spawning of any ships with formations (e.g. Empire ships). So far as I can see there's no good reason for the change.

### DIFF
--- a/dat/scripts/formation.lua
+++ b/dat/scripts/formation.lua
@@ -10,7 +10,7 @@ local function count_classes(pilots)
    return class_count
 end
 
-local formations = {}
+formations = {}
 
 function formations.cross(leader)
    local pilots = leader:followers()


ids1024:
Oh, I see the problem; and I'll fix it. My commit was correct though.

I wrote the script so it can be used with `local formation = include "dat/scripts/formation.lua"`, which avoids polluting the global namespace. I believe this is considered a best practice is Lua, though Naev doesn't really follow it otherwise.

I think the previous lack of a `local` keyword was just a mistake when I wrote it.

It looks like the spawn script is broken because it names the local it defines `formation`, then accesses the `formations` variable. Easy enough to fix.

ids1024:
434052cb0a16537d1d4cb695e1c3ebf68eab7a23 should fix the problem.

onpon4:
Awesome, thanks! :)

Issue #859, opened 2019-12-29
Missile Rebalancing


onpon4:
I'm thinking a bit about the need in Naev to rebalance the way missiles work, especially as it pertains to medium ships. I honestly have no idea where to start with this, so I'd like to start a discussion and see if anyone has any thoughts.

Particular issues I have in mind:

1. Missiles are far more useful for A.I. players than for the human player. This was briefly addressed in #832; it's because A.I. players always warp in with maximum ammo capacity whereas players don't necessarily have full ammunition when they warp in. This is a pretty minor issue that could be solved by missile auto-reloading and a mechanism for randomizing how many missiles any given A.I. ship has when it jumps in from an outer system. I think everyone would be in favor of such a change.

2. Missiles are most useful for small, fast craft which have a sufficient turn rate to get the lock-ons missile weapons need and have a particular need, and capability, to stay out of the range of larger ships' firepower.

3. Medium and large ships really don't have any way to counter missiles. Missiles can't be shot down, only jammed, and jamming is really only effective with a really fast ship (or a reasonably fast ship with slow missiles). So in practice, when piloting a medium or large ship, your only possible recourse if pelted with missiles is to hope that the ships attacking you run out of ammo before you get destroyed. This is especially a problem for medium ships.

4. Because of that dynamic, missiles create a strange situation where in practice, medium ships have no credible advantage over light ships. They are too easily pelted into smithereens by pretty much any kind of lock-on missile.

5. At the same time, even given a solution to #1, countering with your own missiles is not particularly effective because small ships have an easier time dodging them and because you're usually outnumbered by a coordinated group attack (#832 addresses the second half of this, but the first half is a consequence of the fundamental way missiles are designed).

Thinking about it, the whole system is so broken that I wouldn't be too upset if we got rid of the missile system entirely. I don't think that would be optimal, however. Missiles (when done right) can be a major enhancement to the experience. I know I've had a lot of fun with missiles in 3-D space shooters for instance.

I'll post my own thoughts in a separate comment since they're disorganized and I don't feel at all sure about them. Anyone else who has any thoughts or ideas, please let them be known! :)

onpon4:
My thoughts:

* Maybe missiles could be reconfigured to only be usable by medium and large ships? But that could create a bad power balance in the opposite direction.
* Maybe missiles could have their range shortened? But that would make them a lot less useful.
* Maybe remove the lock-on requirement?
[...]

PR #861, opened 2020-01-06
Made missiles auto-refill.

onpon4:
An important part of #859, this is meant to prevent the situation of missiles being more useful for AI than for the player.

This was a fair amount of work and a lot of code changes to make this possible, but essentially only three things were done here (effect-wise):

1. Launchers and fighter bays now gradually refill ammo/fighters at a set interval (defined per missile but currently set to 30 seconds for most of them) if they are not being used and (in the case of fighter bays) if there is not already a fighter deployed by the bay.

2. Active Cooldown now refills launchers and fighter bays at the same rate as it cools down your ship. I decided to refill on a threshold basis here, i.e. it refills your missiles up to a point based on how far along you are in active cooldown. For example, if you are out of missiles, you will start gaining ammo pretty quick, but if you have say 50% ammo, your ammo will not increase until active cooldown has reached (and passed) the point of cooling down your ship by 50%.

3. Ammo outfits no longer show up in the store, and there is no need to buy them. The outfitter automatically equips missile launchers with as many appropriate missiles as possible, no purchase necessary.

Reload times might be adjusted later on, but there's a whole bunch of other unrelated work to do with balancing anyway. For now, this is ready for a merge. 👍


onpon4:
Oh! Whoops, a few merges from the past year somehow ended up in this. Sorry about that, I thought for sure I had gotten rid of those..

PR #862, opened 2020-01-06
Made missiles auto-refill (take 2)

onpon4:
An important part of #859, this is meant to prevent the situation of missiles being more useful for AI than for the player.

This was a fair amount of work and a lot of code changes to make this possible, but essentially only three things were done here (effect-wise):

1. Launchers and fighter bays now gradually refill ammo/fighters at a set interval (defined per missile but currently set to 30 seconds for most of them) if they are not being used and (in the case of fighter bays) if there is not already a fighter deployed by the bay.

2. Active Cooldown now refills launchers and fighter bays at the same rate as it cools down your ship. I decided to refill on a threshold basis here, i.e. it refills your missiles up to a point based on how far along you are in active cooldown. For example, if you are out of missiles, you will start gaining ammo pretty quick, but if you have say 50% ammo, your ammo will not increase until active cooldown has reached (and passed) the point of cooling down your ship by 50%.

3. Ammo outfits no longer show up in the store, and there is no need to buy them. The outfitter automatically equips missile launchers with as many appropriate missiles as possible, no purchase necessary.

Reload times might be adjusted later on, but there's a whole bunch of other unrelated work to do with balancing anyway. For now, this is ready for a merge. 👍

(This is a resubmission of #861, the branch in that one accidentally had a bunch of merge commits in it.)


dat/outfits/launchers/enygma_systems_turreted_fury_launcher.xml
```diff
@@ -5,13 +5,14 @@
   <size>small</size>
   <mass>25</mass>
   <price>80000</price>
-  <description>Enygma Systems has developed a turret mounted Fury launcher. It comes at a steep price, but can be a powerful means of discouraging pirate fighters.</description>
+  <description>The Fury class interceptor missile is a lightweight missile propelled by a powerful engine, delivering its payload to its target quickly and efficiently. Its high agility comes at the cost of destructive power, however, limiting its effectiveness to light, nimble fighters. Enygma Systems has developed a turret mounted Fury launcher. It comes at a steep price, but can be a powerful means of discouraging pirate fighters.</description>
```
ids1024: I think this description is more readable if a blank line is included between the description of the missile and the launcher specific description. So it displays like this:

> The Fury class interceptor missile is a lightweight missile propelled by a powerful engine, delivering its payload to its target quickly and efficiently. Its high agility comes at the cost of destructive power, however, limiting its effectiveness to light, nimble fighters.
> 
> Enygma Systems has developed a turret mounted Fury launcher. It comes at a steep price, but can be a powerful means of discouraging pirate fighters.

This change also results in the same missile description appearing in multiple places. That's generally undesirable. Since the missile `.xml` files still exist, perhaps naev could automatically display both the missile's description and the launchers description?

ids1024:
I'm not sure if we want fighter bays to regenerate in exactly the same manner as missiles... though I'm not sure how we would want it to work.

But eh, good enough for now. I think this is better than the previous way fighter bays worked. Just something to keep in mind.

dat/outfits/launchers/enygma_systems_turreted_fury_launcher.xml
onpon4: I did consider programmatically including the text that way but honestly it wouldn't be a perfect solution anyhow. I can add those newlines though.

onpon4:
I actually tested fighter bays most extensively because there were some major issues I had to fix. What I observed from that is that fighter bays are very very powerful, albeit more difficult to direct. I actually like that; if carriers are supposed to get their power from fighter bays then I'd like fighter bays to be very effective.

I chose to make them regenerate at a rate of 1 every 2 minutes for now. That might be adjusted later.

On a side note, I just realized I forgot to make sure the fighter bay's regeneration timer wouldn't start until *after* a deployed ship was destroyed! I'm going to fix that now.

ids1024:
Yeah, I like the way fighters work now better than how they did. Though escorts are generally fairly WIP.

(I guess replacing on landing wouldn't be ideal if fighters are the ship's main defense; then it might too easily be rendered unarmed.)

dat/outfits/launchers/enygma_systems_turreted_fury_launcher.xml
onpon4: OK, I've tweaked those descriptions. Not all in the same way, mind; some I made two separate paragraphs and some I consolidated in a different way.

onpon4:
OK, so I've pushed in a change to make sure that the reload time for bays doesn't start until the escort is destroyed (quick test suggests that it seems to be working well), and I updated all those descriptions to make them nicer.

onpon4:
OK! So, I'm quite confident that the balance is good WRT escorts. I took an Empire Peacemaker with as much armor, shielding, and armor regeneration I could muster, and just a whole lot of Lancelot fighter bays. Taking that, I tried to make it to Kretogg. I made it to Kretogg, but I didn't make it further than that, the swarms of pirates wiped me out and I wasn't able to respawn Lancelots fast enough to replace them. Which I'd say is perfect! 😄👍

onpon4:
Minor, but I just noticed that the medusa missiles were still showing up in the shop! Fixed now.

onpon4:
One more tiny change: adjusted that NPC line about buying missiles to reference commodities instead (since you don't buy missiles anymore).

src/pilot_weapon.c
```diff
@@ -1164,6 +1163,9 @@ static int pilot_shootWeapon( Pilot* p, PilotOutfitSlot* w, double time )
    /* Reset timer. */
    w->timer += rate_mod * outfit_delay( w->outfit );
 
+   /* Shooting prevents reloading. Reset reload timer. */
+   w->rtimer = 0;
```
bobbens: Should shooting prevent reloading? Not necessarily certain, because it will encourage the player to sort of min-max stuff. Might be better to just let it reload while shooting and adjust the outfit parameters to take that into account.

onpon4: The idea I had isn't so much to limit reload rate as to prevent it from happening while you're using it, if that makes sense? Basically the goal was to create a dichotomy between "in use" and "being replenished" for any given launcher.

It can be removed if you want. I like the effect personally, though. Basically I was trying to make it so that you're encouraged to use missiles in concentrated volleys.

onpon4: One thing I think is a must if that's removed, the faster reload rates for Mace Rockets and that other dumbfire rocket would need to be increased to compensate. Maybe I could kick them all to a standard 60 seconds or something.

onpon4: Testing some changes to this effect now.

ids1024: > Basically the goal was to create a dichotomy between "in use" and "being replenished" for any given launcher.

If we keep this behavior, it might be nice if the current state was indicated in the UI in some way. Certainly not critical though.

I like the idea of not reloading while shooting. But I'm not really sure which will would out better in practice.

onpon4: Admittedly it doesn't make that big of a difference if the reload late is very very slow anyway. But at that point I think it more or less just encourages active cooling.

onpon4: OK, so I pushed a change getting rid of the "no reload while shooting" restriction (d01b1d6) and then a reversion to that (52ec9e8).

Honestly, technically any way would allow for any possibility to work right, it's just a question of what you need to do to achieve that. I mostly like this way better:

1. Because it's easier to design and balance that way (rather than having to go through math to find out what the true ammo capacity is after regen the numbers are pretty straightforward); and
2. Because it encourages using missiles in a different way (missile barrages rather than ammo conservation; the latter is already heavily encouraged for regular weapons by the heat system).

onpon4:
OK, so I've decided on removing the restriction for when reloading happens and I've also adjusted the numbers a bit. More of a deliberate method of choosing numbers that takes into account the stats of the launchers.

src/pilot_weapon.c
onpon4: I did some thinking about this and have decided that honestly, it doesn't really matter all that much. Calculations show that the effective max ammo of nearly all rockets won't be increased at all, and of those whose effective max ammo does increase, it's not that much (e.g. with a reload wait of 6 Mace Rockets have I think around 4 extra missiles in a full volley).

onpon4:
What do you think of it now, @bobbens?

onpon4:
Oversight: it wasn't refilling missiles when you landed and took off from a planet (without modifying equipment; or perhaps this was when there just wasn't an equipment screen). Fixed now; your missiles now refill the same place your fuel does.

onpon4:
I've just noticed a problem with escorts caused by this that causes deployed escorts to not reappear in space when you take off. Investigating this now.

onpon4:
OK, that problem alluded to in my previous comment is fixed now. Caused by me making a mistake while coding in the middle of the night (as I'm doing again now to correct it; I apparently never learn. 😆)

I also noticed that the function where the problem was was being used in player.teleport() when it shouldn't have been (which led to duplicate escorts). Fixed that. And also changed a warning to a debug message as a stepping stone for possibly making undocked escorts a thing in the future.

onpon4:
Another small tweak: made it so that the shop displays the launcher's mass as the mass of the launcher itself + the mass of all the missiles it can carry. Makes the stat more useful.

bobbens:
> Another small tweak: made it so that the shop displays the launcher's mass as the mass of the launcher itself + the mass of all the missiles it can carry. Makes the stat more useful.

Do missiles need mass? If they are regenerating we sort of seem to be violating laws of physics. Maybe it would be better to just make launchers be the max weight and have fighters/missiles be weightless.

onpon4:
They do; the engine won't allow them to be massless. :/ Actually removing mass was one of the first things I tried but the missiles then just disappeared immediately when shot.

bobbens:
> They do; the engine won't allow them to be massless. :/ Actually removing mass was one of the first things I tried but the missiles then just disappeared immediately when shot.

OK, that is due to the fact that they are treated as solids and able to accelerate and stuff (e.g., mace rockets) and need a mass. We can either keep it as is with mass changing on the fly, or do one of the two following:

1. Have the max ammo * their mass be added to the outfit mass and remain constaint
2. Ignore ammo and just have the outfit mass be the base one

Not sure what is best.

onpon4:
I opted to keep it as-is both because I found that I was having trouble figuring out how to get rid of ammo mass and because (thinking about it) I figured it had an advantage of making missiles a little more unique anyway. (I'm not concerned about violating the laws of physics really; the eternal batteries already do that.)

If their mass isn't counted toward ship mass then that would be good in my book (and best thing to do in that case, just use the launcher's mass). It's mainly just the difficulty I was having figuring out what to change where, and I think I ran into a part where getting rid of the added mass messed something up? Not sure.

I can look into this again after I finish the commit I'm working on at present (having to do with redefining the missile balance, but what I'm working on right now is adjusting how AI ships are equipped).

bobbens:
OK. I'm at work so I can't test it yet, but I'll try it this night and merge.

onpon4:
I've pushed one more change for this pull request that lowers the mass of Lancelot fighter "ammo". That should be it for this PR; I'm doing missile rebalancing as well but that I'll put in a separate PR.

bobbens:
Looks good, thanks!

PR #863, opened 2020-01-06
Added default keys for escort commands.

onpon4:
It occurred to me that none were set by default. Well, let's fix that. :)
Default set to:

Attack: END
Clear Orders: HOME
Hold position: INS
Return/dock: DEL

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -253,10 +253,10 @@ void input_setDefault ( int wasd )
    /* Escorts */
    input_setKeybind( "e_targetNext", KEYBIND_NULL, SDLK_UNKNOWN, NMOD_NONE );
    input_setKeybind( "e_targetPrev", KEYBIND_NULL, SDLK_UNKNOWN, NMOD_NONE );
-   input_setKeybind( "e_attack", KEYBIND_NULL, SDLK_UNKNOWN, NMOD_NONE );
-   input_setKeybind( "e_hold", KEYBIND_NULL, SDLK_UNKNOWN, NMOD_NONE );
-   input_setKeybind( "e_return", KEYBIND_NULL, SDLK_UNKNOWN, NMOD_NONE );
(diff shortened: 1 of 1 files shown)


onpon4:
Is there any objection to using these defaults? I chose them because they're a set of 4 keys that are currently not used for anything and are close to each other on the keyboard. I also put the most useful commands (Attack and Clear Orders) onto the keys that are most universally accessible and tried to assign them in a way that makes some intuitive sense:

* END (attack and metaphorically "end" my target)
* HOME (clear orders and come "home" to the circle formation)
* INSERT ("insert" yourself into the position you're in)
* DELETE (Come back to the ship to "delete" your presence)

I'm not deadset on these keys, though; another set of 4 keys would be perfectly fine, the important point is that there are defaults.

ids1024:
On many laptops, you have to use `fn` to access things like `home` and `end`. But it's still probably already, since these keybindings wouldn't be needed at a particularly high rate (unlike the fire key, etc.).

This change seems reasonable for now; though ultimately, I think I like @bobbens suggestion in  #864 to have some kind of UI that doesn't require a separate key for each escort command.

onpon4:
Huh, really? I'm surprised to hear that, I've had Ins on Fn before but I'm used to Home/End/PgUp/PgDn being in a column above the arrow keys.

Let me come up with some better ones.

onpon4:
Hrm, I'm not really able to come up with 4 keys that are all next to each other, guaranteed to show up on all keyboards, and are easy to remember. :/ Honestly Home/End/Ins/Del seems like the best combination.

I do agree that a UI method of using these commands would be better. That said, the keys are still useful even if one is made (just like the L, B, and J keys are still useful for example) and for now they're required to get escorts to do anything except passively defend in a formation, so I think having a default for them (or at least Attack and Clear) is still important.

bobbens:
For now let us use these defaults, until we can get a better interface like #864 . Thanks and merged!

PR #870, opened 2020-01-14
Rebalanced missiles, plus reworked the way outfits are chosen for AI pilots.

onpon4:
Originally I was going to just modify missiles to be rebalanced in a certain way, and that's still a big  part of this pull request (contained in commit fd0d467). But then I found that the way Naev defines what outfits go on what AI ships is confusing and difficult to work with, so rather than updating that system to use the modified rockets in the intended way, I decided to scrap that entire system and replace it with a new one.

Alright, so let's start with the missile balancing, since that's the simplest change: missiles have been rebalanced such that unguided missiles are generally small outfits (i.e. the Banshee launcher is now a small outfit), and guided missiles are now all medium or large outfits (usually medium outfits). The purpose of this is to redesignate missiles as an equalizer and tactical advantage for bombers, corvettes, and destroyers. Of course I haven't tested every possible engagement yet, but so far I'm satisfied with the results. I think every size and ship class is now competitive rather than the galaxy being dominated by very light ships and very heavy ships.

Of course, though, I also had to update the way AI ships were equipped (since a lot of outfits' size and intended ships were changed). Unfortunately I found the existing system for equipping AI pilots incredibly confusing (with vague definitions of "low", "medium", and "high" outfits; and confusing jargon like calling utilities "mediums" and calling structurals "lows"), and even when I kind of figured out the system, I found it to be quite limiting and cumbersome. The existing system depends on categorizing all outfits into rather arbitrary and abstracted categories which a bunch of custom code all uses in different ways, but usually by deciding that class X gets this many "low" outfits and this many "medium" outfits or some such. In general, if you wanted to add an outfit to be equipped to AI ships, you had to read and understand several different faction equipping functions, then try to generalize that in a way that works well for all of them. It's a mess.

So rather than working with that system, I decided to replace it. This new system I've designed uses a much, much simpler process that allows the relationship between outfits and ships to be specified far more easily and far more precisely, and it also streamlines and standardizes equipping so that very few cases will require custom code for individual factions.

This system is based on tables. Very long tables (that's the only disadvantage of this method), but nonetheless much easier to read and modify. You have tables for defining the three kinds of outfits (weapons, utilities, structurals) and the three kinds of cores (core systems, thrusters, hulls) for the various classes, base types, and ships. Ship definitions are prioritized over base type definitions, and base type definitions are prioritized over class definitions. In this way, definitions can start generalized (on the class level) and then be made more and more specific as much as needed.

Exactly how the tables work is explained in a comment in dat/factions/equip/generic.lua. Basically:

* Core outfits are defined as lists that are randomly chosen from.
* Other outfits are defined as lists of lists; the inner lists are cycled through and then randomly chosen from, allowing control over what outfits are prioritized and how many to look for. (The picking behavior is configured slightly by the "num" and "varied" elements of the tables.)

The new system works fantastically. All previous custom equippings have been retained and improved, and in fact some new custom equippings have been added, like by preferring forward-mounted weapons more in the Goddard base type (to take advantage of its forward-mounted weapons bonuses). This system has also allowed previously difficult tasks to be completed; for example, Empire Lancelots and Collective Drones produced by fighter bays are now equipped the same way as in the Empire and Collective factions respectively, made possible just by moving a few tables from one file to another.

I haven't given the Soromid specific outfits yet since the ones that previously existed were generic placeholders anyway. Now they just have different placeholders (the generic equipment); not a huge deal in the short-term, and the long-term solution involves (if I'm not mistaken) adding special bio-weapons of some sort, which is outside the scope of this PR and something I'd like to discuss with others before having a go at it.

This PR is ready to merge. Of course, I'd like to take advantage of this to add customizations to how even more ships are outfitted, but that would be in separate PRs.


ids1024:
> I decided to scrap that entire system and replace it with a new one.

I certainly have no great fondness for the previous design, which I've looked at and perhaps not understood, since it seems rather confusing and arcane. Of course, good design can be a challenge.

It seems a bit awkward listing the outfits separately for each combination of faction and ship. I'm not sure what would be best though...

onpon4:
Not so much for each combination. Most is kept generalized on a per-class basis but more specific stuff can be added on if needed. I actually defined the Drone and Heavy Drone stuff in the collective manner on the class level because I could (and deleted the Collective equip script because it was unnecessary), and I did consider just getting rid of all these class definitions entirely since the exception of the Goddard under the Thurion, all of them are faction-specific ships so far (and the Goddard is just a filler in that case).

And even as is, I didn't define utilities or structurals per-ship in most cases (the exceptions being the Za'lek and Thurion ships).

Allowing faction-level definitions is just an option for additional fine-tuning, really. Even with none of it and everyone using generic outfits the universe looks pretty great.

onpon4:
I thought of a way to improve the definitions: I changed them from ship definitions to type definitions.

Also fixed a typo in one place ("Ragnarok" instead of "Ragnarok Beam").

If you guys are unsure about this PR I'd like some feedback on what you would like to see different. I think I've come up with a pretty efficient and flexible system here. Already I've generalized the Collective equipping of drones, for example (which means drones you spawn have Collective weaponry), while at the same time I've added faction-nonspecific yet type-specific equipping for the Vigilance and Goddard ships. It's also a simple system (easy to read and edit), requires no code additions for new factions, yet still allows custom code if desired for one reason or another.

ids1024:
As written, this doesn't make it possible to have different probabilities for different outfits (not that the previous implementation did), right? I don't know if that's an important feature, but it's a difference from a lot of other parts of naev.

ids1024:
At a very high level, a design like this seems more appealing:
1. For each faction, have a list of outfits that are associated with the faction.
2. Use a single algorithm to determine which of the factions outfits would be appropriate for a ship based on slot size and ship class.
3. Provide some mechanism to add custom behavior to the faction script that overrides aspects of the main algorithm.

Now, honestly I don't know how this would work. Figuring out which outfits fit an appropriate slot is easy enough; I'm not sure where the code would get much information beyond that about what is appropriate for a particular ship or ship class. And I just have no idea what part 3 here would consist of; it just seems nice to have more flexibility.

But if we can come up with a system like this, it should be both concise and flexible...

onpon4:
I don't think that's feasible.

Firstly, Naev doesn't have any way to specify what kind of ship we have other than class, base type, ship name, and stats. If you look at the old code, all it did was split ships into arbitrary categories based on arbitrary mass amounts.

Secondly, just because an outfits *fits* on a ship doesn't mean it *belongs* on the ship. For example you can fit a Tricon Zephyr II engine on anything, but it's too heavy for the optimal function of a Hyena or a Collective Drone. These things really need to be configured on the level of classes at the very very least, and realistically, they have to be configured on the level of individual ships. Anything less will lead to crappy AI ships.

So the generalist behavior you were suggesting: it already existed, and that was exactly the problem I set out to solve. It's *too* general.

I've instead developed a system for generalizing by class - that covers 90% of cases - and by type and ship name where absolutely necessary.

Now, probabilities, on the other hand, might be feasible. Currently the way to do it is to have redundant entries. But maybe I can add an extra inner table specifying what outfits to create redundancies for? Like I could specify that the Unicorp PT-100 Core System has a "probability" of 3, meaning it's included in the list 3 times. How's that sound?

onpon4:
Alrighty, I've implemented support for specifying probability, and it's a good thing I did because the num and varied fields were busted! Time to do some more testing.

ids1024:
Yeah, designing a good general system is difficult. I think the table method is an improvement over the existing code, with saner behavior, so it should be fine for now; I'm just not sure it's the ideal architecture.

> For example you can fit a Tricon Zephyr II engine on anything, but it's too heavy for the optimal function of a Hyena or a Collective Drone. 

I guess the logic would be something along the lines of "use Zephyr for fighters below X weight, and Zephyr II for fighters above X weight"? Though that still requires coding for the specific engine... though perhaps taking into account the engine mass limit would help with that... I suppose ideally it would use a somewhat similar logic to what the player would use; considering how the mass of the ship and all outfits balances against the engine's mass limit.

...

Looking specifically at this implementation, does it do anything to deal with the CPU limit? It doesn't seem to be explicitly addressed. Is it expected that the way the tables are written, it won't try to equip a ship with more than it has the capacity for?

onpon4:
As regards CPU I used pilot.addOutfit with bypass set to false (the default value), so I'm letting addOutfit prevent that from happening (and if an outfit is not equipped for any reason, it's removed from the list and either another is tried or the next set is moved on to depending on num and varied).

It's true that in theory it should be possible to find and equip the ideal thruster by cycling through all of them and finding out which one has the closes mass limit to actual ship weight. That of course would require fitting the thruster at the end. Naev doesn't offer a sane way to cycle through all thrusters, though, at least not on the Lua side.

Anyway, the thrusters thing was just an easy example. It's really not possible to abstract weapons and outfits in such a way; there's more to them than just pure stats, there's also lore to consider for example (such as that House Sirius is a big fan of disabling, House Dvaered is a big fan of bombarding with heavy projectiles, Za'lek is a fan of beams, Soromid has dedicated bio-weapons...).

I suppose a "choose best thruster" argument could be tacked on to make it choose the ideal thruster at the end among those available in the list. I don't feel like this makes sense, though; the purpose of the way we outfit AI ships isn't to create the most ideal combinations, but rather to create a sense of realism. So it would make sense that some ships would have less than ideal combinations due to things like cost, availability of parts in the shops they go to, and how highly a pilot values various stats.

ids1024:
Okay, after some though, I present **The Over-engineered Solution™**. Which is at least interesting to think about.

Fundamentally, determining the best way to equip a ship (for a particular purpose) is an optimization problem. So theoretically, the solution should be optimization. Since there are a relatively small finite number of outfits, the brute-force solution is the obvious approach.

* Have a list of outfits a faction will use
* For a ship, calculate all possible space-worthy permutations of available outfits
* Calculate the resulting ship stats
* Calculate, additionally, some higher level stats, like "firepower" and "asteroid mining ability"
* Per ship class, provide an optimization function to weight these stats; additionally, alternate functions can be faction-specific, ship-specific, specific to a particular task (asteroid miner), etc.
* Choose the highest score from these
* Cache the result, which hopefully addresses the poor time complexity

I'm not sure what the best way to add non-determinism would be; but it's not inherently difficult to provide multiple ways for the same ship to be equipped.

This would be complicated to implement, but done correctly I think it could give quite good results with a lot of flexibility.

onpon4:
All possible space-worthy configurations would be a rather long list.

Suppose a simple example. 4 weapon slots, all small-sized, with 5 possible weapons. That would be 5^4 (625) weapon combinations alone. Ad 5 utility slots with 3 possibilities: 3^5, or 243, and combined with our weapons, that's over 150,000 possible combinations. Now add on 3 structural slots with 3 possibilities: 27 combinations, bringing our total to well over 3 million.

And that's a *small* example. Those numbers are going to get real huge when you consider a ship that has to consider something like 10 slots that could fit 15 weapons realistically (that's over 500 billion combinations just for weapons). So either you're going to spend a whole lot of computation time, or you're going to devour massive amounts of RAM caching trillions of combinations of outfits for the various ship types and factions.

I don't think that's worth it. Not to mention the difficulty of trying to abstract everything about what makes a ship good or bad into a number.

ids1024:
> weapon combinations

That's the calculation for permutations. Come to think of it, since the order doesn't matter, only combinations are needed, which improves the situation. Probably still high in the extreme cases.

It should be possible to reduce the number of things that need to be considered. I might try to see if I can implement some variant of this; it seems like an interesting thing to test.

> Not to mention the difficulty of trying to abstract everything about what makes a ship good or bad into a number.

Perhaps. I find it hard to guess how hard it will be to write satisfactory weighting functions without being about to try.

onpon4:
True. Didn't think of that and I actually don't remember how to calculate the number of non-ordered possibilities.

In any case I have zero interest in something like that. It strikes me as that old-fashioned idea of trying to automate game design. It never seems to work well.

And I'd like to point out that my system that I have developed *can* be overridden by custom code still. I've just designed the system so that you don't have to. It's just as easy with this system as with the previous system to use a simple script to change the way outfits are varied.

PhoenixRiver:
I just thought I better add something here having worked on a number
of insanely huge combinatorial problems. If you want to go with
finding the perfect ships (see later for a discussion about the
problems with that idea) I would suggest that a script be written that
checks if the brute forcing has been done and if not ask the user for
how much to do; light/none, medium or full and then do it before
loading the users game and write a file containing the solutions so
that it only ever has to be done once for each version of the game (or
maybe only each time the ships get changed? I don't know how you'd
check that realiably). For most players this will mean relatively
little disruption, for testers things could get harder though unless
there is a way to turn it off - possibly the solution might be to even
have both systems and switch to the current table system if the user
selects "none". The light, medium, heavy options would involve having
three categories of ships, those essential to the game which are
always optimised (light), those which would be good to optimise
(medium) and then full which optimises everything.

I might note though that none of the numbers we have for specifying
ships optimises anything even faintly accurately; if a player is good
at aiming then beams make sense, turretedon slow ships or not on small
ships... again, is the player good at manouvering is what decides the
ship size choice... sonow you have to do that for the AI; is it good
at aiming, yes, potentially perfect, so all ships should use beams,
and is it good at manouvering, it could be also be perfect, so all
ships should be small, fast ships... the problem here is that
ultimately you are making a single perfect ship that is all the AI
uses because we only have one AI... now maybe you split that into
factions with different predelictions and other factors come into play
but that still leaves only one ship per faction... yes, I am taking
things to the extreme, but if you don't then a combinatorial system is
nothing better than the table system we already have except that it
produces fewer solutions and requires perfection in the numbers used
and how the AI uses them and becomes unrealistic... and also
frustrates the player because Artificial Stupidity is much more fun to
play against that Perfect AI.

I would suggest that if such a script is made it is used by us coders
to do sanity other other checks on changes or overhauls to factions,
modules, cores, ships etc. in the game; I suggest also that the table
system allow essentially flawed setups to be allowed (at least highly
underpowered ones) because real people do that sometimes for good
reason e.g. cargo ships are essentially completely flawed according to
almost all the numbers because they are not meant to fight and most of
the numbers are about fighting - another reason the combinatorial
method may have real issue with variety, it will like prefer combat
ships...)

When it comes to it the user will enjoy a more flawed set of ships to
come up against, one with variety and unexpected setups that test
their mettle, not their patience.


Just my two cents (though maybe I'm still thinking old money ;)
Phoenix

ids1024:
> e.g. cargo ships are essentially completely flawed according to
almost all the numbers because they are not meant to fight and most of
the numbers are about fighting

In principle, the equipping a cargo ship would be based on a different utility function that optimizes for the aims of a cargo ship. Additionally, it might be possible to calculate "best under X credits" or something like that, to account for the fact that the best possible outfits may not be available.

Probably a pure optimization approach isn't the best solution for one reason or another; it could be interesting to investigate though. And see if any hybrid approaches could be fruitful.

And indeed, variety is definitely a goal; and not one that pure optimization is well suited for.

I'm fine with tables, at least for now. It just seems like alternatives are worth thinking about.

onpon4:
If we're all fine with tables any chance I could get a merge? :3

@bobbens

bobbens:
Thanks for the commits. I tried it a bit and it seems to work fine. I wrote my thoughts on the conversation and a possible implementation in #872

PR #871, opened 2020-01-15
Bugfixes

onpon4:
This fixes some wrong text in the second Shark mission, and also _possibly_ fixes an occasional segfault (I know where the segfault occurred, but I can't be 100% sure that what I did fixed it because the crash is/was pretty rare).

### DIFF
--- a/src/pilot_outfit.c
+++ b/src/pilot_outfit.c
@@ -828,10 +828,6 @@ void pilot_fillAmmo( Pilot* pilot )
    Outfit *o, *ammo;
 
    for (i=0; i<pilot->noutfits; i++) {
-      /* Must have outfit */
-      if (pilot->outfits[i] == NULL)
-         continue;
-
       o = pilot->outfits[i]->outfit;
 
       /* Must be valid outfit. */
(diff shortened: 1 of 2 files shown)


src/pilot_outfit.c
```diff
@@ -828,6 +828,10 @@ void pilot_fillAmmo( Pilot* pilot )
    Outfit *o, *ammo;
 
    for (i=0; i<pilot->noutfits; i++) {
+      /* Must have outfit */
+      if (pilot->outfits[i] == NULL)
```
ids1024: Is there any case where this is expected to be `NULL`? Or is any instance in which this condition is satisfied evidence of a bug elsewhere?

onpon4: I haven't a clue. All I know is:

1. The game segfaulted at the next statement.
2. I don't think pilot was NULL.

This was just a shot in the dark. I can't check whether or not it worked because the segfault was very rare. :/ If I get the segfault again I'll remove the check and see if I can figure out the check or fix actually needed. (I figure the worst case scenario, an unnecessary check, isn't too bad.)

ids1024: Additionally, I think this will just move the segfault elsewhere. Grepping for `->noutfits`, you can find other similar loops without this test.

onpon4: I can revert it, that's fine.

onpon4:
Just wanted to ping this. I reverted the C change, so it's just a fixup of the second "Shark" campaign mission.

bobbens:
Cherry-picked the sh01 and sh02 commits. Thanks!

PR #877, opened 2020-01-19
Increased trader escort mission reward.

onpon4:
The trader escort mission is probably the most tedious regular mission
in the game, so I think it's best for it to be a very rewarding one.
Not so rewarding that it's a must-do, but rewarding enough that there's
a point to doing it instead of a bounty hunt.

I've also decreased the difference between small and large convoys.
It's still there, but not as drastic. Previously the largest was 20x
the distance reward of the smallest; now it's only 10x.

### DIFF
--- a/dat/missions/neutral/trader_escort.lua
+++ b/dat/missions/neutral/trader_escort.lua
@@ -76,45 +76,46 @@ function create()
       misn.finish(false)
    elseif numjumps == 0 then
       misn.finish(false) -- have to escort them at least one jump!
-   elseif avgrisk <= 25 then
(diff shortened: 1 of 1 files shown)


onpon4:
Some further refinement. I ended up more or less completely rewriting the formulas the mission uses so that it weights piracy risk a lot higher and (importantly) using the *total* piracy risk rather than the average (the previous formula had the absurdity that you would be rewarded for a shorter "averagely dangerous" route than a longer route which contains that route plus safe systems).

The modification there also has the nice side-effect of making the mission more readily available.

bobbens:
Looks good, thanks.

PR #878, opened 2020-01-19
Added Unicorp PT-900 to techs

onpon4:
It was missing for some reason! Literally the only place you could buy
this core system was in Bon Sebb, meaning you couldn't sell it anywhere
either. Fixed! (It's now found everywhere the Unicorp PT-1000 Core System is found.)


onpon4:
Pushed another bugfix here: some leftover gendered language in reference to the player in the FLF prelude mission.

bobbens:
Looks good, thanks!

PR #879, opened 2020-01-20
Modified Soromid commodities completely.

onpon4:
Previous definitions followed no clear patterns and didn't seem to
follow the lores written for the respective planets. I've addressed
this by going through every Soromid asset and redefining commodities
to make more sense and make Soromid space more diverse as far as its
economy goes.

### DIFF
--- a/dat/assets/arrentil.xml
+++ b/dat/assets/arrentil.xml
@@ -25,8 +25,7 @@
    <commodity/>
   </services>
   <commodities>
-   <commodity>Food</commodity>
-   <commodity>Ore</commodity>
+   <commodity>Industrial Goods</commodity>
   </commodities>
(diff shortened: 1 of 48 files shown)


onpon4:
So, just to be more specific about what I did here:

Previously most Soromid worlds had either no commodities or food, with a hodge-podge of worlds that had industrial goods. So shipping missions were hard to get, and the thing that struck out to me the most was that medicine wasn't readily available in the economy, which makes no sense to me for a faction well-known for highly advanced medical treatments.

When I went through all Soromid assets, I looked at the descriptions and the surrounding planets and stations:

* a small number of planets were given Luxury Goods because they are either vacation spots or have massive populations with thriving economies.
* Many planets described as mineral-rich have ore and industrial goods.
* Food is found far less than before, only in habitable planets and in stations and planets mentioned as food sources.
* Medicine is found scattered throughout Soromid space, particularly on planets with a lot of life on them, as well as all three "Soromid Wards" locations and anywhere genetic treatment advancement or something is mentioned.

Basically I made it so that:

1. Shipment missions are now much more available and much more rewarding;
2. Medicine is now much more readily available; and
3. Asset descriptions now match the actual commodities they have (whereas before there was often a mismatch).

This is a part of my efforts to develop the Soromid further.

bobbens:
Looks good, thanks!

PR #881, opened 2020-01-20
New mission: "Coming of Age" (Coming Out part 2)

onpon4:
This is the second in the "Coming Out" campaign. Another simple one; you take Chelsea to a Soromid planet with a shipyard so she can buy a rusty old Llama and start becoming a pilot. I also fixed up the first "Coming Out" mission so that it stores when you finished that one (and the same is done for Coming of Age).

This concludes the setup of the campaign. Next up will be the more exciting missions (involving working together on missions of various kinds).


bobbens:
Looks good, thanks!

PR #882, opened 2020-01-20
Fixed the problem with escorts (I think).

onpon4:
Not tested thoroughly yet, but I think this should fix the problem of
escorts causing segfaults when their origin ship dies. A cursory test
shows that this seems to work.

This also causes pilot.setLeader to remove "deployed" status from the
origin ship.

### DIFF
--- a/src/pilot.c
+++ b/src/pilot.c
@@ -2110,7 +2110,7 @@ void pilot_delete( Pilot* p )
    if ( p->dockslot != NULL )
    {
       p->dockslot->u.ammo.deployed--;
-      p->dockpilot = NULL;
+      p->dockpilot = 0;
       p->dockslot = NULL;
    }
 
@@ -2430,7 +2430,7 @@ credits_t pilot_modCredits( Pilot *p, credits_t amount )
  */
 void pilot_init( Pilot* pilot, Ship* ship, const char* name, int faction, const char *ai,
       const double dir, const Vector2d* pos, const Vector2d* vel,
-      const PilotFlags flags, Pilot* dockpilot, PilotOutfitSlot* dockslot )
+      const PilotFlags flags, unsigned int dockpilot, PilotOutfitSlot* dockslot )
 {
    int i, p;
 
@@ -2448,7 +2448,7 @@ void pilot_init( Pilot* pilot, Ship* ship, const char* name, int faction, const 
    pilot->dockslot = dockslot;
 
    /* Make sure we have both or neither */
-   assert( (pilot->dockpilot == NULL) == (pilot->dockslot == NULL) );
+   assert( (pilot->dockpilot == 0) == (pilot->dockslot == NULL) );
 
    /* Basic information. */
    pilot->ship = ship;
@@ -2591,7 +2591,7 @@ void pilot_init( Pilot* pilot, Ship* ship, const char* name, int faction, const 
  */
 unsigned int pilot_create( Ship* ship, const char* name, int faction, const char *ai,
       const double dir, const Vector2d* pos, const Vector2d* vel,
-      const PilotFlags flags, Pilot* dockpilot, PilotOutfitSlot* dockslot )
+      const PilotFlags flags, unsigned int dockpilot, PilotOutfitSlot* dockslot )
 {
    Pilot *dyn;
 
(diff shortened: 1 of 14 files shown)


src/pilot.c
```diff
@@ -2903,6 +2909,14 @@ void pilot_destroy(Pilot* p)
 
    /* copy other pilots down */
    memmove(&pilot_stack[i], &pilot_stack[i+1], (pilot_nstack-i)*sizeof(Pilot*));
+
+   /* Clear docks for launched fighters if necessary */
+   for (i=0; i < pilot_nstack; i++) {
```
ids1024: I think this will work...

But it means every `pilot_destroy` call iterates over every pilot in the system, which might not be the most efficient solution.

If instead `dockpilot` is a pilot ID, and `dockslot` an index in the slots, I think that would be both more efficient and easier not to break (since there aren't pointers lying around to potentially become dangling).

ids1024: I think using a pilot ID and slot index should work nicely with a helper function like this (untested):

```c
PilotOutfitSlot* pilot_getDockSlot( Pilot* p ) {
   Pilot* dockpilot;
   if (p->dockpilot != 0) {
      docpilot = pilot_get(p->dockpilot);
      if (docpilot == NULL)
         p->docpilot = 0;
      else
         return docpilot->outfits[p->dockslot];
   }
   return NULL;
}
```

onpon4: That's real fair. It did occur to me that would be cleaner as an implementation, too.

ids1024:
Calling `assert` broke the Travis build. If you want to do that, include `assert.h`.

src/player.h
```diff
@@ -47,6 +47,9 @@ typedef char PlayerFlags[ PLAYER_FLAGS_MAX ];
 /* comfort flags. */
 #define player_isTut()     player_isFlag(PLAYER_TUTORIAL)
 
+/* dt_default grabber helper */
+#define player_dt_default() (player.p != NULL && player.p->ship != NULL ? player.p->ship->dt_default : 1)
```
ids1024: I don't know if it really matters in this case, but I'd probably be inclined to use a function instead of a macro when possible. A modern C compiler can inline the function to have the same effect.

src/pilot.c
onpon4: OK so I'm finding implementing it that way to be a bit cumbersome. Giving up on that for now, I don't have the energy for it.

(Also worth noting, that change isn't going to reduce looping at all, the looping is to find the ships that mark the pilot being destroyed as a dock and changing the way said pilot is marked won't prevent having to loop to do that. The only way to avoid that is if we use a list stored for each pilot that shows all the pilots that use them as a dock, which I guess the escort list *might* work for, but I haven't looked into that yet.)

src/player.h
onpon4: OK, done. It's a function now.

onpon4:
I've added an include for assert.h  and changed player_dt_default to a function.

I'm not sure about switching up the way dockpilot and dockslot are stored, it's not going to improve efficiency any (and in fact can make it worse because it adds more looping). That it involves passing pointers around is true enough but failure to clean up the other values properly can cause bugs too (just bugs that would be less obvious like choosing the wrong ship as a dockpilot and then trying to dock on a slot that it doesn't belong in).

src/pilot.c
ids1024: > that change isn't going to reduce looping at all

With the way I suggested, I *think* this loop could be removed. When the escort's code tries to call `pilot_getDockSlot()`, that will simply return `NULL` since the pilot that launched it wouldn't exist any more; without `docpilot`/`docslot` having to be updated already.

onpon4:
Now, if every single ship that could possibly have the destroyed pilot as a dockslot will always be in the escort list, that can be used to reduce looping a bit. That looks rather flimsy to me, though, and I don't think that's really worth it when the pilots have to be looped through all the time anyway.

src/pilot.c
onpon4: Alright, as per usual after whining for a bit I did the conversion. It's using ints for both dockpilot and dockslot now.

onpon4:
There we go, switched it up to use ints instead of pointers.

ids1024:
Do you see my point about just removing the loop from `pilot_destroy`, or am I missing something? All that code does is setting `dockpilot` and `dockslot` to `0` and `-1`. Without that loop, those variables will be set the same way the next time `pilot_getDockSlot()` is called.

Otherwise, the changes here look good. :+1:

onpon4:
Ahhhhhhh, I see what you're saying. That does make sense come to think of it. Lemme do that...

onpon4:
There we go, I've pushed that through. Seems to work.

PR #887, opened 2020-01-21
Goddard/Krain campaign

ids1024:
I forgot about this campaign that @l0k1 and I were working on way back in 2015. I'll post this here as a draft PR for future reference (for myself and anyone else)...


ids1024:
If I recall correctly, we considered this campaign mostly done, except for having graphics for two new ships. In the current code, it says "Goddard MkII" and "Kestrel MkII", though I later thought it might be better not to just introduce more powerful versions of those ships, but instead to create corvette ships from Goddard and Krain in a similar style to their other designs.

Perhaps adding two campaign-specific ships wasn't the best idea...

l0k1:
Well, good news is I’m now pretty proficient in blender (see https://github.com/l0k1/MiG-21bis). Do we still have the blender to spritesheet export script? I can make the ships if we do.

ids1024:
The scripts for that are in the [naev-artwork](https://github.com/naev/naev-artwork/tree/master/3D) repo.

There's also the challenge of making sure they're nice ships with decent specs, but don't especially overpower everything else. I think currently the faction ships all have higher specs than civilian ships, so making them corvettes that are similar to the best corvettes in the game might be satisfactory...

Then to buy the ship, the player should have a medium combat vessel license. Perhaps the mission should require that; or give the player a free ship without requiring a license (just ignore the regulations a bit).

bobbens:
> Well, good news is I’m now pretty proficient in blender (see https://github.com/l0k1/MiG-21bis). Do we still have the blender to spritesheet export script? I can make the ships if we do.

I would like to point out that I got it all working again, so I should be able to render new sprite sheets. However, the main limitation is that it is done with blender 2.62, so if you use features from a newer blender (not have them compatible with 2.62), it won't work. I do want to reimplement them in the blender LTS, but that will be fairly tricky to get looking good I guess.

bobbens:
Would there be any interest in getting these campaign working with latest Naev? Instead of giving new ships, giving unique cores or something along those lines might be less balance-breaking.

dat/assets/krain_station.xml
```diff
@@ -0,0 +1,33 @@
...
+  <hide>0.250000</hide>
+  <services>
+   <land/>
+   <refuel/>
+   <bar/>
+   <shipyard/>
+  </services>
+  <commodities/>
+  <description>This station is the headquarters for Krain Industries, manufacturer of the Kestrel. Its location in Salvador helps supply raw materials and avoids the Empire bureaucracy.</description>
+  <bar>Krain Station has limited facilities and is focused on the manufacture of ships.  The bar has a utilitarian but not unpleasant atmosphere.</bar>
```
TheGiraffe3, line 28:
```suggestion
  <bar>Krain Station has limited facilities and is focused on the manufacture of ships. The bar has a utilitarian but not unpleasant atmosphere.</bar>
```

TheGiraffe3:
@ids1024 hey there, any updates on this?

bobbens:
Closing as out of date.

PR #890, opened 2020-01-24
Add a small world of Proteron survivors in Iris

ids1024:
This consists of the survivors of the Proteron who were left on this side of the nebula.

According to the lore, the Proteron were a Great House before the incident. Since they were not a secret, unlike the Thurion, there probably should be some mention of them. It seems logical that some would have been stranded on this side of the nebula.

This provides some interest in the Proteron for the player, and could be a useful site for future missions (a Proteron invasion might begin by secretly establishing contact with and control of Osiris).

https://web.archive.org/web/20170813171731/http://wiki.naev.org/wiki/Proteron

### DIFF
--- a/dat/assets/osiris.xml
+++ b/dat/assets/osiris.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<asset name="Osiris">
+ <pos>
+  <x>8165.439424</x>
+  <y>-3379.428504</y>
+ </pos>
+ <GFX>
+  <space>moon-X00.png</space>
+  <exterior>uninhabited-rocky-magenta-01.png</exterior>
+ </GFX>
+ <presence>
+  <faction>Proteron</faction>
+  <value>5.000000</value>
+  <range>0</range>
+ </presence>
+ <general>
+  <class>A</class>
+  <population>100</population>
+  <hide>0.250000</hide>
+  <services>
+   <land/>
+   <refuel/>
+   <bar/>
+   <commodity/>
+  </services>
+  <commodities/>
+  <description>House Proteron was once a formidable Great House of the Empire, before all of its teritories were destroyed in the incident. The Proteron ships that were outside what is now the Sol Nebula gathered at the uninhabited world of Osiris.
+
+In principle, the system is under Empire sovereignty, and House Proteron was dissolved after the destruction of all of its worlds in the incident. In practice, the former Proteron officers and soldiers residing here do not care for imperial control, nor does anyone else have an interest in the world.</description>
+  <bar>The bar has a very utilitarian design, consisting of the bare minimum to keep out the toxic atmosphere. They offer a small selection of drinks imported from neighboring Empire and Soromid space.</bar>
+ </general>
+</asset>
--- a/dat/ssys/iris.xml
+++ b/dat/ssys/iris.xml
@@ -12,6 +12,7 @@
  </pos>
  <assets>
   <asset>Aluben</asset>
+  <asset>Osiris</asset>
   <asset>Tenal-P</asset>
   <asset>Virtual Trader Local</asset>
  </assets>


onpon4:
Not a bad idea, though I have mixed thoughts about the geography as regards a Proteron invasion. It seems to me that the invasion should involve a hidden jump or series of hidden jumps getting them around the nebula, but that particular location for Osiris kind of makes the only sensible invasion path going through the nebula somehow.

Why not a more outer system instead, like Toros or Tormulex? Or maybe a new system beyond Toros?

onpon4:
Just a bit of an addendum: obviously that old wiki article mentions that the Proteron are developing tech to traverse the nebula, but (AFAIK) that was written at a time before the Thurion were conceived and when Protera was a lot closer by (behind a much smaller inner nebula). It seems to me that the Proteron and Thurion should be causing havoc from two separate directions. If the Proteron have to go through the nebula, though, that kind of requires them to conquer the Thurion first, which would be in the background for most players and not especially exciting.

ids1024:
The Thurion are another issue; I think we need some mechanism to hide them better. It seems we expect the player to fly through the nebula without necessarily knowing what they are (so, clearly not being able to simply land on their planets and visit the bar and shipyard).

Perhaps flying through the nebula, the player should see strange things going on, but not know exactly what? I like the idea that there's a clear sense that something's "out there", but you can't lock onto a ship or even necessarily be sure that it is a ship. I don't know how that would work. (Somewhat like the Shadows in Babylon 5.)

I guess we'd need to introduce a Thurion cloaking device and figure out a game mechanic for how such a device will work.

Then we could also have events and missions involving strange sightings in normal space, that no one knows is the Thurion.

ids1024:
Adding hidden jumps around the nebula is definitely an alternative, and Tormulex would be a sensible place for that.  But making the Thurion semi-hidden even in there own systems seems more like the lore I've read. And it seems much more interesting if we could find a satisfactory way to implement it.

onpon4:
Eh, I rather like the idea of having two separate fronts. Like imagine the Thurion coming in with an invasion from the Nebula while the Proteron are invading through Frontier space, and the two forces ultimately colliding in Dvaered space. Plus I like the idea of allowing the player to join the Thurion.

ids1024:
> Plus I like the idea of allowing the player to join the Thurion.

The descriptions I recall, I think on the forum, suggested some kind of mission/event that would bring the player into contact with the Thurion. So this could happen even if they are initially hidden in some way.

> Like imagine the Thurion coming in with an invasion from the Nebula while the Proteron are invading through Frontier space, and the two forces ultimately colliding in Dvaered space.

This could be interesting. Though, from Tormulex, they wouldn't have to go through the frontier (perhaps they want to conquer it first).

A Thurion cloaking device could be interesting in any case; I don't know how we'd want such a thing to work.

BTAxis:
The Thurion were definitely meant to be hidden in the sense that a player who goes in without prior knowledge shouldn't just bump into them like with any other faction. If at some point there was to be an in-game knowledge base containing game lore, they would be omitted from it (at least until discovered through gameplay).

So the way I imagine it is that there might be Thurion ships in the nebula, but they should be few and far between (artificially reduced presence values, maybe?), and if targeted they should remain unidentified. Possibly even disappear before the player can get close enough to scan them, for that "ghost ship" effect. I do think it should take a mission or event to fully open up interaction with them.

BTAxis:
As far as the Proteron invasion goes, last I heard the idea was, more or less, to tie them in with the Incident lore and the hypergate network, the latter of which remains absent from the game. Has that thinking been superceded?

I do like the idea of the Proteron coming completely out of left field, which would preclude a planet this side of the nebula. But it all depends on the direction we want to take this in.

onpon4:
It seems to me the best way to make assets impossible to discover is to simply create them with diffs, like what's done with Sindbad. That could certainly be an interesting and relatively simple thing to implent. The ships could just technically not be there until then, too (since the assets create the presence). Then later on a "Thurion ghosts" effect could be added if someone gets around to it, though it's less clear how that would work.

Regarding Tormulex and Frontier space, what I was thinking is they could approach from Teller. That would land them in the Frontier and the edge of Sirius space, plus Tormulex with the hidden jump between Tormulex and Anger.

onpon4:
By the way, the end of the FLF campaign is definitely 100% intended to get the player in contact with the Thurion. 😛 Right now it just gives you knowledge of the hidden jump, though it's occurred to me that it would be better for some sort of secret contact to be there, reveal the Thurion to you, and introduce you to them. Possibly in an unfriendly manner.

BTAxis:
For the Proteron invasion I still prefer the hypergate angle myself, because the idea is to have the Proteron show up all over the place, not just from one direction. I think if they just came through the jump lanes around the nebula, they would either be too localized for their impact to matter much, or they would render big chunks of existing factions inaccessible. That could really mess up quests and the like. The hypergate approach lets us make the invasion relevant to everyone without necessary having to deny the factions their existing assets.

Also, the idea of the Proteron finding a way around the nebula using conventional travel kind of cheapens the whole idea of the nebula as this forbidding, unsurmountable barrier. A long roundabout route prohibited by fuel constraints, or a route made up of hidden jumps CAN work, but it seems a little lame to be honest.

onpon4:
Ahhhhhh, I see, so a hypergate is more or less like instant teleportation wherever they want? That does sound real neat.

BTAxis:
Yes, the hypergate network is planned to be a secondary form of transportation through the Naev galaxy that allows much faster travel across long distances. This same concept existed in EVN, too. For Naev, we wanted to also tie it in with the Proteron and the Incident lore. Deiz and KAHR-Alpha wrote a lot of words on the subject (mostly the Incident part; the Proteron angle I tacked on myself later). Sadly the forum post for this seems to not be accessible on the wayback machine. I should probably write this up again.

ids1024:
Ah. I don't know exactly how the hypergates are expected to work, but that seems potentially more interesting than either going through the nebula or some convoluted way around it.

> A long roundabout route prohibited by fuel constraints, or a route made up of hidden jumps CAN work, but it seems a little lame to be honest.

Another variation on this would be to have the Proteron develop a technology to construct new jump routes, which is probably more interesting than either of these. But probably still strictly less interesting than hypergates. And we probably don't want both.

This still leaves two options:
* provide no significant hint in the game of the Proteron having ever existed  (despite the fact they weren't a secret pre-incident). Then a completely unexpected invasion can show up in the middle of a each major campaign.
* We can include indications of the Proteron, like this survivor would. But we can locate it anywhere in the galaxy.

BTAxis:
It's not necessary to eliminate ALL references to the Proteron from the pre-invasion play area. They Proteron are a normal part of the Empire's history, so records of their existence would still exist. They can safely be referenced in descriptions, missions, bar banter, etc.

While a complete planet of Proteron would not exist on this side of the Nebula, it's certainly possible for there to be some kind of smaller community of Proteron citizens living somewhere else. After all, not all Proteron people would have been in Proteron space at the time of the Incident. They would be left stranded, cut off from their own region, with no way to know if the Proteron worlds even still exist. Maybe these people could have found each other on some station or somewhere. The station would nominally belong to some other faction, but it could be populated mostly by Proteron.

onpon4:
Incidentally if that's the way such a station exists then it can be basically anywhere, even smack-dab in the middle of Empire space.

Actually, one thing that occurs to me is that I specified faction texts for the Proteron a while back and those texts wouldn't fit with a group of people who were formerly Proteron before the Incident. That of course could be avoided by setting such a station's faction to "Independent", of course (leaving the Proteron reference just in lore text).

ids1024:
> The station would nominally belong to some other faction, but it could be populated mostly by Proteron.

Right. I was considering that it might be better to make it an "independent" asset. It also could be a world that isn't entirely/primarily populated by Proteron, but them in addition to ordinary traders/etc (and can be described as such in the world's description).

ids1024:
> That of course could be avoided by setting such a station's faction to "Independent", of course (leaving the Proteron reference just in lore text).

Right. Having them considered the Proteron faction may be best avoided for one reason or another. It also depends on whether we want Proteron ships to appear there, either ordinarily or in specific missions/events (I'm inclined to think we don't want the player to see *all* of the ships the Proteron have before an actual invasion occurs).

BTAxis:
> Right. Having them considered the Proteron faction may be best avoided for one reason or another. It also depends on whether we want Proteron ships to appear there, either ordinarily or in specific missions/events (I'm inclined to think we don't want the player to see _all_ of the ships the Proteron have before an actual invasion occurs).

Definitely no Proteron ships. We want to present the image that, as far as anybody knows, House Proteron was destroyed in the Incident. Whatever remnant may exist this side of the Nebula should be on the level of refugees.

We can still have missions involving Proteron "survivors", but they should just use civilian ships.

onpon4:
We do of course have one Proteron ship (Baren Sauterfeldt), but that one can be explained as an old ship obtained prior to the Incident. 🙂

BTAxis:
Indeed. One-offs like that aren't a problem.

ids1024:
> We can still have missions involving Proteron "survivors", but they should just use civilian ships.

That works for me.

> We do of course have one Proteron ship (Baren Sauterfeldt), but that one can be explained as an old ship obtained prior to the Incident. 

I think it would be entertaining if we include a couple lines of dialog where Proteron officials are bewildered by how he ended up with a top-tier Proteron military vessel. (Though presumably he did so somewhat before the incident). But provide no specific explanation.

I think some kind of world with Proteron survivors should be a good way to begin a Proteron campaign. A Proteron citizen or a former officer may want the players help for, among potentially other things, investigating some unusual circumstances involving the incident. This could lead to an opertunity to side with the Proteron when they show up.

I guess it depends on whether we want the Proteron invasion to occur before or after the player is committed to a faction (though in either case, missions like this could exist).

onpon4:
Is it really top-tier? The Kahan is a Proteron corvette and not decidedly more powerful than any other corvette.

ids1024:
It's certainly described as such.

onpon4:
I think the description actually doesn't match what the ship is really like. Possibly a leftover from when most medium ships didn't have large weapon slots in the past? Some of these descriptions definitely need to be cleaned up for consistency.

BTAxis:
Or just revisited. The nature of the Baron's ship really isn't that important. We can change how it's referenced, hell, we can change the ship if it's convenient.

ids1024:
So, this leaves me an idea for the first part of a Proteron campaign, if there's no existing plan for one.

* The player meets a Proteron refugee who needs the player's help to obtain some supplies (building materials, medicines, alcohol, weapons, whatever) the Proteron refugee community is in need of. This could provide a few relatively routine missions, but the dialog could introduce some lore about former House Proteron.
* This person, trusting the player as a skilled pilot, refers them to a former officer (an admiral, or a Proteron equivalent). This person was the highest ranking among the Proteron known to have survived. They know some classified details regarding the events leading to the incident, and they have suspicions based on what they know that some Proteron worlds may have been outside the area destroyed.

  They do not reveal all this, of course. But they may want the player's help investigating the nebula in some way to confirm if it's composition/distribution/etc fits with their calculations. They want help spying on the Empire. And perhaps make contact with certain individuals and groups that may have information or resources that could be useful.

* Ultimately, the player is tasked with making contact with the main Proteron forces, who have secretly succeeded at travel/communication (however the hypergates end up working) to the rest of the galaxy. The admiral needs to relay to Proteron command all of the information he has (public knowledge, and secrets he's been able to discover) about what has been going on since the incident among the other factions.

* The Proteron invasion then begins in full force.

If the player instead plays another major faction's campaign, the Proteon invasion will still occur at some point. But the player won't have the background knowledge this campaign provides, and presumably on or more other pilots will have played a similar role.

BTAxis:
I've just typed up the rough plot for the Incident and the invasion via the hypergates, as I remember it. Put it in a paste here: https://pastebin.com/ZfHSgNyM

Let me know what you think.

ids1024:
> Initially, only two gates were constructed: one in a remote system [note: unsure which] and one in Sol.

I seem to recall some version I read (probably on the forum) placed the second gate in what is now Collective space, and tied the Collective to the Hypergate project in some way? It's definitely an out of the way location (especially from Sol) and already known as somewhere the Empire has used for secret projects.

>  Thus, in the political upheaval following the loss of Sol, the Empire moved ahead with the original plan, building a network of Hypergates across what remained of inhabited space. When the game begins, this project is still ongoing. The hypergates exist, but are not yet operational. As the game progresses, the gates come online, providing the fast travel system described above.

Should the gates be visible early in the game, then? Should the player know what they are?

> They have no way of knowing which gate leads to the new seat of power of the Empire (the Proteron do not realize that the Empire has splintered into separate, autonomous states). They take the only course of action that will let them leverage the element of surprise: they invade every destination simultaneously.

This sounds somewhat ridiculous, but in a fun way. I like it.

I suppose this conflicts with my idea to have a mission secretly contacting the Proteron leadership before their invasion. So, perhaps change that to contacting them after the invasion begins (reaching Protera by hypergate in the middle of this war stretching the galaxy could be made into an interesting challenge).

onpon4:
There could be multiple ways to start a Proteron campaign honestly. I'd like to see at least one of those options involving a player who has stumbled into Proteron space after discovering the Thurion (i.e. have a way to start a Proteron campaign from Protera if the player makes it there).

ids1024:
> There could be multiple ways to start a Proteron campaign honestly.

Right. The Proteron refugee compaign could be one option. Beginning the Empire campaign but defecting to the Proteron shortly after they invade could be an option. I don't know exactly how we want to introduce the Thurion and nebula-crossing technology, but that could indeed be an option too.

onpon4:
So, one thing that occurs to me reading that: might it be more dynamic to have the hypergates as a lost technology rediscovered on accident from wreckage or something, only to find out that the Proteron have been around all along, have the tech already, and are now using it to invade the galaxy?

BTAxis:
> 
> 
> > Initially, only two gates were constructed: one in a remote system [note: unsure which] and one in Sol.
> 
> I seem to recall some version I read (probably on the forum) placed the second gate in what is now Collective space, and tied the Collective to the Hypergate project in some way? It's definitely an out of the way location (especially from Sol) and already known as somewhere the Empire has used for secret projects.

You're correct, I remembered the same thing last night before I went to bed. The Incident lore Deiz and KAHR-Alpha was more about the Collective than anything else. The Proteron angle I added on later because the sabotage part of it was left somewhat vague. It was always meant to be perpetrated by the Proteron, but I connected it to their invasion agenda.

BTAxis:
> 
> 
> So, one thing that occurs to me reading that: might it be more dynamic to have the hypergates as a lost technology rediscovered on accident from wreckage or something, only to find out that the Proteron have been around all along, have the tech already, and are now using it to invade the galaxy?

Do you mean only the tech was discovered, prior to the Incident, or the gate network itself was already in place and only needed to be activated somehow, a la the X universe? Either way, we'd need to come up with some plausible origin for the tech, because one of the more basic premises of the Naev universe is "humans only". We want to be a bit careful about sprinkling alien influences in.

onpon4:
That's even cooler actually (not familiar with what the X universe is), though what I was suggesting was more of a "the Incident caused us to forget" scenario because of the top-secret nature. Like, that the Sol system might have wreckage of the hypergates intact discovered by scientists and then they look into it and find out that, hey, we can build this hypergate network thing! And suddenly all the factions are getting into this new tech in a race to the top, when all of a sudden, BOOM! Proteron forces sweep in.

BTAxis:
Well, the tech can't be salvaged from Sol, because Sol is so deep inside the nebula that no ships can get close without being destroyed. The intact gate in Collective space could potentially be reverse-engineered in that manner though, but again it's a dangerous section of space so that seems unlikely. Possibly the player could play a role in retrieving data on it via missions, but that would imply that gate construction doesn't start until the game is already underway.

Alternatively, maybe we can work through House Za'lek, who would no doubt be involved at least in the R&D portion of the hypergate project. They'd still have the tech after the Incident, if not insight into the particulars of the project.

onpon4:
Another possibility that comes to mind is that the hypergates could be naturally occurring phenomena that need to be tapped into but provide helpful shortcuts (like Quasispace in The Ur-Quan Masters is). In that sense they would act like standard jump gates except they would lead into a secondary "hyperspace map" that then has routes back into space. It could even have an instability element similar to but not exactly the same as the Nebula making it dangerous for ships not properly fitted for it.

onpon4:
> The intact gate in Collective space could potentially be reverse-engineered in that manner though, but again it's a dangerous section of space so that seems unlikely.

At one point the Collective is dead though. 😉

Regarding Sol, I was thinking it could be discovered by a special Nebula-resistant scientific probe. Then not collected, but studied from afar the same way we study space today.

BTAxis:
> Regarding Sol, I was thinking it could be discovered by a special Nebula-resistant scientific probe. Then not collected, but studied from afar the same way we study space today.

Again though, that defeats the core idea of the nebula as a completely impassable barrier. If a probe can make it to Sol, it can make it out the other side, thus re-establishing contact with House Proteron. I think it's best to not shoot ourselves in the foot that way.

Besides, for Sol we should do something special. I do think the player should get to visit it during the course of the game, but I don't have a clear idea on how exactly, or on what the system should contain. I'm thinking in the direction of some weird, impossible stuff that defies the laws of physics due to the sheer damage to the fabric of reality that the Incident caused. Maybe even something dark and dangerous like the things the Arilou warn you about in Star Control 2.

onpon4:
> Again though, that defeats the core idea of the nebula as a completely impassable barrier. If a probe can make it to Sol, it can make it out the other side, thus re-establishing contact with House Proteron. I think it's best to not shoot ourselves in the foot that way.

That's fair.

Honestly I like the idea of it being a natural phenomenon better anyhow, now that you've made me think about it. 😛

> Besides, for Sol we should do something special. I do think the player should get to visit it during the course of the game, but I don't have a clear idea on how exactly, or on what the system should contain. I'm thinking in the direction of some weird, impossible stuff that defies the laws of physics due to the sheer damage to the fabric of reality that the Incident caused. Maybe even something dark and dangerous like the things the Arilou warn you about in Star Control 2.

That sounds like it could be incredibly awesome.

Maybe the Incident was an intentional self-sacrifice because otherwise a mixup of hypergate configuration led to something even *worse*? 🤔😉

onpon4:
Ping! I think the consensus is that we don't want this PR, right?

Issue #885, opened 2020-01-21
Improve rendering system


bobbens:
Naev at the current time supports OpenGL 1.1 and handles some extensions up to 1.5. It would be probably worthwhile to either rewrite/add support for more advanced shader-based stuff. Either a modern version of OpenGL or Vulkan would probably be best. This will probably save us headaches and bugs in the long run due to the incredibly old API we are using.

ids1024:
I've started migrating the code to use [shaders](https://github.com/naev/naev/tree/master/dat/glsl). It should require OpenGL 3.0 now.

I hope to update all code that is using the fixed-function pipeline or other legacy features, and switch to OpenGL 3.2 core profile. That should provide pretty good backwards compatibly, without limiting things too much.

Ultimately, it's a 2D game, so I don't think we need crazy modern things (e.g. real-time ray tracing). If we think of any good use for any newer OpenGL features, we can consider (optionally) using them.

Hopefully rewriting the rendering code to be more modern should improve performance and make it easier to implement additional effects.

BTAxis:
These changes have broken compilation on Windows due to undefined references. I'm missing a dependency, but I'm unsure which packages to get. Autotools doesn't catch this either. Any ideas?

ids1024:
It's probably because I added a `#define GL_GLEXT_PROTOTYPES` line to get the shader-related functions. Apparently OpenGL on Windows doesn't include `glext.h`...

I guess I should switch that to use Naev's `opengl_ext.c`, or maybe something like GLEW would be better.

ids1024:
Looking at various [OpenGL loading libraries](https://www.khronos.org/opengl/wiki/OpenGL_Loading_Library), I like the look of [Galogen](https://github.com/google/galogen), which generates a header and C file using the official xml specifications of opengl for any choice of GL version and combination of extensions.

ids1024:
With commit 1744ca0803022b2dd91547138bffbff9047457cf, Naev is now using OpenGL 3.2 Core Profile.

ids1024:
I'm getting around 120-130 FPS under normal conditions, with Intel graphics and a 4k display. (Some things using the toolkit can be worse.)

With a tiny window size, I can increase that to about 400. So I guess the main bound on graphics performance here (which is already good enough, on my system anyway) is the fragment shaders and any related per-pixel processing. I'm not sure there's much possible optimization.

[...]

PR #892, opened 2020-01-26
Remove all code related to SDL1 support

ids1024:
Note: I haven't tested `mkspr`, since the Python render scripts in [naev-artwork](https://github.com/naev/naev-artwork) seem to have various issues with the latest Blender.


ids1024:
Getting rid of this old code should make it easier to work on things like https://github.com/naev/naev/issues/885.

PR #900, opened 2020-01-30
Added factional combat music variation.

onpon4:
This defines a variation of combat music by faction. Faction is
determined by checking the faction with the highest presence in the
current system that you are enemies with.


bobbens:
Looks good, thanks!

PR #901, opened 2020-01-30
Replace opengl_ext.c with files generated using Galogen

ids1024:
Looking at several OpenGL extension loading libraries, I like Galogen because it allows generating files for any choice of GL version and extensions. So when we (hopefully) switch to GL 3.2 Core, we can generate headers for such and it will be a compile error to use features that are removed in that version, or added by later versions.

Generated with:
```
galogen gl.xml --ver 3.0 \
               --profile compatibility \
               --exts ARB_texture_filter_anisotropic
```

https://github.com/google/galogen

@BTAxis Hopefully this will fix the Windows build issue you mentioned in #885.
@bobbens Any objection to getting rid of the old mechanism for GL extension loading, or about including generated files like this?


BTAxis:
It does fix the Windows build, thanks.

ids1024:
It seems to have broken the Mac build on Travis due to `KHR/khrplatform.h` not being found. I'm not sure the best solution.

PR #904, opened 2020-02-01
Fixed the music so it plays combat music correctly.

onpon4:
There were a few parts to this. In a nutshell:

1. Fixed the way enemies are counted so that disabled pilots are properly
   accounted for (previously a disabled and then destroyed enemy would
   decrease the enemy counter twice, for instance).

2. Removed the unnecessary requirement for the enemy becoming hostile
   to be the first enemy before setting combat music. (Lua already takes
   care of this, so the check in C was redundant.)

3. Fixed choose_combat Lua function so that if it finds that it's already
   playing, it still marks itself as the last music played (important because
   otherwise "ambient" is always set to "last" when jumping in, and music for
   missions ends up getting set before "idle" plays but after "ambient";
   if previous system had been playing combat music it would catch this and
   then not record itself as the "last" type, leading to ambient being "last"
   even though combat really was).

TL;DR: combat music now plays correctly, 100% of the time. No strange cases
of ambient music when you're fighting a gigantic horde of pirates.

### DIFF
--- a/dat/snd/music.lua
+++ b/dat/snd/music.lua
@@ -315,7 +315,7 @@ function choose_combat ()
       local cur = music.current()
       for k,v in pairs(combat) do
          if cur == v then
-            return false
+            return true
          end
       end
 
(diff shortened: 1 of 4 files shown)


bobbens:
Looks good, thanks!

PR #905, opened 2020-02-01
Fixed several instances of assuming that planet.faction always returns a faction.

onpon4:
The function actually returns nil if the planet has no faction (i.e. Eirok),
so code that relied on this broke whenever the player landed on such a
planet. I've done a grep for these and I think I've fixed every one,
preventing Lua errors in the News and NPC events most notably but also
preventing possible (albeit unlikely) mission errors.

I also added a note that it can return nil to the Lua documentation of the function to make it less likely for this to happen again.

### DIFF
--- a/dat/events/neutral/news.lua
+++ b/dat/events/neutral/news.lua
@@ -379,7 +379,9 @@ articles["Thurion"]={}
    --create generic news
 function create()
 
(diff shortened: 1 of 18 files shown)


bobbens:
Looks good, thanks!

PR #907, opened 2020-02-02
Modified the OSD to compact itself to hide old information.

onpon4:
This change does a couple of things to make the OSD much more readable
when you have a whole bunch of missions you've completed but haven't
collected your reward for:

1. Earlier "steps" of the OSD are not shown; when you complete an
   objective it disappears, so that the first highlighted objective
   is always the first one listed.

2. Multiple OSD entries that would render exactly the same are compacted
   together, and the number of them you have is shown in the title.

This is mainly helpful for an FLF-aligned player; missions done in
Frontier space aren't "finished" until the player lands on Sindbad, so
previously you would have a huge cluster of old OSDs that were no longer
relevant. With this change, these old missions collapse down into a
smaller number of shorter entries, allowing the player to see the
ones that are more important. But the change can help in other cases as
well, e.g. if multiple cargo missions to the same planet have been
accepted.

### DIFF
--- a/src/gui_osd.c
+++ b/src/gui_osd.c
@@ -403,13 +403,23 @@ void osd_render (void)
 {
    OSD_t *ll;
    double p;
-   int i, j, k, l;
+   int i, j, k, l, m;
    int w, x;
    const glColour *c;
+   int *ignore;
+   int nignore;
+   int is_duplicate, duplicates;
+   char title[1024];
 
    /* Nothing to render. */
    if (osd_list == NULL)
       return;
+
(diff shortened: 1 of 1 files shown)


onpon4:
Just pushed one other fix here: fixed a long-standing bug where the OSD's dimensions aren't calculated when you first load your game (causing OSD background size to be wrong until the OSD changed).

bobbens:
Thanks! Modified it to use strncpy and nsnprintf instead of strcpy and snprintf respectively to avoid possible buffer overflows.

Issue #719, opened 2017-04-30
Swapping ships occasionally causes segfaults


onpon4:
Basically, *sometimes*, swapping ships causes Naev to segfault. It always seems to come out of nowhere, and when I go back into the game and do the exact same thing, it doesn't happen again. So I haven't the slightest idea what's causing it or even how to reproduce it. And it's so rare that running the game in the debugger for a couple hours isn't even going to find it. I think it's happened to me, in total, around 3 or 4 times over the last year or so.

So I don't expect this to get solved any time soon, but just opening a report about it so that everyone is aware. Maybe someone will get lucky and stumble across it while they're debugging something else.

BTAxis:
If it happens only sometimes, that could be an indication the game is reading corrupt data from memory. Something trying to reference an object that doesn't exist anymore, maybe?

bobbens:
I tried to reproduce. Haven't been able to.
I ran naev under valgrind to see if it would catch something (valgrind logs memory and can detect illegal read/writes), but it came up clean also. Only thing I can think of is that at some point, when you get more ships, the vector in which the ships are stored gets realloc'd and then stray pointers become invalid and it blows up. Did you buy a ship before swapping?

onpon4:
Hm... I think so. It always happens so suddenly, though, so I'm not sure.

onpon4:
Huh, I didn't realize this was something I found out since I submitted this bug report, but I've found that this tends to happen only when I transport a ship and then swap it (again, though, not usually, only sometimes). So it's the ship location/transportation system that probably needs to be looked at for corruption.

Honestly, I'm not even entirely sure the ship location system is worthwhile. I wonder if it might be better to just remove that feature and have all ships available wherever you go, like outfits are.

bobbens:
That might be for the best. On a side note, probably getting rid of missile ammunition too would be good. Less micromanagement is good.

PhoenixRiver:
> That might be for the best. On a side note, probably getting rid of missile
> ammunition too would be good. Less micromanagement is good.

Though that brings up the issue of balance as generally the need for
ammunition is the reason that missiles are less appealing than lasers
but pack a bigger punch, especially since I don't think the missiles
add to your thermal load like lasers do... of course if less
micromanagement maybe we should remove the thermal system?

River

onpon4:
[...]

PR #908, opened 2020-02-02
Copy previous ship's target info when swapping ships.

### LINKED ISSUE #866: Swapping ships occasionally (rarely) leads to a segfault
This may seem like a deja-vu moment: no, your eyes do not deceive you, yes, you are seeing #719 again.

Yeah, it's not solved unfortunately. :/ I just experienced one of these crashes again. The thing that sucks about this bug is that it's so rare and impossible to produce that it's impossible to know it's broken until you specifically see it failing.

So, just a recap: the bug is that Naev very very rarely segfaults when you swap ships. I've not been able to gather much info beyond that over the years, though I have noticed that when I swap ships multiple times it's more likely to trigger it.

Previously I thought #836 might have fixed this, but sadly I was wrong on that. So! Back to square 1. 🤷‍♀️

onpon4:
This should finally fix #866 (formerly #719). @ids1024 and I did some investigating when the bug surfaced randomly and we found that it was caused by a target planet index being kept as-is when swapping ships; this would cause an attempt to read a bad location if the planet index it had targeted (landed on) previously was an index that isn't valid for the current system (e.g. if it was the 8th planet, but the current system only has 5 planets).

Just to be safe, I decided to copy *all* target info from the previous ship to the new ship. This might be worthwhile anyway because it's possible to call the Lua player.swapShip function in space; copying target info means you don't have your targets cleared when that happens.

### DIFF
--- a/src/player.c
+++ b/src/player.c
@@ -586,6 +586,13 @@ void player_swapShip( char* shipname )
       /* move cargo over */
       pilot_cargoMove( ship, player.p );
 
+      /* Copy target info */
+      ship->target = player.p->target;
+      ship->nav_planet = player.p->nav_planet;
+      ship->nav_hyperspace = player.p->nav_hyperspace;
+      ship->nav_anchor = player.p->nav_anchor;
+      ship->nav_asteroid = player.p->nav_asteroid;
+
       /* Store position. */
       v = player.p->solid->pos;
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good, thanks!

PR #909, opened 2020-02-02
Some small fixes to the FLF

onpon4:
Two changes:

1. Count the FLF as "friendly" at 30 reputation instead of 70. This doesn't have much of a substantial effect except that they show as "friendly" much sooner (which mainly just makes more sense). Also prevents the absurdity (during the campaign) of the pirates being friendly while the FLF are just neutral.

2. Renamed "Dvaered Patrol" to "FLF Patrol" in the OSD, since it makes it clearer that it's an FLF mission and prevents the weirdness of Empire patrol eliminations being called "Dvaered Patrol".


bobbens:
Looks good, thanks!

PR #910, opened 2020-02-02
Made it unlikely (but not impossible) for double-playing a music track

onpon4:
It's a pretty simple change: now, if it selects the same ambient or
combat music as the last music played, it reselects music up to 3
times (as long as that same track is chosen). This makes it extremely
unlikely, but not impossible, for the same track to play twice (for
example, if 3 music choices are available, this reduces the chances
of playing a track twice from 33% to 1%).

I chose this method both because there's nothing wrong per se with
playing a duplicate on occasion, and in case only one music choice is
defined for some reason.

### DIFF
--- a/dat/snd/music.lua
+++ b/dat/snd/music.lua
@@ -249,7 +249,20 @@ function choose_ambient ()
       end
 
       -- Load music and play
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good, thanks!

PR #917, opened 2020-02-04
Improved the mining system, added more asteroids around the galaxy to mine from

onpon4:
OK, this is a pretty big one. Changes to the asteroid mining system include:

* Removed the "asterokiller" distinction of weapons. Asteroids now simply take physical damage (like hulls do) until they run out of "armour", at which point they are destroyed. This is much more simple and removes the need for specialized weapons.

* Replaced the polygonal system for defining asteroid fields with a much simpler circle-based approach. This makes it much less tedious to define asteroid fields while not really losing much. (To be clear, the horrible tediousness of defining asteroid fields the previous way was my main motivation for doing this.

* Removed the requirement to slow down when trying to collect gatherables. That mechanism didn't really add anything, it just made the process of mining needlessly more tedious.

* Added a few new asteroid fields. I made a point of putting these in dangerous places (i.e. places with pirates in them).


ids1024:
> Replaced the polygonal system for defining asteroid fields with a much simpler circle-based approach. This makes it much less tedious to define asteroid fields while not really losing much. (To be clear, the horrible tediousness of defining asteroid fields the previous way was my main motivation for doing this.

I like the idea of a more flexible system than just circles (though something supporting circles as well would work). It could add interesting detail to systems to be able to have a ring of asteroids around  a planet, a crescent shaped asteroid field (probably unrealistic, but no less so than other things), etc.

I think polygons wouldn't be too bad with proper editor support. Though, I'm not sure I'm volunteering to implement a full vector graphics editor in the system editor...

onpon4:
Well that's incredibly theoretical. Why would asteroid fields need to have complex shapes to begin with? All it does is massively complicate the code and make defining basic circular areas (the kind you most likely want) much more tedious.

To wit, I didn't even notice that the asteroid field in Hakoi was non-circular until I saw what it looked like in the editor. Or any other asteroid field, for that matter. In gameplay there's no discernable difference.

Besides, the code allows asteroids to move between areas no problem, so you can define complex shapes if you absolutely must as a series of overlapping circles.

And circles are also much, much more efficient than polygons. Rather than having to calculate all these polygon boundaries (which takes something like hundreds of lines), you just check the distance from the asteroid field against a radius. Determining initial position is a lot more simple and quick too: choose a random angle between 0 and 2 * M_PI, and a random radius between 0 and radius, and use basic trigonometry to figure out the coordinates from that.

I don't think editor support is a solution. Suppose you did allow editor support so you can click 10 dots in a row to make a circular-ish shape. Well, now suppose you want to make it bigger. Delete, add 10 new points further apart or closer together, then reconfigure to match the old one; or drag all 10 points out one by one to the new size. Whereas if it's a circle all you do is change the radius. Much simpler.

I don't even think complex shapes are necessarily easier with polygons. Some (like, say, circles) are objectively harder. Rectangles, sure, those are easier. (But why would you want a straight line of asteroids?) But when you get into the realm of things like, say, rings of asteroids around a planet, the polygon thing doesn't help at all; in a good GUI interface you just click and drag to make some circles forming a ring shape, whereas with polygons you click 25-ish times to get a polygonal ring-like shape (while trying your best to keep the ends of the "ring" as close together as possible, but not being abe to cross them because, to be clear, this polygon system doesn't allow lines to intersect, it errors out when you try). And I don't even know for sure if that method using the polygon definitions would work (I'm not about to try; I'm not masochistic enough to put in that amount of work with the polygon system).

nenau-again:
I remember that I had first implemented circular fields, and then I was asked to switch to polygonial fields instead (which gave me a few very cute geometrical problems to solve btw). I don't really know what to think about this point right now...

Anyway, the need to slow down to gather ore had proven to be essential to avoid having unrelated ships gather stuff by mistake on their way.

BTAxis:
In regard to belts and such, maybe you could allow defining a "negative" field that erases whatever asteroids would be generated within it, so you could just put a negative circular field inside of a normal one to make a belt.

I'm also still toying with the idea of systems that have endless asteroids in all directions, with pockets of empty space inside it to navigate. I don't have a use case for this though, and it doesn't make much sense at all, it's just an idea.

onpon4:
> In regard to belts and such, maybe you could allow defining a "negative" field that erases whatever asteroids would be generated within it, so you could just put a negative circular field inside of a normal one to make a belt.

That would indeed be a much better way to do that, and I considered that when I saw @ids1024's reply. All it would have to do is forbid spawning in that location and make it a "no asteroids" zone (spawning could still happen the exact same way no problem, just have a check when spawning is all). Simple and yes, makes things like rings a lot simpler to make (as opposed to polygons where you have to define a whole bunch of points to try to approximate the effect and hope that it works).

This would also be another way to define things like narrow strips of asteroids. One big asteroid fields with two exclusions on either side, making a sort of hourglass shape. Add a few more exclusions to straighten it. Pretty simple.

In fact, I think I'll implement this when I get back to my desktop computer, given how easy it will be to do.

> I'm also still toying with the idea of systems that have endless asteroids in all directions, with pockets of empty space inside it to navigate. I don't have a use case for this though, and it doesn't make much sense at all, it's just an idea.

Endless isn't really feasible because of the way Naev generates stuff (you can't make it in such a way that they're endless because spawn locations are absolute and not player location based), but with the circle system you can easily enough just make the radius something like 1.5 times the radius of the system to essentially achieve the effect.

onpon4:
> Anyway, the need to slow down to gather ore had proven to be essential to avoid having unrelated ships gather stuff by mistake on their way.

I did notice that possibility. Honestly I don't see that as a problem. Asteroids respawn anyhow so all it means is that you need to find a new asteroid and shoot it down. No different than if the asteroid was empty to begin with. (Plus I thought it was kind of neat when pirates attacking me just swooped in and grabbed the ore I was mining from right under me.)

I think it's better than the tediousness of having to slow down every time you want to collect some ore.

bobbens:
One thing that worries me about only circular asteroid fields is the interactions with "safe lanes" when implemented. The idea is that safe lines should create safe paths between asteroid fields and the likes. It might be possible with circles, but how do you handle overlapping elements? Just double the density? I like the other changes, but I still thing polygons are probably the best way. I would propose doing the following changes:
1. Implement polygon editor for asteroids
2. When safe lines are implemented, allow the game to automatically (and deterministic) create safe lanes and the asteroid polygons based on high level specifications like system asteroid density and such.

These two changes would make it easy to use polygons without the limitations of a circle-based system. Thoughts?

Zireael07:
Safe lanes would probably just remove any asteroids in the path.

PhoenixRiver:
> * Removed the &quot;asterokiller&quot; distinction of weapons. Asteroids now
> simply take physical damage (like hulls do) until they run out of
> &quot;armour&quot;, at which point they are destroyed. This is much more
> simple and removes the need for specialized weapons.

This was done to make asteroid mining a choice (as I remember) as was
the slowing down thing... like switching to a cargo hauler to more
efficiently do cargo runs... now you can mine in any old ship... (less
interesting if you ask me).

> * Replaced the polygonal system for defining asteroid fields with a much
> simpler circle-based approach. This makes it much less tedious to define
> asteroid fields while not really losing much. (To be clear, the horrible
> tediousness of defining asteroid fields the previous way was my main
> motivation for doing this.

Gah, I have been working on a thing that has broken up planets
circling their star as a ring of asteroids... I even created a script
to do it all with actaul objects before the asteroid field definition
was added, and now it seems I'll have to resurrect it again (and the
terrible number of objects) to make it look right... so count this as
a vote against circle fields (especially since a ring of asteroids
around a star is about as simple as an asteroid field gets in reality
- unless you count a stars accretion disk but then most of the
material is not really asteroids yet at that point...)

> * Added a few new asteroid fields. I made a point of putting these in
> dangerous places (i.e. places with pirates in them).

Definitely appreciated :)

Just my tu'penny,

Phoenix

onpon4:
> These two changes would make it easy to use polygons without the limitations of a circle-based system. Thoughts?

Polygons have limitations to. You can't create a circle with a polygon.

If you really wanted full flexibility you would implement a large variety of shapes. But this is an asteroid field we're talking about. Who's going to notice if a field is a circle rather than a square? Or a circle rather than a trapezoid? It's like worrying about the shape of a level in a platformer and putting a whole lot of work into making it possible to make triangle levels when really, no one's going to notice if all the levels are rectangles (as is the usual case).

It's needless complexity.

> One thing that worries me about only circular asteroid fields is the interactions with "safe lanes" when implemented. The idea is that safe lines should create safe paths between asteroid fields and the likes. It might be possible with circles, but how do you handle overlapping elements? Just double the density?

Asteroids aren't dangerous though? And if you must exclude them from safe lanes it wouldn't be hard to just add a check not allowing asteroids to be within a certain range of safe zones. You don't need to manually define polygon shapes for that.

That ties in with the exclusion idea @BTAxis mentioned.

As for density, that really doesn't matter. Asteroids can travel freely between different fields anyway. A few dense pockets aren't likely to be noticed by the player unless you have such a dense field that the player can see the field boundaries, which would look unnatural anyhow. But even if you're worried about that, as I said, safe lanes can simply act as exclusion zones. If safe lanes are basically just lines that would be very simple to do (just find an asteroid's distance from that line).

What I'm advocating here (and have implemented, though I'll be touching it up with that exclusion zones idea when I get the chance) is simplicity. There's no sense tasking a system designer with painstakingly specifying exact details of how an asteroid field is shaped to get effects that should be consistent.

Heck, if it's desired, we could even add automatic exclusion zones to planets for asteroid fields, too.

Basically, I think you're making a simple task needlessly complex in a way that doesn't enhance gameplay (which is what matters the most) at all.

> Gah, I have been working on a thing that has broken up planets
> circling their star as a ring of asteroids... I even created a script
> to do it all with actaul objects before the asteroid field definition
> was added, and now it seems I'll have to resurrect it again (and the
> terrible number of objects) to make it look right... so count this as
> a vote against circle fields (especially since a ring of asteroids
> around a star is about as simple as an asteroid field gets in reality
> - unless you count a stars accretion disk but then most of the
> material is not really asteroids yet at that point...)

But surely a field with a smaller exclusion zone would achieve this effect?

onpon4:
And just to note, those added asteroid fields _would not have happened_ under the polygon system. I don't have the patience for defining asteroid fields that way.

nenau-again:
As it looks, both polygons and circles both have their own advantages :
* Polygons make it possible to do complex shapes without weird overlapping zones, where the density will not be homogeneous
* Circles are more simple to use and to represent

So what about keeping both : what I propose is to have 2 types of fields : one of them being circular and the other polygonal. That would make it easier to add other types like the ring form (2 radius and a center), the ellipse... Shouldn't be so hard to do.

(I confess I am also trying to save the hundreds of polygon-related lines of code I had made)

onpon4:
I view polygons as just a bad idea. Keep in mind the sunk cost fallacy: just because you worked on something doesn't mean throwing it away is the wrong course of action. In fact lots of old code has been thrown away throughout this project's history.

We could certainly add support for every shape under the sun: circles, polygons, ellipses, diagonal ellipses, parabolas... heck, you could modify it to use equations somehow. You could do all kinds of things. But at the end of the day, these are fields of asteroids that spawn randomly. They're not artistic canvases.

PhoenixRiver:
> I view polygons as just a bad idea. Keep in mind the sunk cost fallacy: just
> because you worked on something doesn't mean throwing it away is the wrong
> course of action. In fact lots of old code has been thrown away throughout
> this project's history.

But my old system was so much more inefficient and the polygon code,
though less efficient than circles is also more versatile and more
efficient than creating a bunch of "planets" to look like asteroids...

> We could certainly add support for every shape under the sun: circles,
> polygons, ellipses, diagonal ellipses, parabolas... heck, you could modify
> it to use equations somehow. You could do all kinds of things. But at the
> end of the day, these are fields of asteroids that spawn randomly. They're
> not artistic canvases.

But, as I mentioned, the default shape for real galaxy asteroid
"fields" is as rings around suns... these circular fields only really
happen near trojan points and even then we are not modeling the
density variation those would have... yes, I understand that Asteroids
the game had a uniform field of randomly moving rocks but that doesn't
make it anything other than just what was reasonable for the game at
the time - it also had screen-wrap... should we be implementing that
in Neav? (It's simple to do annd saves dealing with all the problems
of (literal) edge cases...)

Phoenix

ids1024:
And as an aside...

> (I confess I am also trying to save the hundreds of polygon-related lines of code I had made)

@nenau-again If you like polygons, I think we should change the collision detection code (for projectiles with ships), which currently does pixel perfect collisions, to be polygon based. This should be faster, and can be used for sprites and 3D models (though I guess we'd have to generate multiple polygons for different orientations).

onpon4:
> But, as I mentioned, the default shape for real galaxy asteroid
> "fields" is as rings around suns...

Again, what's wrong with just doing that as a circle but with a smaller circular exclusion zone? Like say a radius 1000 asteroid field with a radius 800 exclusion zone; that would give you a perfect circular ring with a size of 200.

onpon4:
> yes, I understand that Asteroids
> the game had a uniform field of randomly moving rocks but that doesn't
> make it anything other than just what was reasonable for the game at
> the time - it also had screen-wrap... should we be implementing that
> in Neav? (It's simple to do annd saves dealing with all the problems
> of (literal) edge cases...)

I don't really understand what you mean here. What had screen-wrap? Are you referring to an old implementation of asteroid fields? I'm not familiar with any old asteroid system, just the current system.

onpon4:
OK, so I've added exclusion zones, with an example found in Lapis. It was harder than I thought it would be because of the way asteroids are generated, but works just fine now. (Also fixed some magic numbers in there around the "appearing" variable).

onpon4:
I've pushed a change that I mentioned on Discord: I've made asteroids occasionally randomly explode as a way of adding a small sense of danger to asteroid fields.

onpon4:
I've pushed a bunch of changes recently:

* Asteroids now explode more frequently. I'm very  happy with this behavior. It creates a sense of danger in asteroid fields to hear explosions all around you and it adds some tactical nuance (where the asteroid explosions hurt small ships but the ew bonus helps small ships).
* Fixed rendering of debris so that it's never on the front of the nebula overlay.
* Added several new asteroid fields, including a giant asteroid field in Sol to represent destroyed planets. and a couple cases of asteroid rings around planets (Gammacron IV and Nougat I).
* Made it so that turret weapons aim at asteroids if no pilot target is selected.
* Added miner presence to several systems with asteroids.
* Exdended miners to also use Koalas and Mules.

One more thing needs to be done: the miner AI doesn't work properly and needs to be fixed (miners seem to just fly around confused, I think because the AI was coded in an inflexible way that only works with Llamas that have certain equipment).

Other than the problem with the miner AI (which I'm looking into fixing but need some help with), I am very happy with what I have here.

nenau-again:
The problem with miner AI may be because they use the weaponset that was used by asterokiller weapons, but as you suppressed this class, they're maybe simply trying to use an empty weaponset.
Anyway, if you've got questions about that AI, I'll be glad to answer.

onpon4:
I don't think that would be the case? Just because the Yachts often succeed at shooting but the other ships struggle to do anything at all.

In any case the behavior of the AI is partly obsolete because turreted weapons aim at asteroids now and swivel works for cannons. Also if I understand correctly I think it always stops on the right-hand side of the asteroid no matter what? Basically it seems it needs some major clean-up.

onpon4:
As an aside the main thing I need help with isn't that specific AI per se but just that I don't understand mem.Kp and mem.Kd, or what a PD system is or what it does in the context of Naev.

nenau-again:
The PD controller is the way used to make sure that ships manage to follow moving objects at nearly 0 relative speed. Kp and Kd are the two parameters of this controller and I used to determine some suitable values per experience. i had let the possibility to change them for very specific applications, but until now, the case did not show up.
It's true that the miners only stop at the right of asteroids, and that could indeed be changed for a random angle.
About the fact that the AI works for Llamas and not for others, i don't really know right now why it could be...

onpon4:
It seems likely to me that it's because acceleration / speed differences. But that's based on limited knowledge on my part.

It seems to me that given the changes I've made to asteroid mining here it would make sense to have the AI mine in a manner more similar to attacking,

nenau-again:
> It seems likely to me that it's because acceleration / speed differences.

Yep, it's very likely to be true : there is a condition for quitting the poitionning step at line 725 (for me) in basic.lua `if relpos < 30 and relvel < 2 then`. I guess you could try to increase those values (mostly on relvel). I hope it will do the trick...

> It seems to me that given the changes I've made to asteroid mining here it would make sense to have the AI mine in a manner more similar to attacking,

I partially agree : the main way I made this behaviour is to avoid being caught in the explosion of the asteroid, what could be damageable for smaller ships.

onpon4:
That's true. 🙂 (Though of course explosions can also be a risk when attacking pilots, albeit usually less severe than asteroids.)

What I noted in particular re: speed and accel is that some ships seemed to try to turn around to brake a lot later than they should have.

onpon4:
OK, this was a massive headache but I *think* I've managed to fix up the mining AI code so that it works well now. I've also fixed a problem in the C code which would have led to miners having nonsense asteroid targets

I probably wouldn't have managed to do it without @nenau-again giving me pointers on Discord, so thanks for that!

This PR should be all set now. There are lots of asteroid fields, miners which behave in a reasonable way, and mining asteroids is a fun side-job that any player can take on (which I think is the best possible fate for asteroid mining). Asteroid fields are circular, but exclusion zones enable rings (seen on Gammacron and Nougat) and other complex shapes if desired.

onpon4:
I've just pushed a change in response to an upstream removal of a function this PR depended on. Mergeable once again now.

onpon4:
I've added beams aiming at asteroids! Was even harder than I thought, but all set now. As with turrets it prioritizes pilots over asteroids.

Also noticed that the Miner faction didn't have its own standing script, so I fixed that as well.

onpon4:
Also worth noting, there's a whole bunch of unrelated fixes that haven't been merged in here because I happened to find them while I was working on the mining system.

bobbens:
Looks good, thanks!

PR #918, opened 2020-02-08
fix in sh01 mission

nenau-again:
In the second shark mission, I imposed the opponent to jump always from Raelid (which is most probably the jump the player has used) and set the player as visible to make sure the fight will actually occur.
What's more, I added a timer in order to give the impression the opponent has jumped after the player.


bobbens:
Looks good, thanks!

PR #919, opened 2020-02-08
Pirate Koala does not re-spawn in kidnappers mission

nenau-again:
If the children have been rescued, there is no more reason for the kidnapper's Koala to be still there. What's more, I've added the possibility to destroy the Koala after having boarded it (as there are no more children in it)


bobbens:
Looks good, thanks!

PR #920, opened 2020-02-09
adapted sh02 to missile rebalance

nenau-again:
The Hyena  is unable to use interceptor missiles, so I gave the launcher to the Lancelot and upgraded the equipment of Hyenas in the mission


bobbens:
Looks good, thanks!

PR #921, opened 2020-02-09
Reset faction flags when resetting their standing.

onpon4:
This fixes a problem where starting a new game after having loaded
a previous save (in the same session) leads to the "known" status
of the previous save leaking into the new one. It should fix any
other such problems with flags if they crop up again, too.

### DIFF
--- a/src/faction.c
+++ b/src/faction.c
@@ -82,6 +82,7 @@ typedef struct Faction_ {
 
    /* Flags. */
    unsigned int flags; /**< Flags affecting the faction. */
+   unsigned int oflags; /**< Original flags (for when new game is started). */
 } Faction;
 
 static Faction* faction_stack = NULL; /**< Faction stack. */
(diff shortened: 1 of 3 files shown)


onpon4:
Just pushed one more fix: fixed the calculation of range for rockets.

nenau-again:
Wow, nice catch for the calculation of the range of rockets !
Btw, the fact that dumb ammo and seekers don't obey the same physics is rather odd. It could be changed imho. (for example, it results in an odd behaviour of torpedos)

onpon4:
You mean like, making seekers inherit momentum, or not making dumfires do it?

The former isn't really feasible because their speed changes as they chase the target.

The latter would be possible but would make the usage of dumbfire rockets a lot more simplistic and less effective for chasing ships (especially slow rockets like the Banshee). Like for instance you could have a fast ship and it could be moving so fast that the rocket falls behind the ship.

onpon4:
One thing I suppose you could do is have missiles that are currently going faster than their max speed slowly decelerate (as with ship thrusting in Star Control). That wouldn't completely put momentum into them though.

bobbens:
Looks good, thanks!

PR #929, opened 2020-02-16
Added "Fake ID" pirate "mission".

onpon4:
This is the same concept as rehabilitation missions, except that the
fake ID "mission" has no victory condition. It simply gives you a
fake reputation with the major factions that you can use to conduct
your business uninterrupted, but takes that away if you do too much
to get noticed.

Namely, you lose your fake ID if you ever wrong an affected faction
in any way (as with rehab missions), if you gain reputation on an
affected faction after its mission cap has been adjusted (i.e. doing
major campaigns for the faction, since that gets you attention which
means added scrutiny), and occasionally when you gain pirate
reputation within one of the affected factions' systems.

Of course, the effect is reversed when aborting the "mission", too.

This can serve three purposes:

1. To help pirate-aligned players travel and do pirate cargo missions more easily.
2. To allow former pirates or FLF pilots to give up on a life of piracy/terrorism.
3. As a part of some campaigns (like maybe the Thurion and/or Proteron campaigns).

### DIFF
--- a/dat/mission.xml
+++ b/dat/mission.xml
@@ -75,6 +75,14 @@
    <location>Computer</location>
   </avail>
  </mission>
+ <mission name="Fake ID">
+  <lua>pirate/fake_id</lua>
+  <avail>
+   <chance>100</chance>
+   <location>Computer</location>
+   <faction>Pirate</faction>
+  </avail>
+ </mission>
  <mission name="Cargo Rush">
   <lua>neutral/cargo_rush</lua>
   <avail>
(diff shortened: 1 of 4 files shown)


onpon4:
Can I haz merge or at least a review? I understand @bobbens had a different idea for how to do this but this could easily be extended or replaced later if his idea ever gets implemented. And I think this is a nice thing I've got here.

bobbens:
Looks good, thanks!

PR #933, opened 2020-02-17
Renamed pirate standings to something better.

onpon4:
I made a point to note in the standings whether or not you can access clan worlds.


bobbens:
Looks good, thanks!

PR #934, opened 2020-02-18
Replaced jammer functionality with a replacement that uses ewarfare.

onpon4:
What I've done here is change jammers to simple ewarfare-enhancing activated outfits, and replace the existing jam functionality with one that uses ewarfare. I needed some help from @bobbens for this, so thank you. 😛

The "Jam Resistance" value of missiles is now used as a "tracking" value to determine the missile's turn rate. I also updated the way jam resistance is displayed to more accurately represent how it affects jamming.

The numbers I picked are more or less ballparks, but I did test them a bit to make sure they're at least sane. They may need further adjustment. Choosing the right numbers is kind of hard because the way ewarfare is calculated is complicated / poorly documented and the way tracking works is dependent on that (tracking effectively works as a double-divisor for the relevant ewarfare info).

Worth noting: this removes the reduction of speed that jammers had on missiles. The main reason I did this is because I removed the bit of only affecting missiles in a certain range, and without that reducing speed effectively causes missile range to plummet. I decided it wasn't really worth it, so I left it out; jamming missiles via ewarfare is all about making them less maneuverable (slower turn rate).

### DIFF
--- a/dat/outfits/activated/milspec_jammer.xml
+++ b/dat/outfits/activated/milspec_jammer.xml
@@ -10,9 +10,9 @@
   <limit>jammer</limit>
   <cpu>-8</cpu>
  </general>
- <specific type="jammer" group="7">
-  <range>230</range>
-  <power>50</power>
-  <energy>-30</energy>
+ <specific type="modification" group="7">
+  <active/>
+  <ew_hide>90</ew_hide>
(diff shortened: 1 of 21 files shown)


bobbens:
Looks good, thanks!

PR #937, opened 2020-02-19
Removed the "reputation" event.

onpon4:
This event broke the "fake ID" mission and was incomplete anyway.
Removed rather than fixing because it's a confusing practice to have
an event "fix" reputation caps that seem "wrong" and could easily
lead to other difficult to solve problems.


bobbens:
Looks good, thanks!

PR #939, opened 2020-02-19
Bugfixes galore

onpon4:
I forgot to consider the case of the cap being nil in #938. Fixed here. Also tweaked the pirate reputation titles a bit.


onpon4:
One more enhancement: I've made pirate transport missions always show up on pirate worlds that offer missions (if destinations are available).

onpon4:
I've pushed another fix: pirate shipping missions never choose a pirate stronghold as their destination.

onpon4:
Noticed some massive problems with the "foodbiz" campaign, so I almost completely rewrote that. Also fixed a typo in the pirate cargo missions.

bobbens:
Looks good, thanks.

PR #940, opened 2020-02-20
Added text explaining what to press to autohail.

onpon4:
This is in draft mode because, ah, it's not working. I have no idea why, though. With the default autohail bind (Ctrl+Y), I get it displaying at "Ctrl + " (without the "Y").

This also has a bugfix though: I accidentally left the chance of the baron event happening at 100% rather than 4% as it should be (this was a temporary value for testing). Sorry about that.

### DIFF
--- a/dat/event.xml
+++ b/dat/event.xml
@@ -80,7 +80,7 @@
  <event name="Baroncomm_baron">
   <lua>neutral/baron_comm</lua>
   <trigger>enter</trigger>
-  <chance>100</chance>
+  <chance>4</chance>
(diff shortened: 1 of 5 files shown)


onpon4:
So apparently I was getting the string length wrong. Thanks @bobbens. This is good now.

onpon4:
Just to explain: this pull request causes the text alerting you that someone is hailing to tell you how to autohail. For accessibility to new players.

bobbens:
Looks good, thanks!

PR #946, opened 2020-02-26
WIP : polygonal collision algorithm

nenau-again:
Here is what I've got right there.

Many ships and outfits have now their polygon collision equivalents, and collision is computed thanks to these polygons, if availiable. A few limitations right now : 
* Some ship's png don't have a size that makes it possible to split correctly the sprites in order to compute the polygon. Normally, the X size should be dividable by sx and Y by sy.
* Code is rough right now, I'll make it more readable when I have time.
* Some shapes are too complex to compute easily the polygon. I know how to do it, but I feel a bit lazy right now.
* Ammo outfit don't use polygon collision yet for the same laziness reason.
* I did not really compute the CPU gain (or loss) with this algorithm.
* I do not 100% guarantee that my implementation is free of memory leaks or other weird C issues

Please tell me what you think of it !


ids1024:
Great! Though I'll probably need some time if I want to evaluate the exact details of the algorithm and implementation.

Ideally, I'd like to see some test that shows this either improves performance or at least doesn't degrade it. I'm not sure how. I guess calling the old and new collision tests in a loop in a test outside of Naev and seeing its runtime would probably be optimal. A fair bit of work though. Perhaps there's an easy way to just measure the average runtime of the functions while running the game (I don't know if something like gprof provides an easy way to do that).

A minor point that's easy to address: I think it's probably good to have a comment at the top of generated files saying they are generated.

I'm inclined to think the script here should belong in [naev-artwork](https://github.com/naev/naev-artwork) so it can be integrated with the Python scripts for exporting blender models.

ids1024:
I don't know if it really matters, but do we want to avoid duplicating things? For instance, `dat/gfx/ship_polygon/phalanx*.xml` are all identical.

Perhaps it's not worth addressing this until 3D models are in use. At which point, I want to have these ships share the same model (but different textures), so the polygons can be shared by the same mechanism, whatever that is.

nenau-again:
How do you enable the profiler during compilation? Do I have to find the right place to add a `-pg` in Makefile? This file looks quite intimidating...

ids1024:
Try `./configure CFLAGS=-pg LDFLAGS=-pg`

nenau-again:
Okay, I've organized a few battles between 300 empire and pirate admonishers under gprof.

Typical result without polygon collision:
`Each sample counts as 0.01 seconds.`
`  %   cumulative   self              self     total `          
` time   seconds   seconds    calls   s/call   s/call  name `
` 13.31      2.36     2.36   450210     0.00     0.00  weapon_update`
`  9.28      4.01     1.65 94214479     0.00     0.00  CollideSprite`

And with polygon collision:
`Each sample counts as 0.01 seconds.`
`  %   cumulative   self              self     total     `
` time   seconds   seconds    calls   s/call   s/call  name`
` 13.33      2.51     2.51   406785     0.00     0.00  weapon_update`
`  5.84      3.61     1.10 31668735     0.00     0.00  luaS_newlstr`
`  5.58      4.66     1.05  3063309     0.00     0.00  luaV_execute`
`  4.94      5.59     0.93 96347402     0.00     0.00  luaH_getstr`
`  4.41      6.42     0.83 77072858     0.00     0.00  CollidePolygon`
[...]
`  0.53     13.95     0.10 12152786     0.00     0.00  CollideSprite`

(When polygon collision is enabled, seekers still use CollideSprite function for now.)

I did a few runs of each and the results seems to be quite stable : with admonishers at least, one can conclude that the cost of `CollidePolygon` is ~1/2 the cost of `CollideSprite`.
This result is rather positive, but more ships could maybe be tested (the shape is probably quite important). What is more, for some sprites (mainly Za'lek ones, but also some orientations of the Lancelot for example), I'll have to refine the polygon in order that the polygon generation algo gives a result, what could lead to bigger computation time...

nenau-again:
I did it again with more diverse ships and the results are fairly the same. On my laptop, The cost of `CollideSprite` is about 1.85e-8 s/call and `CollidePolygon` is 1.05e-8 s/call.

ids1024:
> The cost of CollideSprite is about 1.85e-8 s/call and CollidePolygon is 1.05e-8 s/call.

Great! I don't know how much of margin for improvement further improvement there is here, but this is a pretty promising result since our highest priority here is just having a collision method that also works with 3D models. So something that reliably performs better is a great bonus (though, that's expected for polygons vs checking each pixel).

In addition to general code cleanup, and improving handling of some of the ships this had issues with, I think it would be good to have a comment at the top of `polygon.py` summarizing the approach it uses.

nenau-again:
> our highest priority here is just having a collision method that also works with 3D models

One thing to figure out : how are we going to generate polygons from 3D models. One way could be to generate sprites from models, and then run the algo that generates polygons from sprites...

> So something that reliably performs better is a great bonus (though, that's expected for polygons vs checking each pixel)

Mm. Actually, it's not that easy because we do only very simple operations with integers in `CollideSprite`, while products and atan are performed on float variables in `CollidePolygon`. By the way, that's why I dont use double, but just float.

> I think it would be good to have a comment at the top of polygon.py summarizing the approach it uses.

OK, it will be done, and btw, as you wrote earlier, this script should rather end in the atwork repo.

ids1024:
> One way could be to generate sprites from models, and then run the algo that generates polygons from sprites...

Sprite generation takes rather a long time, much more so than just exporting a model (and the Python script for it currently seems pretty broken with recent versions of blender). So I'd rather not rely on this, ultimately.

ids1024:
I've written some crude code to show polygons in game: https://gist.github.com/7dee202402051577a19ff7daeabea151

screenshot004

It seems helpful for visualizing. Just some quick code; I'd write it better, make it a configurable option, factor it out into a function and use it for weapons as well, etc. if we wanted to actually include it in naev.

nenau-again:
Hey, that looks cool ! Very good debug feature indeed, I think it's worth adding.
Meanwhile, I've managed to generate polygons for all the ships and ammos. Last thing is to manage the beams (shouldn't be very hard) and I'll push it.

nenau-again:
Here we are. I believe I'm done for 2d ships. Please tell me what I could have forgotten. I'll submit the polygon generation script to naev-atwork.

ids1024:
Seems to be working well, and the polygons look good (including for Za'lek ships).

There seem at least one memory error: https://gist.github.com/ids1024/fd563e904d6ca2a36338c6c26d027838

This is from building with `./configure CFLAGS='-fsanitize=address -fno-omit-frame-pointer -g' LDFLAGS='-fsanitize=address -fno-omit-frame-pointer -g'`.

nenau-again:
That was a stupid off-by-one error. Surprising that the code was actually running... But according to the new configure options you gave me, there are plenty of leaks of the type :
`Indirect leak of 30 byte(s) in 3 object(s) allocated from:
    #0 0x7fdd14fae538 in strdup (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x77538)
    #1 0x7fdd13dcdfd7  (/usr/lib/x86_64-linux-gnu/libfontconfig.so.1+0x24fd7)
    #2 0x6d0000006f0062  (<unknown module>)`
I don't really know how to cure this...

ids1024:
That leak should be unrelated to this change; there's a lot of leaks like that on master. I've started addressing them in some changes in my git stash.

>  I'll submit the polygon generation script to naev-atwork.

Why not PR the script now?

nenau-again:
> Why not PR the script now?

I have a few more cleaning to do. I'll do that tomorrow.

nenau-again:
A few polygons have changed because I was still developing the generation code while generating them. I updated them.

ids1024:
I think I'm happy to merge this.

@bobbens Any objections?

bobbens:
It looks good. Some things that I can see as future improvements:

1. Get rid of `OPENGL_TEX_MAPTRANS` in `ship_loadSpaceImage` and also for outfits when polygons are available (to avoid using more memory).
2. Add the polygon visualization option to the options menu when in `--devmode`
3. Warn if no polygons are available (and possibly mention briefly how to generate them or point at a wiki page)

src/ship.c
```diff
@@ -741,6 +811,9 @@ static int ship_parse( Ship *temp, xmlNodePtr parent )
          /* Load the graphics. */
          ship_loadGFX( temp, buf, sx, sy, engine );
 
+         /* Load the polygon. */
+         ship_loadPLG( temp, buf );
```
bobbens: Should warn if no polygon is available and give a hint or something on how to generate it.

src/outfit.c
```diff
@@ -1067,6 +1168,9 @@ static void outfit_parseSBolt( Outfit* temp, const xmlNodePtr parent )
             temp->u.blt.spin = atof( buf );
             free(buf);
          }
+         /* Load the collision polygon. */
+         buf = xml_get(node);
+         outfit_loadPLG( temp, buf, 1 );
```
bobbens: Should warn if no polygon is available and give a hint or something on how to generate it.

src/outfit.c
```diff
@@ -1413,6 +1517,9 @@ static void outfit_parseSAmmo( Outfit* temp, const xmlNodePtr parent )
             temp->u.amm.spin = atof( buf );
             free(buf);
          }
+         /* Load the collision polygon. */
+         buf = xml_get(node);
+         outfit_loadPLG( temp, buf, 0 );
```
bobbens: Should warn if no polygon is available and give a hint or something on how to generate it.

src/collision.c
```diff
@@ -104,6 +172,192 @@ int CollideSprite( const glTexture* at, const int asx, const int asy, const Vect
...
+      }
+   }
+
+   return 0;
+}
+
+
+/**
+ * @brief Checks whether or not two polygons collide.
+ *  /!\ The function is not symmetric: the points of polygon 2 are tested
```
bobbens: Not sure if it would be worth to make it always swap the polygons so that polygon 2 is the smaller one or have a wrapper function do that.

src/ship.c
ids1024: If we can generate polygons completely reliably, we might want to just completely remove the code for sprite collisions, and error if there isn't a polygon.

ids1024:
Since it's working without issue, I've merged this and https://github.com/naev/naev-artwork/pull/13.

Of course, PRs for any further improvements are welcome!

src/ship.c
nenau-again: > If we can generate polygons completely reliably, we might want to just completely remove the code for sprite collisions

I think it's quite risky for now... I would say that we can generate polygons semi-reliably. Indeed, afaik, there is no algo that works in all cases to generate non-convex polygons.
However, thanks to the fact that in our case, the smaller size is a pixel, I managed to write an algo that always works, at the price of extreme refinement of the polygon (with lines of 1 pixel long). Experience shows that this extreme refinement is not needed more than on 1 or 2 sprites per ship for the most complex shapes, which is enough to ensure good performance of the algo.
But I cannot guarantee it's not possible to design a ship that would request such refined polygons that it would make the algo unusable.

I'd prefer adding a warning.

PR #947, opened 2020-03-04
added more commodities to be mined in asteroids

nenau-again:
This commit adds 3 types of asteroids containing water, gold and diamonds, and makes the asteroid scanner available to purchase outside of Bon Sebb.
For now, only water asteroids are integrated, with the commodity being sellable in various places (mainly stations and desert planets), but I plan to integrate diamonds and gold as well.


onpon4:
Mining a substance as plentiful as water from asteroids seems a bit silly doesn't it? 😉

Also worth noting, when you make a commodity available somewhere that enables you to *buy* the commodity and makes it able to ship the commodity elsewhere. So it seems if you're going to have water as a commodity it should be available on water planets and not desert planets (i.e. the way I understand commodity definitions on planets is that the planets *produce* the commodities they're defined with).

nenau-again:
Well, water is plentiful only on a minority of planets. I would not say that it's a abundant resource in the universe. And for gravity reasons, I believe that it is much more rational to take water from small bodies (like comets or asteroids) than from planets...

> implicitly assuming that all commodities can be sold anywhere unless specifically excluded.
Yep, I believe it's a good idea

Actually I had proposed these water asteroids one only to give an interest to the asteroid scanner. If people believe it's better for gold or diamonds, I can change...

BTAxis:
Also, I'd argue that even on Earth-like planets water might not be an export product. It's important enough to not risk upsetting the balance of the ecosystem by removing it from the planet in such quantities. I'd say only ocean planets would export it. So, mining water from asteroids seems pretty sensible to me.

onpon4:
> Well, water is plentiful only on a minority of planets. I would not say that it's a abundant resource in the universe. And for gravity reasons, I believe that it is much more rational to take water from small bodies (like comets or asteroids) than from planets...

Well think of it this way: Earth has a basically unlimited supply of water in its oceans that can simply be distilled to produce something drinkable. Also, water is a renewable resource, so you don't need to keep buying and buying it.

And the fact of the matter is that in the real world, water exists pretty much everywhere. Of course on Earth it's especially abundant (as well as many other bodies, like Europa). But levels of water are very common in the universe since the elements it's composed of are so common. I can't see any realistic economy on exporting gigantic amounts of water.

> Also, I'd argue that even on Earth-like planets water might not be an export product. It's important enough to not risk upsetting the balance of the ecosystem by removing it from the planet in such quantities. I'd say only ocean planets would export it. So, mining water from asteroids seems pretty sensible to me.

Haha, I think you overestimate how ecologically responsible humans are. :stuck_out_tongue: Just look all throughout history, people just dump toxic waste into lakes and rivers without a second thought.

bobbens:
Mining water makes sense imho. You would actually be mining "ice" and then liquifying it so that it is more compact.

onpon4:
I'm conceding the point. :laughing:

In any case this needs to be modified to have water produced by planets that *do* have enough water to sell, not those that don't. (And I would still argue that an Earth-like planet would do so, ecological responsibility be damned. :stuck_out_tongue:)

Maybe sometime soon I could add that thing I proposed for making commodities sellable on all planets generally.

nenau-again:
> Maybe sometime soon I could add that thing I proposed for making commodities sellable on all planets generally.

All right, and then I'll propose a new version of this one.

onpon4:
Just to note, the changes you'd need would just be to re-assign commodity tags to producers (not consumers), which is already consistent with the rest of Naev. It's not dependent on the system I suggested in any way. So I don't understand why you want to wait for me on this one.

nenau-again:
Because I need places to sell the gathered water, and selling to ocean worlds only seems really... odd.

bobbens:
OK, I want to merge this, but we need more places to be able to sell this and not have deadweights. Could this be incorporated with the current economy system?

nenau-again:
I guess one would have to add the `planet_modifier` tags for the new commodities, and make water be basically buyable/sellable anywhere there is a commodities market.
Maybe one could make a python script for that purpose, but I won't be really available to code stuff until the end of this month.

An other approach could be to assume that some commodities (water, food, industrial goods, medics) are buyable and sellable anywhere...

onpon4:
> An other approach could be to assume that some commodities (water, food, industrial goods, medics) are buyable and sellable anywhere...

We probably ultimately want to do that with all commodities.

nenau-again:
Only with standard commodities (not mission stuff like "four winds informant" or "The Baron's holopainting") That could be a tag in the commodity definition.

onpon4:
Yes, of course.

bobbens:
Do non-standard commodities actually have to be commodities? We could just add an API to add mission cargo and specify the weight, just having the name be whatever you want. It would be saved with the player data and therefore wouldn't be treated as "actual cargo". AFAIK it wouldn't even need an icon.

onpon4:
Yeah, that does sound reasonable to me.

bobbens:
Can you either redo the PR or merge with latest trunk? Now that all assets are available everywhere, we can have more asteroid commodities!

nenau-again:
Yep, I'll redo it by hand.

bobbens:
Closing as implemented in #1110 .

PR #951, opened 2020-03-08
Add new Thurion ships.

V-ktor:
Add new Thurion destroyer and carrier class ships.


ids1024:
Cool!

@V-ktor Could you PR the models to [naev-artwork](https://github.com/naev/naev-artwork)?

Issue #954, opened 2020-03-20
Mission "Sirian Truce" broken...


HansMKern:
Warning: [misn_runFunc] Mission 'Sirian Truce' -> 'jumpin': [string "dat/missions/sirius/achack/achack04.lua"]:187: attempt to call local '_' (a number value)
stack traceback:
	[string "dat/missions/sirius/achack/achack04.lua"]:187: in function <[string "dat/missions/sirius/achack/achack04.lua"]:180>
Warning: [hook_runMisn] Hook [jumpin] '1164' -> 'jumpin' failed

nenau-again:
Should be solved once https://github.com/naev/naev/pull/955 is merged (please complain again if not)

PR #955, opened 2020-03-20
corrected an other underline bug

nenau-again:
To solve https://github.com/naev/naev/issues/954 (that old _ issue).


ids1024:
Thanks! I wonder if having a function in Lua called `_` is not a great idea...

PR #956, opened 2020-03-20
Dogfight helper

nenau-again:
I propose a system that aims at making dogfight easier for the player.

This displays two lines : the green one is the orientation of the player's ship and the red one is the direction the player should aim at to hit the target with the current weaponset.
I have trained a bit with this system and it turns out it help in some cases (however with keyboard I can't aim very well. Maybe if someone uses joystick...)

Please tell me what you do think of the concept, and if you want, try it out in order to decide if it's something we want to add to the game.

If people approve it, I'll have more work to integrate it better :
* Make it an option with a box to cross somewhere
* Make it disappear once the target is too far away
* I've written a gl function for drawing lines, I'd like to make `gui_renderRadarOutOfRange` use it.


src/opengl_render.c
```diff
@@ -665,6 +665,38 @@ void gl_drawCircle( const double cx, const double cy,
...
+   gl_Matrix4 projection;
+   gl_vbo *my_vbo;
+
+   projection = gl_view_matrix;
+   vertex[0] = x1;
+   vertex[1] = y1;
+   vertex[2] = x2;
+   vertex[3] = y2;
+
+   my_vbo = gl_vboCreateStatic( sizeof(GLfloat) * 4, vertex );
```
ids1024: It's probably better to create a global VBO and modify it here; I think creating and destroying VBOs regularly isn't very efficient.

src/player.c
```diff
@@ -1010,8 +1017,33 @@ void player_render( double dt )
     * Render the player.
     */
    if ((player.p != NULL) && !player_isFlag(PLAYER_CREATING) &&
-         !pilot_isFlag( player.p, PILOT_INVISIBLE))
+         !pilot_isFlag( player.p, PILOT_INVISIBLE)) {
+
+      /* Render the aiming lines. */
+      a = player.p->solid->dir;
+      r = 200.;
```
ids1024: I can't think of why at the moment, but the line seems to get shorter the closer the camera is zoomed in (when it should get longer, or stay the same). You can see this clearly if you enable "manual zoom control".

onpon4:
I would note that difficulty or inability to distinguish green and red is the most common kind of colorblindness, so this system would be difficult to use for a lot of people.

Though in truth I can't say I fully understand what problem this is trying to solve. For most weapons some kind of auto-aiming (swivel) is used anyway, so lining up shots with them would be pointless. The only exception is unguided rockets (because launchers don't have swivel). But if this is designed for that, wouldn't it be better to just make launchers swivel too? Not to mention, different weapons on a ship can easily have different speeds and therefore different places they should aim at, which just seems to me like another reason to facilitate auto-aiming for all weapons rather than trying to add indicators for where to manually aim.

nenau-again:
The problem this system is trying to solve is the fact that I believe that dogfight is quite tricky in Naev because we're not using subjective view. (Actually, I don't know if other people have the same feeling as myself about dogfight and that's part of the reason why I opened this).

Imho, the question of swivel for launcher is a separate issue as this proposal also works for cannon. I am actually using the same orientation the AI would use to aim (using a mean of cannon weapons velocities) to facilitate aiming for the player, because swivel is only limited. I didn't want to add an "autoaim" function because it was imo too much help.

I personally tend to only use turrets, or to shamelessly exploit the fact that the AI goes straight at enemy, which prevents me to use a fighter at its full potential, and I wanted to know if I was the only one. I opened this PR because I knew it would be more difficult to explain correctly what I had in mind to help people actually do dogfight, than to actually code it...

> I would note that difficulty or inability to distinguish green and red is the most common kind of colorblindness, so this system would be difficult to use for a lot of people.

Thanks for the info, I'll change the colours then.

onpon4:
> The problem this system is trying to solve is the fact that I believe that dogfight is quite tricky in Naev because we're not using subjective view. (Actually, I don't know if other people have the same feeling as myself about dogfight and that's part of the reason why I opened this).

I don't have any such problem. It's just like playing Asteroids or Star Control or some such thing.

If you have a problem with the view being oriented in a single direction all the time, wouldn't it make more sense to tackle that at the source and add support for a subjective view? All you would need to do in principle is rotate an enlarged version of the rendering area and causing it to choose different ship sprites/orientations that don't get rotated.

> I didn't want to add an "autoaim" function because it was imo too much help.

Just a note, we sort of have that already. It aims directly at the ship and not based on weapons, but by default the Q key with WASD layout (or the A key with arrows layout) generally works ok for that. I use it that way all the time.

> I personally tend to only use turrets, or to shamelessly exploit the fact that the AI goes straight at enemy, which prevents me to use a fighter at its full potential, and I wanted to know if I was the only one.

I think that's more a function of how good your reflexes are and how fast your ship is. I don't see how being told where to aim would help me in any way; if anything it would be a distraction and make the task harder, because exact aim isn't really so much of a concern (thanks to swivel). The main concern is aiming in that general direction fast, then pivoting to dodge the enemy bullet that's definitely there. Which I've done lots of times; it's easiest to do now with the Hyena since it slows down time. But not all players are going to be good at this, and it doesn't work very well with weapons that are designed to be rapid-fired, a problem that really has more to do with the fact that when you're dodging around you don't have much time to fire.

Of course this is all very subjective, though. It could be that allowing the view to make "up" the direction you're facing would be useful to some people (though for me it would be headache-inducing), so maybe experimenting with an option that does that could be worthwhile. I don't think it's going to make dogfighting any better, but it could be useful as an accessibility feature of sorts perhaps.

ids1024:
> It could be that allowing the view to make "up" the direction you're facing would be useful to some people (though for me it would be headache-inducing), so maybe experimenting with an option that does that could be worthwhile.

3D games with rapidly moving first person views tend to make me feel motion sick after a while, so I'm not really sure I could use it either. But it seems like an interesting thing to experiment with.

PhoenixRiver:
Part of the reason I don't tend to dogfight is that I am bad at aiming
and having used many "helper" systems before there is little to no
difference unless it is a true autoaim for me. This is why I run
turrets and ammo-less weapons in almost every game (at least where
they have them) that isn't first person...

>> It could be that allowing the view to make "up" the direction you're
>> facing would be useful to some people (though for me it would be
>> headache-inducing), so maybe experimenting with an option that does that
>> could be worthwhile.
>
> 3D games with rapidly moving first person views tend to make me feel motion
> sick after a while, so I'm not really sure I could use it either. But it
> seems like an interesting thing to experiment with.

...and I don't play fast FPS games because I also get nausia very
quickly with the except of space sims where you have true 3D motion
possible (possibly because my third space game (after asteroids and
arkanoid was Elite!) but even then I twist my body to compensate for
the g-forces I create by moving ingame!! :D

On top of that the 2D rotation solution would also likely cause me to
throw up pretty quickly; I have good periferal vision which I use a
lot so games where the edges of the screen slew a lot (e.g. fast 3D
FPSs, fast 2D top-down where the character sprite always faces up) -
it might also be due to a disconnect between my fingers inputting a
change of direction but the world shifting instead... I don't get the
same problem where I have to rotate the world (e.g. Black & White...
can't think of any I've played more recently that have this with
rotation... just scrolling which is different).

nenau-again:
I've addressed @ids1024 's issues and the few points I had put in the initial comment.
Addressed also the colour issue by using different shapes (you have to put a cross in a circle instead of aligning 2 lines)

onpon4:
Conflicts need to be resolved in this PR, but other than that I think this should be fine now? I don't see any outstanding issue that should make merging this a problem (other than the conflicts, those can be resolved easily enough).

🕷️

nenau-again:
I tested it for about 50 minutes and no crash happened.
It should be OK. Only task left is to inform the player that this exist...

onpon4:
If you want to test for memory leaks, compiling with `./configure DFLAGS="-fsanitize=address"` can be helpful for that.

🕵️

PR #958, opened 2020-03-26
warn when there is no polygon

nenau-again:
...And by the way, I added polygons for the last thurion sprites...


ids1024:
So, the code shouldn't be using sprite collisions for anything any more, right?

Perhaps the sprite collision code should be removed.

nenau-again:
It uses sprites collision for the asteroids. I've heard people speaking about changing the way asteroids will be rendered (with something more dynamic) so I don't yet want to implement much stuff with the system we have right now on asteroids...

ids1024:
Fair enough. I don't really like how the asteroids look currently, so it could be interesting to try procedural generating meshes for asteroids. It might not be too hard to get something that results in nice looking and unique asteroids.

PR #960, opened 2020-03-29
Za'lek space expansion

Lineth:
Added some additional content in Za'lek space.

I never got around to filling out all of Za'lek space. This is a first-wave PR to remedy that. Plus a spoopy Kont-from-EV-Nova expy to possibly act as a basis for future missions.


Lineth:
Yeah, the universe editor edited every system and asset in the game so I had to prune everything. Got a bit to snip-happy.

ids1024:
`Warning: [planet_parse] Planet 'HZNSS-3' missing 'faction' element`

That should be easy to address.

PR #964, opened 2020-04-01
Commodity tweaks and Za'lek Weapon tweaks.

Lineth:
Since I can't find a way to separate the two and the commodity stuff is really minor anyway.

I feel that the Za'lek's focus on beams makes them weak at long range, and their drone tech could be useful for making kamikaze 'missile' drones.

I added two powerful new Za'lek launchers, balanced after the Thurion launchers, as I see the two factions as being on similar footing tech-wise. The launchers are not available for purchase anywhere until the player completes some future missions for the Za'lek. The reaper gfx is a pure placeholder.  I've been fiddling around with the 3D artwork in the art repo, but nothing conclusive as of yet (Because Blender won't load any textures for some reason)


nenau-again:
Hey, you've even done the polygons ! Nice !

PR #967, opened 2020-04-05
Fixed color so that inert object text is legible again.

onpon4:
I made it a bit darker and also made the map neutral color darker so they're still distinct.

### DIFF
--- a/dat/faction.xml
+++ b/dat/faction.xml
@@ -22,6 +22,7 @@
   </allies>
   <enemies>
    <enemy>Pirate</enemy>
+   <enemy>Wanted Pirate</enemy>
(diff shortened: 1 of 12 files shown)


onpon4:
Other fixes pushed to this branch:

* Fixed spacing in some Za'lek launcher descriptions
* Fixed some texts for launchers and some ships not showing up
* Fixed grammar of Seek And Destroy mission
* Made Seek And Destroy mission consider making the player go further distances
* Made "alive" bounties use a custom faction (so that the pirates won't necessarily just be blasted away by police forces).

bobbens:
Looks good, thanks.

Issue #970, opened 2020-04-08
Pass SDL_WINDOW_ALLOW_HIGHDPI when creating the Window


ids1024:
When using Wayland on Linux (in particular, I'm testing with Sway), the compositor can have a per-display scale factor, for handling hiDPI displays. If this flag is not passed, on a 4k display Naev only gets a 1920x1080 drawable (which is automatically scaled up).

If this flag is enabled, `SDL_GetWindowSize` returns the size in "screen coordinates", which on a hiDPI display will be different from the actual size in pixels.

I believe this should work the same on MacOS devices with Retina displays.

Enabling this seems to cause a variety of small problems; or I would just have pushed a commit.

ids1024:
The main problem I seem to come across with this is mouse input. The mouse position is in "screen coordinates", which is half the actual pixels with a 2x scale set by Sway. It seems this may require changes in various parts of the toolkit.

But if this works, devices that support this feature will get hiDPI without any program-specific scale factor (though that can be used as well).

UncombedCoconut:
Note for posterity: According to []()

UncombedCoconut:
Another possible pitfall, depending on how user settings are translated to window dimensions / fullscreen display modes, is [this "ABANDONED" bug](https://bugzilla.libsdl.org/show_bug.cgi?id=3433). If we care whether SDL_DisplayModes' w/h are "screen coordinates" just like hidpi-aware SDL_CreateWindow (units of scale_factor pixels) or pixels, it can depend on the combination of drivers, platforms, and workarounds. SDL2 always seems to *intend* pixels, but on Windows that requires the workaround described in the issue and on Wayland it requires setting SDL_VIDEODRIVER=wayland.

This problem goes away if we opt to split split conf.width and conf.height into separate settings for window dimensions and fullscreen modes. Thinking through some user interactions (like toggling out of fullscreen, resizing the window with the mouse, and toggling back), I think that could make sense.

On a related note, I think having gl_screen.desktop_w and gl_screen.desktop_h is also a source of trouble. (Why should they be constant? What if you drag the game window between monitors?) If it's uncontroversial, I think it makes sense to remove those fields and check on the fly.

bobbens:
I believe this is solved ;)

PR #971, opened 2020-04-08
Pass SDL_WINDOW_ALLOW_HIGHDPI when creating window (#970)

ids1024:
Used on MacOS and (at least some) Wayland compositors.

Without this, the program will be rendered at a low DPI and scaled up.

#970

### DIFF
--- a/src/opengl.c
+++ b/src/opengl.c
@@ -388,13 +388,14 @@ static int gl_createWindow( unsigned int flags )
    /* Create the window. */
    gl_screen.window = SDL_CreateWindow( APPNAME,
          SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED,
-         SCREEN_W, SCREEN_H, flags | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE );
+         800, 600, flags | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE
+                                   | SDL_WINDOW_ALLOW_HIGHDPI );
    if (gl_screen.window == NULL)
       ERR(_("Unable to create window!"));
 
    /* Reinitialize resolution parameters. */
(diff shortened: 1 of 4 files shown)


src/opengl.c
```diff
@@ -388,13 +388,14 @@ static int gl_createWindow( unsigned int flags )
    /* Create the window. */
    gl_screen.window = SDL_CreateWindow( APPNAME,
          SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED,
-         SCREEN_W, SCREEN_H, flags | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE );
+         800, 600, flags | SDL_WINDOW_SHOWN | SDL_WINDOW_RESIZABLE
```
ids1024, line 391: I've marked this PR as a draft because, even if there are no other issues, I'm not happy with this part.

The problem is the width and height passed here are in "screen coordinates", not pixels. So it ended up trying to create an 8k window (since these are multiplied by the scale factor), and generate a nebula at that resolution...

bobbens: Other than that, does it work well? It should be possible to define the nebula resolution somewhere else instead of just using the screen resolution.

Another issue is that font will probably look really bad unless we use some sort of higher resolution rendering.

ids1024: Yeah, it should work other than that.

> Another issue is that font will probably look really bad unless we use some sort of higher resolution rendering.

I think that's already a problem with the current font rendering, when a scale factor is set. I don't know if this really changes that.

My "plan" is to reimplement font rendering using [signed distance fields](https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf), which should scale well without performance issues while also allowing things like outlines to be added in the fragment shader, but I haven't made too much progress with that.

bobbens:
Should be solved with @UncombedCoconut 's magic ;)

PR #973, opened 2020-04-11
Added next "Coming Out" mission.

onpon4:
This mission involves the player assisting Chelsea in an effort to kill
a pirate.


onpon4:
Reordered mission.xml so that the mission will show up (and since I was doing that, I just reordered the entire thing to make more sense).

PR #980, opened 2020-04-14
Improved start-of-game tutorial, replacing the menu tutorial

onpon4:
I decided to do this because the classic tutorial was quite long and had some problems with inconsistency with the rest of Naev as it currently is, and yet the intro text I had previously added was inadequate. This new tutorial is designed to be a nice middle-ground.

As before, the tutorial is optional, offered to you at the start of the game. Unlike before, you go through an entire mission demonstrating all the important game mechanics, rather than just having a bunch of information thrown at you.

I also decided to give the tutorial some character by making the one giving you the tutorial obnoxious with marketing of Melendez Corporation (the manufacturer of the Llama, your first ship).

This makes the old tutorials accessible from the main menu (which I haven't recommended to anyone in years) obsolete, so the "Tutorial" button and old tutorial mission have been removed. Incidentally, this removes the last remaining instance of Naev assuming that the player is male.

Please let me know if I missed any important information in this new tutorial that should be covered. I think I covered everything that matters, but I'm only human and may have overlooked something.


amylaar:
When I came back to naev after a long pause, I found myself lost doing the disabling missions.  How exactly did that work.  There
was a tutorial for it...  I coudn't find the tutorial in the menu.  I started a new game, but the information never appeared.  I ended
starting naev 0.7, it it was perfect with tutorial from the menu and I could pick the lesson that I wanted.

I would suggest to
- add info on disabling in the new tutorial
- log some quick reference text for each lessen performed in the tutorial.  Could go under under 'tutorial' or 'Melendez Shipyards' / 'Melendez Corporation' .
- For disabling, the main point is, if the blue part of the armour is not growing, you're doing it wrong.  You must penetrate or eliminate the shield, then you need weapons with high disable/dps ratio, mostly ion cannons or turrets.  If what remains of the armour is all blue, stop shooting and start boarding.
- be able to re-take (or take it the first time if you skipped it and found you really shoudn't have) the tutorial at any shipyard that sells Melendez ships

PR #988, opened 2020-04-18
Change in .gitignore : ignore datapath.lua and nave-data folder

Mutos:
Make it so that, if someone wants to use a naev-data folder with a datapath.lua file, it is ignored by Git.

Signed-off-by: Benoît 'Mutos' Robin <benoit.robin@hoshikaze.net>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ src/shaders_c_gen
 
 /.cproject
 /.project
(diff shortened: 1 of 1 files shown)


BTAxis:
I wouldn't assume people would always call it naev-data. Mine is called .naev.

PR #989, opened 2020-04-18
Fix bin folder and add naev-data and datapath.lua on .gitignore

Mutos:
Add naev-data and datapath.lua on .gitignore

Signed-off-by: Benoît 'Mutos' Robin <benoit.robin@hoshikaze.net>


PhoenixRiver:
> Add naev-data and datapath.lua on .gitignore

Um... the actual commit has a typo for naev-data (it's written "naev-dta")!

PR #990, opened 2020-04-19
Bugfix mapedit

Mutos:
Fixed excessive length of nsnprintf on map's name, filename and description.


bobbens:
Thanks for the patch! However, I opted to rewrite most of the code to be cleaner and more flexible. Please look at f9c9c464615468b4bad1fff4963350071400d1f6 and 8454fd85b816354dd2326deed87d476e09c1756e . It should not crash anymore and be able to handle arbitrary lengths, although the mapeditor input windows themselves are limited.

Mutos:
Thanks bobbens ! As an occasional dev only, I can only do simple things, and that is assuming I got proper time ^-^

bobbens:
Figuring out where the problem is is 90% of the fight ;)

PR #992, opened 2020-04-24
Implemented a system for bioships

onpon4:
This PR is a variant of the idea @bobbens talked about on Discord a while back, with Soromid ships that upgrade. I did some thinking and found that it would be possible to implement a "leveling up" mechanic simply using the existing outfits system and some Lua code, and that's what I did. :slightly_smiling_face:

Here's how it works: we now have "bio outfits" which work the same as regular outfits, but have separate slot definitions. In the case of core systems, Soromid bioships can *only* equip biocores: Brains (equivalent to Core Systems), Shells (equivalent to Hulls), and Fins (equivalent to Engines). Said biocores start out very weak, around the same as similar low-end Unicorp synthetic cores, but as they level up, they end up in a slightly better place than similar high-end synthetic cores. There's one catch: CPU capacity of fully leveled up brains is much lower, about half the CPU capacity of similar synthetic Core Systems. This is designed to encourage the use of organic weaponry, which will be less demanding on CPU than synthetic outfits, while still allowing Soromid bioships to equip standard synthetic weaponry (it's just that they can't equip as much due to the lower CPU capacity of bioship brains).

Leveling up is done on a per-component level and handled by a new event called "Bioship Upgrade", which always runs. This event catches whenever the player is payed with the player.pay() lua function, and when that happens while player.pilot() is a valid bioship, it adds a certain number of "experience points" (that is the amount paid / 10,000). When experience points reach 100, a random bio component is upgraded to the next "stage". Most bioship components have anywhere from 2-7 normal "stages", and each component has one final stage after that, which is called "Stage X": the fully upgraded stage. This is all defined in a simple way in the Bioship Upgrade event, so any new bio-outfits can be added very easily.

Of course, it would be possible for a station to sell leveled-up bioship outfits. But since that would defeat the purpose of them, stations only sell Stage 1 bio-outfits (with the exception of Bon Sebb, which also sells Stage X outfits).

Currently, as of this pull request, bio-cores (brains, shells, and fins) have been fully implemented, as well as one organic weapon: the BioPlasma Organ, which (in Stage X) is a slightly weaker but longer-range version of the Plasma Blaster MK3. I have also defined the Soromid Brigand to use the BioPlasma Organ, and all Soromid ships to use the appropriate biocores.

More weapons will have to be implemented later, and ideally bio-outfits should have separate images. But for now, the bio-outfit system is implemented and new bio-outfits can be added at leisure. :slightly_smiling_face:


ids1024:
> This event catches whenever the player is payed with the player.pay() lua function, and when that happens while player.pilot() is a valid bioship, it adds a certain number of "experience points" (that is the amount paid / 10,000). When experience points reach 100, a random bio component is upgraded to the next "stage". 

It seems like this mechanism is rather unintuitive, so it could seem kind of random to the player. In some ways, it may be desirable, if it ultimately results in a feeling that the ship just grows organically over time.

It may, however, result in players being confused, and wondering what they should do to advance their outfits, why they don't seem to be advancing, etc.

It's hard to say whether the result is satisfactory without a fair amount of testings. Like all major features, I guess.

> Of course, it would be possible for a station to sell leveled-up bioship outfits. But since that would defeat the purpose of them, stations only sell Stage 1 bio-outfits

I guess this also makes the higher level ones impossible to sell, except by putting them in a ship and selling it? I don't know if it matters a huge amount; that's just an interesting thing to note.

onpon4:
Yes, you indeed can't sell them. I see this as a systemic problem in Naev's shop system itself, not specific to the bioship system. I've thought for quite some time that it would be more ideal for you to be able to sell any outfit anywhere that an outfitter exists, but it's not the system we have right now.

I just want to note that this PR also takes away access to Soromid ships by players, since it now requires a higher standing with the Soromid than is attainable without the (as of yet undeveloped) Soromid campaign. So as far as regular players are concerned, this won't actually affect anything immediately and thorough testing can be done before actual access to Soromid ships by players happens.

> It may, however, result in players being confused, and wondering what they should do to advance their outfits, why they don't seem to be advancing, etc.

That should be solvable by proper text in the Soromid campaign when we start working on that (since that's also when Soromid ships will actually be obtainable by players). As said, this PR isn't trying to get everything perfect in one go. It's a step-by-step process, and getting the system in place is the first step. The next step will be expanding the available organic weapons, and after that will be actually introducing the system (and the ships themselves) to the player.

nenau-again:
This sounds cool. I have a few remarks:

* Using player.pay hook is rather strange, and there could be exploits, like running an entire mission with one setting and then equipping all the biostuff just before the payment. Couldn't it be better to use inflight hooks like attacking hooks that upgrade weapons and armor, jump hook that upgrades engine (and for systems, I don't know...)

* About what you've done with bio-weapons being less CPU-consuming than other weapons. Does it mean that a regular ship can equip bio-weapons ? Isn't it possible to address the issue by using special slots for bioweapons and other for normal weapons ?

* In your system, experience does stack linearly (every 100 points, there is an upgrade). What about experience being logarithmic (first upgrade costs 10 points, next one 20, then 40 and so) ? That's the system that is often used in rpg games afaik.

onpon4:
> Using player.pay hook is rather strange, and there could be exploits, like running an entire mission with one setting and then equipping all the biostuff just before the payment. Couldn't it be better to use inflight hooks like attacking hooks that upgrade weapons and armor, jump hook that upgrades engine (and for systems, I don't know...)

I discussed this with @bobbens some time ago in Discord. He doesn't want to use combat because he doesn't want Naev to reward combat too much. Hooking the player being paid was the solution he was most happy with.

As for the possible exploit, I am aware. This is unavoidable though, unless we directly link experience gain to combat, which bobbens was against.

> About what you've done with bio-weapons being less CPU-consuming than other weapons. Does it mean that a regular ship can equip bio-weapons ? Isn't it possible to address the issue by using special slots for bioweapons and other for normal weapons ?

No, it doesn't. There's a separate non-exclusive "bio_weapon" weapon slot property used for organic weapons; this means that bio-weapons can only be equipped to "bio_weapon" slots, but regular (synthetic) weapons can be equipped to bio_weapon slots as well. I did forget to assign that property to the BioPlasma Organ, though, so I've fixed this now.

> In your system, experience does stack linearly (every 100 points, there is an upgrade). What about experience being logarithmic (first upgrade costs 10 points, next one 20, then 40 and so) ? That's the system that is often used in rpg games afaik.

Yes, but Naev isn't an RPG and with this system, making upgrades logarithmic would penalize the player for having upgraded the ship in the past. Remember that even assuming we change the script to be ship-specific, ships can have components swapped in and out. The amount of payment a player gets also doesn't increase logarithmically as you progress in the game, so that would amount to the player having to do more and more grunt missions. Imagine having to do hundreds, maybe thousands, of missions paying you millions of credits because you've had a ship around for a while. Any competent player would just exchange the ship for a new identical one and start over (which would also make this kind of system completely pointless).

bobbens:
Some of the options we talked about on discord:
1. **Gaining exp from time passing**: either easily exploited by running around with time compression, or too slow and thus doesn't really make sense to implement
2. **Gaining exp from combat**: this sort of conflicts with the fact that I don't think it's good to make combat the central role in Naev (even though it is very important now), having more options for non-combat stuff is good.
3. **Gaining exp from payments** (this PR): this can be seen basically as a proxy for progress, given that you get payments only when you do missions. It would probably be better to have each mission give you exp of certain types, but this is a lower maintenance compromise.
4. **Progress through missions** (new!): just thought of this now, but it would also be possible to have different triggers to have your ship improve/evolve. For example, meeting some requirements like disabling/destroying a ship bigger than you (for weapons), to basically doing lots of cargo delivery quests (engines), or even specific quests just for these purposes. This is a lot more difficult to implement, but could be the most flavorful way of doing it.

I think that using payment is probably the easiest way to get it implemented and working now, with long-term solution being more mission and particular feat-locked being a long-term goal. Thoughts?

onpon4:
I don't know if 4 is really going to improve much, but it would certainly be feasible by having a second optional argument to player.pay() indicating an argument to pass to the pay hook, with some special handling of those strings by the upgrade script. Actually this would be pretty simple to do if that's the direction we want to go. (It would also ensure there are no weird cases in the future of missions giving you payment before completing the mission and that counting as exp; we don't have those now but if we do later, making the call to player.pay() explicitly say that it should grant experience of type X would prevent any weirdness or exploits from such missions.)

So I agree with your thought on the way to move forward. :+1:

PhoenixRiver:
I feel like both the mission progression system and the payment system
would work well together - with the payment system providing only
modest amounts of advancements with mission advancements profiding
more defined, targetted advancements to particular types of system or
indeed only specific items.

I feel like this bioship concept is one though that should require
more buyin by the player given the final result is better than noraml
(though this would require this to be made more clear in explaining
what the bioships are as well but would also allow them to be
introduced later in the game once the standard systems are better
understood by the player)... should one want a bioship then it might
be possible to add systems that are not bioship systems but to add
bioship components to a non-bioship should not necessarily possible
(as with the weapon systems)... maybe an extra component would be
necessary for non-bioships to host boship components (i.e. lifesupport
or such) since a bioship hull or core would normally provide such
things (maybe that would explain the lower processing power of bioship
cores).

As an idea that might expand on this maybe it would be possible to
slowly evolve bioship hull into different ones allowing the player to
grow or shrink their ship over time to "design" it more into whatever
they feel is he type of play they prefer - this might be possible for
other systems too... maybe there could only be one ship hull, core,
weapon, generic component that could be bought and the player would
need to evolve the type of ship they wanted by specifying a larger or
smaller hull, a more efficient or larger core, etc. (I realise weapon
systems would be more involved in this repesct). This is all rather
pie in the sky though... maybe too much for the Soromid... maybe more
appropriate for a Zalek experimental ship or collective type ship?
(With their current ships simply being the result of long term use).
Anyway...

As for specific notes to the code itself, I was wondering if your
random upgrade system took into account systems that had already
upgraded to the max and, if it was randomly picked, whether a reroll
was made (or are such fully upgraded systems removed from the pool of
potentially upgradable systems?)

River

onpon4:
> I feel like this bioship concept is one though that should require
> more buyin by the player given the final result is better than noraml

Bioship cores aren't *better*. They have substantially less CPU capacity than synthetic cores, which limits weapon and utility outfitting options. That's why bioships can *only* equip bio-cores: a bioship with a synthetic core would actually be wildly overpowered because bio-weapons require little CPU, so the CPU limitation would become basically meaningless.

It's true that the final result is marginally better than synthetic cores as far as other attributes go, but it's not a huge difference. Shells have a bit more cargo than the S&K combat plating (but not even close to what cargo hulls have). Brains get the slight increase in stats from both the Milspec Prometheus and Aegis core systems, but have substantially less CPU (limiting weapon options). Fins end up in the same place as the Tricon and Melendez engines (in fact the stats are exactly the same between Stage X fins and similar synthetic engines).

So just to be clear, allowing synthetic cores would make bioships *more* powerful, not less.

As for costs, they're not that much lower, but they're a bit lower to make up for the fact that the components have to slowly level up.

> should one want a bioship then it might
> be possible to add systems that are not bioship systems but to add
> bioship components to a non-bioship should not necessarily possible
> (as with the weapon systems)... maybe an extra component would be
> necessary for non-bioships to host boship components (i.e. lifesupport
> or such) since a bioship hull or core would normally provide such
> things (maybe that would explain the lower processing power of bioship
> cores).

Giving synthetic ships the pluses and minuses of being bioships would make Soromid ships far less special.

> As an idea that might expand on this maybe it would be possible to
> slowly evolve bioship hull into different ones allowing the player to
> grow or shrink their ship over time to "design" it more into whatever
> they feel is he type of play they prefer - this might be possible for
> other systems too... maybe there could only be one ship hull, core,
> weapon, generic component that could be bought and the player would
> need to evolve the type of ship they wanted by specifying a larger or
> smaller hull, a more efficient or larger core, etc. (I realise weapon
> systems would be more involved in this repesct).

That would require an entire RPG level-up and stats system, for each ship and/or outfit, and all the balancing that would require. I think that's a bit too much of an undertaking. I don't see the sense in building such a massive system for such a minor and unimportant part of the game (which many players won't even see). Plus, it would be far too different from the experience everywhere else in Naev; it would be like teleporting from one game into another game.

> As for specific notes to the code itself, I was wondering if your
> random upgrade system took into account systems that had already
> upgraded to the max and, if it was randomly picked, whether a reroll
> was made (or are such fully upgraded systems removed from the pool of
> potentially upgradable systems?)

I'm not sure what you mean by this, but the upgrading takes place via a simple Lua event, in events/bioship_upgrade.lua, so you can read that file if you want to learn more how upgrading works. If you're asking if I made a mistake that would cause the outfits to go from Stage X to Stage 1, no, I didn't.

bobbens:
Looks good, thanks! I think this is a first good step to having more interesting Soromid ships. Possible vectors for improvement are:
1. Make special requirements / events to level up specific items.
2. Add a mini help-message when buying first organic ship (tutorial-ish).
3. Add more lore / dialogues / news for Soromid ships.
4. Organic-ship oriented missions, etc.
5. Possibly add more complicated RPG-like system just for Soromid ships.

bobbens:
Added issue #994 to continue discussion / improve the current system.

Issue #936, opened 2020-02-19
Add Map Decorators


bobbens:
We have an infrastructure for since 0447e6da57a7f44e10a4e676032b781ad2bf5ec1 , however, there is only a simple one being use for the Empire. More map decorators should be created and used. In particular the following would be nice:

1. Nebula
2. All main faction areas
3. Dense asteroid area (with all the related systems having lots of asteroids)

It should be also possible to make thematic areas to add more flavour. Like high pirate density areas and stuff. This would give the universe a better and more natural feeling. Adding them is simple:

1. Add the new graphics to `dat/gfx/map/`
1. Create a new entry in `dat/map.xml`

bobbens:
The original decorator was done by @Kinniken in #685 as a placeholder. I think it wouldn't be an issue to remake it in a new style with whatever font you want given that it is just an example of what can be done.

bobbens:
Closed with #996

PR #996, opened 2020-05-01
added decorators for the major factions

nenau-again:
As requested in https://github.com/naev/naev/issues/936

The decorators consist in ships and logos of the faction. And I reduced the detection radius as I thought it was more natural.

My main concern is about  the readability. Please tell me if you think I should darken the pictures.

If people agree with the concept, I will add smaller ones for interesting places (goddard, krain factory, piracy spots, traderoutes, nebulae...)
For collective, I'm unsure : is it possible to remove the picture with an unidiff once the faction has been beaten ?


bobbens:
Thanks! I looked at it and have a few comments. First, it might be better to make the decorators more simple and larger to compensate, so that when you zoom in you don't notice them too much, only having them appear when scaled out (we could also just hide them when zooming in). Second, I have a feeling they are too bright. Might be worth darkening them (or making them 50% transparent).

> If people agree with the concept, I will add smaller ones for interesting places (goddard, krain factory, piracy spots, traderoutes, nebulae...)

That would be cool, however, I'm not sure if all the small things would necessarily need a decorator. In particular, I think the nebulae would be most important to have a really nice decorator for.

> For collective, I'm unsure : is it possible to remove the picture with an unidiff once the faction has been beaten ?

It should be possible to add this functionality, but given that they are a few systems, I'm not sure to what point it warrants it's own decorator.

Basically one of the design questions I think that is still open is what function these decorators should be. If it is just showing the macro-layout of the universe, then major factions+nebulae should probably be enough, but if it is just some fancier eye-candy, it can be done at lower levels. However, this detailed eye-candy might also be implementable using shaders or something so that when you zoom in you can see if a system has asteroids and stuff by showing little moving things. Thoughts?

nenau-again:
New version of map decorators. I did not tackle the issue of nebula decorator since I believe this one is more subtile (as there is no obvious logo nor ship (except for the ghost ships, but it would spoil to show how they look on the map)).
I don't really understand the idea that decorators should show asteroids as asteroids are per-system stuff. There is no such thing as multi-system asteroid field.
But maybe showing little moving stuff could be cool some way. However I am pretty sure there is nothing in space that moves fast enough to be considered as moving at such a great scale. Maybe flares ?

bobbens:
Looks much better, thanks!

PR #997, opened 2020-05-01
New core system icons

V-ktor:
Add new icons for the specialized core systems.


nenau-again:
Hey, wait... Does it mean each single core outfit has its own gfx now ?
If yes, that's awesome !

V-ktor:
I think each core outfit has a unique icon now, except engines.
There is also an icon for each biocore outfit. There should be another one for the last stage, right?

onpon4:
Yeah, I would say that stage X bioship parts should have separate images from the earlier stage bioship parts. But this is great! 😄

bobbens:
Looks great! Thanks.

PR #998, opened 2020-05-03
added warning for polygon length

nenau-again:
It could avoid troublesome segfaults to contributors.
And I allowed myself to add an entry in changelog about the collision system.


ids1024:
Yep, warning on this seems like a good idea!

Ideally, it should be impossible to trigger a segfault just through what's in ndata, but that's not too important, as long as it's clear what the problem is and is easy to fix.

PR #1000, opened 2020-05-21
Starmap

Mutos:
Added back utils/starmap that had been deleted long ago, and enhanced it  with easier to read color settings.

Two points of notice :

- We probably should first investigate about why it had been deleted and whether it allows us to reintegrate it. I made a simple search in Issues and PR and didn't find any clue.

- The current starmap has a few SoC-specific lines in it, to display systems containing asteroids. It uses my old way to make asteroids fields appear in the map by using assets named "Asteroids Cluster xxxxxx". This was before the new asteroids anchors system. It does nothing adverse when reading NAEV data. I'm currently converting my old asteroids-assets into regular asteroids anchors and I'll remove these SoC-specific lines afterwards and present a follow-up PR.


nenau-again:
Yay, PR#1000 !
For ignorant people like me, could you explain what this script does ?

Mutos:
Hi nenau,

Didn't even notice this was PR#1000 ^-^ thanks for signalling !

It outputs an SVG map of the current universe, see the following examples, converted to PNG to be uploaded on Github :

NAEV :
starmap svg

Stars of Call :
starmap svg

nenau-again:
Mm. Could be useful indeed to have a way to export universe in a clean svg format. Btw, it looks buggy for exit-only jumps (like in Zied).

bobbens:
Looks a bit buggy (some edges are weird like that long one from sioux), but I guess it's a bit better than nothing. If possible some more usability would be nice (--help and allowing running from root directory, readme, etc). However, I think it is better to have than not so merged. Thanks!

Mutos:
Hello bobbens,

Thanks for the appreciation ! I just took it from its previous instalment and enhanced it for better readability, didn't debug it in depth.

Issue #849, opened 2019-08-24
Trade is sorely missing


Mutos:
There is a sore miss in NAEV, player-wise : trading. The commodities exist and are displayed on planets, but they are at the same price everywhere.

I know the commerce and price computing model isn't officially fixed and there are a number of ways it could be designed, so for now, there is no way to make a "default" consistent model.

But the new player discovering NAEV doesn't know of that. They come to an EV-like game, and their basic expectation for such a game includes to be able to make money by trading stuff. So they land on a planet, buy some commodities and fly to another planet they feel fit to sell it. And bang, at that point, they discover that, wherever they land, it's always the same price.

That definitely can throw potentially interested players out of NAEV. It's about the very definition of the NAEV gameplay. We DO PROMISE trade, so delivering it is NOT AN OPTION, while honing it the way we think it should be under the hood DEFINITELY IS.  

So IMHO, it should be a highest-level priority to have at least a tentative, simple model : at least, something  that makes prices be different on different planets...

It could be done by making available a LUA price API on planets and systems, then implementing on LUA the simplest possible model. In the meantime, we could keep on arguing on what should be an ideal model, and having the LUA price API could even help us to prototype it. Multiple prototypes could be implemented and tested, helping us to make a choice. 

Or even further, if the LUA model is not too CPU-expensive, the system could stay like that in the long run, with a choice of multiple pricing system in the options and a default one corresponding to the one chosen by the devs... But that would require a little more effort...

bobbens:
The main issue is it is not clear what model to use at all. Something simple would be do like what original EV did and basically hardcode the prices on all the planets, then use some events to basically move things up and down locally. However, this is not dynamic, and once the player figures out an "optimal trade route" it is fairly grind-enabling. Furthermore, my main worry is that pushing all the economy design to the developers is not something that scales well and also needs balancing. However, it would be fairly simple to implement this. Prices would be assigned to each commodity, and the planet editor would allow setting that. Event API could be added to read/write prices and some common event framework would be written to basically handle the entire price smudging. Should also be fairly straight-forward to integrate with the news so the player can get an idea of what to do.

Other options were a whitepaper I found like 10+ years ago on simulating the economy as an electrical grid. Intensity is amount of trade, voltage is prices, planets and stocks act as capacitors, and trade routes/piracy are resistances. While the idea was fairly interesting, in my intents on implementing it, in never worked out quite too well. I'm still not certain if my implementation sucked or if the idea just does not work.

[...]

PR #1002, opened 2020-05-29
Working pseudo economy model

agb32:
Pull request for a working pseudo economy model.
Upon initialisation, prices are calculated depending on planet/asset attributes, and on the solar system attributes, and then smoothed by neighbouring systems.  A variation scale and time period is also calculated.

Upon landing at a system, and visiting the commodity tab, the current cost of the commodity is calculated (a very simple operation).

This is a deterministic model, with prices varying sinusodally with time.

Benefits: It is now possible to make money by trading.  And the variation in prices gives some feeling of a working economy.

The next steps will be to add some Lua hooks, and some of the other items discussed in issue #849.  However, I am taking small steps in case available times runs out (i.e. COVID19 restrictions are lifted!)


nenau-again:
Hey, that looks interesting ! (however, I guess we should not try to merge such things before the next release, which is supposed to be our next big task).

I've never looked close at this economy problematic, but afaik, there have been previous attempts that did not end up so well... @onpon4 knows more than me about this.

Anyway, what you've got looks like a good start that could imho be tested and then gradually improved. Here is what I would propose to do at mid-term :
* replace the sinusoidal stuff by a wiener process
* Add some way not to have to hardcode the caracteristics of the planets and of the commodities
* Replace the averaging process by a sparse linalg operation (with impedance) @bobbens told me that he was not satisfied with his own attempt for that, but I don't know if the issue is that he has a problem to solve the linalg problem, or if the process works well, but the result is not good looking.

agb32:
Thanks - good ideas.  So, my suggestion would be:
Merge it as it is for the next release - so that people can finally do trade.  Since its entirely bound and deterministic, we already know that the values won't get stupid...

Then start work on a Wiener process, impedance, etc, to give plenty of time for testing until the next release.

The risk is that if we try to do too much, it will never get done - so a "keep it simple stupid" inclusion in the next release would be a great start.

What do others think?

Another thought for later - since currently, its entirely a function of time, it would be possible to have people in the bar who provide information such as "I was on xxx planet nnn STP ago, and the price of yyy then was zzz" where planet xxx would be randomly chosen to be 1-3 jumps away, nnn would be a typical time to travel in a reasonably fast ship, and the commodity (yyy) chosen would be the one that gives the best possible trade (or a good trade) xxx jumps away.  (easy to code - I have it in Python, so simple to convert to Lua).

It would also be possible, by saving the time at which a player lands at planets (so 8 additional bytes per planet visited), to show the average price of a commodity as seen by a traveller, the average price at a given planet, or even a plot of price against time at a given planet...

Might be trickier to do with an impedance based network though if the matrix not invertible (i.e. if we can't get back to a previous time deterministically from the current time).

So, what do people think?  Merge now in time for the release?   (or go without trade for another N years in the release!)

agb32:
I should add - the compute requirements for this merge request are minimal - some calculation upon startup, no calculation in flight, and negligible calculation when landed - basically evaluation of a sine function whenever a commodity is viewed.
Memory requirements are also small: 60 bytes per commodity on each planet, 12 bytes per planet, and 12 bytes extra per system (plus 60 bytes per system during initialisation which are then freed).
So little to worry about it slowing down older computers.

onpon4:
The economy system I worked on was a simplistic and static one made by bariumblue, and that economy system led to a situation where some planets were just better for buying and some were just better for selling.

One other thing I would note is that we currently specify that planets don't produce commodity X by not having that commodity available for trading at all, whereas in a realistic scenario those planets should (usually) be planets you can sell to at a higher price. So trading is always going to be incomplete until a solution for that is added.

Anyway, assuming the price fluctuations aren't easily predictable by the player I'd personally be in favor of what you describe this economy system as (as much as I understand it anyway, I don't know much of anything about sinusoids and wiener processes are completely alien to me).

agb32:
Yes - thats correct - they won't be easy to predict, as they will periodically fluctuate up and down.  So for example, although most of the time, you might make a profit trading food between A and B, occasionally, you might find you make a loss.

I did think a bit about planets that don't produce or require a commodity (hence nothing in the shop) - here, we could just add the ability to sell at a low price (and we can easily determine the lowest price available, so as to be sure that a profit is never made) - and never have any of it to buy.  That way, a player can dump their cargo there, if they see a much better trade they can make.

onpon4:
I was thinking more about commodities that a planet requires but cannot produce, like desert planets and water for example. I suppose this could be handled just by making the price the highest it could be or something.

Ah, I do see a problem here now that I've looked at the changes: you're hardcoding commodity names and asset classes in C code. That makes the system unflexible, all the necessary parameters should be defined in XML along with the commodity definitions themselves.

I also notice that the usage of indentation in your changes is inconsistent with the rest of the Naev project. It should be three space indentation and statements shouldn't be as crammed together as they are (actually most Naev code adds extra spaces, e.g. `if ( x || foo( y, z ) ) {`, though something more like `if (x || foo(y, z))` is commonplace too; in any case your code is much more crammed together than any Naev code I've seen).

onpon4:
Ideally there should also be a way for planets to override commodity pricing parameters as necessary, though that's not essential.

agb32:
Good point about the coding style - I'll try to update that, and add spaces, etc.
No, commodity names are flexible - only taken to match the base commodity in the xml files so that a baseline price can be given.  But I see what you mean about the if statements comparing commodity names.  Would this be better defined in xml?  I could perhaps add something to commodity.xml, and define the various factors in there.

onpon4:
Yes, that's what I meant, the commodities XML can't be extended with new commodities for example without changing the source code. And yes, adding those parameters to commodity.xml would probably be best.

PhoenixRiver:
I'd have suggested that release the new release with some economy
would be better than not... it has been a ong, long time in coming and
maybe this will provide the necesary step to get it started on a road
to being a thing; I remember these discussions happening when I read
the old archives before I even did anything on Naev so it realy is
about time we had _something_ at least then people can work on
something concrete (even if it is bad) I work much better that way,
though I don't know about anybody else?

River *<:@)

onpon4:
I'd rather at least have it be doing all things right before merge. I've had too many cases of having to clean up badly written missions for example. Better to get it right the first time.

That said it doesn't have to be perfect, just designed properly (proper coding style and proper techniques - using XML and Lua where appropriate). As long as the base is good it can be extended in functionality any time.

PhoenixRiver:
Hello :)

> That said it doesn't have to be perfect, just designed properly (proper
> coding style and proper techniques - using XML and Lua where appropriate).
> As long as the base is good it can be extended in functionality any time.

Given more than ten years of people working on this and that we don't
even know how economies should work in the real world... I should also
say that most of my work on Naev has been tiding up, but I like that
and it's something I enojy doing and can do - making the new stuff and
having the good ideas for new missions and getting new stuff off the
ground, that is the hard part; so maybe you can see where I'm coming
from.

River *<:@)

agb32:
ok - I've made the changes, put the configurable stuff in xml, and I think sorted out the indents.  Will create a new pull request.

onpon4:
Why a new pull request? The changes you push to this branch automatically become a part of the pull request.

agb32:
Yes, I've just figured that... so the changes are now here... and I hope ready to go!

onpon4:
Not quite. I've taken another look and noticed you're mixing tab characters and space characters. Simple fix, just replace the tab characters with however many space characters you mean them to represent (find-and-replace can do that easily enough for you).

Not mixing tabs and spaces is important because different editors can interpret the tab character as a different number of spaces (e.g. GitHub interprets it as 8 spaces, which makes a lot of your code hard to read).

onpon4:
(Of course it's ultimately up to the project members to decide if it can be merged or not, I'm just letting you know the issues I see.)

agb32:
Thanks - done.  Emacs is usually pretty good, so I don't know what happened there... confused by 3 spaces I guess!

bobbens:
I think it should be an improvement over what we have and can be used as a base for further development. However, the code is a bit hard to follow. I have made some minor improvements making most of the functions you added static, pushing to the [agb32_economy](https://github.com/naev/naev/tree/agb32_economy) branch. Could it be possible for you to consolidate all the functions of the type `economy_calcSysJumps(...)` into a single `economy_calcCommodityPrice(...)` or something? It would be also beneficial if you put more comments about the calculations you are doing throughout the code.

agb32:
Sure - I'll rationalise it and add comments.

agb32:
bobbens - I have an economy_initialiseCommodityPrices function - do you want all the economy_calcSysJumps(...) etc functions pulling into that?  (so just a single function to initialise the economy?

bobbens:
> bobbens - I have an economy_initialiseCommodityPrices function - do you want all the economy_calcSysJumps(...) etc functions pulling into that? (so just a single function to initialise the economy?

Well if you look at the agb32_economy branch, we have
```
static int economy_calcPriceClass( char *class, Commodity *commodity, CommodityPrice *commodityPrice );
static int economy_calcImg( char *gfx_spaceName, Commodity *commodity, CommodityPrice *commodityPrice );
static int economy_calcSurface( char *gfx_exterior, Commodity *commodity, CommodityPrice *commodityPrice );
static int economy_calcPopulation( uint64_t population, Commodity *commodity, CommodityPrice *commodityPrice );
static int economy_calcFaction( char *faction, Commodity *commodity, CommodityPrice *commodityPrice );
static int economy_calcRange( int presenceRange,Commodity *commodity,CommodityPrice *commodityPrice );
static int economy_calcSysRadius( double radius, Commodity *commodity, CommodityPrice *commodityPrice );
static int economy_calcSysVolatility( double nebu_volatility,double interference,Commodity *commodity,CommodityPrice *commodityPrice );
static int economy_calcSysJumps( int njumps, Commodity *commodity, CommodityPrice *commodityPrice );
```

I think all the functions above should be consolidated into a single `economy_calcCommodityPrice( ... )`, which would make it easier to read and debug, because as it is, they are all called exactly once.

agb32:
ok - I've tidied up and put those into a single function...

bobbens:
Looks good! Now we can use this as a base before improving it.

agb32:
Great, thanks.  My first thought was to add a history of all prices seen by the player into the save file.  That way, when landing at a planet it will be able to show the average commodity price seen globally, and the average price seen at this planet (which will be same as the current price if this is the first time landed).  Save-file memory would grow with the number of landings...
We could also use this info in the map (displayed at the side) to help people choose a route.  Thoughts?

bobbens:
No need to store the entire history. What about just using a moving average? That way you would only need to save the average price, and when you see prices again you update based on the following rule:
```
average_price = last_seen_average_price * theta + current_seen_price * (1-theta)
```
where `theta` would be a value between 0 and 1 (0 means you only use newly seen value and 1 means you never update your values).

BTAxis:
Writing things into a save is risky because you might get unpredictable effects if the system, or the data set, changes.

agb32:
I have a rough trade route planner in Python... but yes, adding something like that to the maps - based on prices already seen would be good.
Also, people at the bar could give prices of nearby systems a short time ago.  That would probably still work with a moving average, if the gain was low.

Mutos:
Reading all the comments, I'm pleased that at least we have a functioning pseudo-economy system ! Thanks a lot, agb32, for having done it !

The only thing that I regret is, that it's hard-coded in C instead of being a Lua module.

In just a week, things are already going in all directions about which model to adopt, etc... I have no opinion on the matter, lacking knowledge about it, but using Lua would have allowed experiments on multiple models until one or a combination would have been chosen and implemented in C.

However, better something working than nothing at all ! So let's drink champagne, as from now on, players will be presented with actual varying prices and the "Commodity" tab will really come to life ^-^

agb32:
Would be interesting to see what the economy is like in SoC.  You'd need to update the commodity.xml file, but I suspect the rest would work just fine...

Mutos:
Oh shit ! I just opened the **_commodity.xml_** and realized everything is based on the asset's **_type_**. I never used this field, so all SoC assets have the default.

I'll dive into what meaning this field bears and update the SoC assets accordingly.... It may take some time, I tell you when it's done...

PhoenixRiver:
> No need to store the entire history. What about just using a moving average?
> That way you would only need to save the average price, and when you see
> prices again you update based on the following rule:

Equally what about a maximum and minimum price the player has seen
with the data being thrown away after a certain length of time of not
visiting, this would allow tips from people in the bar and other
sources to give the player new information or maybe update it and a
mechanism to remove the information should there be a change to a
system's XML data or a progression in the storyline changing things
radically e.g. a planet's destruction or a serious shift in the
disposition of trade or population.
More data would have to be saved but it allows more variety of sources
to interact with the information without depending on a average that
may not reflect what is happening both in universe and out universe.
it also allows the player to gauge how much prices change wihtout them
having to memorise things (something that trade systems are often not
very good at and I suffer from a lot - and there is a lot of data if
you want to trade trade prices acorss many planets on paper, i have
tried many times!)

Hope this helps,

River *<:@)

PR #1005, opened 2020-06-04
Re-added Dummy Systems.

onpon4:
It was removed accidentally in 3db3d851e57829b2416e6cee57ee39a13d89d723


bobbens:
Looks good, thanks!

PR #1011, opened 2020-06-09
Added economy statistics and last purchase price

agb32:
This merge request improves on the basic economy stuff.
When a player visits a planet, the commodity price information is collected, as is the price at which a player last purchased each commodity.
This greatly helps game play as it gives the player an idea of whether a certain commodity is more or less expensive than average (of what they have already seen), and also reminds them of what they purchased the commodity for - so that they decide to sell if it makes a profit or not.
For each commodity on each planet so far seen, 4 bits of data are saved, from which the mean and standard deviation of prices can be calculated.
These are displayed in the commodities tab, in the form of average price +- std, for the current planet, and for the galaxy as a whole.

### DIFF
--- a/src/economy.c
+++ b/src/economy.c
@@ -93,6 +93,12 @@ static int econ_createGMatrix (void);
 credits_t economy_getPrice( const Commodity *com,
       const StarSystem *sys, const Planet *p ); /* externed in land.c */
 
+/*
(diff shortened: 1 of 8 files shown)


bobbens:
Looks good, thanks. I removed some stray printfs and unified the text with the rest of the land windows.

Issue #1006, opened 2020-06-04
Add more diversity to the lineup of NPC portraits


onpon4:
I've been thinking lately about how, erm, white Naev's NPC lineup is. I can't work on this of course (I'm no good at drawing), but it occurs to me that the collection of NPC portraits should be diversified.

Adding more NPCs with different ethnic backgrounds would not only be a nice gesture to people of color, it would also make Naev's lineup of NPCs more varied.

Obviously no-name NPCs should get this treatment, but I would say that unique portraits should get this treatment too. Currently we just have a single black unique NPC; I think more unique NPCs should be people of color (and with a greater variety of apparent ethnic backgrounds).

I would also note that in resolving this issue, we should take care not to let our unconscious biases seep in, such as by portraying black people as criminals or some such thing (which actually seems to have seeped into the lineup of pirate NPC portraits, so working on that might be worthwhile).

agb32:
Perhaps not just known colours, but a whole range of different colours - who knows what evolution will do to us in the next few 000 years...?

onpon4:
Well the Soromid would be a good place to do unusual colors (and blue people - agyria - from silver poisoning is a real thing), but for everyone else I imagine we probably want to be more grounded.

Chances are what'll happen in our continuing evolution is we'll get darker on average. There wouldn't be any evolutionary pressure to color the skin in a weird way, the only pressure that causes more or less pigment in skin is more or less sunlight. If we want fanciful colors (outside of the Soromid) it would make more sense to put those in hair and eyes.

This talk reminds me of another aspect that could use diversifying: the size of NPCs. Different weights, different face proportions, etc.

bobbens:
https://opengameart.org/content/portraits-for-naev should be incorporated! (found by @onpon4 on discord)

amylaar:
> Well the Soromid would be a good place to do unusual colors (and blue people - agyria - from silver poisoning is a real thing), but for everyone else I imagine we probably want to be more grounded.
> 
> Chances are what'll happen in our continuing evolution is we'll get darker on average. There wouldn't be any evolutionary pressure to color the skin in a weird way, the only pressure that causes more or less pigment in skin is more or less sunlight. If we want fanciful colors (outside of the Soromid) it would make more sense to put those in hair and eyes.

If we only think about existing pigment, a logical optimum is dark skin for radiation resistance, with vitamin D supplements
taken as a matter of course.  But not too dark, lest you can't find melanoma in time.  OTOH, if the rich don't breed and the
[...]

PR #1014, opened 2020-06-11
Modified portraits by Nihilos

onpon4:
This was something I came across that I found was really interesting:

https://opengameart.org/content/portraits-for-naev

They're modified portraits for Naev made by using piecemeal parts! It's lower resolution than the original masters by grumbel, but it's laid out in such a way that it's easy to crank out new portraits. It also includes various skin tones, which goes a long way toward solving #1006, though it's still not perfect (a lot of components have outlines of white skin around them, which gives the very unfortunate appearance of blackface if used on dark skin components).

So I've taken the liberty of making some changes here to integrate those portraits into Naev... and make several new ones, including some missing unique portraits! 😄

I like what this results in so far already. Diversity is heavily expanded compared to before, previously missing unique portraits have been created, and this is subjective, but I think these modified portraits look really nice! It would also be very easy to expand the .psd file (which I submitted a pull request for in the naev-artwork repo) to enable more possible looks, which we'll need for the Soromid in particular. (Of course when this happens we'll probably want to convert it to xcf format.)

So of note:

* Lots more generic NPC portraits (4 new male, 5 new female).
* Lots more diversity: more skin tones, and even a non-unique transgender NPC!
* Unique portraits for named characters in the Coming Out and Heretic campaigns, as well as the one-off Crimelord mission.
* A female pirate NPC portrait.

This doesn't completely solve #1006, to be clear, but it's a major step.


onpon4:
Sorry to flood y'all with emails. I thought I was done but decided that I wanted to do more. 😅

So I this is what I did after I made the PR:

* Changed nebu_satellite to use Neil (obviously inspired by a certain famous scientist) instead of a generic scientist portrait.
* Added some unique portraits: Cynthia and her father, the fake sister in traffic_01, the laid back person (from the race missions), Michal and Paddy from the foodbiz campaign, and Arnold Smith from the Nexus Shipyards campaign. I made some of these PoC, so that's more progress on #1006! 😄

PR #1016, opened 2020-06-11
Portraits (take 2)

onpon4:
This is a redo of #1014 which is less wasteful on Git history space (it's a recommit of the same files but with optimizations to PNG compression).


bobbens:
Looks good, thanks!

PhoenixRiver:
Oops, just spotted that female1.png got missed from the ARTWORK
LICENSE file, apologies I can't submit the fix atm :(

River

PhoenixRiver:
Sorry, scratch that, it is in, I wasn't looking at the right commit :(

PR #1017, opened 2020-06-12
Fixed equipping of Soromid Brigand

onpon4:
Was broken because the table for it was malformed.

I also updated the changelog to mention the changed portraits.


bobbens:
Looks good, thanks!

PR #1018, opened 2020-06-12
Modified the OSD title of Patrol and Sightseeing missions.

onpon4:
This change removes insertion of details about the mission into the
OSD titles of these missions, allowing multiple missions from one
planet to be collapsed by the OSD engine into a single entry after
the patrols / sightseeing tours are finished.


bobbens:
Looks good, thanks!

PR #1019, opened 2020-06-13
Updates to NSIS installer, and fixes to allow for cross compile.

ProjectSynchro:
I have updated the NSIS installer script to correctly point to github releases, a new plugin is required.
You can find it [here](https://nsis.sourceforge.io/Inetc_plug-in). It allows for fancy proxy downloads and other methods of downloading files, most importantly over HTTPS, which NSISdl lacks in terms of functionality.

When the installer is built, it is bundled with the output, however it is required to be installed in order to build the installer.

It looks like some preprocessor macros weren't set up right in some of the OpenAL headers and source files. This would cause failures when building without OpenAL.

I also tweaked configure.ac to properly check for OpenAL using the autotools PKGCONFIG modules. (I think, I'm still attempting to learn how it works, the GNU documentation sorta blows)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -361,11 +361,15 @@ AS_IF([test "$used_csparse" = "internal"], [
     CSPARSE_CFLAGS="-I\"\$(top_srcdir)/lib/csparse\""
     CSPARSE_LIBS="\$(top_builddir)/lib/csparse/libcsparse.a"
 ])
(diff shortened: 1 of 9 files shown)


bobbens:
I've merged it in, but done quite a few changes to make things work better and be a bit more robust. Thanks!

PR #1020, opened 2020-06-13
Fixed some bugs in the GUI.

onpon4:
I noticed both that the fleet formation button appears behind the OSD, and that the planet pane was, erm, stretching a thing that's supposed to be tiled. Oops. This fixes both and also makes the fleet formation button both look better and capable of extending with other fleet-related buttons in the future.


bobbens:
Thanks!

PR #1023, opened 2020-06-13
Fixed some bugs in Seek And Destroy mission.

onpon4:
They were causing destroying the target to reset itself if you re-entered the target system, and also caused spawning of the target to not work (or rather crash the mission) if you entered the system by takeoff.


onpon4:
I pushed an expansion of pirate names into this branch as well. Seems to give good, funny results. :slightly_smiling_face:

bobbens:
Looks good, thanks?

PR #1025, opened 2020-06-14
Change Travis to use Homebrew SDL2 packages for MacOS.

ProjectSynchro:
This should fix travis builds for MacOS. 
I keep seeing failed travis builds which pass on linux but fail on MacOS due to missing sdl2, this should at least get builds to actually start on MacOS, to see if there are other issues.


bobbens:
Looks good, thanks!

PR #1026, opened 2020-06-14
Added a whole bunch more choices for pirate names.

onpon4:
I might be getting a bit carried away here. :P


bobbens:
No such think as too many pirate names!

PR #1030, opened 2020-06-14
Replaced player.pilot():pos with player.pos.

onpon4:
Prevents warnings if escorts gain you reputation after you die.

Also added a couple more pirate names.


bobbens:
Looks good, thanks!

PR #1035, opened 2020-06-15
Portraits galore!

onpon4:
OK, I might have gotten a bit carried away with this one. I started
out by adding portraits, but then I thought, hey, this portrait
selection thing is scattered, so I went ahead and centralized that
into the portrait.lua script. :) The NPC event now uses this, as do
all missions that previously either had their own definitions for
random portraits or specified a particular generic portrait.


bobbens:
Looks good, however, some notes:
1. They don't seem optimized in size.
2. Probably should make a pull request to naev-artwork repo too.
3. Some missions chains will get random portraits every time, not sure if it is worth making it consistent, i.e., same portrait every time for the entire chain.

In particular if you could redo it with the images optimized I shouldn't have a problem merging.

onpon4:
1. I think I optimized all of them. Which ones in particular do you think I missed? I used a tool called Trimage which applies a whole bunch of optimizers to the image.
2. I did submit the PSD (later converted to xcf because I made a tweak) these are all based on, before I added any of them into Naev in fact: https://github.com/naev/naev-artwork/pull/20
3. I did consider that, but I think every case I used random portraits in it shouldn't be an issue. Most of them were actually already using random portraits and the rest, unless I made a mistake somewhere, only show the NPC once. Let me know if you see a particular mission which needs adjustment.

bobbens:
Randomly tried some like
```
[ess@akarii (~/naev)] $ optipng dat/gfx/portraits/sirius/sirius_fyrra_m5.png
** Processing: dat/gfx/portraits/sirius/sirius_fyrra_m5.png
200x150 pixels, 4x8 bits/pixel, RGB+alpha
Input IDAT size = 44841 bytes
Input file size = 44898 bytes

Trying:
  zc = 9  zm = 8  zs = 0  f = 5         IDAT size = 44474
  zc = 9  zm = 8  zs = 1  f = 5         IDAT size = 43269
  zc = 1  zm = 8  zs = 2  f = 5         IDAT size = 42907
                               
Selecting parameters:
  zc = 1  zm = 8  zs = 2  f = 5         IDAT size = 42907

Output IDAT size = 42907 bytes (1934 bytes decrease)
Output file size = 42964 bytes (1934 bytes = 4.31% decrease)
```

onpon4:
Well when I ran these through the optimizer they got a lot smaller than just 4% less. Most had a reduction to around 40% of their original size IIRC. In any case this is as small as Trimage can make them. It's the same tool I used before; I guess it doesn't make the images quite as small as what you use, but the images aren't unoptimized. Feel free to re-optimize them if you want, though.

I double-checked the changes and found one case of a mission that probably needed to cache portrait choice, so I added that in. Everything else looks like it should be good.

bobbens:
It's not that much bloat, but if possible I would highly recommend using "optipng" if possible from now on. Thanks!

PR #1036, opened 2020-06-15
Morrrrrrrrrre portraits

onpon4:
Lots more portraits for Dvaered, Empire, and Sirius! And yes, I optimized the images.


bobbens:
Looks good, thanks!

PR #1046, opened 2020-06-16
Improved the wording of those price hint messages.

onpon4:
For what it's worth I actually don't care for these. I think it'll
just add tedium by incentivizing blindly clicking on all the NPCs
you see (which we already have a bit with the jump hints, but since
it's only early on in the game when those make a notable difference,
and their role is easily replaced by just buying the local map, it's
not that big of a deal). But for now I've fixed it up so that it
displays nicer and fixed up the chances of things happening so that
the economy hints get inserted as the tip chance goes down (same
chance as before after the tip chance goes down to minimum, but
without the reduced chance of tips at the start).

Ah, also fixed some bad code formatting.


bobbens:
Looks good, thanks!

PR #1048, opened 2020-06-16
Fixed land override in DV attack on Sindbad.

onpon4:
No arguments defaults to false, not true, so it wasn't allowing me to land on the station to finish the mission.


bobbens:
Looks good, thanks! Maybe it should default to true in this case...

PR #1054, opened 2020-06-17
Subtract viewport position for mouse events in the GUI.

onpon4:
This causes mouse positions to be identical to graphical positions
from a Lua perspective. (Also fixes a bug where buttons in Brushed were offset.)


bobbens:
Looks good, thanks!

Issue #1021, opened 2020-06-13
Bad strings


Wuzzy2:
During translations, I noticed some mistakes in the original strings.

Here's a list of what I found so far (and I explain why it's wrong):

* Factor used to divide the above resolution with
    * Wrong word. The word “factor” is for the operands in multiplication. For division, the first operand is called “dividend”, the second one is “divisor”, and the result is called “quotient”: https://en.wikipedia.org/wiki/Division_(mathematics)#Notation
* Size of the OpenAL music buffer (in kilobytes).
    * “kilobyte” is ambigious. 1000 or 1024? Could use “kibibyte”? https://en.wikipedia.org/wiki/Binary_prefix
* Font sizes (in pixels) for NAEV
    * “NAEV“ → “Naev”
* tonne
    * Game is sometimes saying “tonne“, sometimes saying “ton”. Pick one.
* Jump for system '%s' not found in system '%s' for removal.
    * The grammar sounds kind of weird. Not sure
* Joystick of index number %d does not existing, switching to default 0
    * “does not existing” = grammar failure
* article is invalid.
    * Not capitalized
* Planet '%s' does not belong to a system..
    * Double “.“
* To use a modifier key hit that key twice in a row, otherwise it will register as a modifier. To set with any modifier click the checkbox.
    * Missing comma after first “key”
* \apAutonav continuing until destination (%d jump%s left).
    * What does `%s` resolve to? The English plural “s”? If yes, very very bad idea, it should be obvious that it will not work for other languages. Just use Gettext's plural system here
* Player starts out in non-existent or invalid planet '%s',trying to find a suitable one instead.
    * missing space after comma
* Uknown
    * typo

I will extend the list when I find more strings.

Wuzzy2:
Mor strings:

* The given link exists in the Vorbis data stream, but is not decipherable due to garbacge or corruption.
    * “garbacge” → “garbage”
asteroid field in %d has less than 3 corners.
    * “asteroid” → “Asteroid”

Wuzzy2:
More strings:

* The given link exists in the Vorbis data stream, but is not decipherable due to garbacge or corruption.
    * “garbacge” → “garbage”
* asteroid field in %d has less than 3 corners.
    * “asteroid” → “Asteroid”

onpon4:
I've fixed most of these in #1055 (only excluded a couple I think are ambiguous). Couple comments:

[...]

Issue #253, opened 2015-03-14
Extents of nebulae should be visible in the hyperspace map


naev:
Original [issue 69](https://code.google.com/p/naev/issues/detail?id=69) created by stephan.... on 2010-12-02T17:44:37.000Z:

The hyperspace map would look a lot more interesting if the extents of nebulae (possibly subject to a fog-of-war-style shroud) were drawn... it'd also make it easier to remember what was where by giving extra landmarks.

onpon4:
Hmm, this might be best done with the thing we're  using to indicate faction locations...

onpon4:
Fixed this in #1057.

Issue #317, opened 2015-03-14
Shadow Vigil introduction gives no indication of where Pas is


naev:
Original [issue 133](https://code.google.com/p/naev/issues/detail?id=133) created by stephan on 2011-06-01T22:36:00.000Z:

At the start of the second Shadow mission, an event is triggered which spawns a ship and hails you. When you respond, you are directed to fly to Pas.

However, this is not the actual mission yet, just the event happening beforehand. And Pas is not exactly an obvious system to visit otherwise; I didn't explore it yet at all in my case. This makes it rather difficult to locate, because there is no universe map indicator like on missions.

Another effect of event-vs-mission is that if the player decides to save and quit, it's easy to forget this ever happened when coming back. There's nothing on the info screen.

onpon4:
Just wanted to comment here that some time ago I modified that event's dialog so that the player would know to make a mental note of what system to go to for this exact reason. I think it could still be better, though. Ideal in my opinion would be to have an actual marker, which would require starting the mission with the event, but another idea that comes to mind is a text description of where Pas is.

bobbens:
Maybe have the event be recurring until the player goes to Pas?

onpon4:
That's an interesting thought! I'll have to look into doing that.

onpon4:
I've added that fix to #1057. It recurs every 500 periods.

PR #1057, opened 2020-06-17
Made time units galactic, plus some other small fixes.

### LINKED ISSUE #1039: Lore: Inconsistent use of time
In some of the lore (news text, civilian text), there is talk of “weeks“, “months”, “years” and “decades” in a non-planetary context. I think this does not fit into the Naev universe, which has its own timing system (SCU, “cycles“, etc.). I suggest to rewrite the texts.

Offending strings:

* Technicians at the Omega Station recently returned from a new effort to understand the Sol Incident. They expect the first data to arrive in two months' time.
* In an interview with Bleeding Edge anchor McKenzie Kruft, a researcher at Eureka labs says he has produced a tabletop atomic reaction. He hopes to publish his results in a science journal later this year.
* The Empire agreed terms with shipbuilder Nexus for a new generation of military craft. The deal extends the partnership with the government for another decade.
* For decades used only as menial labourers, now SuperChimps are being widely adopted as domestic companions. Enhanced primates make an affectionate, intelligent pet, or low-cost servant.
* You have attached your docking clamp to the derelict's airlock, but the door refuses to open. A few diagnostics reveal that the other side isn't pressurized. The derelict must have suffered hull breaches over the years. It doesn't seem like there's much you can do here.
* Only two years after the mysterious disappearance of its owner, the galaxy's only deep space bar shut down the generators and cycled the airlock one last time.

Note that there are also other strings which already refer to type via “cycles”.

onpon4:
Fixes #1039


onpon4:
Added a fix to #317 here as well in the form of a reminder for shadowcomm.

onpon4:
Fixed #253 as well. I'd like to add text that says "The Nebula" to this; I couldn't figure out how to do so. But you can see the nebula from the map now.

onpon4:
Found out that it's just baked into the image, added that to the nebula image. Also fixed #1021 stuff.

onpon4:
I've redone the commits in #1059 to remove bloat from the image edits.

PR #1059, opened 2020-06-17
Redo of #1057

### LINKED ISSUE #253: Extents of nebulae should be visible in the hyperspace map

### LINKED ISSUE #317: Shadow Vigil introduction gives no indication of where Pas is

### LINKED ISSUE #1021: Bad strings

### LINKED ISSUE #1039: Lore: Inconsistent use of time

onpon4:
This is a redo of #1057 so that it doesn't have multiple edits of the nebula map image (I just amended the original commit).

Fixes #1039
Fixes #317
Fixes #253
Fixes #1021


bobbens:
Looks good, thanks!

Issue #1028, opened 2020-06-14
Make commodities searchable


Wuzzy2:
I find it very hard to remember which planets offer which commodities, so I suggest to add them to the search (like the System/Planet/Outfit/Ship search).

onpon4:
I was suggesting on Discord that commodities should generally be available everywhere, just with price adjustments. If we were to do that then it would help with this (as well as other issues).

agb32:
These are now basically searchable using the new trade function on the map (currently in a pull request)

bobbens:
I don't think this is an issue anymore.

PR #1060, opened 2020-06-18
Removed unused leftover tutorial stuff.

onpon4:
The tutorial feature has been non-functional for some time now but there were still some leftover functions related to it. This cleans those up. (I made sure that the game still compiles and no problems.)


onpon4:
I've just pushed a change to this branch replacing the price info NPCs with price info in the news feed. It does essentially the same thing, however:

1. It provides prices for all items in the current system and adjacent systems, rather than random prices for random planets.
2. It updates your averages automatically without you having to interact with it.
3. It lists all the info at one time in a convenient place, so you can compare prices with nearby planets easily.

This also has a bonus of helping make it easier to find commodities you need. Not exactly a *fix* for #1028, but substantial progress on that front (the search is a lot easier now).

bobbens:
Looks good, thanks!

PR #1061, opened 2020-06-18
Replace Travis CI with Github Actions

ProjectSynchro:
This set-up should effectively replace Travis CI.
This set-up does not upload artefacts to BinTray, but has them available in the Actions Dashboard after a build, uploads to BinTray can be set-up if required.

Two secrets will need to be added to the repository.

SSH_KEY - this would be an unencrypted PEM
KNOWN_HOSTS - ssh-keyscan to get the known_hosts entry for the server being rsynced to.


bobbens:
I'm not too familiar with Github Actions, but this would be like a straight replacement for travis? It would also say if compiles have failed or something? How are the SSH keys and stuff added?

ProjectSynchro:
This would be a straight replacement. It's configured to be triggered on a PR or on a Push.
When it's triggered it will attempt to compile on Linux, Mac, and Win32/Win64

If a step fails (these can be defined at any point within the build script) it will say at which step it failed at, as well each of the runs will show up as separate checks, so you would know at a glance if lets say MacOS failed to build.

In terms of adding SSH keys and other secrets, you'd have to add them via the settings menu for the Github repo
image
The only requirement would be that they match the variable names in the build script (SSH_KEY, KNOWN_HOSTS) this would be up to @ids1024 to provide due to the fact that he is hosting the Lua documentation.

How the lua documentation is deployed is just copy pasted from the travis config so it should more or less still work. just a matter of having the right SSH_KEY and KNOWN_HOSTS in the git repo's secrets.

ProjectSynchro:
Another benefit is that a separate workflow could be configured for releases, allowing for installers, binaries, etc. to be generated and uploaded to the appropriate release whenever a new release tag is created.

ProjectSynchro:
I'm going to create a new PR for this because it is a mess in terms of cleanly merging this.
Will reopen and reference this in the new one.

PR #1064, opened 2020-06-18
Add German translation

Wuzzy2:
This adds a German translation for the game. Including metadata files.

It has not been tested in-game yet, as I am still unable to actually activate the translation for Naev. So you have to test it yourselves (especially if any button labels are “bleeding over”).


bobbens:
Looks good, have to figure out why it's not being activated...

bobbens:
I pushed some patches and was able to get it to work. Try with latest master and make sure to make install and run it.

PR #1067, opened 2020-06-18
Fixed possibility for duplicate econ articles.

### LINKED ISSUE #1063: Main game tutorial mission has no description or reward text
If you open the missions tab in the info menu during the main game tutorial, the Tutorial mission has no description text nor reward description text.

I know there's no reward, but the text should then just say something like “No reward.” or something.

onpon4:
This just deletes any existing econ articles before adding new ones.

Fixes #1063


onpon4:
I've added name, description, and reward definitions for the tutorial mission (fixes #1063) and also fixed a misspelling in the tutorial mission.

onpon4:
I pushed a reversion of 4cd59cc here because it actually breaks the radar at 1x scaling. Having looked at the changes I don't understand why, but for now getting rid of it since 1x scaling is the "usual" case. (Will look into it and see if I can figure out how to do it properly.)

bobbens:
Looks good, thanks!

PR #1069, opened 2020-06-19
Further refinements to the Brushed GUI

### LINKED ISSUE #1070: More typos
Two more typos I found in 0.8.0 beta:

Short mission HUD description of a PIRACY mission:

> Kill target Trader ship
> Patrol of %s

(the “%s“ was litereally displayed on-screen)

Pirate civilian text:

> I tried to get get into pirate clanworlds, but they wouldn't let me because I'm a \"small-time pirate\"!

Again, literally `\"` instead of `"`.

### LINKED ISSUE #1071: Za'lex Test gives trillions of credits
I did the Za'lex Test mission a few times to build up my reputation. Initially, these missions paid “normal” amounts of money, nothing special.

But after I built up my reputation a little, the Za'lex Test missions eventually paid out absurd amounts. First I have seen a mission that paid 400 MILLION credits. After I completed it, I got another mission in the TRILLIONS. Thanks for all the cash, but I think that was not supposed to happen. xD

It seemed to happen most likely when taking the mission on the Straight Row system.

### LINKED ISSUE #1072: Escort formation button misplaced
The button for escort formation is misplaced somehow. OK, not really, but when you hover it with the mouse, the focus is a little off. You must put the cursor a little BELOW the button to make it clickable.

How to reproduce: Get Lancelot Fighter Bay, launch fighters, get the button.
Tested with `slim` GUI only.

Version: 0.8.0 beta

### LINKED ISSUE #1076: String concatenation in translated game
There are a couple of bugs with the translation. I have tested this with German for reference.

The game concatenates translatable strings. I probably should explain in detail:

* Audio options:
    * Volume sliders have bad translation: “Sound volume” should be “Tonlautstärke” in German but it's “Ton Lautstärke”. The space is forced by Naev, as translator I have no control over it. IIRC “Sound” and “volume” are separate strings …
    * Same for “Music volume”
    * My guess is that you concatenated strings, which should NOT be done for translation because concatenation assumes a word order, but translations frequently like to flip words around (and for various other reasons). ;-)
    * Suggestion: Turn “Sound volume” and “Music volume” into full strings, no concatenation
* Star map:
    * Translation of interference and nebula variants (dense/light/dangerous/...) is tricky because just like in the Audio options, it all seems concatenated. This makes it hard/impossible to translate every possible nebula/interference variant properly in German, because of grammar rules
    * Suggestion: Instead of concatenating strings, create an individual translatable string for each nebula/interference variant. Basically it's the same problem as with Audio
    
Version: 0.8.0-beta1

onpon4:
There were some remaining problems, so I've edited the Brushed GUI once again.

Now, everything on the bottom has been shifted over the bottom bar, and both the bottom and top bars have been shifted inward just enough to leave space for the targeting display around the edges.

I might consider shrinking the bottom bar later, or it might be possible to utilize that space for extra buttons now that the "main" portion is out of the way.

In any case, this prevents a bug where you could accidentally click on the edge aiming thing when you were trying to click on the minimap, and it ensures that nearly all of the edge aiming thing is visible without looking ugly.

This also fixes a bug that would have caused the game to crash (a variable was misspelled), although I've never seen the case happen in practice.

Fixes #1076
Fixes #1072
Fixes #1071
Fixes #1070


PhoenixRiver:
Can I just note here that according to the comments attached to the
commits here that many of these changes do not relate to the Brushed
GUI and as such putting these changes into this pull request makes
things harder to find...

River

Issue #868, opened 2020-01-09
Colorblind accessibility


onpon4:
This is an issue I've thought of for years since I always look out for this sort of thing in games I develop, but I haven't come up with a solution myself. So I think I'll raise it here.

Basically, Naev relies a lot on color to convey information, so it's not colorblind accessible. Here are the examples of requiring color vision I can think of (in order of importance):

* The overlay map and display indicators use only colors to distinguish hostile/neutral/friendly ships, as well as other kinds of space objects and whether or not a given thing is selected (very important)
* Current mission objective in the OSD (as opposed to other objectives) is indicated only by color (very important)
* Mission marker types are distinguished only by color (very important)
* Slot sizes are distinguished only by color (important)
* Some text is differentiated from other text only by color (not very important; differentiation usually unnecessary)

There of course may be others that I missed (being someone with full color vision).

I'd like to note as well that many cases involve having to differentiate red and green; red/green color deficiency is quite common so that's going to prevent even more people from being able to play and enjoy the game.

There's nothing wrong whatsoever with using colors in these situations, and I think we can all agree that those should stay (as they're really nice and useful for those of us with color vision). So what I'm suggesting in all cases is keeping colors, but also adding a secondary method of distinguishing these things for those who cannot see the color differentials. What to do to this end depends on the particular situation. Sometimes brightness levels can do the trick, sometimes adding a symbol does the trick, and sometimes modifying the shape of something does the trick. The point is to make the game as accessible as it possibly can be.

Some of these are tricky, of course. The map in particular poses quite a challenge as so much information is being packed into color right now. Let's discuss some ideas on how to deal with this. :slightly_smiling_face:

ids1024:
Android has a neat feature under developer options for simulating different types of colorblindness. I haven't tried it, but apparently [Color Oracle](http://colororacle.org/) can be used for that purpose on Windows/Mac/Linux.

onpon4:
That could definitely be real helpful for testing purposes.

It also occurred to me that perhaps there could be a "colorless mode" that filters all colors to shades of gray, as a fallback in case a type of colorblindness is missed.

bobbens:
The colors defined in the code could easily be moved to the config file for full configuration. Or just define two sets (one for colorblind one for regular) and have an option to change between them.
[...]

PR #1080, opened 2020-06-19
Economy information on the map

agb32:
I've added economy information to the map, with a trade button next to autonav, which allows player to cycle through commodities (in reverse with shift, or toggle between commodity and none with ctrl).
Information is then displayed on the map as follows:
Prices are calculated at all known planets using known economy information.  They are then compared with the selected system.  A best trade (lowest price in selected system against highest price in other system) and worst trade (highest price in selected against lowest in other) are then computed.  2 circles are drawn around each system, representing best and worst price.  Yellow if a good trade, and blue if a bad trade, with most intense for best/worst, and fading to grey for equal cost.
If the selected system is the current system, and player is landed, then the best/worst trades are compared against prices on the current planet, rather than system as a whole.

I've also added a couple of hints about how this works to the npc.  
Note - separately, I think we should bring back the economy information in npc rather than as news because:
1. Means there is a reason to go to the bar, even late into the game
2. Means information is only gained if the player is actively doing something (approaching) - rather than just being given the information in news by visiting the planet.

Note - which ever way, whether in bar or in news, the extra economy info provided gets added to the players economy info and thus save file.  It isn't just for reading there and then, but helps build up statistics.


onpon4:
I'm not the biggest fan of this method (stretching those images onto a tiny button is ugly and I didn't even know what I was looking at at first). In any case, there seems to be a bug where it doesn't work on all commodities. I found that it worked with Medicine and Luxury Goods, but not Ore or Industrial Goods.

I would also suggest that "Trade" is not really an accurate descriptor of what that button does, when you talked about a "trade button" I assumed you meant like, some kind of trade hub window showing prices and stuff (which would be real nice), but all it does is cycle through commodities (including unknown commodities) and circle systems that have that commodity.

agb32:
It works on ore and industrial too - it just means that you haven't yet got any information about those - either by visiting a planet that sells them, or by seeing it in the news.  (check your save file to see)

True - a different name would be welcome - but something short to keep the button small.  Though it does show you what a good trade would be...

onpon4:
> It works on ore and industrial too - it just means that you haven't yet got any information about those - either by visiting a planet that sells them, or by seeing it in the news. (check your save file to see)

This was from the game that I've been playing for the last week and a half, with most of the map revealed. I didn't get any indications of ore or industrial goods anywhere.

agb32:
The colour gives pricing information - Blue means a -ve trade, Yellow means a +ve trade, gray means nothing - and its a gradual slide from bright yellow to grey to bright blue.
Perhaps not the most intuitive, but something like that was discussed...
However, perhaps we need something better - though at a glance, seeing bright yellow tells you to buy at the selected place, and sell at the bright yellow place.  Seeing bright blue tells you to buy at the blue place and sell at the selected place...

agb32:
It won't show anything for gold, most likely, as I think there is only 1 place that sells that... (last time I checked)

onpon4:
> The colour gives pricing information - Blue means a -ve trade, Yellow means a +ve trade, gray means nothing - and its a gradual slide from bright yellow to grey to bright blue.

That's not accessible to colorblind people.

#868

Also not intuitive to color-sighted people.

I'd recommend replacing this with a separate window showing pricing information in text.

> though at a glance, seeing bright yellow tells you to buy at the selected place, and sell at the bright yellow place. Seeing bright blue tells you to buy at the blue place and sell at the selected place...

That's not what my brain thinks at all. (And as mentioned, that won't work for colorblind people even if you explain it to them.)

> It won't show anything for gold, most likely, as I think there is only 1 place that sells that... (last time I checked)

Gold doesn't exist anywhere in the actual game yet, just in Zied (which is inaccessible to players, that's just a testing ground).

agb32:
How about the price difference in text next to the name?  Would that work?  (its actually 2 values, a best trade and worst trade).
I think showing a price difference with selected system would be better than showing just raw prices - makes it easier to quickly work out whether a trade would be beneficial.

onpon4:
I would recommend organizing it into a separate window. The map itself is already busy as-is and this method you're using doesn't strike me as ideal.

Like for example, maybe make it like the missions screen, with a small map on the left and all relevant pricing info for a selected planet listed out on the right. Or this window could sort it by commodity, with a list to select the commodity and then a list of each planet's price created by that. There's lots of possibilities that would be more intuitive than what you currently have, and would look nicer and be colorblind-accessible too.

agb32:
If we remove the background images, that might make the map less busy!  Though I quite like them!
A separate map could be an option, though its not obvious what to get rid of - the missions map is just as busy, but smaller!
I think having it all on one map, but with the option to turn it off gives best of both.  The only question is how to display the information.  Perhaps an up arrow and down arrow (for more expensive and cheaper - with the lengths giving some indication of by how much) or something like that?  Or we could do something with the filled in yellow circle in the middle - though I don't know what.

(incidently, holding ctrl and clicking the trade button toggles on/off the current commodity - which gives a good visual indication - there was an npc who mentioned that!).
I guess one thing we could to is when the trade button is on a commodity to not show missions.


I was planning on having some sort of solar system screen - by clicking on an already selected system on the map, it would then bring up solar system information - with each planet displayed in a line along the top, and then below this information about economy on that planet, about what outfits sold there and what ships sold there (and other info such as population etc).  (of course, only if the player knows that information)

But that would be in addition to the whole economy information - having it all laid out at a glance is very helpful in making decisions.

onpon4:
> If we remove the background images, that might make the map less busy! Though I quite like them!

No... that was a very recent graphical enhancement and it enabled fixing a long-standing issue. The map has plenty of information on it, we don't need to simplify it to cram even more onto it.

> A separate map could be an option, though its not obvious what to get rid of - the missions map is just as busy, but smaller!

You don't need to get rid of anything, just don't use the map directly. Use it just to help select the planet you're looking at and display all your info elsewhere in the window.

> I think having it all on one map, but with the option to turn it off gives best of both. The only question is how to display the information.

You're really overthinking this. Plain text outside of the map on a separate window is fine.

You don't even necessarily need to use the map at all. You could make it more similar to the outfit search for example. That would work too.

> I guess one thing we could to is when the trade button is on a commodity to not show missions.

I don't see any reason to do this.

> I was planning on having some sort of solar system screen - by clicking on an already selected system on the map, it would then bring up solar system information - with each planet displayed in a line along the top, and then below this information about economy on that planet, about what outfits sold there and what ships sold there (and other info such as population etc). (of course, only if the player knows that information)

That would be doable. You could re-use map editor code for that. Though in all honesty that might be a bit overkill since there's usually only a couple applicable planets per system. A simple list for planets would probably work just fine.

agb32:
I think the main point is we need some way to show advantageous trades, rather than just price information.  So what we have is like a trade route planner.
Having played with this a bit, I had a python script that would show me prices.  But I found that wasn't enough (there are so many of them), and so a trade route planner was better.

Now I agree that price information is also useful - hence the solar system screen.  But having good trade routes available at a glance is a must, I think.  Sort of: "Click a system, and see where you can trade with from there".

In terms of cluttering up the map - it is only when a commodity is selected...

(I wasn't suggesting removing the background images - I like them!)

onpon4:
Showing advantageous trade routes is probably unnecessary but perfectly fine if you want to do that in a separate window. Actually presenting that information would be difficult though.

> In terms of cluttering up the map - it is only when a commodity is selected...

That's not the point really. The point is that the design you chose is ugly, unintuitive, and requires you to cram more information onto an already busy map, which is inevitably going to lead to somethting hard to read. Whereas if you just have a separate window and present the same info outside of the map onto a dedicated text area, it would look much much nicer and be much easier to read. Or you could forego the map altogether if you implement it as a commodity search, not unlike the outfit search. That actually might be better since then you would see a list of planets by distance.

agb32:
ok - new commit - please see what you think of this one.

Now, when you select a system on the map, it compares prices of the commodity with those known in other systems.  Then, if the trade is good (ie buying at selected and selling at another would yield a profit), it displays a yellow circle above the system, with size increasing (up to a max) depending on how good the trade is.  If trade is bad, it shows a blue circle below the system.

So visually:  Circle above means make a profit, circle below means make a loss.
Size of circle represents amount of profit or loss.
Colour of circle also represents amount of profit or loss (from grey to yellow, or grey to blue).
This way, colour blind people can tell by the size, while others get an additional indicator.

So easy to quickly identify where will make a good trade.
Colours and shape could of course be changed if there is something better (e.g traffic light colours, up/down arrows, or something like that).

If people could comment, that would be great.  If its not liked, and we really want to keep this off the map, then I would try doing something textual which would be on a separate window...

onpon4:
That fixes colorblind accessibility but not intuitiveness.

Also, it doesn't work on my end.

agb32:
Sorry - try again now... a couple of zoom/size issues

agb32:
Perhaps the minimum size (for small profit/loss) is too small - I can increase that if you think it would help...

agb32:
Green orange red might help with intuitiveness - i.e. green signifying "buy at selected to sell here..."
Or perhaps up/down arrows instead of circles?

agb32:
(oh, and note - if the currently selected system doesn't sell that commodity, nothing will be shown, since it will be impossible to make a trade!  - perhaps something to indicate that would be needed.)

agb32:
Changed the trade colours from yellow -> grey -> blue (profit -> break even -> loss) to green -> orange -> red (traffic lights).  Size still indicates profit/loss, so okay for colourblind.
Also added messages if no economy info on selected commodity at current system, and added a message in the bar below describing how to interpret the information.

agb32:
Merge conflict sorted.  "message in the bar" refers to a message in the status bar at bottom of map, not in the actual bar.

agb32:
Hopefully this is now good to merge?

agb32:
Hopefully this is now good to merge?

onpon4:
That's ultimately up to bobbens, but I really really think you should reconsider the design of this. Most of these dots are practically invisible and the information they convey is imprecise. And it's needlessly complicated, and the button with squashed-down commodity images looks very very bad.

If you just got rid of this and replaced it with a simple text-based window, it would be much much easier to read (no need to squint at pixel-sized dots that often are the same color as what they're directly behind) and look a whole lot nicer.

agb32:
I agree its not great - but better than nothing, until a better design can be thought of.  The size can easily by changed, but I kept it small, to not clutter the map.  Though of course, small dots mean small profit, so you're really only interested in the larger ones.
A text based one is being worked on, but that won't provide the whole lot of information that can be shown graphically all at once on the map - so the idea would be to have both, and give players the choice... and I think they'll be complimentary...

onpon4:
I actually disagree. This system is bad enough that it just makes Naev look even more unprofessional than it already does with zero benefit.

Screenshot at 2020-06-21 18-08-21

This isn't useful. Even discounting the fact that many of these dots are practically invisible, tthe amount of effort I have to go through to figure out what this means is much too high. And I have to click on the system I intend to... buy from? I honestly don't remember.

This system is making assumptions about how someone would want to use pricing info which might not be correct, anyhow. What if I'm just looking for the most expensive market in a 4-system radius to sell mined ore in? What if I'm just looking for any place at all that sells Food to fulfill a mission? These are problems that would easily be solved by a simple commodity search, but not even remotely helped by this map marking system. And it's ugly; that button looks more like the game's graphics glitched out than like it's showing an image of food.

agb32:
I guess there are 2 things here.  As a trade system, which this is trying to help with, it will tell you (based on the above image) that buying food at Hatter, can then be sold at profit at Longbow, but not anywhere else of note.  On the other hand, buying at a large red dot (e.g. Sarcophogus)  will then make a profit by selling at Hatter, or better, Longbow.  I think it could be very useful...  and of course, the size of the spots can easily be increased - a trivial change to do this.

But - it doesn't tell you actual prices.  That will be better done on a separate system screen (a bit like the trade tab, but without having to be landed).  That way, you can find out where would sell food, or where you could sell mined ore.  I have something in development for this.

Re the ugly button image - yes - I'm not an artist, so used what was available.  This could easily be replaced if someone wanted to draw new icons for it... and I am sure willing volunteers could be found once its in there.

onpon4:
> Re the ugly button image - yes - I'm not an artist, so used what was available.

The image isn't the problem. What you're doing with the image is. It looks like a middle school programming project. Image buttons are basically never a good idea, let alone image buttons with stretched out images (which might, by the way, get stretched out even more depending on the language).

I'm telling you that the *basic idea* of this design is bad. Not that you chose the wrong image. You shouldn't be throwing an image onto a button like that.

agb32:
I could replace it with the name of the commodity - that was my original idea - but something like "Industrial Goods" would give a large button.  Is that okay, or what would be better?

onpon4:
Not using the button that way would be better.

bobbens:
What about implementing it as a different map mode or something like that? Currently we display faction information, but instead it should be possible to display the pricing information or something different? Still not clear how to visualize though. Maybe do a mock-up first?

Some ideas:
1. Hide the colour halos due to the factions.
2. Show the price above the system (while name is shown below)
3. Have the colour of the ball inside the system represent the price (gray for no commodity trading systems)
4. Have the map change per commodity.

agb32:
This I could do.  But the question is, how to get the player to switch into that mode in the first place. We would need a button of some sort (a hidden key press not great), and preferably not a button per commodity - doable at the moment, but not if we expand to lots of commodities.
Currently, the button will cycle through the commodities, showing the image of the current commodity on the button.  But this doesn't look great, and I'm not an artist!
Point 3 wouldn't be great for colour blind people - which is why I tried adding size and location of the balls (size to show amount of profit/loss, and position to show whether a profit or loss).  But having the price difference (profit/loss) as you suggest in 2 would help here.

So what I suggest:
1. Implement what you've suggested.
2. Keep the button as it is for now, until we have some better idea, or at least better images (to allow cycling through the commodities, and back to no commodity info).
3. Whenever the map is displayed, initialise in the no-commodity mode, so that faction colour haloes are shown.
4. Colour of ball inside system from green (profit) -> orange (no profit) -> red (loss) and grey (not traded here)
5. No change in size.
6. Hide faction colour halos when displaying commodity info.
7. Show price difference above system, name below.
6. Show the profit/loss as text above the system, with name below.

agb32:
ok - I've made those changes.
image

This is what it looks like now - much clearer.

bobbens:
Some comments:
1. Instead of using the same button for everything, I would make it so that when you click on the trade thing, another button pops up to select the commodity from a drop-down list. You should probably hide commodities that aren't known about (like Gold). This should reduce number of clicks needed to do stuff.
2. Another option would be to change the "Trade" button to "Mode" which would list modes like "Travel" (default), "Trade - Food", "Trade - Ore", etc... That way there would be no need to add another button. When you click the button you get a drop-down list to select what mode you want (list widget).
3. The price number should be aligned with the system name.
4. Price should probably be in % not in absolute value
5. Instead of displaying the relative price to the selected system, it should probably be displayed with respect to the global average..

agb32:
As suggested, I now use a list widget.  Click "Mode", and this pops up.  Click "Mode" again, and it goes away.
It then lists Travel (default), and the various commodities, 2 modes for each - the raw price, or trade, which shows price difference between selected system.
I kept absolute value rather than %, because I thought that would actually be more useful... I'd like to know that trading medicine at a 5% profit will actually give me more income than trading food at a 20% profit, for example.
Only shows commodities that are known about at the time (so e.g. not gold).
The price is now aligned with system name.
In price mode, price is shown relative to a global average - green is more expensive, red is cheaper - and the global average price is given.
Hopefully this will be good enough?

agb32:
Hang on - slight change required...

agb32:
ok - now ready to go...

agb32:
@bobbens can you accept this please, and then I'll take a look at the unidiff issue... thanks.

dat/events/neutral/npc.lua
```diff
@@ -205,7 +205,29 @@ msg_jmp =                  {_([["Hi there, traveler. Is your system map up to da
...
+                           _("I recently had a failed trade on %s (%s system).  Prices were:%sSince then, I've done some reading around, and found out that if you hold down shift while clicking the Trade button on the map screen, you can scroll through commodities in reverse.  If you hold down control while clicking, you can toggle the current commodity on and off, which is useful if you want to find out how prices compare to the selected system.  Useful to know!"),
+                           _("The Trade button on the map screen will show you price differences between the currently selected system and others around (assuming you have an idea what prices are there!).  It shows the smallest difference and the largest difference, as two separate circles, i.e. from the planet with highest or lowest prices within the selected and other systems.  I was able to make a good trade on %s in the %s a few STP ago, when prices there were:%sIf you are currently landed in the selected system, then the map will show price differences with this planet only - useful if you're deciding what to buy."),
+                           }
+
```
onpon4, line 230: This should be re-removed, these are strings that aren't used anymore.

agb32:
Not sure how they crept back in - now gone.  At some point, I might put them back as tips (i.e. about how the economy works, how to use the map etc without specific prices).

bobbens:
I think it's good enough, just one small change. The "XXX: Cost" map views should make it so that the default colour when price==average is a neutral colour, price>average becomes some colour (not sure if green is best, because it depends if you're selling or buying), and when price<average it becomes an opossite colour. I find that it is now all reddish even when below/above average, or is this with respect to current system? If so, I think it should be absolute, and the "XXX: Trade" should be the relative one.

dat/events/neutral/npc.lua
```diff
@@ -205,7 +205,8 @@ msg_jmp =                  {_([["Hi there, traveler. Is your system map up to da
                               _([["True fact, there's a direct jump from here to %s. Want to know where it is? It'll cost you! Ha ha, just kidding. Here you go, I've added it to your map."]]),
                               _([["There's a system just one jump away by the name of %s. I can tell you where the jump point is. There, I've updated your map. Don't mention it."]]),
                            }
-            
+
```
bobbens, line 208: Whitespace should not be modified if possible.

src/commodity.c
```diff
@@ -149,6 +149,29 @@ Commodity* commodity_getW( const char* name )
...
+/**
+ * @brief Gets a commodity by index.
+ *
+ *    @param indx Index of the commodity.
+ *    @return Commodity at that index or NULL.
+ */
+Commodity* commodity_getByIndex( const int indx )
+{
+   if ( indx < 0 || indx >= econ_nprices )
+      return NULL;
```
bobbens, line 171: Should also spew a warning.

src/map.c
```diff
@@ -51,11 +51,18 @@ static int map_selected       = -1; /**< What system is selected on the map. */
...
+static Commodity **commod_known = NULL; /**< index of known commodities */
+static int nmap_modes = 0; /**< number of map modes (depends on number of commodities) */
+static char** map_modes = NULL; /**< Holds the names of the different map modes. */
+static int listMapModeVisible = 0; /**< Whether the map mode list widget is visible. */
+static double commod_av_gal_price = 0; /**< Average price across the galaxy. */
 /* VBO. */
 static gl_vbo *map_vbo = NULL; /**< Map VBO. */
 static gl_vbo *marker_vbo = NULL;
-
+//static unsigned int listwid=0;
```
bobbens, line 65: If not needed should be removed.

src/map.c
```diff
@@ -287,9 +308,12 @@ void map_open (void)
...
    window_addButton( wid, 80, 20, 30, 30, "btnZoomOut", "-", map_buttonZoom );
    /* Situation text */
-   window_addText( wid, 140, 10, w - 80 - 30 - 30, 30, 0,
-         "txtSystemStatus", &gl_smallFont, &cBlack, NULL );
+   window_addText( wid, 140, 10, w - 80 - 30 - 30 - 60, 30, 0,
+                   "txtSystemStatus", &gl_smallFont, &cBlack, NULL );
+
+   map_genModeList();
 
+   //window_addImage( wid, -23-3*(BUTTON_WIDTH+20), 23, 44, 24, "imgCommod", NULL, 0 );
```
bobbens, line 316: Should be removed.

src/map.c
```diff
@@ -1219,66 +1543,70 @@ static int map_mouse( unsigned int wid, SDL_Event* event, double mx, double my,
    StarSystem *sys;
 
    t = 15.*15.; /* threshold */
-
+
```
bobbens, line 1546: Weird whitespace was added here.

src/map.c
```diff
@@ -1219,66 +1543,70 @@ static int map_mouse( unsigned int wid, SDL_Event* event, double mx, double my,
    StarSystem *sys;
 
    t = 15.*15.; /* threshold */
-
+	
    switch (event->type) {
-
-      case SDL_MOUSEWHEEL:
```
bobbens, line 1225: It seems like the indentation was changed weirdly in this block.

src/map.c
```diff
@@ -1219,66 +1543,70 @@ static int map_mouse( unsigned int wid, SDL_Event* event, double mx, double my,
...
+   case SDL_MOUSEMOTION:
+      if (map_drag) {
+         /* axis is inverted */
+         map_xpos -= event->motion.xrel;
+         map_ypos += event->motion.yrel;
+      }
+      break;
    }
-
+
```
bobbens, line 1609: Added unnecessary whitespace.

src/map.c
```diff
@@ -1312,10 +1640,166 @@ static void map_buttonZoom( unsigned int wid, char* str )
...
+   }
+}
+
+/**
+ * @brief Updates the map mode list.  This is called when the map update list is clicked.
+ *    Unfortunately, also called when scrolled.
+ *    @param wid Window of the map window.
+ *    @param str Unused.
+ */
+static void map_mode_update( unsigned int wid, char* str )
```
bobbens, line 1701: Function name should be "map_modeUpdate" to be consistent with the rest of function names.

src/map.c
```diff
@@ -1312,10 +1640,166 @@ static void map_buttonZoom( unsigned int wid, char* str )
...
+/**
+ * @brief Handles the button commodity clicks.
+ *
+ *    @param wid Window widget.
+ *    @param str Name of the button creating the event.
+ */
+static void map_buttonCommodity( unsigned int wid, char* str )
+{
+   (void)str;
+   //int ncommod = commodity_getN();
```
bobbens, line 1732: If unecessary should be romev.

src/map.c
```diff
@@ -1948,7 +2432,7 @@ void map_show( int wid, int x, int y, int w, int h, double zoom )
       map_selectCur();
 
    window_addCust( wid, x, y, w, h,
-         "cstMap", 1, map_render, map_mouse, NULL );
+                   "cstMap", 1, map_render, map_mouse, NULL );
```
bobbens, line 2435: Indentation has been changed. Should be restored.

onpon4:
> or is this with respect to current system?

Unless it's changed since I last checked it's compared to whatever system you have selected.

agb32:
Sure, I'll make those changes.  Re the "I find that it is now all reddish even when below/above average, or is this with respect to current system?" - I think this is an artifact of using the mean - It is possible for prices to go high (i.e. well above the mean) but not low by the same amount (since we don't have negative prices).  Therefore, one or two high prices pull the mean up, meaning that the rest are below this and appear red.  Could potentially use the median price - though that probably wouldn't tell people what they are expecting to see.
How about blue -> grey -> yellow ? (where blue replaces red, yellow replaces green?)

agb32:
> or is this with respect to current system?
In cost mode, it is absolute, in trade mode, it is wrt the selected system.

bobbens:
In the below example it's all orange. It should show the colour based on the difference with average price.
foo

agb32:
Hmm - but if use grey for a neutral colour, how to differentiate between "not sold here".  Darker grey for that, perhaps.

Zireael07:
Empty circle for not sold here, maybe?

agb32:
Except that is also places with no landable assets...

bobbens:
Usually yellow is being used as neutral in Naev. Grey is used to mean "inert"

agb32:
Orange -> yellow -> blue ?   (since orange is the opposite colour of blue)

bobbens:
Using a diverging palette from https://seaborn.pydata.org/generated/seaborn.diverging_palette.html#seaborn.diverging_palette would probably be good. With regard to grey being planet being asset without commodities, I would just make those hollow. As with respect to commodities, having no planets or no commodities is the same.

agb32:
Ok - now going from orange -> yellow -> blue.  All the whitespace fixes made etc, and changed the scaling a bit, so should be a bit more obvious.

src/map.c
```diff
@@ -1439,21 +1436,19 @@ void map_renderCommod( double bx, double by, double x, double y,
                best = maxPrice - curMinPrice ;
                worst= minPrice - curMaxPrice ;
                if ( best >= 0 ){/* draw circle above */
-                  gl_print(&gl_smallFont, x + (sys->pos.x+11) * map_zoom , y + (sys->pos.y-22)*map_zoom, &cGreen, "%.1f",best);
-                  best = tanh ( best / 100. ) /2 ;
-                  /*ccol.r=best+0.5; ccol.g=best+0.5; ccol.b=0.5-best; ccol.a=1;*//*yellow*/
-                  ccol.r=1-2*best; ccol.g=best+0.5; ccol.b=0; ccol.a=1;/*green to orange*/
+                  gl_print(&gl_smallFont, x + (sys->pos.x+11) * map_zoom , y + (sys->pos.y-22)*map_zoom, &cBlue, "%.1f",best);
+                  best = tanh ( 10*best / curMinPrice );
+                  ccol.r=1-best;ccol.g=1-best;ccol.b=best;ccol.a=1;/*yellow (0) to blue (1)*/
```
onpon4, line 1441: This is practically unreadable. You really should break these up into separate lines and give more space.

```
ccol.r = 1 - best;
ccol.g = 1 - best;
ccol.b = best;
ccol.a = 1;
```

This is a general criticism by the way, I've seen a lot of code from you that compacts down code far too much at the expense of readability, and I've corrected dozens of these. A few bytes are not worth making the code harder to read, especially because (in the words of the Python style guide):

> code is read much more often than it is written.

agb32: Changed to a macro.

bobbens:
Two comments:
1. The blue font you are using is fairly unreadable when dark. (side note, I think we need to make the dark blue of the hyperlanes also easier to see)
2. I find that when the planet does not have the commodity, the other assets are shown in colour that I don't really understand. Furthermore, the "no price info about food here" is fairly easy to miss. Maybe display it near the list window or something like that?
foo

agb32:
Perhaps continual tweaking with the colour could be done after a pull, once others have suggestions?  Anyway, I've changed the fount colour to light blue now.
The "No price info about food here" is on the selected system.  It tells you that you can't trade that item from the selected system.  Seemed to be the obvious place - though I guess it could be moved elsewhere.  But I'd suggest leave it for now, and let others have input.

agb32:
> I find that when the planet does not have the commodity, the other assets are shown in colour that I don't really understand
Ah - these are just the default map colours that show whether you're friendly or not with a faction.  I've only seen them all yellow because within my savefile, I've not yet visited any hostile systems.  I guess they should all just turn black or something.

Zireael07:
> I guess they should all just turn black or something.
This, most likely.

agb32:
ok - in trade mode, if selecting a system without the selected commodity, it now renders all systems black, to signify no trade.

src/map.c
```diff
@@ -1438,17 +1439,17 @@ void map_renderCommod( double bx, double by, double x, double y,
                if ( best >= 0 ){/* draw circle above */
                   gl_print(&gl_smallFont, x + (sys->pos.x+11) * map_zoom , y + (sys->pos.y-22)*map_zoom, &cBlue, "%.1f",best);
                   best = tanh ( 2*best / curMinPrice );
-                  ccol.r=1-best;ccol.g=1-best;ccol.b=best;ccol.a=1;/*yellow (0) to blue (1)*/
+		  setcolour(1-best,1-best,best);/*yellow (0) to blue (1)*/
```
onpon4, line 1475: You used tabs here, should be spaces

src/map.c
```diff
@@ -1438,17 +1439,17 @@ void map_renderCommod( double bx, double by, double x, double y,
...
                   gl_drawCircle( tx, ty /*+ r*/ , /*(0.1 + best) **/ r, &ccol, 1 );
                }else{/* draw circle below */
                   gl_print(&gl_smallFont, x + (sys->pos.x+11) * map_zoom , y + (sys->pos.y-22)*map_zoom, &cOrange, "%.1f",worst);
                   worst= tanh ( -2*worst/ curMaxPrice );
-                  ccol.r=1;ccol.g=1-worst/2;ccol.b=0;ccol.a=1;/*yellow (0) to orange (1)*/
+		  setcolour(1,1-worst/2,0);/*yellow (0) to orange (1)*/
```
onpon4, line 1480: Tabs are here as well, should be spaces

agb32: Now removed

PR #1089, opened 2020-06-22
Added some more NPC hints and increased the chance of getting them.

onpon4:
I was originally going to add an extended tutorial but wasn't able
to figure out a way to do that in a way that it wouldn't just be a
giant, confusing exposition. So instead I added some more NPC hints
explaining the things I want explained with a bit of extra emphasis
during the tutorial that talking to NPCs is a good idea.


bobbens:
Looks good, thanks!

PR #1090, opened 2020-06-23
support for commodities sellable and buyable anywhere

nenau-again:
It's an xml tag in the commodity definition `<standard/>` that makes the commodities sellable and buyable anywhere there is a commodity market.

The goal is to make stuff like https://github.com/naev/naev/pull/947 easier to manage.

I made Ore and Gold standard (but it's only for demonstration). If people agree on this one, I'll make Food, Ore, Medecine, industrial and luxury goods standard and I'll remove them from the assets definitions.
What is more, if people are interested, I can add a tag to prevent a standard commodity to be sellable/buyable at a particular place.

Note : In order to avoid a crash, I had to remove the `<commodity/>` tag from Eiroik and an other (probably mission-related) not-inhabited asset. I hope this will not break things somewhere else. And I added a warning to help people if they get the same crash as I had (it took me hours to figure out the reason)

### DIFF
--- a/dat/assets/eiroik.xml
+++ b/dat/assets/eiroik.xml
@@ -15,7 +15,6 @@
   <services>
    <land/>
   </services>
-  <commodities/>
   <description>Eiroik was a former mining planet under Empire rule, known for its toxic atmosphere and difficult, rough terrain produced through millennia of volcanic activity. The few human colonies were eradicated by the Collective. The generators have failed and the former spaceport lies in ruins. All that remains is dust.</description>
  </general>
 </asset>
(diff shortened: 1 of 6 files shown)


onpon4:
YESYESYESYES

One thing that I'd like to see as well is to redefine commodity tags to mark a commodity as cheap in a particular planet (and have a corresponding way to mark it as expensive on a particular planet). A bit of support for human intervention on the pricing.

An empty commodity definition tag causing a crash on those planets seems to be indicative of a problem though, is there a particular reason that happens? Maybe a memory leak somewhere?

onpon4:
OK, so I looked into this and sent a pull request your way, @nenau-again. Bet you didn't expect that. :slightly_smiling_face:

https://github.com/nenau-again/naev/pull/1

bobbens:
Looks good, thanks! Pulled from @onpon4 's blanch.

PR #1097, opened 2020-06-23
Some ambient factional music for the Thurion.

onpon4:
Comes from: https://opengameart.org/content/motherload

Was just looking through music on OGA and thought this one sounded like a good fit for Thurion space. I just re-exported the wav in Audacity to Ogg Vorbis and made it so that the Thurion would use it.


bobbens:
Looks good, thanks! Should probably be added to naev-artwork

onpon4:
Sure! Added to https://github.com/naev/naev-artwork/pull/22 (you still haven't pulled that :wink:)

PR #1099, opened 2020-06-23
Keep Collective system planets around after Collective dead.

### LINKED ISSUE #1092: Planets in Collective systems disappear after defeating the Collective
After you defeat the final Collective mission (the one that ultimately destroys the Collective faction), almost all planets (not just space stations) disappear from the collective Systems (C-00, C-14, ...), with only one exception: Eirok in C-43.

It makes sense that the space stations disappear, I suppose they are meant to be factories that got destroyed by the Empire.
But it does not make sense that all the planets disappear as well.

f502b70c4c20f517f0bec79de222c1897696ba92

onpon4:
Fixes #1092.

### DIFF
--- a/dat/unidiff.xml
+++ b/dat/unidiff.xml
@@ -2,27 +2,21 @@
 <unidiffs>
  <unidiff name="collective_dead">
   <system name="C-00">
-   <asset name="P0125">remove</asset>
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good ,thanks!

PR #1102, opened 2020-06-23
Don't warn about resolution change being impossible if none attempted

### LINKED ISSUE #1094: Resolution warning when confirming settings without changing anything.
If I open the Options, do nothing and click OK, I always get the warning the resolution cannot be changed while it is maximized.
But I didn't touch the resolution, nor do I play in windowed mode.

For the record, my settings are: fullscreen @ 1920×1080.

Random guess: Could it be that Naev uses a “fake fullscreen” and Naev then thinks it is a maximized window?

f502b70c4c20f517f0bec79de222c1897696ba92

onpon4:
Fixes #1094.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -1379,21 +1379,23 @@ static int opt_videoSave( unsigned int wid, char *str )
          SDL_SetWindowFullscreen( gl_screen.window, 0 );
    }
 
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good, thanks!

Issue #1078, opened 2020-06-19
Game is very very fast with big ships


Wuzzy2:
For some reason, when I have a big ship like the Kestrel the game becomes very fast but no speedup multilpier is displayed. It's so fast that it's pretty hard to react to enemy actions, to battle, and so on. Weird.

0.8.0-beta1

onpon4:
That's intentional. It's the time dilation effect of ships, which you can see in the shipyard. Big ships have time pass faster, small ships have it pass slower. (In fact the Hyena actually makes time run at 75% speed.)

The reason for this is because big ships' slow speed can be very tedious at normal time passage rate, and it's justified in-universe as the big ships being harder to control than smaller ships. (So the being more difficult to react effect is also intentional; it's an added benefit of flying smaller ships.)

The Kestrel actually doesn't have the highest time dilation, carriers like the Peacemaker do. Kestrels have a time compression rate of 250%. (You can look at and compare time dilation rates in the shipyard, as mentioned.

Anyway, I would recommend the Hyena if you need more reaction time, or possibly a Corvette class ship if you want something bigger (which only has 150% time dilation).

onpon4:
I've adjusted the Kestrel's time dilation a bit in #1069. Made it just 225% (instead of 250% like other Cruisers) since the Kestrel is generally designed to be faster than other Cruisers. I think this helps make it feel more natural.

Wuzzy2:
Ugh. Then this is just a horrible mechanic IMO.

I mean, speeding up THE ENTIRE GAME by 250% or even more, at all times, even if you're not moving, just because your ship is big, is absurd.
This makes direct combat basically impossible. You could maybe still fight 1-2 enemies, but taking out large numbers of enemies will be very hard, even if you had more than enough firepower to defeat them all, simply because your reaction time is not fast enough. In other words, the benefit of having a large warship is nullified. Not good.
Also note that 250% is a very extreme number. Heck, even 110% would probably be noticable.
That it also goes the other way is … weird.
If you have like 5 enemies, how are you supposed to concentrate even on a single target at this game speed? And then you told me that 250% was not even the end of the line … Oh boy.

Seriously, where is the fun in that? Combat is already hectic enough at 100% speed, thank you very much. I mean, the whole concept of speeding up LITERALLY THE ENTIRE GAME depending on the size of your ship ... it just doesn't make any sense to me.

[...]

PR #1104, opened 2020-06-24
Added a quick tutorial event for when you first buy a large ship.

onpon4:
This tutorial event briefly explains how time dilation works when
you buy your first large ship. It's only a single message window,
so it shouldn't be a major bother to those who already know. I've
also integrated it in as a concerned salesperson trying to provide
good customer service to explain the sudden message away.

The message also explains that larger ships should generally use
turrets, since that's related to the Time Dilation issue.

This may help with #1078; the Time Dilation was not explained
particularly well and this ensures that an explanation is always
given when it becomes an important point.

### DIFF
--- a/dat/event.xml
+++ b/dat/event.xml
@@ -1,12 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Events>
- <event name="Time Dilation Tutorial">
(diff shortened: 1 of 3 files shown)


bobbens:
Looks good, thanks!

PR #1105, opened 2020-06-24
Changed rep gains for the FLF campaign.

onpon4:
Made rep boosts smaller so the player has to do more of the minor missions (more in-line with how the main factions operate).


bobbens:
Looks good, thanks!

PR #1107, opened 2020-06-24
Added access to the jump nearby Kretogg for the pirates.

onpon4:
Seems to make sense as a point for getting into Za'lek space.


bobbens:
Looks good, thanks!

PR #1108, opened 2020-06-24
Standardize commodities

nenau-again:
With this PR, the main commodities are standard. Only gold is non-standard and only sellable on Bon Sebb. I did this mainly in order to have a remaining occurence of the commodity tag somewhere.

I used the map editor to change the files, and some systems files have been modified as well (it's just a question of format apparently). By the way, the system editor created a `virtual_simbad.xml`. I did not keep it because I remembered there was mess around this one.

### DIFF
--- a/dat/assets/aillis.xml
+++ b/dat/assets/aillis.xml
@@ -24,10 +24,7 @@
    <missions/>
    <commodity/>
   </services>
-  <commodities>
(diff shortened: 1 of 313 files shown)


bobbens:
Can you get gold from asteroids? If so it should be also made standard.

PR #1109, opened 2020-06-24
made kidnapped mission avaliable later

nenau-again:
Now that the medusa launchers are medium size, this mission is really hard on a Llama. Maybe it should be proposed to the players only once they have started to get interested enough in the weaponry to buy a combat vessel license ?


bobbens:
Looks good, thanks!

PR #1110, opened 2020-06-24
added more types of asteroids

nenau-again:
Basically a reimplementation of https://github.com/naev/naev/pull/947 with the standard commodity system.
I added a few water and diamond asteroid fields here and there.
Golden asteroids are only in Zied and Sol. gold is still non-standard because I wanted to keep at least one non-standard commodity.
I guess the price of commodities should be rebalanced. Right now, ore, water and diamond have similar prices, but I don't know what could be the expensive commodity... An obvious answer could be diamond, but I found it funny to think that maybe one day water will be more expensive than diamond...


bobbens:
Looks good, thanks!

PR #1111, opened 2020-06-24
Added use of the intro music for the Thurion intro.

onpon4:
Drama! Also, you *are* being introduced to the Inner Nebula, so justified!

Also fixed a bug that would have caused the event to crash if you bailed on the Thurion discovery.


bobbens:
Looks good, thanks!

PR #1112, opened 2020-06-24
Made it so that right-click cannot select a new target.

onpon4:
This fixes an annoyance introduced by the ship-following before: it
usually just gets in the way of autonav since you aren't concerned
about the the pilots all around the planet you're trying to land on.
This fixes that: in order to select a pilot you must first
left-click it to indicate your intention to target it (after which
it has priority over planets for the right mouse button).


bobbens:
Looks good, thanks!

PR #1113, opened 2020-06-24
Removed the Commodity Run missions.

onpon4:
These missions don't work anymore because commodities are available
everywhere (just at different prices). There is no way to salvage
this; it will always be possible to take advantage of jacked-up
prices by buying a commodity at the same planet, taking off, and
landing (this is why they don't spawn anymore, the mission itself
prevents this by not allowing the mission in places where the
commodity is available).

This should probably be replaced by some sort of mechanism to make
commodities specially cheap or expensive, just as an event. But for
now, the Commodity Run mission is an unsalvagable deadweight, so
this removes it.


nenau-again:
What about making the commodity run availiable to planets without the commodity service ?

onpon4:
In theory possible, though that only works in areas where no commodities are available in the current system. I wouldn't be opposed to that though if we want to intentionally make commodities unavailable in some systems.

bobbens:
Looks good, thanks!

PR #1120, opened 2020-06-26
Various small fixes

onpon4:
* Un-gendered the pirate informer in the ship stealing mission (so female pirates can still be chosen without misgendering)
* Removed references to STU/STP/SCU in the source code (with the exception of their use in save files, that can't be changed for obvious compatibility reasons)
* Fixed some bad whitespace


bobbens:
My heart is heavy at the loss of my old terminology, but I guess it does make things more simple. Thanks!

PR #1121, opened 2020-06-26
Improved the way ambient music is selected.

onpon4:
Specifically this makes the ambient music get chosen in the same way combat music does: based on presences rather than who "owns" the system (meaning you can hear factional ambient music in empty but factionally dominated systems), and with support for explicitly marking which factions get neutral music too (rather than the Collective being a specialized exception).

This also adds some factional ambient music for the pirates. Sourced from:

https://opengameart.org/content/pirates-orchestra-208
https://opengameart.org/content/pirate-theme


bobbens:
Looks good, thanks!

PR #1123, opened 2020-06-27
Added some unique Pirate battle music.

onpon4:
At first I thought this would be tough to use but turns out Naev
has a credits screen I didn't know about. xD So yeah, here's some
unique pirate music!

c.f. https://opengameart.org/content/blackmoor-tides-epic-pirate-battle-theme


bobbens:
Looks good, thanks!

PR #1124, opened 2020-06-27
Various text improvements throughout missions and stuff.

onpon4:
Started off with getting rid of references to "minutes" and "hours",
but I decided to also go through and remove a bit of gendering, plus
remove some sanist language (I basically just did grep searches
for problematic terms and replaced them with alternatives, most
notably "insane" and "crazy"; luckily there wasn't that much and
a lot of what was there was my fault).


bobbens:
Looks good, thanks?

PR #1125, opened 2020-06-27
added zalek scientific chatter at bar

nenau-again:
I hope what they say is not too confusing...


PhoenixRiver:
> I hope what they say is not too confusing...

I like them :)

River

bobbens:
I like them, but I would probably drop the reference to NLGMRES , or at least do something more general. I don't want to scare off people too much :P

nenau-again:
Mm. I actually thought such a method would not exist, and I put a random name. But apparently, after checking on scholar, the nonlinear GMRES exists... And now I want to try it out in my research.
Anyway, what about HO4H-GMRES (for High Order Hyperreduced, Hyperaccelerated, Hypertruncated, Hyperreduced once more Generalized Minimization of RESidual) ? This acronym sounds too stupid to be real, no ?

There could even be some joke with some labs using HO4H-GMRES, while others use PMLKQR-GMRES and each claim their method is obviously better than the other one. And complain that Zlk central has created one post of professor in the lab that develops HO4H-GMRES and no post in the other lab, which is clearly unfair because last cycle, the zalek spacenavy computation center already recruted someone to work on HO4H-GMRES.

And so on...

nenau-again:
Removed mention to GMRES, and added 3 other lines.

Wuzzy2:
> It's not only trivially because of the question of local minima, but also [...]

Shouldn't it be “It's not only *trivial* because [...]”? I'm not a native English speaker, however, so I might be wrong.

nenau-again:
> I'm not a native English speaker

Neither am I. I guess we're doomed.
I'm pretty sure 'trivially' exists in english (in german, it would be 'offensichtlich'). At least, it exists in the international ersatz of english that people use in scientific communication. It's mainly used in proofs of theorems. For example : 

"\begin{proof}
 The lemma is trivially derived from theorem 4.1 in \cite{my_supervisors_paper}
\end{proof}"

Wuzzy2:
Well, at least I know that “trivially” is an adverb (“trivial” + “-ly”) so it must stand before a verb. Your example is correct and has nothing to do with science. In the sentence I attacked, the word “trivially” does not stand before a verb, that's why it looks wrong to me.

nenau-again:
Mm. I guess I will simply remove the word then, as it makes the sentence longer while barely adding anything. (I just wanted the zlk to sound like those unbearable people in conferences, who use the 'trivial' word 3 times in each sentence)

Wuzzy2:
Let's wait for a native English speaker to appear. :D

PhoenixRiver:
Hello :)

> Let's wait for a native English speaker to appear. :D

Apologies I took a while! ;)

I assume it is this sentence you are talking about?

"Do you know why many optimization algorithms require your objective
function to be convex? It's not only trivially because of the question
of local minima, but also because if your function is locally concave
around the current iterate, the next one will lead to a greater value
of your objective function. There are still too many people who don't
know this!"

TLDR; In spoken context this is entirely good grammer (and why I
didn't say anything about it), in spoken context between those sorts
of people that annoy people who are not up-to-date with the rubric,
jargon and language usages of their particular miniture niche of a
discipline it is rather wordy - but then it is a complaint so the
person being spoken to might not be so specifically in their niche of
a discipline and maybe the speaker also likes the sound of their own
voice... (guess who was attendant to too many of these converssations
wihtout being a specialist in the discipline of topic! ;)

TL; The grammer, standard replacements and contractions of English
mean that this use of "trivially" is both correct but confusing... The
full sentence in question in a textbook should be rendered as "It is
not only trivially _so_ due to the topic of local minima but also ...
blah blah"

The "topic" is replaced by "question" because it more specifically
refers to what the sentence is dealing with, so much so this
replacement has become standard even though it would be a little
suspicious (proofreading term to refer to words that are not strictly,
100%, always usable in written non-dialect text) in textbook English.
"due to" rather than "because of" is the correct strict grammer term
here, however standard usage would allow both, especially as the term
"because of" is used later with an also for emphasis and that emphasis
works better if the term "because of" is reused - this is a fiction
writing technique that should not be used in a written scientific
paper but outside of that context is entirely standard. The "so" is
dropped because it is easily assumed but is also the verb that the
adverb "trivially" is referring to hence causing confusion amongst
people who know English grammer better than standard usage in English
speaking countries ;) - who also probably know English grammer better
than native speakers too, for that matter! :D

Hope this helps :)

River *<:@)

nenau-again:
If I understand well, the sentence is correct, but weird, which is what was intended.

> that annoy people who are not up-to-date with the rubric

It has to be decided if we really want to annoy the player... I don't know in that case if it sounds funny or just... well... annoying.

Wuzzy2:
I don't think the sentences are annoying. I, the player, chose to talk to the people, after all.
In fact, I think these sentences are quite hilarious. Content-wise, I have 0 problems with it.

PhoenixRiver:
> It has to be decided if we really want to annoy the player... I don't know
> in that case if it sounds funny or just... well... annoying.

It amused me... :)

onpon4:
I've fixed all typos here:

https://github.com/nenau-again/naev/pull/2

🕷️

PR #1131, opened 2020-06-29
re-compute economy after each unidiff

nenau-again:
I could not reproduce the -\infinity problem. The only issue I had was that the commoditie's prices at Sindbad was 0 (and when selling I gained 0 as well). Now, Sindbad's prices are computes the same way as the other ones.


nenau-again:
Strangely, this seems to have been commited while no-one pulled it. Not sure what happened...

bobbens:
Looks good, thanks! Still have to clean up the economy stuff a bit.

PR #1132, opened 2020-06-29
Added some colonies for the Proteron

onpon4:
This adds a series of post-Incident colonies surrounding Protera. Makes the idea of a Proteron invasion seem far less absurd and also will enable pro-Proteron missions better.

Also adjusted the Kahan to have large-sized rather than medium-sized slots, since being all medium-sized made it markedly less powerful than other Destroyers. I might look into making other tweaks to the Proteron ships later but this was the most egregious problem.


bobbens:
Looks good, thanks! Side note, proteron need map graphic!

PR #1133, opened 2020-06-29
Fixed the way the game determines if commodities should be added, restored Commodity Run

onpon4:
The previous way it was doing things, commodities were added to planets if (and only if) they had a commodities list, whether it had entries or not, regardless of whether the commodity service was available. There are a lot of problems with that, so I changed it so that commodities get added to any asset with the commodities service, instead.

Also restored the Commodity Run mission for use where commodities are not available at all.


bobbens:
Looks good, thanks!

PR #1134, opened 2020-06-29
Adjusting commodity prices

onpon4:
This is an attempt to adjust the various price variation factors to something that makes sense. This is a draft right now though because the results aren't actually what I'm going for.

I'm going to need some help understanding what sorts of numbers to pick to get the expected result. I assumed that I would just want to use e.g. 1.2 for a 20% increase and 0.6 for 60% the base price. That sort of thing. But the changes I'm actually seeing are way out of proportion of what I'm expecting.

### DIFF
--- a/dat/assets/bon_sebb.xml
+++ b/dat/assets/bon_sebb.xml
@@ -26,9 +26,7 @@
    <outfits/>
    <shipyard/>
   </services>
-  <commodities>
-   <commodity>Gold</commodity>
-  </commodities>
+  <commodities/>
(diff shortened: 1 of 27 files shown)


onpon4:
Turns out there was a bug that caused repeated recalculations to compound. Fixed that, and now this is ready for merge. :slightly_smiling_face: Prices seem to vary in a reasonable manner from what I can tell.

bobbens:
Looks good, thanks! Maybe we need some sort of sanity checks to make sure everything doesn't get too weird.

PR #1138, opened 2020-06-30
Tweaked the commodity modifiers a bit.

onpon4:
I noticed that the population modifier I gave for food led to incredibly small values for stations with low population, so I changed it from 1 to 0.75. This still leads to very low prices with low populations, but in the realm of dozens rather than easily going as low as 1 or 2.

Also added a thing to make sure the price will never go below 1.


bobbens:
Looks good, thanks!

PR #1139, opened 2020-06-30
Added Virtual Thurion Local assets to inter-Thurion Nebula systems.

onpon4:
These assets add a tiny presence (10) to these systems, causing
Thurion music to be played and also making these systems seem less
empty.


bobbens:
Looks good, thanks! What we would need is some more nebula music also.

PR #1140, opened 2020-06-30
Added expensive minerals to more asteroid fields.

onpon4:
There were a bunch that *only* had ore, so I changed that. All
fields now have at least two kinds of commodities that can be
extracted, making the Asteroid Scanner more useful and the mining
experience a bit more dynamic in these systems.

### DIFF
--- a/dat/ssys/amatiri.xml
+++ b/dat/ssys/amatiri.xml
@@ -32,6 +32,7 @@
    <type>default</type>
    <type>default</type>
    <type>massive</type>
+   <type>golden</type>
    <radius>3000.000000</radius>
    <pos x="4000.000000" y="-1000.000000"/>
    <density>1.000000</density>
--- a/dat/ssys/arcanis.xml
+++ b/dat/ssys/arcanis.xml
@@ -34,6 +34,10 @@
  </jumps>
  <asteroids>
   <asteroid>
+   <type>default</type>
+   <type>default</type>
+   <type>default</type>
+   <type>massive</type>
    <radius>4000.000000</radius>
    <pos x="4000.000000" y="6000.000000"/>
    <density>1.000000</density>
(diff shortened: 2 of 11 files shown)


bobbens:
Wasn't gold not a standard commodity? If you add gold it should become a standard commodity, althou @nenau-again 's reason for not having gold is that he wanted at least one non-standard commodity.

nenau-again:
> he wanted at least one non-standard commodity.

True. I think we should still use at least one `commodity` tag somewhere. But that could be anything else. Actually, I think we should have at least one minable commodity that is much more expensive than ore.

onpon4:
Is there a particular reason to have gold (specifically) as non-standard? Wouldn't it make more sense to have something like Industrial Goods or Luxury Goods as non-standard?

I also don't really like - in our current system - the meaning of "non-standard". There's currently no way to define someone as *only* a buyer or *only* a seller. My best guess for how to achieve this is instead with tags that increase or decrease price. But if you consider gold, for example, just defining certain planets as trading gold doesn't give any meaning to who *uses* and *produces* gold.

If we do want manually defined commodities then I would think the reason for that should be that it's a commodity that's only traded at all in certain places. Like contraband, for instance. That could make sense (available only at pirate strongholds and certain Independent planets).

But I don't get the idea of making a commodity non-standard *for the sake* of having non-standard commodities. That doesn't make sense to me.

onpon4:
Ah, by the way, this PR shouldn't have any bearing on the issue of gold being standard or not, no? If you find gold in an asteroid it doesn't mean you have to sell it, you can throw it away if it's of no use to you. The main point I had with this PR is making the Asteroid Scanner tool useful.

bobbens:
Looks good, thanks!

PR #1142, opened 2020-06-30
Added support for cargo missions on Proteron worlds.

onpon4:
Also made it so that cargo missions won't choose Thurion assets
unless the mission comes from another Thurion asset. (This prevents
Proteron from shipping goods to the Thurion, who they don't even
know exist.)

### DIFF
--- a/dat/ai/proteron_dissident.lua
+++ b/dat/ai/proteron_dissident.lua
@@ -1,0 +1,37 @@
+require("dat/ai/tpl/generic.lua")
+require("dat/ai/personality/civilian.lua")
+
+
+mem.shield_run = 100
+mem.armour_run = 100
+mem.defensive  = false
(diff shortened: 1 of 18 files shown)


bobbens:
This opens up a question. Do we want Proteron to be just another faction? I was always under impression that they would never be treated as a standard faction, and thus no standard missions or whatever. Of course it's not much of an issue as the player shouldn't be able to make it there anyway, and I believe they are initially hostile either way.

onpon4:
If they're initially hostile would you like to change their default standing with the player? That would be simple enough.

I would think that the Proteron would need deliveries just like anyone else. Unless you envision them not hiring private contractors for anything? In which case a pro-Proteron campaign would be difficult.

bobbens:
Looks good, thanks!

PR #1145, opened 2020-07-06
Resolved unidiff issue

agb32:
The unidiff system was wiping player economy knowledge information.  This pull resolves it.

### DIFF
--- a/src/economy.c
+++ b/src/economy.c
@@ -625,10 +625,10 @@ static int economy_calcPrice( Planet *planet, Commodity *commodity, CommodityPri
    commodityPrice->price *= scale;
    commodityPrice->planetVariation = 0.5;
    commodityPrice->sysVariation = 0.;
-   commodityPrice->sum = 0.;
+   /*commodityPrice->sum = 0.;
    commodityPrice->sum2 = 0.;
    commodityPrice->cnt = 0;
-   commodityPrice->updateTime = 0;
+   commodityPrice->updateTime = 0;*/
    /* Use filename to specify a variation period. */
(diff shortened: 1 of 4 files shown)


nenau-again:
I did not think about that! The economy info was erased at https://github.com/naev/naev/blob/master/src/economy.c#L869, right?
If yes, why not splitting `economy_initialiseCommodityPrices` in two. One part that initialize the prices, and the other one to erase the economy info, and call only the first one after the unidiff.

Because I believe that it's saner to re-generate all prices after each unidiff. As a matter of fact, prices on other assets can be impacted by the appearence or the disappearance of a planet, and switching to a longer-range price computation system would require to recompute everything each time any asset changes.
What is more, unidiffs can do much more than creating planets. One can for example change faction of planets, or suppress assets... 

What do you think ?

agb32:
No, the economy info was erased at   [https://github.com/naev/naev/blob/master/src/economy.c#628](url) and for the next 4 lines.
So, yes, I've now split it - the unidiff reinitialises the whole economy, but the player info is saved.  This is only reinitialised upon load game or new game.

nenau-again:
> reinitialised upon load game

Does it mean that each time I load a game, the player's price knowledge disappear ?
Shouldn't such info be stored in the save file ?

agb32:
No - this info is stored in your save file, so loads correctly.
But, it reinitialises it because if you load a different save file, it all needs to be reset, except for the stuff in the save file...
So - it will work as you expect.  And if you create a new game, if starts with no knowledge...

PR #1149, opened 2020-07-11
make the ambush in seekndestroy actually noticeable

nenau-again:
... and make the scared pilots want to run away after giving an info.


bobbens:
Looks good, thanks!

PR #1150, opened 2020-07-11
corrected problem at landing and jumping for the trader escort mission

nenau-again:
The function `timer_traderSafe` was causing the ai goal to be reinitialized any 2 seconds, and was causing the ai to have problems when landing or jumping.


bobbens:
Looks good, thanks!

Issue #1053, opened 2020-06-16
[Proposal] Ship's Log


ids1024:
It seems to be relatively common in RPG and adventure games to have some kind of "journal" recording important events and information. It could make sense to have something like this in Naev. Calling it a "Ship's Log" seems like it would fit well.

I'm not sure what the best interface would be. But while a chronological log of information could be part of it, it would be nice if there's a quick way to be reminded that the player knows of an odd baron in the remote Ingot system. And things like that, which the player might forget.

bobbens:
Could be something as simple as just active/cleared quests by time with some small comments. Might be a way to try to generate the automatically from OSD, but I guess manual would be better.

onpon4:
Probably the most difficult thing to do would be to prevent useful info from getting lost in clutter about random cargo missions or some such thing, so maybe it would be best, if we're going to do something like this, to have missions define when something should be logged? That way you'll only get something logged when it's important (assuming mission authors use the feature correctly).

bobbens:
I would argue for adding a new API for this. Copying other games (Baldur's gate et al. comes to mind), would mean that the following would probably be decent:

1. Allow missions to create a log
2. Allow missions to add information to the log (it is time stamped, maybe location stamped)
3. Allow missions to finalize the log (either complete or aborted)

One thing that has to be taken into account is that missions can be failed and restarted. Therefore I would propose having the log be tied to the ID of the mission name, so that if you create a log and one already exists with the same ID, it gets recreated and the old one is erased.

I would also argue that it is not probably necessary to be able to edit the log entries. Just have it be additive.

Finally, they should be saved in the save game.

ids1024:
> I would argue for adding a new API for this.

Definitely. That's the only way it will really provide the most useful kind of information.

> I would also argue that it is not probably necessary to be able to edit the log entries. Just have it be additive.

Ideally, we would be able to have typo fixes and such apply to earlier log entries, and switching language would switch the language of the entries. That may be more effort than its worth.

agb32:
So, something like (in C, with lua hooks):
int shiplog_create(char *logname,int overwrite); /* returns a log ID, for log with specified title.  If overwrite set, and title already exists, the log will be cleared and previous log ID returned, otherwise a new log ID will be created. */
shiplog_append(int logid,char *msg);/* Add a message to the log */
shiplog_close(int logid); /* Actually, this probably isn't necessary.*/
The shiplog functions would capture current time and system/planet (if landed) and record these.
[...]

Issue #1077, opened 2020-06-19
Broken news reel in translated game


Wuzzy2:
If you play in a translated version of the game, the news reel goes bonkers.

In the News reel, the News caption text is duplicated in English and German. Also, several news entries themselves appear in English AND German. Some only in one language. It's very chaotic and this doesn't happen if you disable translation.

I haven't found a logic behind this oddidy, but something about the newsreel translation system is definitely very broken. This needs testing.

It's hard to explain, see for yourselves. I tested in Sirius space on planet Pauldur in Nixon system.

0.8.0-beta1

onpon4:
That's because the news event looks for previously published articles and unpublishes them if they match the name of an article that's being published. The code unfortunately doesn't have any way of dealing with that, but it'll only show up if you played under one language and then switched to another language.

(This applies to the news caption too because under the hood that's actually an "article".)

I'm not sure if there's a viable fix for this since the news feed is something that's built on, not just generated. Previously submitted articles stay up until a certain expiration time. And the script only has access to the current language, not other languages that may have been used in other sessions.

Wuzzy2:
Does this mean the savegames save the literal text of the news, and not just a list of something like “news ID” or something like that?

If yes, this is bad design even if we ignore language. If you ever want to change the text of an existing news, it won't apply to old savegames.

onpon4:
> Does this mean the savegames save the literal text of the news, and not just a list of something like “news ID” or something like that?

Yes, that's how it works. There's no ID system for news articles (if there were the text would have to be contained in an XML file somewhere and there would be little flexibility for posting news articles; the economy news articles would be impossible for instance).

> If you ever want to change the text of an existing news, it won't apply to old savegames.

That's not quite true actually, the event only compares titles. Any article with an identical title will be removed and replaced with the new one. The duplication problem only happens when two articles are a version of the same article but have different titles (as would happen if you were to switch languages in the middle of a save, for instance).

bobbens:
This is non-trivial to handle in a super nice way. What could be done is to save the language and remove (or hide) if the game is set in another language.

onpon4:
I just made some changes to t he news event in #1178. It doesn't do everything, but it uses tags instead of article titles to prevent duplicate articles. This means that articles in other languages will remain, but you won't have, say, both the German version and the English version of the same article. So the worst of what you experienced is solved.

[...]

PR #1153, opened 2020-07-14
Shiplog implemented

agb32:
Implemented a shiplog as per proposal #1053 
Lua bindings created, and examples added to a few missions.
Others should feel free to add to other missions once completed.

### DIFF
--- a/dat/events/neutral/npc.lua
+++ b/dat/events/neutral/npc.lua
@@ -248,8 +248,8 @@ function create()
 
    -- Chance of a jump point message showing up. As this gradually goes
    -- down, it is replaced by lore messages. See spawnNPC function below.
-   jm_chance_min = 0.25
-   jm_chance_max = 0.35
+   jm_chance_min = 0
+   jm_chance_max = 0.25
    jm_chance = var.peek( "npc_jm_chance" ) or jm_chance_max
 
    local num_npc = rnd.rnd(1, 5)
--- a/dat/events/neutral/npc.lua
+++ b/dat/events/neutral/npc.lua
@@ -248,8 +248,8 @@ function create()
 
    -- Chance of a jump point message showing up. As this gradually goes
    -- down, it is replaced by lore messages. See spawnNPC function below.
-   jm_chance_min = 0
-   jm_chance_max = 0.25
+   jm_chance_min = 0.25
+   jm_chance_max = 0.35
    jm_chance = var.peek( "npc_jm_chance" ) or jm_chance_max
 
    local num_npc = rnd.rnd(1, 5)
--- a/dat/missions/neutral/cargo.lua
+++ b/dat/missions/neutral/cargo.lua
@@ -104,7 +104,6 @@ function accept()
       tk.msg(full[1], full[2]:format(amount, amount - player.pilot():cargoFree()))
       misn.finish()
    end
-   origin_p, origin_s = planet.cur()
    misn.accept()
    misn.cargoAdd(cargo, amount) -- TODO: change to jettisonable cargo once custom commodities are in. For piracy purposes.
    misn.osdCreate(osd_title, {osd_msg:format(destplanet:name(), destsys:name())})
@@ -117,7 +116,6 @@ function land()
       -- Semi-random message.
       tk.msg(cargo_land_title, cargo_land_p2[rnd.rnd(1, #cargo_land_p2)]:format( cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], cargo ))
       player.pay(reward)
-      dest_p, dest_s = planet.cur()
       misn.finish(true)
    end
 end
--- a/dat/missions/neutral/cargo.lua
+++ b/dat/missions/neutral/cargo.lua
@@ -104,9 +104,7 @@ function accept()
       tk.msg(full[1], full[2]:format(amount, amount - player.pilot():cargoFree()))
       misn.finish()
    end
-   misn.createLog(buildCargoMissionDescription( nil, amount, cargo, destplanet, destsys ),"Trade",0)
    origin_p, origin_s = planet.cur()
-   misn.appendLog("Cargo accepted on "..origin_p:name() .. " for " .. destplanet:name())
    misn.accept()
    misn.cargoAdd(cargo, amount) -- TODO: change to jettisonable cargo once custom commodities are in. For piracy purposes.
    misn.osdCreate(osd_title, {osd_msg:format(destplanet:name(), destsys:name())})
@@ -120,15 +118,11 @@ function land()
       tk.msg(cargo_land_title, cargo_land_p2[rnd.rnd(1, #cargo_land_p2)]:format( cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], cargo ))
       player.pay(reward)
       dest_p, dest_s = planet.cur()
-      misn.appendLog("Cargo delivered to " .. dest_p:name() .. ", payment " .. math.floor(reward) .. " credits")
-      misn.setRemoveLog(0)
       misn.finish(true)
    end
 end
 
 function abort ()
-   misn.appendLog("Cargo delivery abandoned!")
-   misn.setRemoveLog(0)
    misn.finish(false)
 end
 
--- a/dat/missions/neutral/cargo.lua
+++ b/dat/missions/neutral/cargo.lua
@@ -64,7 +64,6 @@ function create()
    
    -- Calculate the route, distance, jumps, risk of piracy, and cargo to take
    destplanet, destsys, numjumps, traveldist, cargo, avgrisk, tier = cargo_calculateRoute()
-   
    if destplanet == nil then
       misn.finish(false)
    end
@@ -105,6 +104,9 @@ function accept()
       tk.msg(full[1], full[2]:format(amount, amount - player.pilot():cargoFree()))
       misn.finish()
    end
+   misn.createLog(buildCargoMissionDescription( nil, amount, cargo, destplanet, destsys ),"Trade",0)
+   origin_p, origin_s = planet.cur()
+   misn.appendLog("Cargo accepted on "..origin_p:name() .. " for " .. destplanet:name())
    misn.accept()
    misn.cargoAdd(cargo, amount) -- TODO: change to jettisonable cargo once custom commodities are in. For piracy purposes.
    misn.osdCreate(osd_title, {osd_msg:format(destplanet:name(), destsys:name())})
@@ -117,11 +119,14 @@ function land()
       -- Semi-random message.
       tk.msg(cargo_land_title, cargo_land_p2[rnd.rnd(1, #cargo_land_p2)]:format( cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], cargo ))
       player.pay(reward)
+      dest_p, dest_s = planet.cur()
+      misn.appendLog("Cargo delivered to " .. dest_p:name() .. ", payment " .. math.floor(reward) .. " credits")
       misn.finish(true)
    end
 end
 
 function abort ()
+   misn.appendLog("Cargo delivery abandoned!")
    misn.finish(false)
 end
 
--- a/dat/missions/neutral/cargo.lua
+++ b/dat/missions/neutral/cargo.lua
@@ -121,12 +121,14 @@ function land()
       player.pay(reward)
       dest_p, dest_s = planet.cur()
       misn.appendLog("Cargo delivered to " .. dest_p:name() .. ", payment " .. math.floor(reward) .. " credits")
+      misn.setRemoveLog(0)
       misn.finish(true)
    end
 end
 
 function abort ()
    misn.appendLog("Cargo delivery abandoned!")
+   misn.setRemoveLog(0)
    misn.finish(false)
 end
 
--- a/dat/missions/neutral/cargo_rush.lua
+++ b/dat/missions/neutral/cargo_rush.lua
@@ -145,7 +145,6 @@ function accept()
          misn.finish()
       end
    end
-   origin_p, origin_s = planet.cur()
    misn.accept()
    intime = true
    misn.cargoAdd(cargo, amount) -- TODO: change to jettisonable cargo once custom commodities are in. For piracy purposes.
--- a/dat/missions/neutral/cargo_rush.lua
+++ b/dat/missions/neutral/cargo_rush.lua
@@ -145,6 +145,9 @@ function accept()
          misn.finish()
       end
    end
+   misn.createLog( buildCargoMissionDescription(cargosize[tier], amount, cargo, destplanet, destsys ), "Trade",0)
+   origin_p, origin_s = planet.cur()
+   misn.appendLog("Cargo accepted on "..origin_p:name() .. " for " .. destplanet:name())
    misn.accept()
    intime = true
    misn.cargoAdd(cargo, amount) -- TODO: change to jettisonable cargo once custom commodities are in. For piracy purposes.
@@ -161,10 +164,12 @@ function land()
       if intime then
       -- Semi-random message.
       tk.msg(cargo_land_title, string.format( cargo_land_p2[rnd.rnd(1, #cargo_land_p2)], cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], _(cargo) ))
+      misn.appendLog("Cargo delivered on time to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
    else
       -- Semi-random message for being late.
       tk.msg(cargo_land_title, string.format( cargo_land_p3[rnd.rnd(1, #cargo_land_p3)], cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], _(cargo), reward/2, reward ))
       reward = reward / 2
+      misn.appendLog("Cargo delivered late to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
    end
       player.pay(reward)
       misn.finish(true)
--- a/dat/missions/neutral/cargo_rush.lua
+++ b/dat/missions/neutral/cargo_rush.lua
@@ -145,9 +145,7 @@ function accept()
          misn.finish()
       end
    end
-   misn.createLog( buildCargoMissionDescription(cargosize[tier], amount, cargo, destplanet, destsys ), "Trade",0)
    origin_p, origin_s = planet.cur()
-   misn.appendLog("Cargo accepted on "..origin_p:name() .. " for " .. destplanet:name())
    misn.accept()
    intime = true
    misn.cargoAdd(cargo, amount) -- TODO: change to jettisonable cargo once custom commodities are in. For piracy purposes.
@@ -164,14 +162,11 @@ function land()
       if intime then
       -- Semi-random message.
       tk.msg(cargo_land_title, string.format( cargo_land_p2[rnd.rnd(1, #cargo_land_p2)], cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], _(cargo) ))
-      misn.appendLog("Cargo delivered on time to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
    else
       -- Semi-random message for being late.
       tk.msg(cargo_land_title, string.format( cargo_land_p3[rnd.rnd(1, #cargo_land_p3)], cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], _(cargo), reward/2, reward ))
       reward = reward / 2
-      misn.appendLog("Cargo delivered late to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
    end
-   misn.setRemoveLog(0)
    player.pay(reward)
    misn.finish(true)
    end
@@ -199,7 +194,5 @@ function tick()
 end
 
 function abort()
-   misn.appendLog("Cargo delivery abandoned!")
-   misn.setRemoveLog(0)
    misn.finish(false)
 end
--- a/dat/missions/neutral/cargo_rush.lua
+++ b/dat/missions/neutral/cargo_rush.lua
@@ -171,8 +171,9 @@ function land()
       reward = reward / 2
       misn.appendLog("Cargo delivered late to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
    end
-      player.pay(reward)
-      misn.finish(true)
+   misn.setRemoveLog(0)
+   player.pay(reward)
+   misn.finish(true)
    end
 end
 
@@ -198,5 +199,7 @@ function tick()
 end
 
 function abort()
+   misn.appendLog("Cargo delivery abandoned!")
+   misn.setRemoveLog(0)
    misn.finish(false)
 end
--- a/dat/missions/neutral/pirbounty_dead.lua
+++ b/dat/missions/neutral/pirbounty_dead.lua
@@ -155,7 +155,6 @@ function accept ()
    osd_msg[3] = osd_msg[3]:format( paying_faction:name() )
    misn.osdCreate( osd_title, osd_msg )
 
-
    last_sys = system.cur()
    job_done = false
    target_killed = false
@@ -321,6 +320,7 @@ function hunter_hail( arg )
 
    local text = share_text[ rnd.rnd( 1, #share_text ) ]
    tk.msg( share_title, text:format( name ) )
+
    player.pay( credits )
    misn.finish( true )
 end
--- a/dat/missions/neutral/pirbounty_dead.lua
+++ b/dat/missions/neutral/pirbounty_dead.lua
@@ -160,9 +160,6 @@ function accept ()
    job_done = false
    target_killed = false
 
-   misn.createLog(misn_title:format( misn_level[level], missys:name() ), "Pirate hunt",0)
-   misn.appendLog("Pirate mission in " .. missys:name() .. " accepted on " .. last_sys:name() )
-
    hook.jumpin( "jumpin" )
    hook.jumpout( "jumpout" )
    hook.takeoff( "takeoff" )
@@ -189,7 +186,6 @@ function jumpout ()
    jumps_permitted = jumps_permitted - 1
    last_sys = system.cur()
    if not job_done and last_sys == missys then
-      misn.appendLog("Mission failed, pirate escaped")
       fail( msg[3]:format( last_sys:name() ) )
    end
 end
@@ -205,13 +201,10 @@ function land ()
    if job_done and planet.cur():faction() == paying_faction then
       local pay_text
       if target_killed then
-         misn.appendLog("Payment of " .. math.floor(credits) .. " received for death of pirate")
          pay_text = pay_kill_text[ rnd.rnd( 1, #pay_kill_text ) ]
       else
-         misn.appendLog("Payment of " .. math.floor(credits) .. " received for capture of pirate")
          pay_text = pay_capture_text[ rnd.rnd( 1, #pay_capture_text ) ]
       end
-      misn.setRemoveLog(0)
       tk.msg( pay_title, pay_text:format( name ) )
       player.pay( credits )
       paying_faction:modPlayerSingle( reputation )
@@ -235,14 +228,12 @@ function pilot_board ()
       local t = subdue_text[ rnd.rnd( 1, #subdue_text ) ]:format( name )
       tk.msg( subdue_title, t )
       succeed()
-      misn.appendLog("Pirate ship boarded")
       target_killed = false
       target_ship:changeAI( "dummy" )
       target_ship:setHilight( false )
       target_ship:disable() -- Stop it from coming back
       if death_hook ~= nil then hook.rm( death_hook ) end
    else
-      misn.appendLog("Failed to board pirate ship") 
       local t = subdue_fail_text[ rnd.rnd( 1, #subdue_fail_text ) ]:format( name )
       tk.msg( subdue_fail_title, t )
       board_fail()
@@ -274,7 +265,6 @@ function pilot_death( p, attacker )
    if attacker == player.pilot() then
       succeed()
       target_killed = true
-      misn.appendLog("Successfully killed the pirate")
    else
       local top_hunter = nil
       local top_hits = 0
@@ -295,7 +285,6 @@ function pilot_death( p, attacker )
       if top_hunter == nil or player_hits >= top_hits then
          succeed()
          target_killed = true
-	 misn.appendLog("Pirate killed, and you did most damage!")
       elseif player_hits >= top_hits / 2 and rnd.rnd() < 0.5 then
          hailer = hook.pilot( top_hunter, "hail", "hunter_hail", top_hunter )
          credits = credits * player_hits / total_hits
@@ -307,7 +296,6 @@ function pilot_death( p, attacker )
          hook.timer( 3000, "timer_hail", top_hunter )
          misn.osdDestroy()
       else
-         misn.appendLog("Pirate killed, but not by you")
          fail( msg[2]:format( name ) )
       end
    end
@@ -315,7 +303,6 @@ end
 
 
 function pilot_jump ()
-   misn.appendLog("Pirate escaped")
    fail( msg[1]:format( name ) )
 end
 
@@ -334,15 +321,12 @@ function hunter_hail( arg )
 
    local text = share_text[ rnd.rnd( 1, #share_text ) ]
    tk.msg( share_title, text:format( name ) )
-   misn.appendLog("Payment of " .. math.floor(credits) .. "made for helping to kill the pirate")
    player.pay( credits )
-   misn.setRemoveLog(-1)
    misn.finish( true )
 end
 
 
 function hunter_leave ()
-   misn.setRemoveLog(-1)
    misn.finish( false )
 end
 
@@ -397,7 +381,6 @@ function spawn_pirate( param )
          pir_jump_hook = hook.pilot( target_ship, "jump", "pilot_jump" )
          pir_land_hook = hook.pilot( target_ship, "land", "pilot_jump" )
 
-	 misn.appendLog("Target pirate ship sighted")
          local pir_crew = target_ship:stats().crew
          local pl_crew = player.pilot():stats().crew
          if rnd.rnd() > (0.5 * (10 + pir_crew) / (10 + pl_crew)) then
@@ -406,7 +389,6 @@ function spawn_pirate( param )
             can_capture = false
          end
       else
-         misn.appendLog("No sign of pirate in expected location, you were too late")
          fail( msg[1]:format( name ) )
       end
    end
@@ -452,6 +434,5 @@ function fail( message )
          player.msg( "\ar" .. message .. "\a0" )
       end
    end
-   misn.setRemoveLog(-1)
    misn.finish( false )
 end
--- a/dat/missions/neutral/pirbounty_dead.lua
+++ b/dat/missions/neutral/pirbounty_dead.lua
@@ -211,6 +211,7 @@ function land ()
          misn.appendLog("Payment of " .. math.floor(credits) .. " received for capture of pirate")
          pay_text = pay_capture_text[ rnd.rnd( 1, #pay_capture_text ) ]
       end
+      misn.setRemoveLog(0)
       tk.msg( pay_title, pay_text:format( name ) )
       player.pay( credits )
       paying_faction:modPlayerSingle( reputation )
@@ -314,6 +315,7 @@ end
 
 
 function pilot_jump ()
+   misn.appendLog("Pirate escaped")
    fail( msg[1]:format( name ) )
 end
 
@@ -334,11 +336,13 @@ function hunter_hail( arg )
    tk.msg( share_title, text:format( name ) )
    misn.appendLog("Payment of " .. math.floor(credits) .. "made for helping to kill the pirate")
    player.pay( credits )
+   misn.setRemoveLog(-1)
    misn.finish( true )
 end
 
 
 function hunter_leave ()
+   misn.setRemoveLog(-1)
    misn.finish( false )
 end
 
@@ -448,5 +452,6 @@ function fail( message )
          player.msg( "\ar" .. message .. "\a0" )
       end
    end
+   misn.setRemoveLog(-1)
    misn.finish( false )
 end
--- a/dat/missions/neutral/pirbounty_dead.lua
+++ b/dat/missions/neutral/pirbounty_dead.lua
@@ -155,9 +155,13 @@ function accept ()
    osd_msg[3] = osd_msg[3]:format( paying_faction:name() )
    misn.osdCreate( osd_title, osd_msg )
 
+
    last_sys = system.cur()
    job_done = false
    target_killed = false
+
+   misn.createLog(misn_title:format( misn_level[level], missys:name() ), "Pirate hunt",0)
+   misn.appendLog("Pirate mission in " .. missys:name() .. " accepted on " .. last_sys:name() )
 
    hook.jumpin( "jumpin" )
    hook.jumpout( "jumpout" )
@@ -185,6 +189,7 @@ function jumpout ()
    jumps_permitted = jumps_permitted - 1
    last_sys = system.cur()
    if not job_done and last_sys == missys then
+      misn.appendLog("Mission failed, pirate escaped")
       fail( msg[3]:format( last_sys:name() ) )
    end
 end
@@ -200,8 +205,10 @@ function land ()
    if job_done and planet.cur():faction() == paying_faction then
       local pay_text
       if target_killed then
+         misn.appendLog("Payment of " .. math.floor(credits) .. " received for death of pirate")
          pay_text = pay_kill_text[ rnd.rnd( 1, #pay_kill_text ) ]
       else
+         misn.appendLog("Payment of " .. math.floor(credits) .. " received for capture of pirate")
          pay_text = pay_capture_text[ rnd.rnd( 1, #pay_capture_text ) ]
       end
       tk.msg( pay_title, pay_text:format( name ) )
@@ -227,12 +234,14 @@ function pilot_board ()
       local t = subdue_text[ rnd.rnd( 1, #subdue_text ) ]:format( name )
       tk.msg( subdue_title, t )
       succeed()
+      misn.appendLog("Pirate ship boarded")
       target_killed = false
       target_ship:changeAI( "dummy" )
       target_ship:setHilight( false )
       target_ship:disable() -- Stop it from coming back
       if death_hook ~= nil then hook.rm( death_hook ) end
    else
+      misn.appendLog("Failed to board pirate ship") 
       local t = subdue_fail_text[ rnd.rnd( 1, #subdue_fail_text ) ]:format( name )
       tk.msg( subdue_fail_title, t )
       board_fail()
@@ -264,6 +273,7 @@ function pilot_death( p, attacker )
    if attacker == player.pilot() then
       succeed()
       target_killed = true
+      misn.appendLog("Successfully killed the pirate")
    else
       local top_hunter = nil
       local top_hits = 0
@@ -284,6 +294,7 @@ function pilot_death( p, attacker )
       if top_hunter == nil or player_hits >= top_hits then
          succeed()
          target_killed = true
+	 misn.appendLog("Pirate killed, and you did most damage!")
       elseif player_hits >= top_hits / 2 and rnd.rnd() < 0.5 then
          hailer = hook.pilot( top_hunter, "hail", "hunter_hail", top_hunter )
          credits = credits * player_hits / total_hits
@@ -295,6 +306,7 @@ function pilot_death( p, attacker )
          hook.timer( 3000, "timer_hail", top_hunter )
          misn.osdDestroy()
       else
+         misn.appendLog("Pirate killed, but not by you")
          fail( msg[2]:format( name ) )
       end
    end
@@ -320,7 +332,7 @@ function hunter_hail( arg )
 
    local text = share_text[ rnd.rnd( 1, #share_text ) ]
    tk.msg( share_title, text:format( name ) )
-
+   misn.appendLog("Payment of " .. math.floor(credits) .. "made for helping to kill the pirate")
    player.pay( credits )
    misn.finish( true )
 end
@@ -381,6 +393,7 @@ function spawn_pirate( param )
          pir_jump_hook = hook.pilot( target_ship, "jump", "pilot_jump" )
          pir_land_hook = hook.pilot( target_ship, "land", "pilot_jump" )
 
+	 misn.appendLog("Target pirate ship sighted")
          local pir_crew = target_ship:stats().crew
          local pl_crew = player.pilot():stats().crew
          if rnd.rnd() > (0.5 * (10 + pir_crew) / (10 + pl_crew)) then
@@ -389,6 +402,7 @@ function spawn_pirate( param )
             can_capture = false
          end
       else
+         misn.appendLog("No sign of pirate in expected location, you were too late")
          fail( msg[1]:format( name ) )
       end
    end
--- a/src/land.c
+++ b/src/land.c
@@ -48,7 +48,7 @@
 #include "nlua.h"
 #include "nluadef.h"
 #include "nlua_tk.h"
-
+#include "shiplog.h"
 
 /* global/main window */
 #define LAND_WIDTH   800 /**< Land window width. */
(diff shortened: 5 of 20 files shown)


dat/events/neutral/npc.lua
```diff
@@ -248,8 +248,8 @@ function create()
 
    -- Chance of a jump point message showing up. As this gradually goes
    -- down, it is replaced by lore messages. See spawnNPC function below.
-   jm_chance_min = 0
-   jm_chance_max = 0.25
+   jm_chance_min = 0.25
+   jm_chance_max = 0.35
```
onpon4, line 252: This seems unrelated to the topic of the pull request, no? I disagree with the change; I intentionally made it so that jump messages disappear both to minimize the damaging "clickfest" effect it can create (confine it to the start of the game) and so that you don't get bothered with unnecessary garbage later on in the game. You've in fact changed it here so that jump messages are more common than they ever used to be even at its lowest IIRC.

dat/missions/neutral/cargo.lua
```diff
@@ -117,11 +119,14 @@ function land()
       -- Semi-random message.
       tk.msg(cargo_land_title, cargo_land_p2[rnd.rnd(1, #cargo_land_p2)]:format( cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], cargo ))
       player.pay(reward)
+      dest_p, dest_s = planet.cur()
+      misn.appendLog("Cargo delivered to " .. dest_p:name() .. ", payment " .. math.floor(reward) .. " credits")
```
onpon4, line 123: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/cargo.lua
```diff
@@ -105,6 +104,9 @@ function accept()
       tk.msg(full[1], full[2]:format(amount, amount - player.pilot():cargoFree()))
       misn.finish()
    end
+   misn.createLog(buildCargoMissionDescription( nil, amount, cargo, destplanet, destsys ),"Trade",0)
+   origin_p, origin_s = planet.cur()
+   misn.appendLog("Cargo accepted on "..origin_p:name() .. " for " .. destplanet:name())
```
onpon4, line 109: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/cargo.lua
```diff
@@ -117,11 +119,14 @@ function land()
...
       tk.msg(cargo_land_title, cargo_land_p2[rnd.rnd(1, #cargo_land_p2)]:format( cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], cargo ))
       player.pay(reward)
+      dest_p, dest_s = planet.cur()
+      misn.appendLog("Cargo delivered to " .. dest_p:name() .. ", payment " .. math.floor(reward) .. " credits")
       misn.finish(true)
    end
 end
 
 function abort ()
+   misn.appendLog("Cargo delivery abandoned!")
```
onpon4, line 129: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/cargo_rush.lua
```diff
@@ -145,6 +145,9 @@ function accept()
          misn.finish()
       end
    end
+   misn.createLog( buildCargoMissionDescription(cargosize[tier], amount, cargo, destplanet, destsys ), "Trade",0)
+   origin_p, origin_s = planet.cur()
+   misn.appendLog("Cargo accepted on "..origin_p:name() .. " for " .. destplanet:name())
```
onpon4, line 150: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/cargo_rush.lua
```diff
@@ -161,10 +164,12 @@ function land()
       if intime then
       -- Semi-random message.
       tk.msg(cargo_land_title, string.format( cargo_land_p2[rnd.rnd(1, #cargo_land_p2)], cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], _(cargo) ))
+      misn.appendLog("Cargo delivered on time to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
```
onpon4, line 167: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/cargo_rush.lua
```diff
@@ -161,10 +164,12 @@ function land()
...
    else
       -- Semi-random message for being late.
       tk.msg(cargo_land_title, string.format( cargo_land_p3[rnd.rnd(1, #cargo_land_p3)], cargo_land_p1[rnd.rnd(1, #cargo_land_p1)], _(cargo), reward/2, reward ))
       reward = reward / 2
+      misn.appendLog("Cargo delivered late to " .. destplanet:name() .. ", payment " .. math.floor(reward) .. " credits")
```
onpon4, line 172: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -155,10 +155,14 @@ function accept ()
...
    misn.osdCreate( osd_title, osd_msg )
 
+
    last_sys = system.cur()
    job_done = false
    target_killed = false
 
+   misn.createLog(misn_title:format( misn_level[level], missys:name() ), "Pirate hunt",0)
+   misn.appendLog("Pirate mission in " .. missys:name() .. " accepted on " .. last_sys:name() )
+
```
onpon4, line 165: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -185,6 +189,7 @@ function jumpout ()
    jumps_permitted = jumps_permitted - 1
    last_sys = system.cur()
    if not job_done and last_sys == missys then
+      misn.appendLog("Mission failed, pirate escaped")
```
onpon4, line 192: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -200,8 +205,10 @@ function land ()
    if job_done and planet.cur():faction() == paying_faction then
       local pay_text
       if target_killed then
+         misn.appendLog("Payment of " .. math.floor(credits) .. " received for death of pirate")
```
onpon4, line 208: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -200,8 +205,10 @@ function land ()
...
          pay_text = pay_kill_text[ rnd.rnd( 1, #pay_kill_text ) ]
       else
+         misn.appendLog("Payment of " .. math.floor(credits) .. " received for capture of pirate")
```
onpon4, line 211: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -227,12 +234,14 @@ function pilot_board ()
       local t = subdue_text[ rnd.rnd( 1, #subdue_text ) ]:format( name )
       tk.msg( subdue_title, t )
       succeed()
+      misn.appendLog("Pirate ship boarded")
```
onpon4, line 237: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -227,12 +234,14 @@ function pilot_board ()
...
       target_killed = false
       target_ship:changeAI( "dummy" )
       target_ship:setHilight( false )
       target_ship:disable() -- Stop it from coming back
       if death_hook ~= nil then hook.rm( death_hook ) end
    else
+      misn.appendLog("Failed to board pirate ship")
```
onpon4, line 244: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -264,6 +273,7 @@ function pilot_death( p, attacker )
    if attacker == player.pilot() then
       succeed()
       target_killed = true
+      misn.appendLog("Successfully killed the pirate")
```
onpon4, line 276: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -284,6 +294,7 @@ function pilot_death( p, attacker )
       if top_hunter == nil or player_hits >= top_hits then
          succeed()
          target_killed = true
+	 misn.appendLog("Pirate killed, and you did most damage!")
```
onpon4, line 297: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -295,6 +306,7 @@ function pilot_death( p, attacker )
          hook.timer( 3000, "timer_hail", top_hunter )
          misn.osdDestroy()
       else
+         misn.appendLog("Pirate killed, but not by you")
```
onpon4, line 309: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -320,7 +332,7 @@ function hunter_hail( arg )
 
    local text = share_text[ rnd.rnd( 1, #share_text ) ]
    tk.msg( share_title, text:format( name ) )
-
+   misn.appendLog("Payment of " .. math.floor(credits) .. "made for helping to kill the pirate")
```
onpon4, line 335: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -381,6 +393,7 @@ function spawn_pirate( param )
          pir_jump_hook = hook.pilot( target_ship, "jump", "pilot_jump" )
          pir_land_hook = hook.pilot( target_ship, "land", "pilot_jump" )
 
+	 misn.appendLog("Target pirate ship sighted")
```
onpon4, line 396: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

dat/missions/neutral/pirbounty_dead.lua
```diff
@@ -389,6 +402,7 @@ function spawn_pirate( param )
             can_capture = false
          end
       else
+         misn.appendLog("No sign of pirate in expected location, you were too late")
```
onpon4, line 405: Assuming the log is supposed to be a useful reference for important past events, this is just cluttering it up with unnecessary information.

src/land.c
```diff
@@ -999,7 +999,10 @@ void land_genWindows( int load, int changetab )
       if (planet_hasService(land_planet, PLANET_SERVICE_BAR))
          npc_generate(); /* Generate bar npc. */
    }
+   /* add to the ship log */
+   shiplog_append(0,land_planet->name);
```
onpon4, line 1003: Why do we need this? This again seems to be useless info that just clutters up the log. Travel seems much too mundane to be recording.

agb32:
I was putting in examples of how to use the log - I expect they'll get taken out - but I didn't want to interfere with the main missions and put stuff there, because I'd be treading on people's toes.
So once people have started integrating it into the main missions, we can remove these ones.  To be honest, I wouldn't know what is to be considered important in the main missions so didn't want to interfere there.  I figured the simple missions was the simplest way for me to implement, test, debug, and show people how to do it...

onpon4:
One other thing I didn't notice: you shouldn't use string concatenation, use %s formatting instead. And make the strings translatable by adding `_()` around the string.

> but I didn't want to interfere with the main missions and put stuff there, because I'd be treading on people's toes.

It wouldn't be treading on people's toes to add useful logs to the many unique missions we have. It would be helpful. If you don't do it then someone else will have to. I don't see why Reynir would need to come back to add logs to the Hot Dogs From Space mission when you could just do that here for example. They're just logs.

> To be honest, I wouldn't know what is to be considered important in the main missions

I assume you've played the game before, right? So you're familiar with how at least some of these missions go? Add logs for what's important in some of those missions.

> I figured the simple missions was the simplest way for me to implement, test, debug, and show people how to do it...

All it does is create unnecessary busywork for someone else if you're just planning for someone else to remove it.

You know that the docs support giving a usage example, right? Put the example under the `@usage` tag in the comments for the Lua function. Take a look at the other Lua functions for an example. Actually that's another shortcoming here that I didn't notice before. The `@usage` tag is how you're supposed to give examples, not throwaway garbage in actual missions that you expect others to clean up.

onpon4:
Hey, just wanted to follow up on this. I appreciate the improvement you've made with the last commit; while I haven't tested it myself, it seems to be a good system. That said, I would still like to see the "examples" and unnecessary travel log removed, and a proper example given for the function under the `@usage` tag. You don't have to implement proper usages if you don't want to or don't feel confident enough to do so, but the usages you did implement absolutely should not be there (as you yourself admitted).

I would also like to see the unrelated change removed, though that's a personal preference.

Obviously I don't have the power to merge pull requests anyway, but those are changes I would personally suggest.

🕵️

onpon4:
Ah, one more thing I have to note: you should add "Fixes #1053" to the description / OP of this pull request, that'll auto-close #1053 with the merging of this PR.

🕷️

agb32:
Fixes #1053 

ok - I've been thinking a bit about this, and have implemented the following.
There is now a way for mission scripts to remove a log after a certain time.
For cargo logs, the log is therefore now removed after time has moved on from completing the delivery, e.g. when you take off.  So - once completed, its still there for you to read, until you take off.
For pirate logs, the same thing, if you succeed.  However, if you fail, the log hangs around for 1 period, and is then removed.  
I have been finding the pirate logs actually useful for telling me whether I have killed the pirate, or whether someone else got there first - so in that case, useful for them to hang around for a bit.

Journey logs are also now limited to the last 6 journeys.  I think this is useful, because if you come back to the game after a few days or weeks, you can then look at the log and remind yourself where you've come from...

I've also changed the jump point thing back to 0 and 0.25 - not sure how that got in here - I must have forgot to reset it after playing around trying to see why I wasn't getting jump messages anymore...!

What this now means is that the logs won't get filled up unnecessarily, but that useful info is still there for a bit.  Also for new starters, nice for them to have something in their log initially...

Hope everyone agrees with this?

agb32:
And have added the @usage tags to give examples.

onpon4:
Ship logs that disappear are useless. This isn't a solution to the problem. Travel shouldn't be logged because it isn't useful to log travel.

Frankly, the ability to delete logs at all is a problem in my book.

> Also for new starters, nice for them to have something in their log initially...

Empty logs aren't a problem. Why are you treating them as one? At the start of the game you shouldn't have any logs. The first logs you get should be from the first unique mission you get, probably the Empire recruitment.

🕵️

agb32:
ok - logs removed from those missions...

agb32:
Still need to remove the travel log though - that requires a bit more thought, and its late now...

onpon4:
I've submitted a PR to you, @agb32. I think I've removed the travel log successfully, but this isn't fully tested. Please check. https://github.com/agb32/naev/pull/1

On a side note, I've noticed that the log system generally only works for missions. This is a problem since events might need to log stuff, too. I'd suggest at least adding support for events logging stuff, or better, relocating the log stuff outside of misn and not tying it to mission types (which isn't what we want to do anyway).

Of particular note, I know that the hint for the Shadow Vigil mission - which is in an event - will need to use the ship log so the player can look back on what system they need to go to. I think that was a motivation for the proposal actually.

🕵️

onpon4:
Thanks for the merge. Of course as I mentioned on the PR, but just mentioning it here again: one change that needs to happen here is all those uses of printf you've added need to be changed to the WARN macro (or the DEBUG macro if that is more appropriate).

🕵️

agb32:
printf statements now removed and replaced with WARN

onpon4:
Great! Thanks.

Any thoughts on how to extend this to work outside of missions?

🕵️

agb32:
The text type can be anything - so I think it would work - nothing is pulled in from the mission its called from.
The idea was that logs can be categorized by campaign.  e.g. the shadow runs might have a type "Shadow", and then a different name for each of them - the player would then be able to select by type (showing all of the shadow mission logs), and then select one of these to get the log entries just by name.  So unless I misunderstand, it already has that...

agb32:
I've started on the shiplog in shadowvigil as suggested.  However, I'm sure there is more to be done here, but I'm probably not the person to do it - I don't feel I know the mission well enough to understand the subtleties.

agb32:
But its a start, and once pulled, easier for others to work on.

onpon4:
> The text type can be anything - so I think it would work - nothing is pulled in from the mission its called from.

Ah, cool! That should be good then, it just needs to be moved out of the misn module (since that's only available to missions).

🕷️

onpon4:
Ah, have you given any thought into how to make translation work with this? As it is you don't seem to be doing any gettext translation but logs almost certainly should be translated, just preferably in a way that it won't cause problems like #1077.

🕷️

onpon4:
OK, I'm a bit confused by this "log ID" thing you've got going. So, like, you have a local log ID within missions, and does that determine what gets grouped together or does it use text?

It seems that missions should be ignored altogether and the system should just have a list of string identifiers or something. I'm not sure. Maybe it would be easier to implement without the categories thing and just basically be a list of strings? That would mean you wouldn't have to worry about titles changing and needing to group them together and whatnot. Not sure how to solve the text itself but worst case at least you would have everything consistent and as you saw it the first time.

agb32:
Summary:
A logset is created, with a given type and name.  This will then create a unique log ID for this logset (except for if the overwrite flag is set, in which case, any previous logset with the same type and name will be erased, and the log ID for that reused).

This log ID is then used within a chronological linked list of messages, to identify the log messages that belong to this particular log ID, and hence log name and type.

Within a lua mission, it needs to first create a logset, and then save the log ID for this set.  This is written to the players savefile, as part of the mission, so that when a player reloads, they can still append to this logset.  

Once you have created a logset with a name/type, you only need to know the log ID to be able to append log messages to it - you don't need to know the name/type again (though that can be obtained from the log ID if needed).  And if the title changes in the lua script, it doesn't matter - you just keep the log ID, so the save file is still valid for this mission.  It keeps it consistent.

The categories (type) thing is so that its easy for the player to sort through the logs, and select the one they want.  I think we should keep that.  Would get tiresome otherwise, if searching through a lot of logs.

As for translation - no, I couldn't figure out how to do that.  The log messages are saved as text strings.  The obvious place for translation is within the lua script, while writing the logs.  But then they'd remain in that language, even if the player changed their language.

To move it out of the nlua_misn.c module, I think we would need:
createLog()
appendLog()
and possibly deleteLog() and setRemoveLog().

The createLog would return the logID.  The lua script would then need to save this log ID.
So for example, in a mission script, you'd somehow need to put the log ID into the logid element of the Mission struct defined in mission.h.  Currently, this is done within the misn_createLog function.  But perhaps this should be changed to something like:

In a lua mission script:
logid = shiplog.createLog("log name","type",0)
misn.setShipLogID(logid)                   <- which does mission->logid=logid;
shiplog.appendLog(misn.getShipLogID(), "message" )

Or something like that.  Currently, the getting and setting of the logID is hidden within the nlua_misn.c, so a bit more user friendly.

The alternative is to keep nlua_misn.c as it is, but then add a similar functionality to e.g. nlua_evt.c.  Though I have to admin, I'm not entirely clear what the difference between missions and events it - everything I've looked at (shipping, pirate, shadowrun) seems to be a mission.

Hope thats not too confusing (I've just realised I've written a lot!)

onpon4:
So, those IDs are mission-specific. What about replacing that with a global string-based identifier? Strings that would always be untranslated and wouldn't be displayed to the player.

Events and missions are similar but different. There's also Lua code outside of both. Only missions import the misn module, and only events import the evt module. The log should be somewhere else. Maybe the player module? That could be a good fit.

🕷️

agb32:
No, the IDs aren't mission specific.  They just increment from 0 up to 2 billion (and I figured no one would ever get 2 billion missions!).

I guess what you mean is that a mission log should be identifiable from the name and type - and therefore there would be no need to save a log ID in the save file.  Rather we could have something such as:
misn.appendLog("shadow","shadow vigil","some log message here")
(where shadow is the type, shadowvigil is the mission name).
Then that way, there would be no need to save the logID in the mission file.

onpon4:
If the log ID is global, one major problem with using numbers is it's hard to coordinate. How do you find out that numbers 1-20 are taken and 21 is the first available one? For that reason strings would be better than numbers (easier to make sure those are unique).

🕷️

onpon4:
Just to clarify, I do think that IDs is a good practice to help with translation support. That can at least allow categorization to be consistent.

🕷️

agb32:
To find out the first available one, it just scans through the list, takes the max, and increments,  Means that gaps can appear, but we have 2Bn numbers to play with, so not worried.

Note - different players doing the same mission (e.g. shadow vigil) will likely have different log IDs for that mission.  Or indeed, the same player doing a different save file.

One problem with my above suggestion for the lua script is it won't work if titles are mission situation/circumstance dependent.  For example, if we choose to have a mission log for seek and destroy missions, they might have a type "Seek and destroy", while the name might be "Search for pirate xxx, starting from yyy" with xxx being the pirate name, yyy being the system on which the mission was accepted.  In which case, since there isn't a generic title, and this lua script doesn't generate unique missions, it won't be able to get the logid unless it can reconstruct the title (which may be possible).  But simpler to keep the currently implemented approach, and save the logid - less likely to go wrong I think.

So, what I propose:  keep the nlua_misn.c as it is - missions are now done.  But also add a nlua_shiplog.c which can be called be events and other things, which has:
shiplog.createLog(name,type,overwrite)   <- which returns a logID
shiplog.appendLogByID(logid,message)
shiplog.appendLogByName(name,type,message)

Or something like that.

And this is something that can be done at a later stage, once we have an example that requires it... (if you can point me to one, I can look at implementing it).

src/nlua_misn.c
```diff
@@ -1020,3 +1029,95 @@ static int misn_claim( lua_State *L )
...
+}
+
+
+/**
+ * @brief Deletes the shiplog.
+ *
+ * @luafunc deleteLog( )
+ * @usage misn.deleteLog()
+ */
+static int misn_deleteLog( lua_State *L )
```
onpon4, line 1094: I think you should remove this Lua function, too. Deletion of logs should always be done by the user if it's done at all. Missions shouldn't be in the habit of making missions disappear automatically.

🕵️

onpon4:
> So, what I propose: keep the nlua_misn.c as it is - missions are now done. But also add a nlua_shiplog.c which can be called be events and other things

Why have two separate APIs for the same thing? Just keep it simple and move the log stuff out of misn. It doesn't have to ever have any connection to missions.

agb32:
ok - I've mostly done the above.  Have created nlua_shiplog.c/h, which now contains the createLog and appendLog functions.
I've also updated the shadowvigil to use these.
And removed these functions from the nlua_misn.c.  
deleteLog function removed. 

I've not changed the messages created in shadowvigil, because I don't think I'd be able to do those very well - better to leave that to someone who knows what they're doing, once this is accepted (likewise for all the other missions and events - given that I liked having a log for killing pirates and shipping, shows that I'm out of touch!).

src/nlua_shiplog.c
```diff
@@ -0,0 +1,134 @@
...
+   return 0;
+}
+
+
+/**
+ * @brief Creates a shiplog for this mission.
+ *
+ *    @luatparam string name Name for this log.
+ *    @luatparam string type Type of log (e.g travel, trade, etc, can be anything.)
+ *    @luatparam number overwrite Whether to overwrite existing mission with this logname and logtype.  Warning, removes previous entries of this logname and type if 1, or logs of this type if 2.
```
onpon4, line 91: The meaning of this is unclear; reading this I have no idea what this argument does. Also, you shouldn't use numbers that could be 1 or 2, it would be better to just have two separate booleans (which would also mean that `nil` would be valid).

🕵️

src/nlua_shiplog.c
```diff
@@ -0,0 +1,134 @@
...
+   const char *logname;
+   const char *logtype;
+   int overwrite,logid;
+   /* Parameters. */
+   logname    = luaL_checkstring(L,1);
+   logtype    = luaL_checkstring(L,2);
+   overwrite = luaL_checkint(L,3);
+
+   /* Create a new shiplog */
+   logid = shiplog_create( logname, logtype, overwrite );
```
onpon4, line 109: It looks like you're using the display names of the log name and log type for differentiation? This is a bad idea, it would be a repeat of the problem we have with news articles. It would be better to have separate strings - ID strings, an additional two arguments - that identify categorization on a programmatic level. The pretty version, which would be translated, should only be used for display.

🕵️

agb32: No - the logid is used for differentiation, once a log has been created.
The log type and logname are displayed to the user in the info widget, but thats all - and these should be translated.
No other differentiation is required.

src/nlua_shiplog.c
agb32: ok, I'll split it into 2 bools.
Basically, if a user starts a mission and then aborts it, the logset will remain (which is what we want to happen).
If they then restart it at a later date, the existing log can then be replaced with the new attempt, if overwrite=1.  Or if overwrite=0, then the old logset will remain, and a new one will be created.

Overwrite=2 means that all logsets of this type will be replaced by the new logset.  This is more of an edge case, so less likely to be used.  In fact, I could probably remove it for now.

src/nlua_shiplog.c
agb32: The logid is independent of the logname and logtype...

onpon4: Then how do multiple, separate Lua scripts use the same "log"? I've told you this already, there needs to be a way for the Lua scripter to specify *where* the log goes. A numeric, auto-generated "log ID" doesn't cut it because multiple scripts (whether that's multiple missions or any other situations) cannot easily share these. You need *string* identifiers.

🕵️

src/nlua_shiplog.c
onpon4: > If they then restart it at a later date, the existing log can then be replaced with the new attempt, if overwrite=1.

Wait, what? I don't think we want this. Logs should *never* be automatically deleted.

> Overwrite=2 means that all logsets of this type will be replaced by the new logset.

Same here. I think you should remove these options altogether. Logs should be permanent.

🕵️

onpon4:
> I would suggest that the number makes more sense in this context and is consistent with how this sort of thing is treated in Unix/Linux system utilities and the like

Lua isn't a Unix/Linux system utility. Using a type-checked, numeric value for this sort of thing goes completely against dynamic typing.

> they would indeed help me recover where I am when I have not been able to play in a while

No they wouldn't, these are throwaway missions you were just doing to make some money. You can see how much money you have and what minor missions you can do now. What minor missions you were doing before is totally irrelevant - except for the fact that different *ships* are suited for different missions, which you can find out about easily on the Info screen.

> possibly the failed version could be wiped once a successful run of the mission/event has been completed?

I would much prefer if it doesn't get wiped at all.

🕵️

src/nlua_shiplog.c
agb32: ok - now I see - I wasn't aware of the need for this before.  I have a solution - will push it this evening, but basically, a new createNamedLog function which will create a logID if needed, or return it if it already exists, and that will solve it.

src/nlua_shiplog.c
agb32: Except if we ever implement logs for throwaway missions.  Might happen in the future, we just don't know.  I'll sort something out.  But for the non-throw-away missions, we can just set overwrite=0...

Is there a concept of flags in the lua?  i.e. in the same way as in c we would & flags.  If so, I could replace overwrite with flags, and then would allow for future expansion, as @PhoenixRiver mentions below.

PhoenixRiver:
> Lua isn't a Unix/Linux system utility. Using a type-checked, numeric value
> for this sort of thing goes completely against dynamic typing.

Are you suggesting that the reasoning for it being done with *nix
system utils is falacious? I don't see you arguing with the actual
point of the increasing complexity it involves for dealing with
different modes for the same utility (log system in this case).

> No they wouldn't, these are throwaway missions you were just doing to make
> some money. You can see how much money you have and what minor missions
> you can do now. What minor missions you were doing before is totally irrelevant
> - except for the fact that different *ships* are suited for different
> missions, which you can find out about easily on the Info screen.

I would hope that you are not trying to tell me how to play the game
and certainly not how my memory works. I am an explorer and I rarely
do missions, esepcailly combat heavy ones. I also have associative
memory where things are chained together so if I last did a series of
commodity runs in one direction I was a) travelling through known
space b) with a direction in mind which gives me a hint as to where I
was going. If my commodity logs are empty then I am exploring local
space and if there are combat logs then I know I should be careful
because clearly there are people who attack me around here. In both of
these cases I have learnt the majority of my situation from those
"useless" logs on one screen in one place and am now back to speed
when otherwise it does take me some time wandering the local systems
and trying to find something I recognise as the last thing I was
interested in - actually a lot of the time these days I do not find it
because I have even less time and am less inclined to spend my time
wondering what it was that I forgot rather than playing the game or
just abandoning, what becomes, many things half way through...
missions included.

>> possibly the failed version could be wiped once a successful run of the
>> mission/event has been completed?
>
> I would much prefer if it doesn't get wiped at all.

This is clearly something that needs to be a option for the player
along with the trade, pirate and travel logs (though since they are
seperately categorised they can already be ignored easily by the
player). Would an binary option to stop the logs from autodeleting
anything be possible in the settings menu? Or maybe a seperate system
on the log screen to be able to change the mode choice for each
category being choosable by the player (with a reasonable default
being set of course). The mode system already exists and adding more
modes would not be so difficult since it is numerical - the hardest
bit would be the wording to convey each mode to the player, maybe a
list to choose from:

Never delete logs in this category
Overwrite duplicate events/missions/etc. when they start
Overwrite duplicate events/missions/etc. when they end
Delete after one [long] time unit
Delete after one [short] time unit
Delete after jump
Do not log this category (takes effect after you close the log screen)

I'm not sure about how difficult all these might be to implement and
it might be necessary to add something to note that changes will only
be reflected from that point on - a help page might be necessary but
I'd suggest that the default err on the side of caution to avoid
necesitating most players having to change anything at all.

Hope this helps,

Johann *<:@)

src/nlua_shiplog.c
onpon4: No, even if you did implement them for throwaway missions, they should be permanent.

Flags aren't something you should be doing in Lua. It's technically possible, but it would be a mess. You can't think of Lua like C, code is written for Lua completely differently. If you needed multiple conditions in one arg, the rest of Naev handles multiple cases or whatever by strings. But in a case like this, booleans would make the most sense.

Again though, I don't think logs should *ever* be removed. They should be permanent. If you're having to remove logs, that says that the logs are unimportant and shouldn't have been made in the first place.

🕵️

src/nlua_shiplog.c
onpon4: That would work, and I would suggest just making that the only logging function, so that everything uses that. Having one without the feature doesn't really add anything, it would just create confusion.

agb32: The reason to keep it would be for missions that can be done more than once - and we don't know whether these will implement logs in the future or not - so for now, I think best to keep it.  However, I'll get rid of it if you disagree.

onpon4:
> Are you suggesting that the reasoning for it being done with *nix system utils is falacious?

No, I'm suggesting that you're comparing apples to steak.

> I don't see you arguing with the actual point of the increasing complexity it involves for dealing with different modes for the same utility

No specific examples given, but I don't think those options should be there anyway. As it is, you're just appealing to fear.

> I would hope that you are not trying to tell me how to play the game and certainly not how my memory works.

Not at all, simply that your memory doesn't matter for the cases you're talking about.

> I am an explorer and I rarely do missions, esepcailly combat heavy ones.

So?

Riddle me this: suppose you don't remember a single thing about the game. You're starting on, say, Secondus Station, you see your map, and you have no idea where you got there. Log shows you've not done anything important.

How is this any worse than just starting a new game and having the exact same circumstance, except that you're on Em 5 instead?

No, really, it doesn't matter. Even if you have complete amnesia, you can make your decision regarding what you're going to do next based on what you see. If all you're doing is exploring the game, then fine, have at it. Just look at your map and decide where you're going to explore next. If you go back to a place you've been before, so what as long as you're having fun?

> if there are combat logs then I know I should be careful

The starmap already does that inherently with the presences display.

> actually a lot of the time these days I do not find it because I have even less time and am less inclined to spend my time wondering what it was that I forgot rather than playing the game

That sounds like exactly what you should be doing. What's the problem?

> This is clearly something that needs to be a option for the player

If you want to let the player delete logs, do so by having an interface for them to do it themselves. The option you suggest for "when to autodelete" is far, far more complicated than is necessary. Logs should not be designed to be deleted. They should be designed to be permanent.

src/nlua_shiplog.c
onpon4: Why can't the missions just make a string up that they'll use? It could be unique and that would be fine. If you have one function it unifies everything and makes it crystal-clear to scripters where commonalities lie as far as categorization.

agb32: Right, I'll do that... will push later.

PhoenixRiver:
*sigh*

> Not at all, simply that your memory doesn't matter for the cases you're
> talking about.

If I was heading for one par tof the universe to do something but have
to stop midway through due to real life then a log of "unimportnat"
things tells me which way I was going... I fail to understand why you
can't see that.

>> I am an explorer and I rarely do missions, esepcailly combat heavy ones.
> So?

So logs adding only missions adds nothing at all to my game and
incereases complexity for no benefit to me. I thought you were against
that.

> Riddle me this: suppose you don't remember a single thing about the game.
> You're starting on, say, Secondus Station, you see your map, and you have no
> idea where you got there. Log shows you've not done anything important.
> How is this any worse than just starting a new game and having the exact
> same circumstance, except that you're on Em 5 instead?

Because I don't have amnesia ever time I play... I have order
associative memory and I don't like spending my time wandering around
the same planets all the time with my time limited play time. Also you
use an extreme example to deny my clear examples (clarified above).

>> if there are combat logs then I know I should be careful
>
> The starmap already does that inherently with the presences display.

The starmap is no replacement for a log of the ships I have recently
fought; now you are comparing apples to oranges as well as suggesting
that graphical and written information are equal... if so why do we go
to the trouble of translating all this written information in game
when we could just draw pictures for it all, aside from people
processing the two types of information differently and with different
competance...

>> actually a lot of the time these days I do not find it because I have even
>> less time and am less inclined to spend my time wondering what it was that
>> I forgot rather than playing the game
>
> That sounds like exactly what you should be doing. What's the problem?

That I want to play the game... implicitly above I failed to be
explicit enough to add and this makes me not want to play the game!


>> This is clearly something that needs to be a option for the player
>
> If you want to let the player delete logs, do so by having an interface for
> them to do it themselves. The option you suggest for "when to autodelete" is
> far, far more complicated than is necessary. Logs should not be designed to
> be deleted. They should be designed to be permanent.

I have an archivist friend who would agree with you and a systems
administrator friend who would disagree with you just as vehemently
(but, to be fair, on his system the logging takes up more drive space
than the user information by a factor of two or three).

You opinion is clearly not shared by everyone but you insist that only
your should be accomodated... I argue for accomodating as many
opinions as possible... I know which would generally be considered the
more reasonable opinion but the decision is the code writer's and the
those with the ability to merge that code.

River

onpon4:
This is giving me a headache. :dizzy_face: Sorry Emmy has been kind of rude. Honestly I just want to get this discussion overwith and have an acceptable solution for everyone. That said, I do think design matters and Emmy is trying to push for good design even if they're not doing so in a particularly tactful way.

So, let's go back to the purpose of the log: the log is supposed to be a reference for important information from the past, right? So, that's mainly useful for recalling past dialog, or in one specific example, it can be used to remind the player of where you need to look to start the Shadow Vigil mission, or of plot details they've learned throughout various campaigns (like information about the Collective).

If I'm understanding, @PhoenixRiver, you're saying that you also find it helpful to have a log of what you were doing last time you were playing, correct? Which isn't in-line with what the ship log system was originally proposed for. That said, there's really no reason I can think of why we can't have both - though I would note that it should be implemented in Lua via an event, probably all contained in a log category called "travel" or something.

As for log deletion, I'm not sure automatic log deletion is particularly helpful. Maybe automatic log pruning for some logs (like limiting the log to X entries) could be useful though? That way you could have a "travel" log (including system entries and combat and such) there, and it would only go back X events, then when new events are added it would get rid of the oldest one (which would prevent it from taking up way more space than it needs to). Emmy has a concern that that would create mistrust or something like that, but we could alleviate that fear by showing in the log when it's a "pruned" log. In practice it would be specific to the travel log.

By the way, just to clarify, I don't want to advocate a return of the C-implemented travel log, rather instead implementing a new travel log via a Lua event (which could also log other interesting things like combat and such).

🕷️

src/nlua_misn.c
```diff
@@ -1020,3 +1029,95 @@ static int misn_claim( lua_State *L )
 
 
 
+
+/**
+ * @brief Creates a shiplog for this mission.
+ *
+ *    @luatparam string Name for this log.
+ *    @luatparam string Type of log (e.g travel, trade, etc, can be anything.)
+ *    @luatparam number Whether to overwrite existing mission with this logname and logtype.  Warning, removes previous entries of this logname and type.
```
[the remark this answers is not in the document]
agb32: Not sure how to do the boolean?  There is no luaL_checkbool or anything like that.
Perhaps I should implement it as a string? e.g. "yes" or "no" etc.

agb32:
I've added a string ID - so missions can now create logs with a name, e.g. "shadow", which identifies this logset.

src/nlua_shiplog.c
agb32: ok - now pushed - updated the createLog function to allow separate lua scripts get the same logset.

agb32:
Unclear how lua bools work - there is no luaL_checkbool function or anything like that - what should I be looking for?

onpon4:
`lua_toboolean` is what you want. It's not called "check" because any value is valid as a boolean, just interpreted differently.

🕷️

src/nlua_shiplog.c
onpon4: Awesome, thanks!

🕷️

onpon4:
Oh! By the way, you can mark comments/reviews as resolved once you finish them. Should help with the long length of this PR page. 😅

🕷️

agb32:
ok - lua_toboolean now done, and I think we're probably there.  
I was looking for something like luaL_toboolean, which is why I didn't find it!  
I'll mark the comments as resolved.

agb32:
Right, I think that's all done.  
I've not updated the messages in shadowvigil.lua, because I don't think I'd do that well.  Better for someone who knows more about it to do it... though I've left them there as a placeholder.
Perhaps, once this is pulled, then someone could work on that straight away.

src/nlua_misn.c
```diff
@@ -109,6 +112,8 @@ static const luaL_Reg misn_methods[] = {
    { "npcAdd", misn_npcAdd },
    { "npcRm", misn_npcRm },
    { "claim", misn_claim },
+   { "setLogID", misn_setLogID },
+   { "getLogID", misn_getLogID },
```
onpon4, line 116: Are these needed for something? I see that you used them in the Shadow Vigil logs you specified but they seem to be unnecessary, i.e. it seems simple variables would do just as well.

🕵️

src/nlua_shiplog.c
```diff
@@ -0,0 +1,140 @@
...
+}
+
+
+/**
+ * @brief Creates a shiplog for this mission.
+ *
+ *    @luatparam string idstr ID string to identify this log, or empty string for unnamed logsets.
+ *    @luatparam string name Name for this log.
+ *    @luatparam string type Type of log (e.g travel, trade, etc, can be anything.)
+ *    @luatparam boolean overwrite Whether to overwrite existing mission with this logname and logtype.  Warning, removes previous entries of this logname and type if 1, or logs of this type if 2.
```
onpon4, line 92: Some of this explanation is outdated, particularly the thing about a mission or whatever and the mention of "1" and "2".

🕵️

src/nlua_shiplog.c
```diff
@@ -0,0 +1,140 @@
...
+ *    @luatreturn number The logid of the mission.
+ * @luafunc createLog( logname, logtype, overwrite )
```
onpon4, line 94: Args list is out of date (should be `idstr, name, type, overwrite`)

🕵️

src/nlua_shiplog.c
```diff
@@ -0,0 +1,140 @@
...
+   if ( idstr != NULL )
+      free( idstr );
+   lua_pushnumber(L, logid);
+   return 1;
+}
+  
+/**
+ * @brief Appends to the shiplog.
+ *
+ *    @luatparam string message Message to append to the log.
```
onpon4, line 124: logid param is missing. That's still needed, right?

🕵️

src/nlua_shiplog.c
```diff
@@ -0,0 +1,140 @@
+/*
+ * See Licensing and Copyright notice in naev.h
+ */
+
+/**
+ * @file nlua_shiplog.c
+ *
+ * @brief Handles the shiplog Lua bindings.
+ */
```
onpon4, line 9: You could use this comment section to show a full example of log usage in between `@code` and `@endcode` tags. Take a look at the other modules for an example.

🕵️

onpon4:
Would it be possible to implement that pruning mechanic I mentioned? Not essential but would be useful for the travel logs thing

🕷️

src/nlua_misn.c
agb32: Yes, required to set/get the logid for the mission, which will ensure that it is saved correctly.

src/nlua_shiplog.c
agb32: Done...

src/nlua_shiplog.c
agb32: Done (and added maxLen

src/nlua_shiplog.c
agb32: Yes, added.

src/nlua_shiplog.c
agb32: Example added.

agb32:
I've added the pruning mechanic.   Now with createLog(), you can specify a maximum length for the log, and if it gets longer than this, the earliest entries get removed.

src/nlua_shiplog.c
```diff
@@ -0,0 +1,149 @@
...
+
+
+/**
+ * @brief Creates a shiplog for this mission.
+ *
+ *    @luatparam string idstr ID string to identify this log, or empty string for unnamed logsets.
+ *    @luatparam string name Name for this log.
+ *    @luatparam string type Type of log (e.g travel, trade, etc, can be anything.)
+ *    @luatparam boolean overwrite Whether to overwrite existing mission with this logname and logtype.  Warning, removes previous entries of this logname and type.
+ *    @luatparam int maxLen Maximum length of the log (zero for infinite) - if greater than this length, new entries appended will result in old entries being removed.
```
onpon4, line 98: The type specified should be "number" rather than "int" (Lua doesn't differentiate integers from floats), and `nil` should be used for infinite rather than zero, and the argument noted as optional (`nil` is the default of any argument / variable in Lua, so making it infinite with `nil` makes the argument optional and causes a maximum length to be recorded only if specified).

src/nlua_shiplog.c
```diff
@@ -0,0 +1,149 @@
...
+   const char *logtype;
+   const char *nidstr;
+   char *idstr=NULL;
+   int overwrite,logid,maxLen;
+   /* Parameters. */
+   nidstr      = luaL_checkstring(L,1);
+   logname    = luaL_checkstring(L,2);
+   logtype    = luaL_checkstring(L,3);
+   overwrite = lua_toboolean(L,4);
+   maxLen = luaL_checkint(L,5);
```
onpon4, line 117: I believe (to do what I commented on the doc line), you would do something like:

```
maxLen = 0;
if ( lua_gettop(L) > 4 )
   maxLen = MAX(0, luaL_checkint(L,5));
```

src/nlua_misn.c
onpon4: Oh, is it because it can change between saves? Hm... it would be better for that to be transparently handled in the background (as it is for pilots for example). Would there be any way to do that, so that you could just store the log ID or whatever in a variable and rely on it continuing to work through saves? That's how we do things in the rest of the game, requiring this different way is confusing and prone to errors.

Alternatively it might be better to merge "creating" the log and "appending" the log into one function? So that it would just look up the proper log by name at the appropriate time. Basically you would just have appendLog include the identifier string as an argument instead of log ID and it would find out the log ID purely through C code (and not expose that to the Lua scripter).

Anyway, just some thoughts. If no change is feasible this quirk at least needs to be documented.

🕵️

onpon4:
OK, so just to expand on something I said elsewhere: would there be any particular reason to not do it so it works like this?

`appendLog( idstr, message )`

And then it could use idstr to grab the appropriate ID or whatever internally? That would avoid the potential problems misn.setLogID and misn.getLogID are trying to solve (and make sure that variables are robust).

🕵️

agb32:
Its already implemented with the logid, which will be more efficient (int comparison, rather than string) - (I know this is only a small effect).
And its easier to generate the next in sequence for a logid, than for an idstr, i.e. for missions that don't want an idstr.
So, logid is used to define the logset, while idstr is used to couple different missions together (and thus create or obtain the logid).

src/nlua_shiplog.c
agb32: ok - I've changed to nil using the code below.

src/nlua_shiplog.c
agb32: Thanks - I've added this.

onpon4:
More efficient yes, but variables are supposed to be reliable between saves. You said that misn.setLogID and misn.getLogID are essential to prevent problems, yes? That goes against how the Lua is supposed to be handled. Lua scripters should essentially be unaware of the idea of sessions; it should be as if saving, leaving, and then loading the game just brings the Lua code back to how it was. If you need to recapture a log ID, that goes against that.

True that it's more efficient but it's probably not worth the potential for errors and confusion.

🕵️

agb32:
Right - so originally, I had it all transparent in the nlua_misn, but had to remove that and put it in nlua_shiplog, so that it could be used for events.  So, what I need is, in nlua_misn, a misn.createLog() that creates the log and saves the logid.
appendLog could then also be in misn.  And we'd have it separately for events.  But again, same problem, we need to save the logid.

We could do it all with idstr, but that has the potential to create confusion by different missions or events using the same idstr unintentionally.

However, I really want to get this wrapped up now, so if you're sure that forcing a idstr is the way to go, I'll do that.  That way, the event or mission would save nothing specific, and appendLog() would be appendLog(idstr,msg), rather than (logid, msg).

onpon4:
> We could do it all with idstr, but that has the potential to create confusion by different missions or events using the same idstr unintentionally.

I don't think that's a problem. You're honestly overthinking this, we've been using the `var` module with plain strings for a long time with no major problems. If there's a conflict we can always just fix said conflict. It's a lot better than making the system overly complex in my opinion.

> However, I really want to get this wrapped up now, so if you're sure that forcing a idstr is the way to go, I'll do that. That way, the event or mission would save nothing specific, and appendLog() would be appendLog(idstr,msg), rather than (logid, msg).

Yeah, I think that would be ideal.

🕵️

agb32:
ok, I'll do that...

agb32:
ok - done - updated the shadowvigil.lua too.  Its now just:
    shiplog.createLog("Shadow","Shadow vigil","Shadow mission",true);
    shiplog.appendLog("Shadow",misn_desc)

(or for pruning, add a number after the true).

onpon4:
Awesome! Sounds good! 😄

So now the project members just need to get caught up on the pile of pull requests that have built up. 😛

🕷️

agb32:
(fixed an old comment)

PR #1158, opened 2020-07-17
Some bugfixes

### LINKED ISSUE #1155: Dark Shadow quest end doesn't make much sense if the Genbu is destroyed
I ended up destroying the Genbu and presumably killing Giornio as I was in a very slow capital ship. However, this final message of the mission ("You find yourself back on the Seiryuu...") doesn't make as much sense if Giornio is dead.

### LINKED ISSUE #1156: Mission "The one with the Shopping" broken
On speaking to the Trader I get the "With what can I help you..." message, but on pressing OK the conversation ends and the mission is not updated. The same behavior repeats it I talk with the trader again, making it impossible to proceed.

### LINKED ISSUE #1159: Misson "The one with the Visit": Boarding the Tokera does nothing
I disabled and boarded the Tokera, but nothing happened and the mission did not progress. Subsequent attempts to board are met with "Your target cannot be boarded again".

This happened on the master branch with the commits from #1158 applied.

onpon4:
Fixes #1155 
Fixes #1156 
Fixes #1159

Also fixes a portrait that looked pretty bad with some image editor magic.


onpon4:
I fixed up a bunch more portraits in the same way as the first one (with some image edits, mostly just the smudge tool albeit supplemented by other things as well).

🕵️

PR #1163, opened 2020-07-21
Update German translation

Wuzzy2:
This PR updates a significant portion of the German translation. Ca. 50% completion.

I plan to post multiple translation PRs in future in irregular intervals, to make sure my local `.po` file and `master` don't diverge too much.


ids1024:
Merging, under the assumption this wouldn't be gibberish if I knew German.

PR #1169, opened 2020-07-26
No consideration about known or unknown JP with argument show_hidden …

nenau-again:
…in map_getJumpPath

Yes, that sounds very technical. Basically, it reverts a commit by @Deiz : https://github.com/naev/naev/commit/2a81121bfcfaa1de463237a338c2742d9aeb1dc1 (but I couldn't find out why that was done)

The problem that is solved with this PR is that standard cargo and escort mission don't assume anymore that you will use the hidden jumps that you know. (and I checked that it does not make the pathfinder to ignore hidden jumps while traveling). Of course, with this commit, the number of jumps displayed for cargo missions can be overestimated, but actually, the people that give the missions are not supposed to know the hidden jumps, so...

(to have the problem I described, buy a pirate hidden JP map and go to Eridani, and look at escort and cargo mission there)

So 3 possibilities :
* This change was a mistake back then
* This change was relevant back then and no longer now
* What I did breaks other things, but I didn't see what.

### DIFF
--- a/dat/gui/brushed.lua
+++ b/dat/gui/brushed.lua
@@ -379,7 +379,7 @@ function update_nav()
          navstring = _("Unknown")
       end
       if autonav_hyp then
-         navstring = (navstring .. " (%s)"):format( autonav_hyp:jumpDist() )
+         navstring = (navstring .. " (%s)"):format( system.cur():jumpDist(autonav_hyp, true, true) )
       end
    else
       navstring = _("none")
@@ -477,7 +477,7 @@ function renderBar( name, value, light, locked, prefix, mod_x, mod_y, heat, stre
       if name == "fuel" then
          show_light = player.jumps() <= 0
          if autonav_hyp ~= nil then
-            show_light = show_light or player.jumps() < autonav_hyp:jumpDist()
+            show_light = show_light or player.jumps() < system.cur():jumpDist(autonav_hyp, true, true)
          end
       else
          show_light = value < 20
@@ -875,7 +875,7 @@ function render( dt )
       if not autonav_hyp:known() then
          name = "Unknown"
       end
-      renderField( name .. " (" .. tostring(autonav_hyp:jumpDist()) .. ")", fields_x + fields_w + 12, fields_y, fields_w, col_text, icon_nav_target )
+      renderField( name .. " (" .. tostring(system.cur():jumpDist(autonav_hyp, true, true)) .. ")", fields_x + fields_w + 12, fields_y, fields_w, col_text, icon_nav_target )
    else
       renderField( _("None"), fields_x + fields_w + 12, fields_y, fields_w, col_unkn, icon_nav_target )
    end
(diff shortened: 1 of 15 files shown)


PhoenixRiver:
> Yes, that sounds very technical. Basically, it reverts a commit by @Deiz :
> https://github.com/naev/naev/commit/2a81121bfcfaa1de463237a338c2742d9aeb1dc1
> (but I couldn&#39;t find out why that was done)

It may no longer be a problem but off the top of my head I seem to
remember there was something that had problems with this so it was
changed but I seem to remember that there was discussion that it might
break other things... the problem is, I think, that this
routine/function got used for things that need to make opposite
assumptions about the existence of hidden nodes and that the users of
it were not necessarily aware of the other use case... sorry, that's
all I remember... there might have been something in the IRC logs?

Either way, I don't know whether this is still a problem but it might
be worth working out where this is used:

src/map.h
src/map.c
src/nlua_system.c
src/map_find.c
src/space.c

I'm unsure about where the problem list but this is used by jumpDist
(which is used in a number of missions of which the cargo missions
seems to be the only one that sets a flag to ignore hidden jump points
- maybe this is a problem) but it is also used for getting faction
planets and various other things... I wonder if (if the problem you
have identified is only related to cargo runs and escorts) would it
not be better fixed in those specific places by specifying the correct
flag to pass to here rather than making this change that could have
unintended consequences?

Hope this helps,

River

P.S. If you're doing this could you also update the usage list of
parameters for map_getJumpPath - it doesn't include the hidden flag -
oops!

nenau-again:
All right, after checking, I found at least 1 place where it pauses problem : in the gui Brushed, when going from Edriani to Draconis, it says that the length of the travel is 4 instead of 1 through the hidden jump. I'll investigate further, but I feel the right usage for the function is still to separate the hidden and known condition (as I did). I believe one could rather change the arguments in `jumpDist` in the gui.

>P.S. If you're doing this could you also update the usage list of parameters for map_getJumpPath - it doesn't include the hidden flag - oops!

True, I had remarked that and forgotten to correct it.

nenau-again:
>src/map.h
src/map.c
src/nlua_system.c
src/map_find.c
src/space.c

Checked. The only one where `show_hidden` is true (and thus my change has consequences) are in `nlua_systems` il lua functions `jumpDist` and `jumpPath` which usages I've manually checked (and by the way, some of them were wrong independently of my change due to start and goal systems being mixed up in the function description. I'll try to untangle all that.

PhoenixRiver:
> e8a4bfcb4d7d183c534e99cfa5c9ac93beb5ab62  adaptations and corrections in lua
> scripts and nlua doc

Cleanly done and nice catch in the pirate bounty mission - I'm pretty
sure I'd have missed that error!

River

nenau-again:
Actually, there are others ocurrences of reversed arguments in the function. I'll make an other commit about it soon.
But actually, the function with only one argument is not applicable, because it gives the distance from sys to the current system, which is very rarely used.

onpon4:
So is this ready to merge then?

🕷️

nenau-again:
>So is this ready to merge then?

I think I've nothing more to add on this branch.

PR #1170, opened 2020-07-26
[WIP] Replace Travis CI with Github Actions

ProjectSynchro:
This is a cleanup of [PR#1061](https://github.com/naev/naev/pull/1061) (I go over what this actually is and how to configure it)

As for merging this with the master branch, in it's current state the configuration file that I whipped up should work side by side with Travis, but can effectively replace it. I commented out the LuaDocs portions as a new SSH key will be needed to get the rsync portion of that action working.

I just need to iron out some missing dependencies at build-time and this should be ready to go, assuming this that is done.

### DIFF
--- a/.github/workflows/naev_ci.yml
+++ b/.github/workflows/naev_ci.yml
@@ -1,0 +1,186 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  win32:
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    steps:
+    - name: Setup MINGW32 Environment
+      uses: eine/setup-msys2@v1
+      with:
+        msystem: MINGW32
+        update: true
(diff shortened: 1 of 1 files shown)


PhoenixRiver:
> c0163fd6116cae180b8b023b768d910e0e1b1e3c  I think this is all of the missing
> deps.

I don't think it makes a difference to the result but you have
autoconf-archive in there twice - once after autoarchive and then
again at the end of the list (in the first dep list)...

River

PhoenixRiver:
Also some of the messages sent to email have a "noreply" address
rather than the address referring to the PR!? Maybe a recent bug
introduced by Github? Or something related to the "master/slave"
changes? The latter seems unlikely though...

River

ProjectSynchro:
@PhoenixRiver Looks like that was definitely a typo, I must've missed that before I committed it.

I'm currently trying to figure out why XML::Parser isn't installed / being detected on MacOS 10.15.
IIRC it should be there, but perhaps some recent changes to XCode are to blame. 

I'm not really that accustomed to MacOS so it looks like compiling the missing module might just end up being required unfortunately, unless I find some breakthrough.

PR #1173, opened 2020-07-27
[WIP] Improve CI build-times

ProjectSynchro:
Previously, MacOS builds were hindered by INTLTOOL requiring the XML::Parser perl module.
Because of this there was the need to compile this missing module at build time. 

It turns out that MacOS' native perl has XML::Parser installed, so removing the version of Perl that is preinstalled in the GitHub Actions MacOS image (via Homebrew) seems to fix this. 

This should speed MacOS test build times by approx 5 minutes.

On top of this, I have changed the MSYS action to the officially supported one. (Same developer, new repository name)
I have also sped up windows builds by using the version of MSYS that is preinstalled on Windows System images.


ProjectSynchro:
The failed checks for win32 and win64 should be fixed by [PR#1174](https://github.com/naev/naev/pull/1174)

PR #1178, opened 2020-07-28
Retooled News Event

### LINKED ISSUE #1082: Economy info: News, NPC, purchase or some combination?
A topic for discussion - now that we have an economy, it would be good to provide information to the player in ways other than just visiting the planet.
Initially we had NPC messages (people in bar), who would tell the prices on a nearby planet (up to 5 jumps away), at some time in the past, as if they had just come from there (Njumps x 2 STP ago).
This was changed to be a News item - which would give all prices on all planets in all neighbouring systems.
Another option has occurred to me - the player could also buy economy information, rather like they do maps.  e.g. if they're taking a courier package somewhere and have some space, they might want to buy economy information about that place so that they can fill up their hold.

So what do we think - do we want some/all of these, and which ones?  Or something else.

I'll post my opinion below... (so as not to clutter up this main post)

onpon4:
This retooled news event completely changes the way the news script works so that it's much more robust and nice.

Firstly, the event no longer tries to remove and then add a whole bunch of articles. Instead, it uses the news system in a more normal fashion, simply adding articles that expire after a while.

Secondly, to avoid posting the same article twice, it now uses tags rather than article titles. This prevents duplicates when in other languages (#1077).

Thirdly, economy news articles (for distant-ish planets) have been added. These are news articles, but they also give you pricing info for a random time in the past at a distant-ish planet.

Fixes #1082


dat/events/neutral/news.lua
```diff
@@ -1,238 +1,115 @@
...
-
-articles={}
+articles = {}
 
 articles["Generic"] = {
    --[[
       Science and technology
    --]]
    {
+      tag = "Techs Probe for Sol",
```
bobbens, line 17: I have added gettext.gettext_noop which is basically a way to get gettext to detect the sentence for translation, but doesn't translate it. This should let you do something like
```
tag = gettext.gettext_noop("Tech Probe for Sol")
title = _(tag)
desc = _("Technicians...")
```

Should be possible to just make the titles be untranslated, and translate it when adding the news. Or even translating it C-side.

dat/events/neutral/news.lua
```diff
@@ -1,238 +1,115 @@
...
--- Event for creating generic news
+-- Event for creating news
 --
 --]]
 
 require "numstring.lua"
 require "jumpdist.lua"
 
 
-header_table={}
```
bobbens, line 10: Why were the headers removed?

dat/events/neutral/news.lua
onpon4: Ah, yeah, that should be useful!

Actually there might be some other improvements we could make with that... hold that thought, we'll look into that today.

🕵️

dat/events/neutral/news.lua
onpon4: We removed those because it seemed we couldn't guarantee them to show up on top, or at least based on how it was put together before that's how it looked. We'll do some experimenting with that.

🕵️

dat/events/neutral/news.lua
onpon4: Alright, I've added in a use of something like this. Not in the way you described because that way's actually impossible, but by using tag to generate the title if title is unset.

🕵️

dat/events/neutral/news.lua
onpon4: Figured it out, so it's the date, huh? Added the headers back in, this time in a more robust fashion (no more duplicates... at least for new saves, I'll mention more on that in the main conversation).

🕵️

onpon4:
OK, so I've resolved the two issues @bobbens had with this PR, so merging.

Before I do that, I want to explain one thing that might look like a regression, but actually isn't: the old news system used *titles* to prevent duplicate articles, which was flimsy, and it also didn't make the articles automatically expire in a reasonable timeframe. So in existing saves *old titles and headers will remain*. I considered adding in a thing to remove those articles, but it would be weird and messy, and since *new* saves aren't affected, it seemed to not be worth it. But you can fix your own save thusly:

1. Open the console with F2.
2. Type the following: `for i, a in news.get() do a:rm() end`

This will simply erase all articles and reset you to a start point (as if you were in a new game). The news will work as normal after that.

Issues like this shouldn't show up anymore in the future both because articles have designated expiration times in the near future and because they use tags to prevent duplicates rather than titles.

PR #1184, opened 2020-07-31
Update Lua Documentation to have a common look and feel to the rest of the Naev site

ProjectSynchro:
I saw this and had a couple free hours to spare so I took a shot at this
I am not a professional web developer by any means..

I have updated:
- [x] Header
- [x] Footer
- [x] Adjusted build script to merge parts of the old CSS with the new CSS

I also have changed the 'about' div to actually show the version of LDoc used to generate the documentation


ProjectSynchro:
You can take a look at what it looks like by downloading the luadocs artefact if you're curious.

ids1024:
> I have minified the CSS, due to the 20k+ lines of the file.

Hm, do we really want it minified in the repository?

If it's minified, that should probably be done as part of the process of building the documentation.

ids1024:
Testing it, this generally seems like an improvement.

Screenshot from 2020-07-30 20-56-38

It doesn't look quite right to me to have the text here almost at the very right of the screen. I'm not really an expert designer either, but it seems there should be some extra spacing there in some form or another.

ProjectSynchro:
I have "De-minified" the minified CSS

> Testing it, this generally seems like an improvement.
> 
> Screenshot from 2020-07-30 20-56-38
> 
> It doesn't look quite right to me to have the text here almost at the very left of the screen. I'm not really an expert designer either, but it seems there should be some extra spacing there in some form or another.

Looks like I missed the margins, I have fixed this. 

image


> > I have minified the CSS, due to the 20k+ lines of the file.
> 
> Hm, do we really want it minified in the repository?
> 
> If it's minified, that should probably be done as part of the process of building the documentation.

This can be added to the Lua-Documentation check, I think I saw some sort of node package that can do this. 
I think that change should be made whenever the actual updates for api.naev.org are configured.

ids1024:
So, before this change `ldoc.css` is 503 lines. With this PR, it grows to 21988. I was kind of confused at where this extra css came from, but I guess this is https://naev.org/style.css de-minified?

It looks like the css on the Naev website is generated from scss, and I guess has some fairly heavy framework dependencies: https://github.com/naev/naev-website/tree/master/content/scss

Minified or not, this is a lot of generated code to add to the repository, that will fall out of sync with the website. Not sure how best to deal with that... perhaps it should just include a second `<link>` tag to include the css from naev.org. This also means a browser has one less file to download/cache.

And is it old fashioned of me to not like that simple websites require such massive css files? But that's an issue with the https://github.com/naev/naev-website, and I guess we can blame @bobbens....

ProjectSynchro:
It was my understanding that you can only have one CSS stylesheet per page.
You're definitely not alone in not wanting such a massive css file.

I didn't want to cherry pick out of [https://naev.org/style.css](https://naev.org/style.css) because as you said, it can fall out of sync very quickly.

Perhaps I can write something into [luadoc.sh](https://github.com/naev/naev/blob/master/docs/luadoc.sh) that generates the CSS file at build time.
The only parts that need to be added and modified are around 200 lines near the start of the file. Those could be included in plaintext somewhere. 
This [line](https://github.com/ProjectSynchro/naev/blob/c0fe045e0a0b91f8e2318a27beabc6e06a8e7c47/docs/lua/ldoc.css#L10880) would always need to be modified otherwise the nebula effect won't work correctly.

I changed this to a draft as it seems some more work is needed.

ids1024:
> It was my understanding that you can only have one CSS stylesheet per page.

No, you can definitely use multiple CSS files in one page.

ProjectSynchro:
> > It was my understanding that you can only have one CSS stylesheet per page.
> 
> No, you can definitely use multiple CSS files in one page.

Well that sure goes to show that I am by no means a web developer..
You learn something new everyday I guess.

Well 1 line fixes the issue of having a massive CSS file, Any outstanding issues should be fixed/finished now.

PR #1186, opened 2020-07-31
Update German translation

Wuzzy2:
Another translation update for German. 70% is reached.

Please, can you update the POT files after merging?


onpon4:
POT files actually aren't stored on the repo anymore. I believe they're generated with the normal ./configure && make process.

PR #1191, opened 2020-08-01
Added a system "map"

agb32:
Not yet quite ready (needs tidying up), but I thought I'd create a pull request to get comments.

This creates a new dialogue screen that allows the player to view information about individual systems.

From the map screen, select a system, and then click on it again - this will pop up a new screen showing information about this system.  Some is purely for interest (population, star type etc), but it also allows you to navigate planets, and see more about them - e.g. what the shipyards sell, what the outfitter sells, currently known cost of commodities there, whether it has missions available etc.

In particular, useful if deciding on which mission to select, e.g. for a shipping mission, you can then view the destination (if known), and see what facilities it has - and for example, if you see it doesn't have missions, you might decide not to accept the shipping mission there...

Some screenshots:
tmp
tmp2
tmp3


onpon4:
> Not yet quite ready (needs tidying up)

OK, so when a PR isn't ready yet, you can mark it as a draft PR. This helps communicate more easily that it's not ready to merge yet and when it's ready to merge. I'm going to mark it as a draft for you; feel free to mark it as ready whenever it can be merged.

🕵️

src/map_system.c
```diff
@@ -0,0 +1,1076 @@
...
+  return 0;
+}
+
+/**
+ * @brief Destroys the map subsystem.
+ */
+void map_system_exit (void)
+{
+   for( unsigned int i=0; i<nBgImgs; i++ ){
+      gl_freeTexture ( bgImages[i] );
```
onpon4, line 140: OK, so, a couple style comments (these won't prevent a merge but just want to let you know):

* `for` statements (and other things that aren't functions) should have a space between them and the parentheses following them.
* Function calls (such as `glx_freeTexture` here) should have the opening parenthesis immediately after, with no space.
* This one's minor, but there should be a space between a closing parenthesis and an opening angle bracket.

So this particular section of code should instead be:

```
   for ( unsigned int i=0; i<nBgImgs; i++ ) {
      gl_freeTexture( bgImages[i] );
```

nenau-again:
Looks amazing !
One tiny remark: the way you mark an item as selected (for example Arcturus Gamma on the two last pictures), it doesn't match the way used everywhere else, with round corners (for example if you select something in space).

src/map_system.c
agb32: Yeah, thanks - that's mostly the tidying up I was talking about - old habbits die hard (what I started programming, I was saving to floppy disks, and every byte counted, as I had to buy them with my pocket money!

agb32:
thanks - good point about the marked item - I'll change this... also a few niggles still to sort.

onpon4:
Ah, can I suggest changing the colors to white for highlight and either gray or some other dark color for no highlight? It helps with colorblind accessibility (that bright green is actually darker than white for some forms of colorblindness).

🕵️

agb32:
ok - looking better now - I've changed the faction colour text from dark grey to yellow.
However, I've not changed the ship stat text to white, because this is just how ship->desc_stats formats it.  So if we decide to change that it will affect the shipyard and here.

onpon4:
Alrighty, looks good!

🕷️

PR #1192, opened 2020-08-01
fix in the collision algo

nenau-again:
Yeah, I've been a bit too much in a hurry with https://github.com/naev/naev/pull/1188.
And about the previously mentioned decreasing in performance, I'm not too sure anymore as my laptop seems to be quite not so deterministic CPU-wise right now (maybe it's due to the heat wave in my country)

I mark as a draft for now because I want to test it a bit more than last time, but I wanted people to be aware of the fact that there was a problem with https://github.com/naev/naev/pull/1188, and that it's probably solved. I'll remove the tag once I'm sure.

(the issue was that the collision often happened without any explosion, the ammo was simply removed and the health substracted)


nenau-again:
It seems all right

PR #1193, opened 2020-08-02
Added outlines to text and improved the tk look.

onpon4:
This has been a long time coming, and we've finally managed to do it! I think I've achieved something that looks fantastic and finally, completely solves the issue of text contrast forever.

The first thing we did was Emmy added text outlines in an admittedly crude fashion; the method is borrowed from Project: Starfighter.

Then we changed the text colors from black to white (and dark to light).

And then we noticed something: tk having a bright background was creating some eye strain, so we set out to change that and thought, you know, since we've gotten our hands dirty with the toolkit, we might as well do something we've wanted to do for months: flatten the colors and make it look much nicer!

Screenshot at 2020-08-01 20-52-48
Screenshot at 2020-08-01 20-56-28

Behold, the new look of Naev! It's nice and simple, so it doesn't look like a product of the early 2000s, and it's dark, so it's easy on the eyes. We also made it so that it would be easy to change the colors in the future, as well, removing most custom cGrey color usages and replacing them with the three standard colors.

I believe everything looks good and this should be merged ASAP. However, I'd like to make sure we agree that this is an improvement first since I've changed the entire look of Naev. 🙂

🕷️

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -570,18 +570,20 @@ void gl_printRaw( const glFont *ft_font,
     * ultimately replaced with use of signed distance fields or some
     * other more efficient method (signed distance fields seem to be
     * the "right" solution). */
-   gl_printRawBase( ft_font, x - 1, y, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x + 1, y, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x, y - 1, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x, y + 1, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x - 2, y, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x + 2, y, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x, y - 2, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x, y + 2, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x - 1, y - 1, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x - 1, y + 1, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x + 1, y - 1, &cBlack, text, 1 );
-   gl_printRawBase( ft_font, x + 1, y + 1, &cBlack, text, 1 );
+   if ( (c == NULL) || (c->a >= 1.) ) {
+      gl_printRawBase( ft_font, x - 1, y, &cBlack, text, 1 );
+      gl_printRawBase( ft_font, x + 1, y, &cBlack, text, 1 );
+      gl_printRawBase( ft_font, x, y - 1, &cBlack, text, 1 );
+      gl_printRawBase( ft_font, x, y + 1, &cBlack, text, 1 );
+      gl_printRawBase( ft_font, x - 2, y, &cBlack, text, 1 );
+      gl_printRawBase( ft_font, x + 2, y, &cBlack, text, 1 );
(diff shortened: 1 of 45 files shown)


AmbientCyan:
A lot of this feedback is going to be very nitpicky so you can adopt what you want and ignore the rest:

* The font against the gray background is very weird because at a glance it seens to have different brightness with thinner letters than thicker letters. I think that a slightly heavier font weight would help in this case. Maybe not full bold but definitely more than it is now.
* I would recommend rounding the buttons rather than having them square, to match the look and feel of the dialog window. Don't have to make it as pronounced as the window though, 4-6px should be enough
* I would recommend making the hotkey typehint in the buttons an actual color instead of dark gray to make it stand out a bit. It also matches the background too closely so it makes it look like it's transparent almost. Maybe use the red of our favorite pirate ships?
* The tabs that aren't selected have a light bevel, but it should probably be dark since the deselected tabs are "recessed" behind the active one. Try a dark gray, maybe `#222`.
* I'm not sure if it is trick of the eye but the light background appears larger int he disabled button in your first screenshot. I'd recommend removing it or changing it to a more contrasting color.

onpon4:
We've made some changes to this, partly from AmbientCyan's comments and partly just from some things we thought about. Let's show some updated screenshots...

Screenshot at 2020-08-02 00-56-51
Screenshot at 2020-08-02 00-57-09
Screenshot at 2020-08-02 00-58-12
Screenshot at 2020-08-02 00-58-28
Screenshot at 2020-08-02 00-59-12

The screenshots show most of the changes clearly enough. One other change I made was to add accessibility symbols to accompany color indications! Currently used to show the status of planets and highlight ship problems in the outfitter.

If there are no objections I think I'll merge this tomorrow.

🕷️

onpon4:
OK, just one more tweak for now. 😅

I tweaked the tabs to get rid of an oddity that may have created an undesirable optical illusion:

Screenshot at 2020-08-02 01-27-52

🕷️

src/font.c
```diff
@@ -547,11 +546,43 @@ void gl_printRaw( const glFont *ft_font,
...
+ *    @param ft_font Font to use
+ *    @param x X position to put text at.
+ *    @param y Y position to put text at.
+ *    @param c Colour to use (uses white if NULL)
+ *    @param str String to display.
+ */
+void gl_printRaw( const glFont *ft_font,
+      const double x, const double y,
+      const glColour* c, const char *text )
+{
```
bobbens, line 568: Don't think it's a problem, but font rendering is fairly inefficient and slow, and I think this will become much slower. However, don't think it's going to be a real issue (hopefully) on anything except insanely old systems. Might be worth thinking about rewriting this to use shaders.

onpon4: Yes, that would certainly be worth thinking about. Actually @ids1024 mentioned using signed distance fields to do this the proper way.

I'll add a TODO comment to document the shortcoming. It shouldn't be a problem, just something to improve in the future.

🕵️

onpon4:
Alright, so I'm going to figure out a solution to one particular problem before I merge this: because text is effectively larger than it used to be, the fact that text is printed right on top of assets on the overlay map thing in many cases is worse than it used to be because in some cases, it can completely obscure small assets that aren't selected. I'm going to look into finding a way to make sure that asset name text never obscures another asset, which might entail an overhaul of how that looks.

onpon4:
Alright, I've fixed the issue I mentioned simply by making an exception for text that has semi-transparency, and making sure that overlay text is always semi-transparent to some extent. This is a very hacky fix and I'm not happy with it, but it at least leaves the overlay looking the same as before. Ultimately it should be modified so it isn't clobbering text together in the first place. Perhaps I'll open an issue on that.

Anyway, that was the only major issue, so merging this now.

🕵️

PR #1198, opened 2020-08-02
added solar system map to CHANGELOG

agb32:
More to CHANGELOG.
We probably should change the text about "Brushed Gui" too.


onpon4:
Sorry we missed this, we actually added something similar to this to the changelog I believe some time after you opened this PR. Redundant now, so closing.

PR #1199, opened 2020-08-03
Fix ldoc generation to make bobbens happy

ProjectSynchro:
Title. 
Turns out you need to configure in order to make :P

Also of note, ldoc should be included in the PATH.
In the case of the runner VM; ldoc is installed locally, therefore the local install path is added to the PATH.

### DIFF
--- a/.github/workflows/naev_ci.yml
+++ b/.github/workflows/naev_ci.yml
@@ -76,7 +76,7 @@ jobs:
       run: |
         sudo apt-get update
 
-    - name: Install Build Dependencies
+    - name: Install Additional Build Dependencies
       run: |
         sudo apt-get install \
           build-essential \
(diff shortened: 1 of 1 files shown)


ProjectSynchro:
One issue with the lua generation.
Just need to fix quickly one commit

onpon4:
L:Hahaha, yes, I was noticing the doc build failure.

ProjectSynchro:
Should be good now :)

onpon4:
Looks like it. Thanks!

🕷️

ids1024:
Would @bobbens be too offended if I suggested it would be better to just call `ldoc.sh` without make to avoid the dependencies? :laughing: 

Otherwise, it would be nice if there's a good way to not have to duplicate the list of dependencies in both this section and the Linux build.

Apparently Ubuntu [packages `ldoc`](https://packages.ubuntu.com/search?searchon=names&keywords=lua-ldoc), so I guess we can at least simplify installation of that. (Perhaps I set up the Travis code before they packaged it? Or at least before Travis defaulted to a version that did...).

PR #1200, opened 2020-08-03
Cleanup the rest of the other unused CI files

ProjectSynchro:
Seeing as #1199 seems to have worked without any failures (aside from Travis), it is safe to remove what's left of Travis from the mix.

As for a replacement for the BinTray stuff...
I am close to finishing something that should provide nightly builds of Naev on Github.


onpon4:
Looks good. Thanks!

🕵️

PR #1203, opened 2020-08-05
fixed pilot.setNoDeath

nenau-again:
Previously, a ship with this option was set to 1hp any time it was hit.


nenau-again:
Damn. There is also a problem with `setNodisable`. I'll put it on the same PR (if I manage to solve it) and I'll change the name for `setNoDisable`

onpon4:
Just want to point where you missed some calls in the name change: one in 02_sciwrong, and several in dv_antiflf03.

🕵️

nenau-again:
Now it should be allright. Both function do what is intended if used separately, and together as well.

onpon4:
There's one more I think you missed, but I'll fix it. Thanks!

🕷️

PR #1205, opened 2020-08-07
Conversion of traders to a static faction (like civilians)

onpon4:
This is just a bit of a proposal and a demonstration of what that entails. I don't really care either way whether or not we go with this, but just want to make my case that honestly, I don't feel like traders should be a faction proper with reputation boosts and such. They should be a static faction like civilians. There's a couple reasons for this:

1. Traders are all over the galaxy, so the idea of traders having their own position in the galaxy seems a bit silly.

2. Traders don't fly military ships and the trader AI isn't designed to defend systems from piracy in any way.

3. The player theirself can be a trader and yet not be a part of the "space traders guild". Wouldn't the simple act of trading make you a space trader?

4. I can't see any direction traders as a "real" faction could go that would make sense and be fun.

Anyway, let me know what you think!

🕷️


PhoenixRiver:
I don't know how others have thought about traders as a faction but my
idea is of them being a loose association (as you mentioned) merchant
guild where merchant companies work together to be a larger entity for
commercial purposes, they are not territorial (except maybe against
each other?) and as traders generally just run from pirates. As for
possible missions, any of the missions that already exist could be
made faction missions for the traders faction with advancement being
linked to missions that involve sorting out the differences between
different trading companies within the guild, benefit to completing
these missions could involve being imparted better market information,
knowledge of hidden jump points, cheaper escort hire etc.. I'd see
advancement with the trader faction being parallel with any other
faction advancement since most other factions wouldn't have a problem
or might even like a player with greater standing in the trader
faction - the exception would be the pirate faction I guess.

Mechanically doesn't the trader faction being dynamic allow the
spawning of trader ships in systems using the simple faction presence
mechanism? If so, then wouldn't removing it mean trader ships wouldn't
spawn any more? (This would make being a pirate faction associate very
much harder, for one).

Just my tu'penny woth of thoughts,

River

P.S. It could be made that rewards are dependent on which other
factions the player is in good standing with - for example if the
player is in good standing with the Soromid faction hidden jump points
that are given as rewards could be limited to those within Soromid
faction space.

onpon4:
> Mechanically doesn't the trader faction being dynamic allow the spawning of trader ships in systems using the simple faction presence mechanism?

No, faction standing is determined by the existence of assets (which for both traders and civilians takes the form of "virtual" assets). So that's a separate issue.

All making a faction static does is mean that there's no reputation changes with the player, like civilians; the player is always neutral with civilians, and this PR makes that so for traders as well (and miners, which I added here after initially opening the PR).

🕷️

PhoenixRiver:
> No, faction standing is determined by the existence of assets (which for
> both traders and civilians takes the form of "virtual" assets). So that's a
> separate issue.

Oops, I had forgotten that that change had been made!

Though this doesn't negate the potential benefits to more commercially
minded players that a trader faction mission set could have...

River

onpon4:
> Though this doesn't negate the potential benefits to more commercially minded players that a trader faction mission set could have...

That's a separate issue to I think, the only difference here would be that said mission set wouldn't provide a reputation bonus for the Trader faction. What this change would really affect is if you wanted a large Trader campaign progression (as with the FLF), rather than decentralized missions / campaigns (as with the various neutral missions).

🕷️

PhoenixRiver:
> That's a separate issue to I think, the only difference here would be that
> said mission set wouldn't provide a reputation bonus for the Trader faction.
> What this change would really affect is if you wanted a large Trader
> campaign progression (as with the FLF), rather than decentralized missions /
> campaigns (as with the various neutral missions).

which is what I was refering to as per my previous message...

onpon4:
Ah ok, I thought you were just talking about campaigns in general.

🕷️

onpon4:
OK, I'm closing this in favor of keeping the Traders Guild as an organization. Supposedly the idea is to make the Traders Guild like a secretly evil capitalist corporation faction, which I really like the sound of!

🕵️

PR #1207, opened 2020-08-09
added Lua function to make the camera shake

nenau-again:
`camera.set` was a bit lonely, I found her a friend.


bobbens:
Looks good, thanks!

PR #1208, opened 2020-08-09
added spfx utility containing only sound play for now

nenau-again:


src/nlua_spfx.c
```diff
@@ -0,0 +1,105 @@
...
+   return 0;
+}
+
+
+/**
+ * @brief Lua bindings to interact with the Special effects.
+ *
+ * An example would be:
+ * @code
+ * @endcode
```
onpon4, line 52: Did you mean to put an example here?

🕵️

onpon4:
Everything looks good here other than the weirdness of an empty example. Merging. Thanks!

🕷️

Issue #1181, opened 2020-07-29
Let's invade the Frontier


nenau-again:
Recently, @agb32 and I had a conversation on discord about the beginning of the Dvaered main campaign. I guess people agree it should involve an invasion of the frontier. So I'm creating this issue to put the ideas together before processing to the coding.

In my idea, the player should be given the following tasks:
* Help negotiations between warlords and Dvaered High Command to settle the battle plan
* Prevent the said battle plan to leak
* Help diplomatic negotiations with the Empire, House Goddard and House Sirius
* Enforce Dvaered positions (the idea was a base building mission in Zacron)
* Take part to sabotage missions to weaken the Frontier defences
* Take part to the invasion itself

Then I recalled that @BTAxis has scenarios for almost everything in store. Does it exist a frontier invasion campaign draft ?

About the timing, I propose to start merging stuff only after the definitive .8 release. At this time, I'll be busy with the safe lane implementation (and with the end of my holidays probably). So if someone else is interested to code the mission, it would be perfect. Otherwise, I'll do it myself.

In the mean time, I'll take any proposal, suggestion, ~insult~, correction...

onpon4:
So I think the Empire shouldn't get involved in the invasion of the Frontier, but one idea that could be interesting is if the Sirii get involved somewhere. Like maybe they want to invade the Frontier themselves, or maybe they want the Frontier as a buffer, or maybe they just want galactic conquest and see the Dvaereds as a first step. Then you would have the Empire indifferent about the Frontier, but have to pick a side between the Dvaereds and Sirii to try to keep the fractured Empire together, only for their efforts to fall apart and them to give up entirely.

Just a thought.

🕵️

nenau-again:
All right, I have written the synopsis of the campaign. I guess I'll start the coding today or tomorrow, and I'll make a draft PR once, well I have a draft campaign.

agb32:
Sounds good - happy to help if it can be of use...

nenau-again:
Help is welcome : 

* Proofreading once the texts are done (if you're a native speaker)
* Comments on the attached synopsis (obvious spoilers) [frontier.txt](https://github.com/naev/naev/files/5018274/frontier.txt)
* If you feel like coding some of the missions (not mission 1 and 2, I'm already working on them)

I guess I'll work on my repo, so if you want to contribute, I'll make the draft PR so that you'll get the adress to PR on my repo.

nenau-again:
[...]

PR #1210, opened 2020-08-10
Frontier invasion: missions 0 to 4

nenau-again:
As in issue https://github.com/naev/naev/issues/1181
The first two missions so far.


agb32:
Looks good - I've spotted a few typos, so I'll do a pull request at some point.

nenau-again:
As said on discord, the beginning of the campaign is ready for review.
Note that mission 4 has not been proofread for language.

bobbens:
Some feedback quick feedback on the first mission:
1. First mission should make it clearer you need a fast ship to keep up. I just tried it with an Admonisher and got mission failed. It is further frustrating because there is no shipyard so I can't change ships either, so maybe also limit that. Alternatively, make them slower.
2. The "Tam is lost" messages are confusing even when saying he jumped somewhere. You should probably change the order so it says what he did last. Something like "Tam is lost!\nTam has jumped to the XXX system" instead of the opposite order as it is now.
3. When the game reloads, it fails saying "You should be escorting Tam, what are you doing here?". This is probably because I changed it so the land hooks run on load
4. To give the player a fighting chance, maybe add a delay between fight starting and Tam boarding the Goddard. I basically got creamed by a vendetta next to me (scripted cinematic scene so can't choose where to be), and vaporized before I had a chance to do anything.

PR #1211, opened 2020-08-15
draft of mission map generator

nenau-again:
With subgraphs and something to differentiate factions and difficulty, it could become decent...


bobbens:
Looks good, thanks!

PR #1213, opened 2020-08-15
Za'lek campaign

V-ktor:
I already started working on a Za'lek campaign but actually gave up on writing texts and dialogs. Anyway, maybe I should just write down what I came up with. I don't think this is ready to be merged right now.

Story wise the first problem is: Why would the Za'lek entrust important missions to the player in the first place? My answer is that, at least for the moment, they don't. Instead the first mission is very dangerous and badly paid.

**Mission 1: Novice Nebula Research**

Description: A Za'lek student asks the player to help him with his research. The player has to visit Doeston and Arandon.
Some minor complications on the way have to be expected.

Difficulty: Easy

Note: The player will lose control over his ship and shields will fail for a moment (causing some hull damage).
The payment for this mission should be substantially lower than usually, possibly compensated by better paid follow-up missions.

**Mission 2: Advanced Nebula Research**

Description: This time the post doc asks the player for help. Reference data of further nebula are needed to understand the Sol nebula. The player has to escort a transport ship through pirate and Dvaered space into the PSO nebula.

Difficulty: Medium

Note: A random Dvaered patrol will attack the transporter as they consider a ship of another faction stuffed with advanced sensory as highly suspicious. This mission may takes too long time as the path is rather long.

**Mission 3:**

Description: The player has to impersonate a scientist and give a scientific talk.

Difficulty: Easy

Note: Just to make it more ridiculous. A good performance will increase the standing significantly.

**Mission 4:**

Description: The player has to help acquiring funds for the shielding protoype. This is going to be hilariously complicated.

Difficulty: Easy to Medium

**Mission 5:**

Description: The nebula shielding prototype is ready and the player has to test it. The first stages go well, but in the final stage the prototype fails. At his point a Thurion ship appears. It is unclear whether this ship sabotaged the shielding prototype or helped the player. The player has to abort the experiment and retreat.

Difficulty: Easy (requires at least a corvette with some extra shields or a larger ship to survive the loss of the shielding prototype)

Note: There will be an outfit that provides some nebula resistance for shields which will be removed and replaced by a 'broken' one after this mission. It is unclear whether it is a malfunction of the prototype or sabotage. Some groups within House Za'lek suspect afterwards that the player had contact to the Thurion (they probably don't know for sure what the Thurion even are) and thus the player becomes interesting to them.

The next part should actually happen after the second campaign. It is probably too early at this point.
After the second campaign the player is stranded in Thurion space. They give the player the freedom to do whatever he wants (it's impossible to leave anyway). They offer the player a mission, though. I guess it has to involve the Proteron as otherwise no other faction is accessible at this point. Maybe the Proteron are about to find out about the Thurion (because they captured a scout or something?). Anyway, the reward will enable the player to leave Thurion space. Joining the Proteron could also be an option. In this case the player has to help them to acquire Thurion shields so that they can send out scouts into the rest of the galaxy.
Afterwards the player has the choice to either keep the Thurion secret or tell the Za'lek about it and proceed with the next Za'lek campaign (the ship's logs have been replaced with faked ones so there are actually no proofs). In the former case the player can start the Thurion campaign (which could also be possible via other campaigns where the player makes contact with the Thurion).


dat/missions/zalek/neburesearch/neburesearch_00.lua
```diff
@@ -0,0 +1,221 @@
...
+title[1] = _("Bar")
+title[2] = _("Please...")
+title[3] = _("What a mess...")
+title[4] = _("Ready")
+title[5] = _("Technical problems")
+title[6] = _("Shadows in the darkness")
+title[7] = _("Mission Success")
+text = {}
+text[1] = _([["Hello Captain! You are a pilot, right? For my project I require a ship heading to the Sol Nebula. Certainly you must be interested in the proposal of researching the phenomenon that cut us off from mankind's patrimony.
+    As this is the point where any other pilots I asked backed out I should start with my problem; due to some unfortunate circumstances the payment for this mission will be only 50000 credits. But rest assured, you will be mentioned in the acknowledgment section of my next paper!"]])
```
onpon4, line 36: Numbers should be written with commas, so this should be "50,000 credits". Or better yet, use the creditstring() function from numstring.lua.

🕵️

dat/missions/zalek/neburesearch/neburesearch_00.lua
```diff
@@ -0,0 +1,221 @@
...
+
+t_sys = {}
+t_sys[1] = "Doeston"
+t_sys[2] = "Arandon"
+
+-- localization stuff, translators would work here
+bar_desc = _("You see a young scientist talking with some pilots, apparently without success.")
+mtitle = {}
+mtitle[1] = _("Novice Nebula Research")
+misn_reward = _("%s credits and the gratitude of a student")
```
onpon4, line 23: You should replace `%s credits` with `%s` and use the `creditstring` function found in numstring.lua. This will make sure that it can be translated properly.

🕵️

dat/missions/zalek/neburesearch/neburesearch_00.lua
```diff
@@ -0,0 +1,221 @@
...
+mdesc = {}
+mdesc[1] = _("Go to the %s system.")
+title = {}
+title[1] = _("Bar")
+title[2] = _("Please...")
+title[3] = _("What a mess...")
+title[4] = _("Ready")
+title[5] = _("Technical problems")
+title[6] = _("Shadows in the darkness")
+title[7] = _("Mission Success")
```
onpon4, line 33: This is minor and you don't need to change this since you already did it this way, but I just want to recommend against this old style that used to be encouraged of throwing everything into ordered tables. It makes the names non-descriptive and confusing. Like I said, no need to change this here, but I just wanted to strongly recommend avoiding this style for future missions.

🕵️

dat/missions/zalek/neburesearch/neburesearch_00.lua
```diff
@@ -0,0 +1,221 @@
...
+text[6] = _([[As you enter your ship you notice dozens of cables of various colors stretched across your ship's corridors. It is a complete mess. Even weirder is the fact that it is less than one period ago that you met the student in the bar. What has he done to your ship?
+    You follow the direction where most of the cable seem to lead to. Finally you find the culprit.]])
+text[7] = _([["Oh, hello again, Captain! I'm done just by now here, so we can take off whenever you're ready. I have to calibrate the sensors during the flight so there is no need to rush. Our first destination is %s."
+    After asking what he has done to your ship he just answers, "Why, I just installed the sensors as appointed. It should have no unwanted side effects on your ship."]])
+text[8] = _([["Just a complete mess? Haven't you noticed the color coding? Well, I have my own system so it may appears confusing to you. Don't worry, I know exactly what I'm doing!"
+    His last words are supposed to be reassuring but instead you realize accepting this mission was not the best idea...]])
```
onpon4, line 47: Haven't verified this, but it seems like you're breaking up text[6], text[7], and text[8] a bit more than is necessary here. Actually it sort of looks like all three of these should be one message box.

I would also note that these paragraph breaks in all three of these seem unnecessary. This looks to me like it should be a single string  consisting of three paragraphs, or if it's big enough to cause size concerns, two paragraphs in a first one and one in a second one. Though that's an English style thing and easy to fix, so I'm not too worried about that.

🕵️

dat/missions/zalek/neburesearch/neburesearch_01.lua
```diff
@@ -0,0 +1,307 @@
...
+
+]]--
+
+include "dat/scripts/nextjump.lua"
+include "dat/scripts/numstring.lua"
+
+
+bar_desc = _("You see a scientist who is apparently looking for someone.")
+mtitle = _("Advanced Nebula Research")
+misn_reward = _("%s credits")
```
onpon4, line 17: Remove this and just use the `creditstring` functiton directly.

🕵️

dat/missions/zalek/neburesearch/neburesearch_00.lua
```diff
@@ -0,0 +1,221 @@
...
+    "I will now start with the measurements. The density of the nebula is lower in this sector so that it is less volatile. For the real measurements we have to enter %s. I will tell you once we're done here."]])
+text[10] = _([[Suddenly you lose control over your ship. Apparently most core systems were shut down. Something drains your ship's energy and there are black outs in several parts of your ship.
+    You realize that your shields are down as well. In an environment like this... That's it, your going to die here! You knew accepting this mission was a mistake from the very first moment.]])
+text[11] = _([[It seems you are still alive and all core components are back online!
+    There is a certain someone you should have a serious talk to.]])
+text[12] = _([["Sorry for causing trouble. I'm not quite familiar with the electronics of this ship type. You really should fly a Za'lek ship instead. It wouldn't have possibly happened!"
```
onpon4, line 54: Worth noting that the Za'lek ships are military designs, so I don't think the Za'lek would expect civilians to fly those (and in fact you can't without already being a Za'lek ally).

🕵️

dat/missions/zalek/neburesearch/neburesearch_00.lua
```diff
@@ -0,0 +1,221 @@
...
+    After telling him that your ship was almost destroyed he just replies "That's interesting! So the nebula is corrosive as well? I should investigate the damage it caused to the armour once we landed. But first bring us to the %s system! Don't worry the blackout will not occur again!"
+    It would probably be the best to just drop him off on the next planet...]])
+text[13] = _([["Thank you for your trust! I will try to make it quick. No, wait, I'll rather try to make it without another blackout. Haha!"
+    You start to think it was a big mistake to come to the %s system. "Anyway, just fly a circle or something. I will tell you once the scan is complete."]])
+text[14] = _([["And that's it! Now, let's head back to %s."
+    Meanwhile while looking out of your ship's window you think you saw something passing by, like a ship - but no one you've ever seen before. Nothing showed up on your radar although the object passed by very close. You could ask the student about it.]])
```
onpon4, line 60: "no one" -> "not one". ("No one" always refers to people, not objects. I assume this is a simple typo.)

🕵️

V-ktor:
I guess I really should change the ending. The shielding prototype will fail earlier so that the player can just fly back. This would also be the point where the ghost ship will appear. If the failure actually was sabotage it also explains why the Thurion ship appears here and not closer to the hidden jump to Thurion space.
I'll save the original idea for the end of the second campaign.

dat/missions/zalek/neburesearch/neburesearch_00.lua
V-ktor: I'm not sure if the student actually knows that. The point here is a lame excuse anyway.

dat/mission.xml
```diff
@@ -1041,6 +1041,29 @@
...
+   <planet>Jorla</planet>
+  </avail>
+ </mission>
+ <mission name="Advanced Nebula Research">
+  <lua>zalek/neburesearch/neburesearch_01</lua>
+  <flags>
+   <unique />
+  </flags>
+  <avail>
+   <cond>player.misnDone("Novice Nebula Research")</cond>
```
nenau-again, line 1061: There is a better way to do that with the `<done>` flag. I'll recommend using that for readability.

dat/missions/zalek/neburesearch/neburesearch_00.lua
onpon4: Ah, naïve upper-class person? Sounds legit.

🕵️

bobbens:
I looked over the concept and I think it is fine. However, I am not sold on the Mission 5 idea. I think it might be better to just have this campaign indicate that the Thurion exist, but not have the player directly interact with them yet. Maybe have a cutscene where the player is saved by "something" and appears again in Empire space or something? Or just have an Empire ship find them. That way the Thurion are hinted to exist, but not directly met with.

The currently implemented missions are fine I think, but maybe changing Arandon with Behar might be better.

V-ktor:
The nebula in Behar is actually weaker than in Arandon. I think Iris is better.
I updated the original proposal. I guess at this point it is too early to let the player make contact with the Thurion.

bobbens:
Some points. I was not against seeing the ghost ship, only that further down the line, the player shouldn't be saved by the Thurion. Hinting at their existence is no problem though. I wouldn't have a problem reverting that change.

I have updated the missions to use the new API (got rid of `dat/mission.xml`) and have merged. Look forward to see the rest of the missions!

PR #1215, opened 2020-08-18
CI Improvements

ProjectSynchro:
I have made some changes to how the SSH Key and LDoc deployments work, allowing repo forks to use CI without it failing every time. (was annoying asf for me so why not)

- [x] LDoc is now installed via apt, as opposed to luarocks. (less potential things to break)
- [x] Lua documentation is only pushed to the webserver; assuming all other jobs finish successfully.

I have also made some other improvements to the nightly release workflow:

- [x] Source bzip2 is now generated for every release.
- [x] Debugging is now disabled on Linux64 builds as the version of bfd packaged on Ubuntu 20.04 differs from many distros.


ProjectSynchro:
Technically `make distcheck` could be done in an additional step, but for now it works :sweat_smile:

bobbens:
Looks good, thanks!

PR #1222, opened 2020-08-24
Added a note on new game creation about the incompleteness of the game.

onpon4:
This is just a message shown at the start of a new game letting the
player know that the game is incomplete.

🕵️

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -447,6 +447,7 @@ int conf_loadConfig ( const char* file )
       conf_loadBool("devmode",conf.devmode);
       conf_loadBool("devautosave",conf.devautosave);
       conf_loadBool("conf_nosave",conf.nosave);
+      conf_loadBool("firstran", conf.firstran);
 
       /* Debugging. */
       conf_loadBool("fpu_except",conf.fpu_except);
@@ -1126,6 +1127,10 @@ int conf_saveConfig ( const char* file )
 
    conf_saveComment(_("Save the config everytime game exits (rewriting this bit)"));
    conf_saveInt("conf_nosave",conf.nosave);
+   conf_saveEmptyLine();
+
+   conf_saveComment(_("Indicates that the game has been run before, so it won't show first-start messages"));
+   conf_saveBool("firstran", conf.firstran);
    conf_saveEmptyLine();
 
    /* Debugging. */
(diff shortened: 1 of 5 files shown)


onpon4:
As suggested by @BTAxis, I've changed this so that it shows the message (slightly tweaked) the first time the game is run, instead.

🕵️

src/naev.c
```diff
@@ -489,6 +489,29 @@ int main( int argc, char** argv )
    /* flushes the event loop since I noticed that when the joystick is loaded it
     * creates button events that results in the player starting out acceling */
    while (SDL_PollEvent(&event));
+
+   /* Incomplete game note. */
+   if (!conf.firstran) {
```
bobbens, line 494: It might be worth making it be some sort of value like the major release version so we can display the message every major release.

src/conf.c
```diff
@@ -447,6 +447,7 @@ int conf_loadConfig ( const char* file )
       conf_loadBool("devmode",conf.devmode);
       conf_loadBool("devautosave",conf.devautosave);
       conf_loadBool("conf_nosave",conf.nosave);
+      conf_loadBool("firstran", conf.firstran);
```
bobbens, line 450: Why firstran and not firstrun? ;)

src/naev.c
```diff
@@ -489,6 +489,29 @@ int main( int argc, char** argv )
...
+      dialogue_msg(
+         _("Welcome to Naev"),
+         _("Welcome to Naev, and thank you for playing! We hope you enjoy"
```
bobbens, line 497: Maybe display the version?

bobbens (approved):

src/conf.c
onpon4: Because "firstrun" wouldn't be accurate, it would suggest that this is the first time the game is running. What it actually records is that the game *already* ran for the first time in the past.

🕵️

src/naev.c
onpon4: Sure, I can do that.

🕵️

src/naev.c
onpon4: Sure, why not?

🕵️

onpon4: Done!

🕷️

src/naev.c
onpon4: We never change the major release. 😉 So I used the short version number string instead.

🕷️

PR #1228, opened 2020-08-31
proposal on how to display extra info with mission_map.py

nenau-again:
via a `note_done` node on the xml header.
I made only one example. If this looks right, I'll extend to all the missions I know
@bobbens 
And btw, I'll try subgraphs

### DIFF
--- a/dat/missions/shark/sh02_tapping.lua
+++ b/dat/missions/shark/sh02_tapping.lua
@@ -6,7 +6,8 @@
   </flags>
   <avail>
    <priority>3</priority>
-   <cond>player.misnDone("Sharkman Is Back") and player.misnDone("Operation Cold Metal")</cond>
+   <cond>diff.isApplied("collective_dead")</cond>
+   <done>Sharkman Is Back</done>
    <chance>3</chance>
    <location>Bar</location>
    <faction>Dvaered</faction>
@@ -18,6 +19,9 @@
    <faction>Traders Guild</faction>
    <faction>Za'lek</faction>
   </avail>
+  <note_done name="Operation Cold Metal">
+   Collective must be dead
+  </note_done>
  </mission>
  --]]
 --[[
--- a/dat/missions/shark/sh02_tapping.lua
+++ b/dat/missions/shark/sh02_tapping.lua
@@ -19,9 +19,10 @@
    <faction>Traders Guild</faction>
    <faction>Za'lek</faction>
   </avail>
-  <note_done name="Operation Cold Metal">
-   Collective must be dead
-  </note_done>
+  <notes>
+   <requires name="The Collective is dead and no one will miss them"/>
+   <campaign>Nexus show their teeth</campaign>
+  </notes>
  </mission>
  --]]
 --[[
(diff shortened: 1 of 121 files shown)


utils/mission_map.py
```diff
@@ -11,8 +10,9 @@
 names = []
 dones = []
 uniques = [] #
+extra_links = [] # List of links given by extra info
 
-for missionfile in glob.glob( './dat/missions/**/*.lua', recursive=True ):
+for missionfile in glob.glob( '../dat/missions/**/*.lua', recursive=True ):
```
bobbens, line 15: I always run it from root. It should be possible to autodetect if it is run from root or utils.

nenau-again: aah. That's why. I'll take care of that

dat/missions/shark/sh02_tapping.lua
```diff
@@ -18,6 +19,9 @@
    <faction>Traders Guild</faction>
    <faction>Za'lek</faction>
   </avail>
+  <note_done name="Operation Cold Metal">
```
bobbens, line 22: If we are checking for `diff.isApplied('collective_dead')`, it might be better to do something like use meta-nodes. Have something like "provides" and "requires". You could have the "Operation Cold Metal" "provide" the `collective_dead` meta-node and have "Sharkman Is Back" "require" the `collective_dead` node. It is a bit trickier, but might be more flexible.

nenau-again: Sounds good.
I guess this simple stuff could be used when it's really simple, like when there is simply an event inbetween

bobbens:
I think it is a good idea. The thing is it can get pretty tricky due to how conditionals can be arbitrary, so I think it might be better have missions `provide` meta-nodes and missions `require` meta-nodes representing arbitrary conditions. Down the line some simple and/or type notation might be necessary.

On a side note, I was thinking it would also be useful to group missions into campaigns using some sort of `<campaign>whatever</campaign>` type notation. This could allow making more compact graphs (only show campaigns + lone missions) and would be useful down the line if we do something crazy like interactive graphs (generate a website using `d3.js`?).

Although we don't have it yet, meta-data also indicating what chapter it can appear in will also be good, but we can think of that down the like.

nenau-again:
* I've remarked that events are very important. For example, the FLF campaign ends up with an event, so I'll integrate them to the graph. For interactions between missions and events, I'll use this kind of xml tags
* I'll implement that meta-node idea. But if we want to be logically clean, I guess we need to add `and`, `or` and `not` somehow (I guess it should be at the end of the edges, and it'll need many meta-nodes for complex logic)
* I'll tackle this issue of mission-grouping. I guess subgraphs is what we need, and there is probably a way to display only one subgraph with graphviz.
* I guess I'll add a xml tag `notes` that contains all this data, that is not read by naev, but by the mission map script

nenau-again:
Allright. I'll have a low-avaliability period soon so I won't touch this for some time. Here are the problems so far:
* I have not implemented the function (that should be quite straightforward) to display only one campaing or a group of campaigns.
* One can not activate the "tiers" and "campaigns" simultaneously. The layout gets too constrained and graphviz has a high probability to crash. Apparently, such layout is actually an open problem. So I simply added an option for layout per-campaigns or per-tiers.

For info, the (pngified) graph I obtain with per-tiers layout:

missions

bobbens:
Looks good, thanks!

PR #1229, opened 2020-09-01
Added logic to allow autonav to on top of stationary or disabled ships to aid boarding

agb32:
Autonav will still follow moving targets, but if they stop or become disabled, will move on top of them.  This will then allow the player to board more easily.


nenau-again:
Seems legit.
What could be done as well is to enable the player to change `mem.radius` in order to be able to follow from far away if needed

agb32:
Good idea - not sure what the tool would be to allow that to be changed dynamically (i.e. what on-screen widget...)

nenau-again:
It could be simply on the same window as "only shoot in range" or "dogfight helper" and that kind of stuff.
And if we're on fire, we can imagine to add a dynamic stuff that appears while following someone, with cursors that allow to change the angle and radius in real time... (No idea how to do that however)

agb32:
Yes - perhaps just pressing + or - while following will allow to increase or decrease the distance.  (e.g. those being the default key binding or something)

PR #1233, opened 2020-09-02
[Feature] Soundtrack automation

ProjectSynchro:
Should build and deploy the soundtrack when a release is triggered.


ProjectSynchro:
false alarm, 
But it looks like updating the default branch isn't possible before the soundtrack is released. (probably something to do with certification etc.)

PR #1235, opened 2020-09-02
Fix ndata link in Windows installer

ProjectSynchro:
The Windows installed runs into issues with the installer 404ing while trying to download ndata.

This is because the RELEASE variable was still pointing to naev-#.#.#-beta# instead of v#.#.#-beta#


ProjectSynchro:
Windows installers will need to be regenerated for v0.8.0-beta.3

Issue #1224, opened 2020-08-27
[Discussion] Player Growth


bobbens:
One thing that is not present in Naev, but could play an important role is player growth. Currently, the player can only get things similar to NPCs and can't really be better than NPCs. The idea would be to incorporate different mechanisms to have the player slowly get stronger so that missions can ramp up in difficulty as time progresses. This would be more important if we adopt the "chapter" view of Naev where the game would be split into 4ish chapters: early-game, mid-game, late-game, end-game. The challenges and missions would basically ramp up in difficulty and thus the player would need to be stronger to compensate.

Some mechanisms I can think off the top of my head.
1. **Unique outfits/ships**. These are planned with the new overlay stuff to be able to identify them. They would be better than default versions
   * I'm not entirely sold on the idea of unique ships because it would sort of "force" the player to chose them instead of normal ships. Something would need to be done so the player can fly their ship of choice. This is not an issue with outfits.
1. **Additional special slot(s)** for each ship in the form of "accessories".
   * These would be reward-only outfits that you could equip on different ships and be an alternative to innate bonuses. They would allow the player to get stronger, but since the number you can equip is limited, you would never grow "too" strong.
1. **Innate bonuses**. Have the player get sort of passive bonuses from doing missions and campaigns. These would stack implicitly on the ship.
   * It might be hard to balance as the player could get too strong just by questing, so not a big fan.
1. **Fleets** (Discussed in #832 )
   * While not making the player stronger by themselves, it does give more firepower overall.
   * A concept of fleet power would be needed to avoid having the player be able to farm for larger fleets (or just temporarily hire to trivialize missions).

Overall I think **unique outfits**, **accessory slots**, and **fleets** does seem like the way to go down the line. **Unique outfits** is already implementable, just would need many outfits created and added as mission rewards. **Accessory slots** would need a bit of code, but it should be fairly "straight forward". Easiest implementation would just be to add a special core slot to utilities/structural/weapons and would only require XML changes. **Fleets** is a really large change though and is not trivial to doo.

agb32:
Also how about some categories of "high scores" - depending on what route the player takes.  Could be high scores for e.g. number of ships destroyed, number of jumps made, number of commodities delivered, number of missions completed, etc.  Perhaps a new tab in the info window.

bobbens:
[...]

PR #1237, opened 2020-09-03
Removal of all unique outfits and the "traveling merchent".

onpon4:
I've made my case on #1224 as to why, but this PR demonstrates exactly what I want to happen:

I want *all* of this "unique outfit" stuff removed. I don't like it and don't think it should be a thing. The functionality in the engine being there is fine, but I don't want us to use it.

Naev isn't an ARPG, and I don't think we should treat it as one. "Traveling merchants" make sense in RPGs, but in Naev you get the outfits you want by searching for them and going to the planets that sell them. I don't want to buy "unique outfits" from a "traveling merchant". I want to buy entirely ordinary outfits from all over the galaxy.

You may notice that there's no replacement here, and that's because I disagree not with the implementation, but with the basic premise. I don't want unique outfits or "player growth" of any kind beyond just amassing a bigger and more diverse collection of ordinary ships.

🕵️


bobbens:
While I could understand issues with the **accessory slot** and its ramifications (which I'm still not entirely convinced of), I don't agree with the removal of the traveling merchant and all unique outfits in general. Closing.

PR #1238, opened 2020-09-03
Next Za'lek campaign mission.

V-ktor:
Add the next mission of the Za'lek campaign. I split it into two short missions.
It is far from perfect.


bobbens:
Looks cool, thanks!

bobbens:
By the way, there are quite a few typos. If possible you should try using a spellchecker. With Vim you can enable it with `:set spell`.

PR #1243, opened 2020-09-09
fix of gui when the speed is superior to max

nenau-again:
The red bar used to be added on the player speed bar instead of the target pilot's one. (for example when targeting a pilot who is jumping out)
This should fix that.


bobbens:
Looks good, thanks!

PR #1244, opened 2020-09-10
added advertisers

nenau-again:
They browse the system and broadcast stupid slogans. Of course, much more can be added...
I made sure they are not too numerous.

### DIFF
--- a/dat/ai/advertiser.lua
+++ b/dat/ai/advertiser.lua
@@ -64,7 +64,7 @@ function create ()
    if system.cur():presence(faction.get("Empire")) > 1 then
       msg[#msg+1] = _("Do you love your Emperor as much as he loves you?")
       msg[#msg+1] = _("You're quick and dependable? The Emperor needs you in the Armada!")
-      msg[#msg+1] = _("Help the police help you: denunce the dissidents.")
+      msg[#msg+1] = _("Help the police help you: denounce the dissidents.")
    end
 
    if system.cur():presence(faction.get("Dvaered")) > 1 then
--- a/dat/ai/advertiser.lua
+++ b/dat/ai/advertiser.lua
@@ -1,0 +1,91 @@
+require("dat/ai/tpl/generic.lua")
+require("dat/ai/personality/advertiser.lua")
+
+
+-- Sends a distress signal which causes faction loss
+function sos ()
+   msg = {
+      _("Mayday! We are under attack!"),
+      _("Requesting assistance. We are under attack!"),
+      _("Civilian vessel under attack! Requesting help!"),
+      _("Help! Ship under fire!"),
+      _("Taking hostile fire! Need assistance!"),
+      _("We are under attack, require support!"),
+      _("Mayday! Ship taking damage!"),
+      string.format(_("Mayday! Civilian %s being assaulted!"), string.lower( ai.pilot():ship():class() ))
+   }
+   ai.settarget( ai.target() )
+   ai.distress( msg[ rnd.int(1,#msg) ])
+end
+
+
+mem.shield_run = 100
+mem.armour_run = 100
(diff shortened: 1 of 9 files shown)


agb32:
Looks good - change denunce to denounce maybe?

dat/ai/advertiser.lua
```diff
@@ -0,0 +1,91 @@
...
+      standing = ai.getstanding( p ) or -1
+      mem.refuel_msg = string.format(_("\"I'll supply your ship with fuel for %d credits.\""),
+            mem.refuel);
+   end
+
+   -- Selects an advertiser message
+   msg = {
+      _("Fly safe, fly Milspec."),
+      _("Reynir's Hot Dogs: enjoy the authentic taste of tradition."),
+      _("Everyone is faster that light, but only Tricon engines are faster than thought!"),
```
onpon4, line 56: that -> than

onpon4:
Is there any particular reason for duplicating civilian.lua rather than just extending that by using `require "dat/ai/civilian.lua"`?

🕵️

dat/factions/spawn/civilian.lua
```diff
@@ -4,15 +4,27 @@ require("dat/factions/spawn/common.lua")
...
-    local civships = {{"Civilian Schroedinger", 8},
-                      {"Civilian Llama", 8},
-                      {"Civilian Gawain", 8},
-                      {"Civilian Hyena", 13}
-                     }
-    
-    local select = rnd.rnd(1, #civships)
+    local r = rnd.rnd()
+
+    if r < 0.5 then
```
bobbens, line 9: This should probably be based on presence. I don't think very unsafe (pirates, etc.) systems should have too many advertisers. Should only somewhat exist in systems that are very populated.

dat/ai/advertiser.lua
```diff
@@ -0,0 +1,91 @@
...
+      _("Dare excellence! Dare Teracom rockets!"),
+      _("Because being unique is hard at times, Nexus invented the Shark fighter."),
+      _("Never take off without your courage. Never take off without your Vendetta."),
+      _("Unicorp: low price and high quality!")
+   }
+
+   -- Faction specific messages
+   if system.cur():presence(faction.get("Empire")) > 1 then
```
bobbens, line 64: Seems like a lot of messages are being duplicated. Might be better to try to organize it into a common file.

nenau-again:
>Personally I'm also not sold on 50% of civilians becoming these "advertiser" ships. That seems like far too many; it'd get old pretty quick I think.

Actually, I've been surprised how low the civilian presence is. As a consequence, 50% is what I needed to get a quantity of message comparable to the quantity of imperial ships messages.

>Might be better to try to organize it into a common file.

I don't understand. The messages are already in the same file.

bobbens:
> > Might be better to try to organize it into a common file.
> 
> I don't understand. The messages are already in the same file.

I mean that we have lots of messages duplicated in different files like `dat/ai/civilian.lua` and such. It might be better to finally condense them all into a helper script like `dat/ai/chat.lua` or something so we don't have things like `_("Mayday! We are under attack!")` repeated 4-5 times like we do now ;)

onpon4:
> As a consequence, 50% is what I needed to get a quantity of message comparable to the quantity of imperial ships messages.

That makes sense, but I don't think trying to be as frequent as Empire ships should be the goal. The idea works a lot better if it's a rare occurrence I think.

🕵️

nenau-again:
Allright. I think I've corrected everything, except for the 50% question. I'll let you test and modify if needed.
(If you test in Zied, there will be many messages because there are many civilians in few space, but on other systems, it's fairly sparse)

bobbens:
Looks good, thanks! We can tweak the values as needed.

PR #1245, opened 2020-09-13
Don't allow the ship computers to access info the player doesn't have

nloewen:
Namely, 
1. AutoNav shouldn't disable until a hostile ship's identity is no longer hidden
2. Targeting hostile ships shouldn't select ships who's identity is hidden.


bobbens:
Looks good, thanks!

PR #1247, opened 2020-09-14
All done with our hooks after the player is disabled

nloewen:


bobbens:
Changed 'disaabled' to 'disabled', but other than that looks good, thanks!

PR #1249, opened 2020-09-14
Renamed Boarding Androids outfits to comply with typical naming conventions.

onpon4:
This is a change that breaks compatibility with saves, so while Naev should definitely do it at some point (so that naming of outfits is consistent), I'm not sure whether it should be done now or later (perhaps after version 0.8).

🕵️


bobbens:
I think now is a good timing to merge. Not like they are ultra-rare nor expensive outfits. Furthermore, given that boarding is sort of "half-implemented", I don't think too many people are running boarding droids.

PR #1251, opened 2020-09-14
Armour -> Armor

onpon4:
This removes the last user-visible instance of British English (that I'm aware of). It breaks savegame compatibility, however. Whether American or British English is preferable is up to taste, of course, so if Naev is going to revert the American English back to British English then this should *not* be pulled. But if American English is to be kept, then this should be pulled at an appropriate time to minimize damage from savegame compatibility breakage.

🕵️


bobbens:
I was under the impression that Naev was in British english by default, and not American english. Were changes done to convert it to American english?

onpon4:
Yeah, we write exclusively in American English and have written a huge portion of the text for Naev in the last five years.

🕵️

agb32:
I favour British English!  Much of the code is in that...

onpon4:
Yes, I don't mind British English. Problem being I can't write in British English properly; my native dialect is U.S. English and I have very little knowledge of the differences in rules for other dialects.

Anyway, I think what's probably best is that I continue using U.S. English for stuff that I can write (and make that the dialect used in Naikari), and others will have to modify it to British English later on for Naev's dialect preference. (That's what I suspected.) I do wish there was a better way to coordinate this stuff, though. The weirdness you get when someone who's not one of the original authors, and speaks a different dialect, ends up writing and editing such a huge portion of the dialog. :wink: Could be worse: at least it's the same language.

Closing this PR both because it seems (as I suspected) that British English is strongly preferred for Naev, and it's got merge conflicts now anyhow.

🕵️

Issue #1106, opened 2020-06-24
[Proposal] Hypergates


onpon4:
Hypergates are a major part of Naev's story, but we don't have them implemented yet.

Story for hypergates is here: https://github.com/naev/naev/wiki/Spoilers#The-Hypergate

So far as I understand (@bobbens can correct me wherever I'm wrong):

* Hypergates won't appear until later on in the game, roughly the second half.
* You can warp from any hypergate to any other hypergate, like an instant teleportation network.
* Hypergates are supposed to be controlled by the major factions. You can't use them without the permission (i.e. good graces) of the respective faction they're located in, possibly for a nominal fee.
* The Proteron will be launching an invasions against all of the factions utilizing their own hypergates. (I assume this would end factions' ability to control who warps out of their hypergates? Or possibly this would be an exclusive feature of the Proteron hypergate.)

Thoughts on how hypergates could be implemented:

* They could open a menu allowing you to select where to warp to, then warp you there as normal (other than the direction you go in possibly being different than the direction you come out). This would be not very immersive, but simple.
* A special "hub" system could be implemented, which has no location and contains a collection of hypergate entry points, not unlike QuasiSpace in The Ur-Quan Masters. More immersive, but a bit more complex.

ids1024:
> (I assume this would end factions' ability to control who warps out of their hypergates? Or possibly this would be an exclusive feature of the Proteron hypergate.)

This seems like it would provide about the dynamic we want:

* A gate is a powered device that must be activated to jump out, which is under control of the owning faction.
* It is not, however, possible to control who can jump in. Once constructed, the gate cannot easily be disabled without damage to space-time and a new incident.
* The Proteron are thus stranded on this side of the nebula... unless they are able to seize control of gates and reprogram them. The same applies to other factions who attempt to enter Proteron space.

This last point seems like it could provide for quite interesting narrative and missions. As long as we make sure the player can't strand themself in Proteron space. This is easy enough; either the faction controlling the gate won't let them, or they find their ship destroyed if they attempt to jump to Protera.

> A special "hub" system could be implemented, which has no location and contains a collection of hypergate entry points, not unlike QuasiSpace in The Ur-Quan Masters. More immersive, but a bit more complex.

In addition to the implementation complexity, this would probably prevent factions from controlling where someone can jump to from their gates. Which could be an interesting dynamic.

onpon4:
Yes, that's true!

[...]

PR #1252, opened 2020-09-15
Hypergates

BrknRobot:
Rebased my old hypergate branch. I've cleaned it up a little to the point that it loads again, but it will need a good bit of refactoring and testing before it's ready. I'll also need to catch up on any changes to the proposal for these (#1106).

Lesson learned, don't wait 8 years before finishing changes :grimacing:


bobbens:
Thanks for the PR, however, given the incredible amount of change in the code in the 14 months since this PR was last updated and the fact we probably want a more general approach for system objects ( #1862 ) that would also integrate hypergates, I think it's best to close this PR for now. Thanks again!

PR #1254, opened 2020-09-16
Jump point convenience fields

nloewen:
Add some convenience fields to `JumpPoint`.

`JumpPoint.returnJump`
Adding a pointer to the return jump allows us to avoid looping over jump points in a number of places. The looping still needs to happen, but we only need to do in once after our memory allocations are finished.

`JumpPoint.from`
The reference to the containing system is not yet used, but will be in an upcoming PR.The idea is that you'll be able to perform an operation such as an adjacency search, and get access to the relevant `JumpPoint` and `StarSystem` in one struct. This will be a bigger deal once hypergates are added, which will introduce adjacent systems that aren't directly reachable through `system->jumps[i].target`.

### DIFF
--- a/src/space.h
+++ b/src/space.h
@@ -172,9 +172,12 @@ typedef struct SystemPresence_ {
 /**
  * @brief Represents a jump lane.
  */
-typedef struct JumpPoint_ {
-   StarSystem *target; /**< Target star system to jump to. */
+typedef struct JumpPoint_ JumpPoint;
+struct JumpPoint_ {
+   StarSystem *from; /**< System containing this jump point. */
    int targetid; /**< ID of the target star system. */
+   StarSystem *target; /**< Target star system to jump to. */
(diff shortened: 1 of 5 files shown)


src/space.h
```diff
@@ -184,7 +187,7 @@ typedef struct JumpPoint_ {
    double sina; /**< Sinus of the angle. */
    int sx; /**< X sprite to use. */
    int sy; /**< Y sprite to use. */
-} JumpPoint;
+};
```
onpon4, line 190: Why did you remove the usual `typedef struct Foo_ { ... } Foo;` idiom? It's the usual way to do these.

🕵️

nloewen: It needs to be forward declared in order to use it in the struct definition

bobbens:
Looks good, thanks!

PR #1256, opened 2020-09-16
Space pilots

nloewen:
Restructure some includes so that pilot.* can reference space.h structures.
This is important for upcoming hypergate changes. Hypergates don't have a single target, so functions such as pilot_choosePoint must be able to return a reference to the target JumpPoint (hypergate) rather than an index in order to work with them.

Depends on #1254 and #1255. I'll rebase on master once they are merged.

### DIFF
--- a/src/fleet.h
+++ b/src/fleet.h
@@ -6,6 +6,8 @@
 #ifndef FLEET_H
 #  define FLEET_H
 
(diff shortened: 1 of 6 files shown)


bobbens:
Looks good, thanks! Next time you add a new file, remember to add it to `src/Makefile.am`

PR #1257, opened 2020-09-17
Add VSCode configuration

nloewen:
Includes
 - Build tasks
 - Launch with GDB
 - Auto complete with the C/C++ extension

Unfortunately, the only way I can get VSCode to understand the includes set up by `configure` is by running `make` through `bear`. `bear` captures the commands sent to `gcc`/`clang`/etc and creates a `compile_commands.json` database that can be understood by VSCode and other Clang tooling. Hopefully anyone wanting to develop in VSCode is okay with installing that.

References
- https://clang.llvm.org/docs/JSONCompilationDatabase.html
- https://github.com/rizsotto/Bear

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ utils/starmap/*.svg
 utils/starmap/*.png
 utils/starmap/*.jpg
 utils/starmap/*.log
+/compile_commands.json
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,14 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -91,3 +91,4 @@ utils/starmap/*.svg
 utils/starmap/*.png
 utils/starmap/*.jpg
 utils/starmap/*.log
+/compile_commands.json
```
bobbens, line 94: shouldn't this be a relative path?

bobbens:
Looks good, thanks!

.gitignore
nloewen: Maybe? That's the only spot it should be showing up though

PR #1259, opened 2020-09-23
zigzag AI

nenau-again:
With this commit, the AI is aware of the fact that there are ammos aimed at its ship. This is not supra-clean as theoretically it should be possible to shoot at someone while being out of sensor range but I supposed it's unlikely enough.

When being shot at, light ships use zigzag trajectories in two occasions:
* While running away
* While approaching for dogfight (and still being out of range for their own weapons)

This suppresses the main AI exploit I was using, namely running away from target, and then turning around and shooting with long-range weapons while the enemy was kindly flying straight towards me (and towards the ammos on the way)

/!\ Warning: In some occasions, the AI ships possibly become noticeably harder to kill (or better said less trivial to kill) with this change. I'm not sure if it doesn't unbalance some missions /!\

### DIFF
--- a/dat/ai/include/atk_drone.lua
+++ b/dat/ai/include/atk_drone.lua
@@ -55,6 +55,11 @@ function _atk_drone_ranged( target, dist )
    -- Check if in range
    if dist < ai.getweaprange( 4 ) and dir < 30 then
       ai.weapset( 4 )
+   else
+      -- First test if we should zz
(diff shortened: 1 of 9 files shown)


bobbens:
Looks good, thanks! Seeing it only affects drone/fighters, I don't think it'll have *too much* impact on missions.

PR #1260, opened 2020-09-24
Meson build system

### LINKED ISSUE #1253: Make Install does not install dat/ correctly.
dat/ is not correctly installed when a `make install` is run.

I can't seem to figure out recursively installing all of dat/ when you run make install
I can't seem to use globs to achieve it
I can't use the output of `find` to achieve it (argument limit is hit, and yes I increased the max)
I tried a recursive wildcard `rwildcard=$(foreach d,$(wildcard $(1:=/*)),$(call rwildcard,$d,$2) $(filter $(subst *,%,$2),$d))` this also doesn't appear to work either
so I'm at a loss until I get some more sleep and continue to read up on stuff

This is directly related to #1234, and occurs after commit https://github.com/naev/naev/commit/5f569c75c8c3891eb71c4e3abcd12c188cab068e

nloewen:
Fixes  #1253

This PR does not break AutoTools. They can be used side by side until all required functionality exists in Meson.

Outstanding work:
- [ ] Nightly/prerelease/release builds
- [ ] Building installers/distributables


ids1024:
This seems to require a newer version of `meson` than Ubuntu 20.04 has:

```
meson.build:1:0: ERROR: Meson version is 0.53.2 but project requires >=0.55.0
```

Seems to build fine Manjaro ARM. I think we probably want to at least support building on the latest Ubuntu (and similar distros) though.

bobbens:
Quick things:
1. Doesn't seem like openal is detected on my system
2. Is gettext being handled? (there is some custom stuff in `Makefile.am`)
3. How is Lua being handled? Does it use luajit as default unless overwritten?
4. Is soundtrack also buildable?
5. Is it possible to build naev from root? Like `meson compile -c builddir` or something like that?

bobbens:
> This seems to require a newer version of `meson` than Ubuntu 20.04 has:
> 
> ```
> meson.build:1:0: ERROR: Meson version is 0.53.2 but project requires >=0.55.0
> ```

That is definitely an issue. Ideally we want support for at least Ubuntu 18.04. If possible 16.04.

ids1024:
> That is definitely an issue. Ideally we want support for at least Ubuntu 18.04. If possible 16.04.

Yeah, ideally we probably want to target the oldest supported Ubuntu LTS (I suppose we should be testing that on CI).

And I believe it's always necessary to have meson to build; unlike autotools where release tarballs have a generated `configure` file that in principle will run on any Unix from the past few decades.

meson.build
```diff
@@ -0,0 +1,471 @@
...
+
+   install_data(
+      'naev-confupdate.sh',
+      'extras/logos/naev.png',
+      install_dir: get_option('datadir') / 'naev'
+   )
+
+   install_subdir('dat', install_dir: get_option('datadir') / 'naev')
+
+   if host_machine.system() == 'linux'
```
ids1024, line 463: This should probably apply to Unix systems other than Linux.

meson.build
```diff
@@ -0,0 +1,471 @@
...
+      'c_std=gnu11',
+      'werror=false'
+   ],
+   meson_version: '>=0.55.0')
+
+
+####
+# Source lists
+####
+source = files(
```
ids1024, line 15: Would it be better if some of this is in a `src/build.meson` file? That's how the autotools configuration is currently set up. Looks like the same convention is at least sometimes used with Meson: https://gitlab.gnome.org/GNOME/mutter/-/blob/master/src/meson.build

nloewen: Maybe? It's largely a question of how we want to organize it I think. Meson organizes the build results according to the structure of the meson.build files. At present, the naev binary is placed into the root of the source tree, so I put all the binary stuff at the root so that Meson would put the naev binary into the root of the build directory. I could move some of the declarations into src/meson.build without affecting that for cleanliness.

docs/ldoc.sh
```diff
@@ -6,61 +6,12 @@ test -d lua || mkdir lua
...
-         d
-         ' $F | awk '
-	 # Skips all comment blocks without an @ tag
-         BEGIN {
-            RS="\n\n"
-         }
-         /@/ {
-            print $0, "\n"
-         }' > lua/"$(basename $F)".luadoc
+   ./c2luadoc.sh $F lua/"$(basename $F)".luadoc
```
ids1024, line 9: Hm, this seems to not work, since `c2luadoc.sh` isn't in `docs/`, it's in `docs/lua/`

nloewen:
>     1. Doesn't seem like openal is detected on my system
OpenAL and all other optional dependencies should get picked up now.
 
>     2. Is gettext being handled? (there is some custom stuff in `Makefile.am`)
The basics of gettext should be handled now, but there's some stuff I'm not sure of. Like what are we doing with the intltool and it's files?
 
>     3. How is Lua being handled? Does it use luajit as default unless overwritten?
I've added a new option `-Duse_luajit` (default: true). You can add this to your `meson setup` command, or `meson configure`.
If luajit isn't found, it will download and compile it from source. If this fails for some reason, it will fall back to regular lua. If that's not found, it will download and compile it from source.

>     4. Is soundtrack also buildable?
It is now! `meson compile soundtrack`

>     5. Is it possible to build naev from root? Like `meson compile -c builddir` or something like that?
That's exactly the right syntax. Meson is set up for out of tree builds, and you can point it at any build directory with `-c`, allowing you to run meson from any location.

nloewen:
> > This seems to require a newer version of `meson` than Ubuntu 20.04 has:
> > ```
> > meson.build:1:0: ERROR: Meson version is 0.53.2 but project requires >=0.55.0
> > ```
> 
> That is definitely an issue. Ideally we want support for at least Ubuntu 18.04. If possible 16.04.

Not sure what to do with this yet. 0.55 has some nice new features that I'm using. I'll look at supporting older versions, but the latest version can always be downloaded with pip, or we can use a helper to download the latest version to our source directory.

docs/ldoc.sh
nloewen: Good catch

meson_options.txt
```diff
@@ -0,0 +1,9 @@
+option('paranoid'    , type: 'boolean', value: false    , description: 'Add file and line number to WARNING messages.')
```
bobbens, line 1: Default should be file and line number. Paranoid with autotools sets -Werror which means the game will error out at the first warning.

src/meson.build
```diff
@@ -0,0 +1,341 @@
...
+)
+
+mac_source = files('glue_macos.m')
+
+naev_source = [
+   source,
+   lua_source
+]
+
+# Headers aren't needed for running the builds,
```
bobbens, line 171: However, meson does trigger rebuilds on header changes right?

src/naev.c
```diff
@@ -411,6 +419,8 @@ int main( int argc, char** argv )
    /* Start menu. */
    menu_main();
 
+   LOG( _( "Reached main menu" ) );
```
bobbens, line 422: Should be DEBUG not LOG.

src/nopenal.h
```diff
@@ -10,8 +10,8 @@
 
 #include "ncompat.h"
 
-#include "AL/alc.h"
```
bobbens, line 13: Looking at openal soft examples, it does seem like `"AL/*.h"` is the proper way to include them (https://github.com/kcat/openal-soft/blob/master/examples/alhrtf.c)

bobbens (commented):
Looks good, just some minor things that need clearing or can be improved.

src/meson.build
nloewen: Yes, but internally it builds that list by parsing #include statements.

src/nopenal.h
nloewen: Apparently, the reason I changed this is because the mac build fails when AL/ is added.

src/naev.c
bobbens: Maybe add a flag like "--verbose"? or "--logmenu"?

src/nopenal.h
bobbens: With autotools it works with "AL/" though, is this some difference in implementation?

nloewen: It turned out this was because Meson was picking up OpenAL.framework. It should stop doing that and use openal-soft now.

ProjectSynchro:
imo this seems ready to merge
I have a couple changes on my fork which I'll make an additional PR for, but for the time being I've been too busy to port a lot of the release scripts.
Since autotools still would work, I don't see any issues with waiting a bit to move things over.

bobbens:
Looks good, merging! Still needs some usability issues like being able to run from the build directory (doesn't find ndata), but it does seem to work. Version is also 0.8.0 not 0.8.0-beta.3.

PR #1262, opened 2020-09-27
Rework po generation a bit. This removes the dependency on itstool.

nloewen:


bobbens:
Looks awesome, thanks!

PR #1263, opened 2020-09-30
ndata is dead! Long live ndata!

### LINKED ISSUE #1234: [Proposal] Remove ndata
We've talked about this a bit on Discord before, but I just wanted to make an issue since it's something I very much believe we should do. I can't think of any particular advantage ndata gives us rather than avoiding the redundancy of copying data into all different builds. That's worth considering but the use of ndata also has a number of downsides:

* End-user difficulty: adding separate downloads the user has to pair together adds complexity for the user downloading the game. It's a lot easier to download one thing than to download two things and put them together following instructions.
* Problems with gettext: gettext needs to be in a directory to bind to. This has been discussed before and while solving the issue would be very hard with ndata, it would be a non-issue without ndata.
* Discouraging modding / contributing: since the ndata is effectively a monolith, it sort of presents the game as "closed off". I know from personal experience that this can have an effect because it discouraged us from even looking at the data until we were sure enough we wanted to contribute that we checked out the repo. This is likely to be a small effect but still worth mentioning.
* Sucking away development time: it's not uncommon for us to have to help troubleshoot problems with ndata because of the first point. We'd have a much easier time giving support when the game doesn't work if we don't have to first check to make sure that the ndata was placed in the proper location.

As for the one advantage ndata has, there's a simple way to solve that: just pack all binaries together in one zip along with the data, like Xonotic does. That said, considering we automate builds now I don't see how even that is necessary; the extra downloads shouldn't cause a problem if a developer doesn't have to manually upload them. So I think either just having separate builds all with the dat directory included or packing all binaries in one zip would be fine.

🕷️

nloewen:
Remove all vestiges of bundled ndata from the C code.

Resolves #1234

Todo:
- [ ] Update packaging scripts


src/ndata.c
```diff
@@ -118,28 +95,61 @@ int ndata_check( const char* path )
...
+      ndata_source = NDATA_SRC_SEARCH_START;
+
+      switch ( ndata_source ) {
+      case NDATA_SRC_CWD:
+         ndata_filename = ndata_findInDir( "." );
+         if ( ndata_filename != NULL ) {
+            ndata_source = NDATA_SRC_CWD;
+            break;
+         }
+         __attribute__( ( fallthrough ) );
```
bobbens, line 125: Apparently these fallthrough are breaking the mac os x compiles. Might be worth doing something like https://stackoverflow.com/questions/45349079/how-to-use-attribute-fallthrough-correctly-in-gcc instead. See also https://clang.llvm.org/docs/AttributeReference.html#fallthrough

src/naev.c
```diff
@@ -422,11 +422,13 @@ int main( int argc, char** argv )
    /* Tell the player to migrate their configuration files */
    /* TODO get rid of this cruft ASAP. */
```
bobbens, line 423: Probably safe to get rid of this no? It was introduced for a breaking change, but it's been like this forever, so we might as well remove it.

src/ndata.c
```diff
@@ -57,57 +56,35 @@
...
-#define NDATA_SRC_LAIDOUT        0
-#define NDATA_SRC_DIRNAME        1
-#define NDATA_SRC_NDATADEF       2
-#define NDATA_SRC_BINARY         3
+#define NDATA_SRC_CWD     0 /**< Current working directory. (debug builds only) */
+#define NDATA_SRC_USER    1 /**< User defined directory. */
+#define NDATA_SRC_DEFAULT 2 /**< Default derectory. (Set at compile time) */
+#define NDATA_SRC_BINARY  3 /**< Next to the Naev binary */
+
+#if DEBUGING
```
bobbens, line 64: Should be DEBUGGING

src/ndata.c
```diff
@@ -57,57 +56,35 @@
...
 /*
  * Prototypes.
  */
 static void ndata_testVersion (void);
 static char *ndata_findInDir( const char *path );
-static int ndata_openFile (void);
-static int ndata_isndata( const char *path, ... );
+static int ndata_isndata( const char *path );
 static int ndata_prompt( void *data );
 static int ndata_notfound (void);
```
bobbens, line 87: ndata_notfound is defined but unused.

bobbens (changes requested):

bobbens:
Naev fails to run with this when using `./naev -N` from the compiled directory. However, `./naev .` does work. Haven't tried the github actions compiled binaries but that would need checking too. Furthermore, as mentioned in the code comments, Mac OS X build seems to have been broken due to the fallthroughs. Once these issues are solved, I do think it will clean up stuff a lot.

src/naev.c
nloewen: Good riddance.

src/ndata.c
nloewen: Looks like it doesn't like the fallthrough attribute on case with only a comment. The rest are just warns. I'll make sure all warns are cleaned up in the Meson builds

nloewen:
> Naev fails to run with this when using `./naev -N` from the compiled directory. However, `./naev .` does work. Haven't tried the github actions compiled binaries but that would need checking too. Furthermore, as mentioned in the code comments, Mac OS X build seems to have been broken due to the fallthroughs. Once these issues are solved, I do think it will clean up stuff a lot.

Should be fixed. Mac build is failing due to homebrew issues now. Compile issues should be fixed though

bobbens:
Looks good, thanks!

PR #1265, opened 2020-10-01
Add .clang-format

nloewen:
I tried to match they style guidelines in the wiki.

There's a known issue with `PointerAlignment: Right` though. It's doing this, which isn't great.

    int *       a;
    myLongType *b;

This works great with the following command, which formats all your staged lines.
`git clang-format`


bobbens:
Thanks!

PR #1266, opened 2020-10-01
Update German translation

Wuzzy2:
Another German translation update. I have reached 100% in the current PO version, but I believe the translation file is outdated again, so it's likely a bit lower than that.

Also, still needs some proofreading and testing, so expect to see more PRs in future.


bobbens:
Thanks! Next version we should hope to have the data stuff sorted out so players on steam and the likes will be able to use your german translation ;)

PR #1270, opened 2020-10-03
More ndata cleanup

nloewen:
- Adds AppImage path handling
- Changes references to the ndata "file", "zip", or "archive" to "directory"
- Removes libzip from docs, dependency searches, and CI.


bobbens:
Looks good, thanks!

PR #1275, opened 2020-10-04
PO changes required for Meson.

nloewen:
I've tested and verified that Meson and AutoTools produce identical output with this patch.
In addition, I've confirmed that no substantial number of translations are being commented or fuzzied.
Updating the PO files after patch produce a small number of meaningful changes, but I believe they are all related to code and data changes that happened since the last PO update.

- update-po.sh generates a combined POTFILES.in for Meson to consume, since Meson doesn't seem to have any performance issues that justify the extra complexity of two generation steps.
- update-po.sh always puts XML files at the end of POTFILES*. This removes sorting as a point of variance between AutoTools and Meson.
- update-po.sh takes the source root directory as a parameter (default cwd). This supports Meson's out of tree builds.
- Moves `po/translation.its` into `po/its/`. This allows the use of an XML locator file required for Meson's combined code/XML gettext processing.
- Adjusts AutoTools to produce file names relative to the source root in .po files. Aside from being cleaner, this matches Meson's behaviour making it easier to compare the output of the two systems.


bobbens:
Looks good, thanks!

PR #1277, opened 2020-10-06
Update German translation

Wuzzy2:
A huge German translation update, now including translations for ships and planet descriptions. 100%

I don't know if the string collection script is perfect yet, so strings that the collector script fails to collect are not included.

Sadly, I cannot currently test the translation in-game at the moment. A couple of commits ago, the German translation worked fine in Naev after compiling, but this has stopped working, so I always get an English Naev.

Expect to see future updates for corrections of the translation (when I have been able to apply heavy testing on the translation).


bobbens:
Thanks! By the way, you should be able to run translations more easily now.

PR #1279, opened 2020-10-09
Remove fontconfig from dependencies

nloewen:


bobbens:
Looks good, thanks! However, Hopefully one day fontconfig will be back from the grave ;)

Issue #1291, opened 2020-10-09
first_steps_into_space.png not found


Houkime:
Arch linux, freshly cloned master .
Crashes shortly in the intro when starting a new game.

```
 Naev v0.8.0-beta.3 (x86_64-pc-linux-gnu)
 git HEAD at fcb73daf7
AppImage not detected.
BFD backtrace catching enabled.
Found ndata: ./dat
Reset language to ""
 Sea of Darkness

SDL: 2.0.12 [compiled: 2.0.12]

OpenGL Window Created: 1024x768@24bpp window
r: 8, g: 8, b: 8, a: 0, db: yes, fsaa: 0, tex: 16384
vsync: no, mm: yes, compress: no, npot: no
Renderer: Radeon RX550/550 Series (POLARIS12, DRM 3.38.0, 5.8.13-arch1-1, LLVM 10.0.1)
Version: 4.6 (Compatibility Profile) Mesa 20.2.0

Using default font 'fonts/Cabin-SemiBold.otf'
OpenAL started: 44100 Hz
Renderer: OpenAL Soft
Version: 1.1 ALSOFT 1.20.1 with EFX 1.0

Loaded 42 Sounds
Loaded 54 Songs
No nebula found, generating (this may take a while).
Generating Nebula of size 1024x768x16
Nebula Generated in 259 ms
Loaded 16 Nebula Layers
Loaded 50 Commodities
Loaded 38 Factions
Loaded 40 AI Profiles
Loaded 127 Missions
Loaded 26 Events
Loaded 443 Outfits
Loaded 79 Ships
Loaded 150 Fleets
Loaded 62 tech groups
Loaded 380 Star Systems
       with 1181 Planets
Loaded 12 UniDiffs
Loaded 8 map decorators.
Warning: [ndata_rwops] Unable to open file 'dat/gfx/intro/first_steps_into_space.png': not found.
Warning: [gl_loadNewImage] Failed to load surface 'dat/gfx/intro/first_steps_into_space.png' from ndata.
Naev received SIGSEGV (address not mapped to object)!
./naev(+0xcdf0e) [0x55b380ad4f0e] ??(...):0 ??
/usr/lib/libpthread.so.0(+0x140f0) [0x7f6d0f95c0f0] ??(...):0 ??
./naev(intro_display+0x95a) [0x55b380ab8aaa] ??(...):0 ??
./naev(player_new+0x2f1) [0x55b380b20f21] ??(...):0 ??
[...]

PR #1292, opened 2020-10-09
remove dat/ prefix from inside dat/intro

Houkime:
Fixes https://github.com/naev/naev/issues/1291

Without this change, intro crashes on Linux because intro pictures are not found (see the issue).


onpon4:
Cool! Thanks for the fix.

🕵️

PR #1295, opened 2020-10-14
Add .clang-tidy and fix warnings

nloewen:
`(void)x;` indicates we know the last value set to x is unused and it's intentional.


bobbens:
Looks good, thanks!

PR #1296, opened 2020-10-17
fix shipstats modification procedure for integers

Houkime:
Before that, i was erroneously pointing into a forbidden region of
memory and writing to it caused a segmetation fault.
This went undetected because of a relative lack of integer stats
tickling this branch of logic.

I discovered the problem when playing with Hidden Jump Detector.
(it segfaulted on activation, and now it works)

Also, the problem is obvious if one compares to the same code for
doubles.


onpon4:
Thanks for the fix!

🕵️

PR #1299, opened 2020-10-17
Prefix array sentinel to avoid future conflicts

nloewen:


src/array.h
```diff
@@ -46,7 +46,7 @@
 #endif /* HAVE_STDALIGN_H */
 
 #ifdef DEBUGGING
-#define SENTINEL ((int)0xbabecafe) /**< Badass sentinel. */
+#define _ARRAY_SENTINEL ((int)0xbabecafe) /**< Badass sentinel. */
```
onpon4, line 49: That leading underscore shouldn't be there:

http://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html
http://web.archive.org/web/20040209031039/http://oakroadsystems.com/tech/c-predef.htm#Table

> In addition to the names documented in this manual, reserved names include all external identifiers (global functions and variables) that begin with an underscore (‘_’) and all identifiers regardless of use that begin with either two underscores or an underscore followed by a capital letter are reserved names. This is so that the library and header files can define functions, variables, and macros for internal purposes without risk of conflict with names in user programs.

onpon4 (changes requested):

PR #1303, opened 2020-10-17
Switch scroll buffer from looped linked list to array with looped ind…

nloewen:
…exing. The primary motivation for this change is to simplify code analysis.


bobbens:
Looks good, thanks!

PR #1305, opened 2020-10-19
Add the next Za'lek campaign mission

V-ktor:
Add the next Za'lek campaign mission. My story makes no sense, right? Also my writing style is somewhat poor.
Anyway, here is the next mission. There will be just one more in the first campaign.


bobbens:
Pulled with @ProjectSynchro 's English checking. Thanks!

PR #1308, opened 2020-10-23
CI: Fix CI Failures, disable macOS temporarily and fix documentation uploads

ProjectSynchro:
I tried, but I'm nearing the end of my rope with macOS. 
Right now the main failures in terms of CI are down to issues with homebrew mirrors, or build issues due to something failing to download.

This PR simply removes macos from the CI build matrix and release workflows (for now) until a more solid solution can be figured out. 
There's no need to block a release from happening if homebrew is unreachable (which has been happening more frequently than when this was initially added) 

I also fixed the upload to api.naev.org due to some paths not quite matching up with the upload step.

Back to macos.. we don't really have the testing capacity or hardware available to give players decent experience using the releases that are generated right now, and I don't think we should be pushing out something half-baked if it can be avoided.

For now I say we stop building for macOS at least until a better solution presents itself.
I tried to get macports working to no avail (some tinkering with the build system might be needed to get testing working.)

I'm (still) working on a more robust release solution that simplifies things quite a bit, opts to use appimages on linux and should allow us to release on itch.io but I've been getting slammed with irl things and work things so.. time and energy are not readily available from me.
Hopefully with that we might be able to get macos stuff going again, but I'm still not 100% on serving something out without proper testing.

### DIFF
--- a/.github/workflows/naev_ci.yml
+++ b/.github/workflows/naev_ci.yml
@@ -54,8 +54,6 @@ jobs:
             shell: bash
             config: macos.ini
             pkg-config-path: '/usr/local/opt/openal-soft/lib/pkgconfig'
-            env:
-              PATH: "/opt/local/bin:/opt/local/sbin:$PATH"
 
     defaults:
       run:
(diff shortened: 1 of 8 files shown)


ProjectSynchro:
Fixed Windows CD as there was a change in the package name..

bobbens:
Looks good, thanks!

PR #1311, opened 2020-10-25
Free speed option text returned by dialogue_runChoice().

UncombedCoconut:
I spotted some minor resource leaks. After talking it over on Discord, I'm raising separate PRs so the fixes which are obviously safe/correct (from Naev devs' perspective) can be merged first.
These changes are all battle-tested (in space!) -- but on this project that's a low bar. :)


bobbens:
Looks good, thanks! However, it would be easier if you grouped up all these PR into a single one next time ;)

PR #1312, opened 2020-10-25
Free opt_video() result strings when we skip them as duplicates.

UncombedCoconut:


bobbens:
Looks good, thanks! However, it would be easier if you grouped up all these PR into a single one next time ;)

PR #1316, opened 2020-10-26
CI/CD: Release and CI Improvements // Windows cross-compile build improvements

ProjectSynchro:
Completed work:
- [x] Windows Installer script Improvements
- [x] DLL collection improvements
- [x] Steam Release Improvements
- [x] Linux AppImage releases


extras/windows/packageWindows.sh
```diff
@@ -1,158 +1,113 @@
...
-    echo "copying $fp to staging area"
-    cp $fp extras/windows/installer/bin
-done
-else
-    echo "Aw, man, I shot Marvin in the face..."
-    echo "Something went wrong while looking for DLLs to stage."
-    exit -1
-fi
+echo "Populating staging area with Meson subprojects' DLLs (if any)"
+find "${SOURCEROOT}/subprojects" -iname "*.dll" -exec cp -v "{}" "$STAGING" ";"
```
nloewen, line 77: I think some of the subprojects might build more dlls than we use, not sure if there's a smarter way to do this though.
Maybe `ninja -t query naev.exe` would be helpful?

UncombedCoconut: Hey, that code's my fault! :)
Yes, we can instead pass these roots to the extract_dlls script, and have it add them to its search.
I'd expect a small slowdown, nothing noticeable given how much work the script's already doing. A bit more complexity too.

The Ninja approach is quite cool. In general, if Meson winds up as The build system it will open up technically superior ways of doing the work.

ProjectSynchro:
Itch.io stuff is a bit out of scope for now..

ProjectSynchro:
After some testing, release workflows are ready,
I also optimized CI, and added additional checks to ensure that autotools is working correctly and added support for building appimages with either meson or autotools (CI checks included)

Linux releases served via github are now appimages (thanks for working on the support @nloewen!) This should allow for users to run Naev without using distro-packaged libraries, in an all in one 'app in a box'.

The windows packaging script (well almost all of the shell scripts) have been improved by @UncombedCoconut and myself.. apologies for the crappy syntax..

The Steam deployment script has been cleaned up, the 2fa script has been cleaned up (this was backported to master)

I also have gotten `make dist` to the state of allowing builds without issues.. `make distcheck` hasn't worked in a while and is down to something with gettext.

Next up is getting the work for butler (itch.io) in place. Getting AppImages building/working was a prerequisite to that.

PR #1317, opened 2020-10-28
Memory fix

UncombedCoconut:
This is a somewhat bolder PR which tries to plug memory leaks. (There are more commits, and this PR is less shy about changing interfaces in a way that hopefully isn't trouble.) With these changes in, it almost isn't insane to run Naev under Valgrind and expect to see very little "definitely lost" memory. Almost.

### DIFF
--- a/src/unidiff.c
+++ b/src/unidiff.c
@@ -945,7 +945,7 @@ void diff_remove( const char *name )
 
 
 /**
- * @brief Removes all active diffs.
+ * @brief Removes all active diffs. (Call before economy_destroy().)
  */
 void diff_clear (void)
 {
@@ -956,6 +956,21 @@ void diff_clear (void)
       diff_removeDiff(&diff_stack[array_size(diff_stack)-1]);
 
    economy_execQueued();
+}
+
+
+/**
+ * @brief Clean up after diff_loadAvailable().
+ */
+void diff_free (void)
+{
+   diff_clear();
+   for (int i = 0; i < array_size(diff_available); i++) {
+      free(diff_available[i].name);
+      free(diff_available[i].filename);
+   }
+   array_free(diff_available);
+   diff_available = NULL;
 }
 
 
--- a/src/unidiff.c
+++ b/src/unidiff.c
@@ -202,7 +202,7 @@ int diff_loadAvailable (void)
       }
 
       diff = &array_grow(&diff_available);
-      diff->filename = strdup( diff_files[i] );
+      diff->filename = diff_files[i];
       xmlr_attr(node, "name", diff->name);
       xmlFreeDoc(doc);
       free(filebuf);
--- a/src/conf.c
+++ b/src/conf.c
@@ -721,8 +721,11 @@ void conf_parseCLI( int argc, char** argv )
    }
 
    /** @todo handle multiple ndata. */
-   if (optind < argc)
+   if (optind < argc) {
+      if (conf.ndata != NULL)
+         free(conf.ndata);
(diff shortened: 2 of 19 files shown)


ids1024:
I wonder if instead of `if (ptr != NULL) free(ptr);`, it would be better to just use `free(ptr);`. The C standard guarantees that it's valid to free null pointers, and it's the preferred style in at least some style guides. But I guess the more verbose version may be the style Naev's code is already using.

Reading through the diff this all seems correct, though issues of this sort can be subtle.

src/unidiff.c
```diff
@@ -202,7 +202,7 @@ int diff_loadAvailable (void)
       }
 
       diff = &array_grow(&diff_available);
-      diff->filename = strdup( diff_files[i] );
+      diff->filename = diff_files[i];
```
ids1024, line 205: Maybe just to be cautious, this line should be followed with `diff_files[i] = NULL;` so it isn't possible to try to use the pointer after it's been "moved".

Not that I see a significant danger of that here, but it may be a good practice.

UncombedCoconut: > I wonder if instead of `if (ptr != NULL) free(ptr);`, it would be better to just use `free(ptr);`. The C standard guarantees that it's valid to free null pointers, and it's the preferred style in at least some style guides. But I guess the more verbose version may be the style Naev's code is already using.

This is validating to see; I considered raising that point on a previous round. (While compatibility is very important, if Naev can use variadic macros, I think it can rely on this part of ISO/IEC 9899 too.)
For the moment, though, avoiding free(NULL) is a pervasive pattern in this code base.

> Maybe just to be cautious, this line should be followed with diff_files[i] = NULL; so it isn't possible to try to use the pointer after it's been "moved".

I like that pattern. (In this file, we free diff_files a few lines below, so I'm kind of indifferent. Would be happy to edit that in tomorrow though.)

onpon4:
Haha, since I learned C by contributing to Naev I didn't even know that `free(NULL);` was valid.

🕵️

UncombedCoconut:
That speaks very well of Naev and its community! :)
In the interest of clarity (and going down a fun rabbit-hole):
The standards explicitly made free(NULL) a no-op starting with C89. [1]
This didn't mean everyone suddenly had a conformant compiler and libc,
though. Various sources allege it was broken on 3BSD [2], NextStep [3], and
PalmOS [4]. Anecdotally, I haven't heard of it being an issue in the past
20 years. Mainstream advice (reflected in those links) is to simplify one's
code by relying on this standard. It's possible a Naev dependency like SDL2
already does, but I don't know. In any case, I'm not here to press the
point, and I wouldn't want to change the style unless bobbens was convinced.
[1] http://port70.net/~nsz/c/c89/c89-draft.html#4.10.3.2
[2] https://www.winehq.org/pipermail/wine-patches/2006-October/031544.html
[3]
https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Function-Portability.html
[4]
https://stackoverflow.com/questions/1938735/does-freeptr-where-ptr-is-null-corrupt-memory/1938758#1938758

ids1024:
The C11 feature I snuck in recently apparently was an issue for the Steam build environment: 77be7e0dac8b1c31944cff063b2af4c5770201bd

C99 is probably fair to assume though, let alone C89.

bobbens:
I'm fine with either way, however, I do recall there were some obscure memory issues back in the day because of either some misunderstanding of the C spec or some OS not following the spec, so it might be better to be explicit.

Merging, thanks!

PR #1318, opened 2020-10-29
Window color, tab, etc. changes, many misc. UI tweaks

djvs:
Couple UI touchups (WIP sorta - see screenshots that go along with commits)

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -162,7 +162,7 @@ void conf_setDefaults (void)
    conf.font_size_console = 10;
    conf.font_size_intro   = 18;
    conf.font_size_def     = 12;
-   conf.font_size_small   = 11;
+   conf.font_size_small   = 11.5;
    conf.font_name_default = NULL;
    conf.font_name_monospace = NULL;
 
--- a/src/conf.c
+++ b/src/conf.c
@@ -162,7 +162,7 @@ void conf_setDefaults (void)
    conf.font_size_console = 10;
    conf.font_size_intro   = 18;
    conf.font_size_def     = 12;
-   conf.font_size_small   = 11.5;
+   conf.font_size_small   = 11;
    conf.font_name_default = NULL;
    conf.font_name_monospace = NULL;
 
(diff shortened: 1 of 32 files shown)


src/conf.c
```diff
@@ -162,7 +162,7 @@ void conf_setDefaults (void)
    conf.font_size_console = 10;
    conf.font_size_intro   = 18;
    conf.font_size_def     = 12;
-   conf.font_size_small   = 11;
+   conf.font_size_small   = 11.5;
```
bobbens, line 165: This is a bit scary because there's probably a whole amount of minor visualization glitches that may appear when changing the font size. Is the difference between 12 and 11.5 significant enough to warrant two different fonts?

onpon4:
I must admit I'm attached to the way outlines currently work in Naev and don't care much for the changes to outlines here. I also don't really feel the "small" font size needs to be increased. Everything else looks ok to me though, I guess (I must admit I might have a preference for the theme we came up with, but that could just be me being overly attached to my work 😅)

🕷️

src/font.c
```diff
@@ -554,27 +554,39 @@ static void gl_printHalo( const glFont *ft_font,
...
+      func( ft_font, width, height, bx, by - halfRadius, bg, text, 1 );
+      func( ft_font, width, height, bx, by + halfRadius, bg, text, 1 );
+      func( ft_font, width, height, bx - radius, by, bg, text, 1 );
+      func( ft_font, width, height, bx + radius, by, bg, text, 1 );
+      func( ft_font, width, height, bx, by - radius, bg, text, 1 );
+      func( ft_font, width, height, bx, by + radius, bg, text, 1 );
+      func( ft_font, width, height, bx - halfRadius, by - halfRadius, bg, text, 1 );
+      func( ft_font, width, height, bx - halfRadius, by + halfRadius, bg, text, 1 );
+      func( ft_font, width, height, bx + halfRadius, by - halfRadius, bg, text, 1 );
+      func( ft_font, width, height, bx + halfRadius, by + halfRadius, bg, text, 1 );
```
onpon4, line 589: Ah, it looks like indentation here was messed up again with your recent changes (the inner stuff is indented by 2 instead of 1 and the comment has the little asterisks misaligned)

🕷️

djvs:
For posterity

onpon4:
OK, I think all concerns have been addressed and this does look real nice, so merging. Thank you~!

🕷️

PR #1322, opened 2020-10-30
Feat/ui tweaks 2

djvs:
Tabwin background + ui fixes, cGrey for key/value lists, land_outfits prettying up


onpon4:
Looks good to me. :+1: I like the use of gray for headers, that looks much better than the dark green we used to have.

🕵️

PR #1323, opened 2020-10-30
put map overlay under gui and shrink contents a little

djvs:


onpon4:
Regarding this, I'd really like the GUI code (Lua code) to be involved in this, make it so that the GUI designates what space to use. Hardcoding it makes compatibility with various GUIs more difficult (both because the size/location becomes implicit and because it puts design limitations on the GUIs that wouldn't be there otherwise).

🕵️

djvs:
Yup, basically it would just impose a 140px top & bottom limitation on GUI size.  I'm open to suggestions (@bobbens ?) on how to handle this with a Lua hook.

bobbens:
You could add a Lua function to `src/nlua_gui.c` to designate the available overlay area or something, and have it fit into that area. Would be fairly straight-forward. Default could be trying to use everything. However, what is the rationale behind putting overlay under GUI? I like it over GUI :P

djvs:
The main issue with having it over GUI was the high opacity overlay which halfway blacks out the GUI.  However, I also actually removed that on this branch so at this point it could go over or under - if it goes over, it doesn't have to have a size limitation at all, but then having it with no background necessitated an outline, which raised an objection from @onpon4 - however, it was originally a 2px outline, and I could easily shrink it to 0<n<=1, which would be nearly the same as nothing in terms of the visibility of markers behind it.

djvs:
I much prefer it without it ^ that's a 2px outline iirc

djvs:
Lua hooks added with default remaining at 140px vertical margin

bobbens:
I'm not sure the overlay under the GUI is an improvement. Please see attached image in which a bright star basically makes the map illegible. I suppose it could be compensated in part by added "halos" to all the objects in the minimap, but if we go down that route it might just end up reproducing the current implementation while being much more complex.
screenshot058

djvs:
@bobbens You mean the markers or the text?  Personally I'd prefer that small change over having the GUI obscured.

djvs:
Possible to simply use an uncommon color for the markers also.

djvs:
Current state

bobbens:
It looks better than before, however, mouse clicking seems to be broken. It is not being translated to the correct positions so the right click GOTO is all FUBAR. Another issue is how to deal with this interaction? If you can click on the overlay + normal view, it might be tricky now that I think about it. An option would be to have tab cycle between 3 modes, no overlay, under gui overlay, over gui overlay or something like that, but then it feels like it does get a bit redundant.

djvs:
@bobbens Fixed, and gave a sliver of opacity back to the overlay

djvs:
The "order of operations" for click events hasn't been changed, and since the overlay is back on top visually, the behavior should be same as before, it appears to just fall back to primary space display if nothing on the overlay matches.

bobbens:
I want to merge this, but would it be possible to add a slider to options to control the opacity of the background? Sorry for taking to long, seems to have a few conflicts... T_T

bobbens:
I think it looks better now. Thanks!

PR #1324, opened 2020-10-31
[Mission] Spellcheck the next Za'lek campaign mission.

ProjectSynchro:
This is a grammar/spelling checked version of #1305 can close both of these when this is merged.

I must say I am definitely feeling some Za'lek icecream right about now..


bobbens:
Thanks!

Issue #1288, opened 2020-10-09
Ship description cut off


Wuzzy2:
In the shipyard on the planetary screen, when you click on a ship with a long description, the description is sometimes cut off. Test this in German. (I don't know if it's an issue in English.)

Commit: fcb73da

onpon4:
One possible solution would be to change the translation to something shorter.

🕵️

bobbens:
We still do need a general solution for text getting cut off. I would argue in favour of having scrollbars appear allowing scrolling when the text exceeds the bounding area. Tricky part will be working with things like the shipyard view where we use two text columns that would have to be aligned. Maybe using another widget that encompasses both would be a possibility.

Wuzzy2:
This could probably be solved even without scrollbar. The cut-off point is way above the actual window border, so the remaining space is currently just empty.

UncombedCoconut:
can confirm this occurs in English gameplay too
![naev hyena description cut off](https://user-images.githubusercontent.com/1224111/97389709-62280500-18b1-11eb-9507-0821261cd7be.png)

On my machine it's "fixed" by a low-effort rect-size change like this. It would be embarrassing to stop there, but maybe it's worth putting in, pending a general solution? (Maybe it's a good change anyway, at least if similar screens are consistent?)

```diff
diff --git a/src/land_shipyard.c b/src/land_shipyard.c
index e5987ef1d..fba51ee68 100644
--- a/src/land_shipyard.c
+++ b/src/land_shipyard.c
@@ -145,7 +145,7 @@ void shipyard_open( unsigned int wid )
          w-(40+iw+20+100)-20, th, 0, "txtDDesc", &gl_smallFont, NULL, NULL );
    y -= th;
    window_addText( wid, 20+iw+40, y,
-         w-(20+iw+40) - 180, 185, 0, "txtDescription",
+         w-(20+iw+40) - 180, 20-y, 0, "txtDescription",
          &gl_smallFont, NULL, NULL );
 
    /* set up the ships to buy/sell */
```

onpon4:
Ah, yeah, that sort of thing can be a pervasive problem in Naev at times. I do think we should be determining height in a more dynamic way. (like in your diff).

🕵️

UncombedCoconut:
Putting on my bureaucrat hat for a moment -- should we close this and make "need a general solution for text getting cut off" a separate issue? Fair to say the "v0.8.0" work item is resolved because ship descriptions aren't cut off in practice anymore?

PR #1326, opened 2020-10-31
Resize text boxes (quick fix to #1288)

UncombedCoconut:
Recalculate the text widgets' ideal size and position based on the current layout. This serves as a tactical fix to #1288.


bobbens:
Looks good, thanks!

PR #1327, opened 2020-10-31
Unify three procedures for displaying faction presence on maps.

UncombedCoconut:
The goal is to stop them from having their own separate problems, now or in the future. The issue I noticed was the editor view. (Its format string was different -- instead of \a escapes it had something else that rendered incorrectly.)
One of them also had extra guards around size_t underflow. Why not all of them... :)
There were a few ways of doing this. This way exposes a UI function (rather than an interface that refers to the text buffer) and needs to add a function for measuring the content height of a text box (which seems like a useful capability).

### DIFF
--- a/src/dev_mapedit.c
+++ b/src/dev_mapedit.c
@@ -656,7 +656,6 @@ void mapedit_selectText (void)
    int i, l;
    char buf[1024];
    StarSystem *sys;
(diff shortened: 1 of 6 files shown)


bobbens:
Looks good, thanks!

PR #1328, opened 2020-10-31
[CI] Disable meson testing temporarily, update meson wrapper

ProjectSynchro:
This PR includes additional build tests (all ubuntu and windows runners) and updates the meson wrapper with what I have in #1316 

Build times are roughly the same, we just spin up more runners.


bobbens:
Thanks!

PR #1329, opened 2020-10-31
Fix edge case in ship-stats tooltips caused by initial newline in the text.

UncombedCoconut:
Not a great situation to be in, but it's way less confusing if we draw it. :)


bobbens:
Thanks!

PR #1331, opened 2020-10-31
More UI tweaks/cleanup/fixes

djvs:
- Simplify tab button logic
- Simplify widget outlines and make highlight color lighter
- Fix tab width calculation issue
- Change image widths in map_system from 64 to 96
- Fix input box alignment on tab windows, remove one border from the input boxes, harmonize border with widget highlight, lighten outline
- Use define'd variables for tab sizes 
- Fix text input color (green-on-black instead of black-on-white), which also fixes  associated text outline glitch
- Removed mass/CPU bar outlines and expand their height
- Fix tab background logic for non-bottomed tab bars

### DIFF
--- a/src/equipment.c
+++ b/src/equipment.c
@@ -542,7 +542,6 @@ static void equipment_renderMisc( double bx, double by, double bw, double bh, vo
    double percent;
    double x, y;
    double w, h;
-   const glColour *lc, *dc;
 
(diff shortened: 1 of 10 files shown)


djvs:
List prettifying

bobbens:
Thanks!

PR #1333, opened 2020-11-03
Sdl2 simplification

UncombedCoconut:
This introduces a mild merge conflict with ids1024's HiDPI work (https://github.com/naev/naev/pull/971), but is basically independent and is a baby step toward easily getting it right.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -36497,10 +36497,6 @@ msgstr "Naev im Vollbildmodus ausführen"
 msgid "Use video modesetting when fullscreen is enabled"
 msgstr "Video-Moduseinstellung benutzen, falls Vollbild aktiviert ist"
 
-#: src/conf.c:982
-msgid "Minimize on focus loss"
-msgstr "Bei Fokusverlust minimieren"
-
 #. FPS
 #: src/conf.c:987
 msgid "Display a framerate counter"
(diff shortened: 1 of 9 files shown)


UncombedCoconut:
Hmm, the CI break looks legit, though I didn't get it locally. Please do not merge until I investigate.

UncombedCoconut:
My bad, did not read *all* of the fine manual. (nloewen explained what I needed to check in Readme.md.)

po/de.po
```diff
@@ -36494,12 +36494,8 @@ msgid "Run Naev in full-screen mode"
 msgstr "Naev im Vollbildmodus ausführen"
 
 #: src/conf.c:978
-msgid "Use video modesetting when fullscreen is enabled (SDL2-only)"
-msgstr "Video-Moduseinstellung benutzen, falls Vollbild aktiviert ist (nur für SDL2)"
-
-#: src/conf.c:982
-msgid "Minimize on focus loss (SDL2-only)"
-msgstr "Bei Fokusverlust minimieren (nur für SDL2)"
```
bobbens, line 36502: It's not clear to me why you removed this translation (only the SDL2 part should be removed no?)

UncombedCoconut: Hi, it's because the string "Minimize on focus loss" already had a translation in the same po file, for options.c. (The SDL2 disclaimer was only written into conf.lua; the options UI knew it didn't need it.)

bobbens: I see, thanks!

PR #1336, opened 2020-11-04
Refinement of ids1024's hidpi branch (CAUTION)

UncombedCoconut:
[Rebased 2020-11-05]
At this point, I believe all the major issues are sorted, but it's going to take more testing to be sure.

Potential gotcha which is not a bug: the dimensions for windowed modes will now be "screen coordinates" (so if your scale factor is 2, "1600x900" gives you 3200px x 1800px. This is the convention adopted by WM's/compositors/SDL2 so that the results are sane when one drags a window between displays with different scale factors.

This PR depends on a gross hack at the end of gl_setupFullscreen. Without it, early init (load screen and nebula generation) will misunderstand the window/drawable size. The basic issue is that SDL2 (and the desktop compositor) might not immediately size our window the way we asked. To get SDL2 to process a resize, we must pump events and trigger the internal Wayland_HandlePendingResize function via SDL_GL_SwapWindow. (A comment in SDL2 says "Wayland-EGL forbids drawing calls in-between SwapBuffers and wl_egl_window_resize".)

This PR does not include the Windows/Mac stuff I mentioned in #970 comments. AFAIK, testing those platforms is the main thing left to do. I can add speculative platform-specific stuff and log whether SDL2 reports SDL_HINT_VIDEO_HIGHDPI_DISABLED -- that's about it.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -248,18 +248,19 @@ void conf_setAudioDefaults (void)
 void conf_setVideoDefaults (void)
 {
    int w, h, f;
+   SDL_DisplayMode resolution;
 
    /* More complex resolution handling. */
    f = 0;
-   if ((gl_screen.desktop_w > 0) && (gl_screen.desktop_h > 0)) {
+   if (SDL_GetCurrentDisplayMode( 0, &resolution ) == 0) {
       /* Try higher resolution. */
       w = RESOLUTION_W_DEFAULT;
       h = RESOLUTION_H_DEFAULT;
 
       /* Fullscreen and fit everything onscreen. */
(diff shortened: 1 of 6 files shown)


UncombedCoconut:
When I first raised this PR, I mentioned beam weapons appearing jagged. I've since realized that they always looked that way -- it just never bothered me until I went looking for things I might have broken. :)

bobbens:
> When I first raised this PR, I mentioned beam weapons appearing jagged. I've since realized that they always looked that way -- it just never bothered me until I went looking for things I might have broken. :)

Don't worry, everything needs fixing in Naev.

This seems to work locally, I'll merge and if win/mac issues arise we'll fix them as they come ;). Probably have a few betas left before release. Thanks!

PR #1338, opened 2020-11-06
backport hidpi fixes that can already benefit mainline

UncombedCoconut:
A few nasties can be reproduced without any hidpi patches -- just set an extreme config like width=3072; height=1728; scalefactor=2.4;
There's also some UI code which mixes an unscaled-distance threshold with a scaled-distance behavior. This is evil, and I've enforced some consistency, but it's an aesthetic judgment that the simplest correction (what I did) looks fine at all scales.
If we merge this I'll rebase https://github.com/naev/naev/pull/1336/

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -1182,7 +1182,7 @@ static void input_mouseMove( SDL_Event* event )
 static void input_clickevent( SDL_Event* event )
 {
    unsigned int pid;
-   int mx, my, mxr, myr, pntid, jpid, astid, fieid;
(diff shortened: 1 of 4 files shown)


bobbens:
Every time I touch this stuff, everything breaks, but it seems to not break anything. Thanks!

ids1024:
As long as it breaks slightly less than it fixes... progress!

UncombedCoconut:
Fair warning. I broke something (specifically signaling -G/--generate through magic numbers). LOL.

Issue #1038, opened 2020-06-15
Minor graphical issue with slider


Wuzzy2:
If you click on a slider (widget), a thin rectangle appears to mark it. This is OK.

But if you move the slider all the way to the left or right, the slider does not really neatly fit into the rectangle. This looks slightly odd.

090051c25a68fe7f0153f0e74f25558981030261

Houkime:
The issue looks like this
![image](https://user-images.githubusercontent.com/32805122/96364777-cfa69980-112b-11eb-8d2a-ec44e1fa214f.png)
![image](https://user-images.githubusercontent.com/32805122/96364795-e6e58700-112b-11eb-8b71-3561f4d28d7f.png)

Note that it is not symmetrical. It looks like the "pivot point" of slider knob is offset to the left.

So probably the fix should be two-part.

Center the knob so that it can hit the edges symmetrically.
enlarge focus indicator (a thin rectangle) x-wise a bit or shrink the sliding area.
Good measure might be to have the same rectangle-knob spacing from knob's edge at maximum-left position as it is now from slider track on the right to the rectangle (x1) 

![image](https://user-images.githubusercontent.com/32805122/96365121-e1893c00-112d-11eb-8485-0ad4541e25b8.png)

Note that on this picture y-wise frame vs slider is also not symmetrical and text is colliding with the focus frame.

Houkime:
In src/toolkit.c in window_render() (https://github.com/naev/naev/blob/master/src/toolkit.c#L1315)  
One can  see that focus rendering is hardcoded to widget's w and h.

```
if (w->focus != -1) {
      wgt = toolkit_getFocus( w );
      if ((wgt == NULL) || wgt_isFlag(wgt, WGT_FLAG_KILL))
         return;
      x  += wgt->x;
      y  += wgt->y;
      wid = wgt->w;
      hei = wgt->h;
      toolkit_drawOutline( x, y, wid, hei, 3, &cBlack, NULL );
```
So, since there are no per-widget settings, all changes should concentrate inside widget code.
The fader widget code is in src/tk/widget/fader.c
The rendering code is fad_render() (https://github.com/naev/naev/blob/master/src/tk/widget/fader.c#L87)

```
static void fad_render( Widget* fad, double bx, double by )
{
   double pos;
   double w,h;
   double tx,ty, tw,th;
   double kx,ky, kw,kh;

   /* Some values. */
[...]

PR #1340, opened 2020-11-06
Fix faders

Houkime:
Mouse is a bit desynced since rendering does not correspond exactly to
how mouse movements are interpreted, but otherwise fine.

faders_relatively_fixed

Codestyle needs work to blend in though.

Fixes https://github.com/naev/naev/issues/1038

### DIFF
--- a/src/tk/widget/fader.c
+++ b/src/tk/widget/fader.c
@@ -12,6 +12,10 @@
 #include "tk/toolkit_priv.h"
 
 
+#define KNOB_DEF_THICKNESS 15.
+#define TRACK_DEF_THICKNESS 5.
+
+
 static void fad_render( Widget* fad, double bx, double by );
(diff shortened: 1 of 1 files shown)


bobbens:
I tried to fix some of the mouse motion issues with custom widgets. However, it does seem like the movement is fairly desynced with the mouse movement. Not entirely sure where the issue can be.

onpon4:
I think this sort of issue is why we gave up on this one.

I like that you're getting rid of those magic numbers, though I would note that consts probably shouldn't be given all-caps names (those are usually used for preprocessor macros). But I'd bet applying the same to mouse click location will eventually lead to solving the mouse click discrepancy.

Houkime:
Changed mouse handling to mirror changes in rendering.
Seems to work flawlessly now.

faders-glorious

bobbens:
Looks good, thanks!

PR #1351, opened 2020-11-10
Typo policing! (Sorry if I've "corrected" anyone's intentional style choices; I tried not to!)

UncombedCoconut:


PhoenixRiver:
Hello :)

Nice catches, I spent a lot of time doing this too... though I would
argue with you about the correct spelling of targetting... especially
given your spelling of "unequipped" ;) - but who said language was
consistent ;) and the "s" in "Sirius's" is definitely old, if not
archaic from today's perspective... ( dat/assets/ember.xml ), ditto
"wintry" ( dat/assets/harnaar.xml ) though the use of the word is
slangy at best anyway, I might suggest "snowy" instead which is both
more used and less ambiguous spelling-wise! :)

Might also have suggested "past time" rather than "pastime" in
dat/assets/stormgard.xml as being more common (and also more
consistent with flip side, eye candy, grace note and light year) but
good catch on "fare" (one I would probably have missed!) and also
"gargle blaster" - I think that might have been my mistake in the
first place! :D - I should have checked!

"Astronomists" might be the only stylistic choice word that someone
might have intended but I seem to remember that outside of spoken text
it had been decided that we would be staying with (relatively)
standard language...

Oh... "Terminus's" - my complain is similar to Sirius' - archaic (
dat/assets/terminus.xml )

Erm... "in case the attacker was parallelly killed" - arg, much better
"in case the attacker was killed in parallel" (
dat/events/sirius/preach.lua ) !!

Hmmm... another stylistic one ( dat/landing.lua ) : return false,
"Nobody expects the Spanish inquisition!" - I would be tempted to
argue that stylistically this _should_ be a small "s" for Spanish
since although we know it is a proper noun, this is both "speech"(ish)
and we can pretty safely say that people in universe would have no
idea at all what "Spanish" is ... but if you prefer it to be
capitalised then you've got to also capitalise "Inquisition" since
that is their full proper noun name at now...

And again "Urnus's" ( dat/missions/flf/flf_pre02.lua ) - archaic

I'd argue for a change to "initiation" here : (
dat/missions/neutral/crimelord.lua ) because I don't want to start a
discussion on the relative merits of "s" / "z" ! :D

And again "Countess's" ( dat/missions/neutral/drunkard.lua ) - archaic

( dat/missions/neutral/seek_n_destroy.lua ) - "OK, OK" might work
better in that spoken form as "Okay, okay"...

( dat/missions/neutral/spacefamily.lua ) - "Harrus's" - archaic...

( dat/missions/trader/anxiousmerchant.lua ) - "Jeez"; really not
convinced that's the right spelling an definitely a stylistic choice
given it's spoken...

Can't comment on the .po files as it won't show me the diffs... I'm
kinda assuming that it's the same as the above though?

( src/nfile.c ) A further change is needed as it should be hyphenated:
"non-existent"
( src/npc.c ) Another further change; here should be "Non-existent"
( src/nxml_lua.c ) Aaaannd for consistency "non-existent" here too...
or at least these three should also be the same! :D
Also in ( src/space.c ) - in src/tech.c they are already correct...

I hope these help and thank you for spending so much time on a
seemingly small thing, they really do add up - I remember making
hundreds of typo fixes and then getting entangled in correctly
classifying planets and and whole bunch of making planets match their
images, descriptions and types due to it... I'm not sure I ever
finished but someone else did I think - I got too side-tracked by new
assets and then life :/

All the best,

River *<:@)

UncombedCoconut:
Hi!
I fully agree. with some of these. (One is actually something I *meant* to do but forgot among all the edits!)
Where I don't agree, it matters very much to me if you are a contributing author who made some of these choices.
(A few of these are very clearly resolved by dictionaries or style manuals, though. For example, it's simply standard to form the possessive of a singular noun by adding -'s, even if it's a singular noun ending in -s. Though apparently the AP style makes an exception if the following word begins with s-, and the APA style makes an exception for singular nouns ending in a silent -s. Messy!)
I basically went in with the attitude that an actual spacefaring civilization would speak nothing recognizable as current English and rely heavily on translators / Babel Fishes. So just translate using current diction/spelling/grammar.
For points of spelling and the question of whether it's valid for "nonexistent" to be a single word, I refer you to a dictionary.

It probably is a good idea to treat the spoken "ok[ay]" and the word "non[-]existent" consistently.
You have me sold on "okay, okay".
"Nonexistent", I'm less sure about. The unhyphenated form is definitely valid and commonly used, but so is the hyphenated one.

UncombedCoconut:
BTW, I think this PR is huge enough already, and unless I've made a disastrous mistake the follow-up can happen separately.

onpon4:
> though I would
> argue with you about the correct spelling of targetting... especially
> given your spelling of "unequipped"

Wiktionary describes "targetting" thusly:

> targetting
> 1. (British spelling) Obsolete spelling of targeting
> 2. Misspelling of targeting.

So regardless of logic, the standard spelling is "targeting".

"Unequiped", meanwhile, doesn't seem to be a word at all. "Unequipped" is the standard spelling.

> and the "s" in "Sirius's" is definitely old, if not
> archaic from today's perspective...

I would disagree. Some style guides call for the extra "s", some don't. Either is acceptable. I personally don't care one way or the other, and I use both. (This same goes for the rest that you refer to as "archaic" for this reason. It won't be archaic until it's an old style that's no longer really in use.)

> ditto "wintry"

"Wintry" is the standard spelling and most in-line with its roots in Old English. "Wintery" is an alternate spelling. and actually comes up as an error on my spell checker.

> though the use of the word is
> slangy at best anyway, I might suggest "snowy" instead which is both
> more used and less ambiguous spelling-wise!

Wintry is a very old word so I don't think it counts as slang. In any case the spelling isn't ambiguous; "wintry" is the standard spelling but "wintery" is an alternative spelling accepted by many. It also doesn't mean exactly the same thing as "snowy". I don't see a reason to avoid it.

> Might also have suggested "past time" rather than "pastime" in
> dat/assets/stormgard.xml as being more common

Um, no, that's wrong. "Pastime" is the standard spelling, and "past time" would be etymologically wrong because the term comes from the term "pass" (as in, something you use to *pass time*), not "past". "Past time" does not occur in any English dictionary I can find and I've never seen anyone use that.

> I'd argue for a change to "initiation" here : (

"Initialization" or "initialisation" is the standard technical term for that context (initialization of a variable in programming). The former is standard in American English, while I think (but don't know for sure) that the latter is standard for British English. The spelling to use would depend on the dialect of English being used. "Initiation" I don't think is a standard alternative term for "initialization", and it has a very different meaning. So I don't agree with keeping "initiation" as a term for that sort of context. (I will admit, though, that since it's a comment it doesn't matter all that much as long as it's understood by programmers.)

> "Jeez"; really not
> convinced that's the right spelling

https://en.wiktionary.org/wiki/jeez

It's a right spelling. Not *the* right spelling; there are a few variants that are in use.

> A further change is needed as it should be hyphenated:
> "non-existent"

That rule is actually archaic. Non-hyphenated forms of a lot of words (including "nonexistent") have been standard for quite some time now (though I don't remember exactly for how long). The hyphenated form is acceptable too, but the non-hyphenated form is correct and I'm pretty sure the most common form nowadays. ("Nowadays", on a side note, also originated as the hyphened form "now-adays", a spelling which is now considered archaic and isn't in use. In fact a very large number of common words originated as hyphenated words, but lost their hyphens.)

Personally, I much prefer "nonexistent" over "non-existent". "Non-existent" looks old-fashioned to me. But either is valid.

🕵️

onpon4:
Just looked through the changes and all look good on my end. Regarding spellings of "OK", that should probably be made consistent, though I don't think it matters what spelling is chosen; "OK", "okay", and even the less standard "ok" are all forms that I'm... ok with.

Style differences on hyphens and the spelling of "ok" aside, everything looks good. We can resolve the style differences separately. Whatever style is chosen (I think most developers prefer some form of British English, though I'm partial to American English myself), we'll want to make it consistent throughout the game. But since the game is far from complete, I'm not going to fret too much about not having picked a style yet. Merging.

🕵️

PhoenixRiver:
Agree with most of the above as long as I don't mind changing hat a
lot in terms of which version of English is being referred to... And I
might note that although a lot of dictionaries like writing British
English (Obsolete) or variations thereof you will find that It is
almost never both of those things: British English is not Obsolete and
the spellings it uses are far from obsolete (e.g. the letter "z" is
almost never used in British English but almost always used in the
postfix "ize" in American English - which causes some strange
inconsistencies!)

But like I said, I'm not getting into an argument about the objections
my (American English) spellchecker has with native British English
writing / spells / grammar / whatever. I am certainly not about to
propose one of the Englishes over another since my career has depended
on knowing the difference between as many of them as possible -
Scottish English (different to Scots; which is either a different
language or is a dialect), Kiwi English, Aussie English, American
(USAian) English, Canadian English, British English, International
(Chinese) English, International (Indian) English, International
(South-East Asian English) English, International English, Pigeon
English (also of many stripes) - you are welcome to argue all you like
about using "standard" English but unless anyone is willing to define
what that means then it's going to be a lot of swings and roundabout!
:)

Also agree that with so much writing still left to write, nevermind
translations, it seems a little premature to argue about such
things... maybe a consistent approach can be decided upon at some
point but I would suggest that by that time it is more likely that
there will be multiple translations for Englishes as well as for other
languages.

River *<:@)

onpon4:
> British English is not Obsolete

That's not what the dictionary entry I referred to means. No dictionary claims that British English is obsolete, because it isn't. It's saying that "targetting" is an obsolete *spelling* which was a British English spelling.

> you are welcome to argue all you like
> about using "standard" English but unless anyone is willing to define
> what that means then it's going to be a lot of swings and roundabout!

None of the examples I gave vary between dialects as far as I can tell. You did see the part about "initialization" where I referred to the British spelling "initialisation" as standard too, right? In cases like those what's standard depends on the dialect spoken. But I don't see how you could argue that there's no such thing as standardized spellings in English. Spellings have been standardized for over a hundred years.

PhoenixRiver:
> That's not what the dictionary entry I referred to means. No dictionary
> claims that British English is obsolete, because it isn't. It's saying that
> "targetting" is an obsolete *spelling* which was a British English
> spelling.

> None of the examples I gave vary between dialects as far as I can tell. You
> did see the part about "initialization" where I referred to the British
> spelling "initialisation" as standard too, right? In cases like those what's
> standard depends on the dialect spoken. But I don't see how you could argue
> that there's no such thing as standardized spellings in English. Spellings
> have been standardized for over a hundred years.

Then you best take this up with my dozen or so English and History
teachers (blame my dad) as to why targetting should be written with
two "tt" if that is not standard British... standard means in use does
it not? Well in the classrooms I was taught in this was in use by the
class as enforced by our teachers.

As for standardised spellings it is entirely apparent to me that
spellings change with the generations as non-hyphenation to
hyphenation to prefixation of words has occurred within one generation
for many dozens of words, nevermind the adoption of "Americanisms" and
pushes to standardise spellings (particularly with regard to the
doubling of consonants before endings...) - these pushes sometimes
work together across versions of English and sometimes against with
the concomitant success or lack thereof of change (quite apart from
other influences).

Anyway, way off topic here ;)

River *<:@)

PR #1352, opened 2020-11-10
As suggested by @nloewen, use an environment object in luajit/meson.build.

UncombedCoconut:
Also clean up the "PREFIX" bit, which was needed for "make install" but isn't for "make amalg".

I feel compelled to make fun of Meson for saying:

1. All objects are immutable!
2. You may add to env, an environment object, by calling env.set(...).

(even though  it's not actually a contradiction...)

I've been thinking about whether to leave it here. This guy has demonstrated that translating the whole luajit build to Meson isn't too daunting (though I'm not sure he covered all cases): https://github.com/franko/luajit
The idea is attractive because the current wrapped "make" builds in-tree, so it's easy to get into Big Trouble trying to do native builds and cross-compiles of the same repo. My understanding is that the pure Meson version would build out of tree and have no such problem.


bobbens:
Thanks!

PR #1353, opened 2020-11-10
Include extras/resource.rc in Meson Windows builds

UncombedCoconut:
(as it is in Autotools builds).


bobbens:
It would be nice to have all these random extras and such files that use Naev's version and stuff to actually generate that from the real version. Would be easy to do with autotools (now that I know a bit of how it works), but wouldn't be sure how to do it with meson. Either way that's besides the scope of this PR. Thanks!

PR #1357, opened 2020-11-10
Hooray, we don't need to roll our own setenv()!

UncombedCoconut:


bobbens:
Since when was there an SDL_setenv!?! I'm pretty sure I searched for it first... Thanks!

PR #1358, opened 2020-11-11
Add a Windows app manifest that declares us HiDPI-aware. Really aware. The Doors of Perception are open.

UncombedCoconut:
Also, I have no proof this crap is actually going to improve Windows users' experiences. 😭
But I've been through mountains of docs implying that (a correct version of) this is what we need for SDL_WINDOW_ALLOW_HIGHDPI to do its thing without extra OS-specific C code.

Although it's not strictly necessary, I figured this would be a good time to move resource.rc.in to extras/**windows**/ since it's irrelevant to everyone else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,10 +46,11 @@ stdout.txt
 *.mode1v3
 extras/.dirstamp
 extras/macos/Info.plist
-extras/resource.rc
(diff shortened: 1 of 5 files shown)


bobbens:
Thanks!

PR #1361, opened 2020-11-12
Fix for Steam builds, which are stuck with ancient autotools packages.

UncombedCoconut:


configure.ac
```diff
@@ -75,7 +75,8 @@ AC_PROG_OBJC
 AC_PROG_LN_S
 AC_PROG_RANLIB
 AM_PROG_AR
-AX_PROG_CC_FOR_BUILD
+# Should be AX_PROG_CC_FOR_BUILD -- except Steam builds occur in a paleozoic environment where that doesn't work.
```
bobbens, line 78: Wouldn't it be possible to provide this macro to the steam environment or is that even more hackey?

UncombedCoconut:
I think so.
If it's not urgent to unbreak the builds, I can look into that before we
push a half-baked solution, but I have to call it quits for today.

bobbens:
I'll pull for now (fixing builds is priority), but I wonder if there's a better way of doing it. What happens with meson and steam?

UncombedCoconut:
On Discord, Synchro mentioned he felt close to having Meson+Steam working
(but still one "ominous error" to solve). If switching to that is
successful, we probably won't have to worry about bugs in 9-year-old
Autotools packages anymore.
TBH, I'm surprised Meson isn't even harder to get working on Ubuntu 12.04,
but it's a happy surprise.

PR #1366, opened 2020-11-13
Make size parameters to array_create/array_resize size_t. Fixes #1365.

### LINKED ISSUE #1365: Galaxy-brain gcc emits a complicated warning for release builds
Here's the tail of a release build. (That means: DEBUGGING not set, optimization level high enough to set say -ftree-vrp, link-time optimization enabled.)
```
[151/151] Linking target naev
In function ‘_array_resize_container’,
    inlined from ‘_array_resize_helper’ at ../naev/src/array.c:43:4,
    inlined from ‘npc_clear.part.0’ at ../naev/src/npc.c:437:4:
../naev/src/array.c:33:11: warning: argument 2 range [18446743764471906312, 18446744073709551336] exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
   33 |       c = realloc(c, sizeof(_private_container) + e_size * c->_reserved);
      |           ^
../naev/src/array.c: In function ‘npc_clear.part.0’:
/usr/include/stdlib.h:550:14: note: in a call to allocation function ‘realloc’ declared here
  550 | extern void *realloc (void *__ptr, size_t __size)
      |              ^
```
What's happening is:
- It's inlining an array_resize(..., 0)
- It's considering the branch where the array doesn't have enough space reserved
- It's Very Smart, so it knows the only way the doubling process goes from <0 space to >= space is an integer underflow, which means we're about to pass realloc something insane. (It's not Very Smart enough to know the branch cannot be taken in the first place.)

The purpose of this issue is to document something funny. Also because it's unclear which fix strategy/strategies will be desirable.

UncombedCoconut:
This approach is good enough for the venerable reallocarray(), but could theoretically make underflow bugs harder to catch immediately.

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -5,7 +5,7 @@
 
 #include "nstring.h"
 
-void *_array_create_helper(size_t e_size, int capacity)
+void *_array_create_helper(size_t e_size, size_t capacity)
 {
    if ( capacity <= 0 )
       capacity = 1;
@@ -19,7 +19,7 @@ void *_array_create_helper(size_t e_size, int capacity)
    return c->_array;
 }
 
-static void _array_resize_container(_private_container **c_, size_t e_size, int new_size)
+static void _array_resize_container(_private_container **c_, size_t e_size, size_t new_size)
 {
    assert( new_size >= 0 );
    _private_container *c = *c_;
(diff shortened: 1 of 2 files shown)


bobbens:
Thanks!

PR #1371, opened 2020-11-13
Fix Occasional Steam Release Failures // Fix Release Artefact Names

ProjectSynchro:
Simplified the steam 2fa script a bit
I had much more work going on but I need to have a sit down and think plan this stuff out.


ProjectSynchro:
I will have a more all encompassing solution this weekend.. I just don't have enough time at the moment to rewrite/ merge everything I planned to cleanly. For now this should have releases looking legit enough, and should keep steam working without issues.

bobbens:
Thanks!

Issue #1281, opened 2020-10-09
Untranslatable strings


Wuzzy2:
List of untranslatable strings:

- "+X Cargo" (where X is a number)
- "+X Fuel" (where X is a number)
- "Civilian", "Empire Civilian", etc.
- "This system is landable"
- Control names
- Settings menu tabs
- Info screen tabs
- Outfit type like "Armour Modifier" or "Cargo Expansion"
- The entire intro screen when you start a game
- The Credits screen
- Large faction names in star map background
- Modes in star map
- (brushed GUI) When you marked a Jump Point to an Unknown system, the target shown at the top says "Unknown (1)"
- (brushed GUI) When no planet is marked, the planet target shown at the top says "None"
- The bottom left text of Star Map in Cost or Trade mode
- System names
- "X jumps" in slimv2 GUI. It's also always in plural form like "1 jumps"
- The word "here" in "No price info for [system] here"
- Slot type letters W/U/S in shipyard
- Slot type letters W/U/S in equipment screen
- "All" (ship log)
- "Add a log entry"
- "Add an entry to your diary:"
- "Your Diary"

Strings which are translatable, but the game seems to ignore the translation currently:
- Outfit tooltip in outfit screen
- The outfit name in the outfit list (button only, the description field to the right is actually fine)
- Commodity names
- Planet names
- Planet description
- Bar description
- "This slot is used to equip accessories that can enhance the ship's performance."
- Commodity names and descriptions in Commodity tab
- (brushed GUI) Service names on the "TARGETED" overlay when a planet is targeted
- Faction names (long and short)
- 2nd line of tooltip of core slots in Equipment screen (e.g. "Core Systems" in "Slot Core Systems")
- "Accessory"
- Special ship names like "Advertiser Hyena"

Commit: fcb73daf7f9f7f51b8024359b6551a3f0d3c66d9

Wuzzy2:
More untranslatable strings can be found in the "legacy" GUI:
* Nav
* 2nd
* Off
* Target
* No Target
* Creds
* Cargo Free
* Cargo
* S (Shield)
* A (Armor)

Other bugs:
* I also noticed the ship names are not translated in the shipyard (on the buttons), but they ARE translated in the long description at the right side. The ship name is also in untranslated form in the "Trade-In" dialogue.
* If you search for outfits or ships in the star map and view the result list, you always see "X jumps" even when X=1.
* The search dialog always displays and searches ship and outfit names in English even when there is a translation available

[...]

PR #1373, opened 2020-11-14
Translatability updates.

UncombedCoconut:
Tracker: https://github.com/naev/naev/issues/1281

### DIFF
--- a/src/outfit.c
+++ b/src/outfit.c
@@ -1240,7 +1240,7 @@ static void outfit_parseSBolt( Outfit* temp, const xmlNodePtr parent )
          "%.0f CPU\n"
          "%.0f%% Penetration\n"
          "%.2f DPS [%.0f Damage]\n"),
-         outfit_getType(temp), dtype_damageTypeToStr(temp->u.blt.dmg.type),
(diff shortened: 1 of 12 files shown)


UncombedCoconut:
Based on the Discord convo, I'll revise history here so the changes are clearer.

UncombedCoconut:
Obviously, the work isn't done, but it will probably be easier for all of us if I stop accumulating changes into this PR.

src/outfit.c
```diff
@@ -768,27 +768,27 @@ double outfit_cooldown( const Outfit* o )
...
+         N_("Ammunition"),
+         N_("Turret Launcher"),
+         N_("Ship Modification"),
+         N_("Afterburner"),
+         N_("Fighter Bay"),
+         N_("Fighter"),
+         N_("Star Map"),
+         N_("Local Map"),
+         N_("GUI"),
+         N_("License"),
```
onpon4, line 791: Bit of a personal note (not directly related to this change), I'm not a fan of us having these outfit types strictly defined like this with canonical names. Not sure if it can be avoided though. 😅

:spider:

onpon4 (approved):
This all looks like sound changes. Thanks! :slightly_smiling_face:

:spider:

PR #1377, opened 2020-11-15
Fun with strings

UncombedCoconut:
Mostly bug fixes. Touched a few loops that merely offended my sensibilities. ;)
~~BTW, I verified the btn_updateHotkey function works as intended, but the actual highlighting doesn't occur. I'm not sure why yet.~~ (Oh, I just lack the colour vision to perceive it!!)

### DIFF
--- a/src/dev_mapedit.c
+++ b/src/dev_mapedit.c
@@ -550,7 +550,7 @@ char *mapedit_nameFilter( char *name )
 
    out = calloc( 1, (strlen(name)+1)  );
    pos = 0;
-   for (i=0; i<(int)strlen(name); i++) {
+   for (i=0; name[i]!='\0'; i++) {
       if (!ispunct(name[i])) {
          if (name[i] == ' ')
             out[pos] = '_';
--- a/src/dev_outfit.c
+++ b/src/dev_outfit.c
@@ -230,7 +230,7 @@ void dout_csvAmmo( const char *path )
 
       /* Get AI name, in lower case. */
       ai = strdup( outfit_getAmmoAI(o) );
-      for (j=0; j<(int)strlen(ai); j++)
+      for (j=0; ai[j]!='\0'; j++)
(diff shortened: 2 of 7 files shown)


src/dev_mapedit.c
```diff
@@ -550,7 +550,7 @@ char *mapedit_nameFilter( char *name )
 
    out = calloc( 1, (strlen(name)+1)  );
    pos = 0;
-   for (i=0; i<(int)strlen(name); i++) {
+   for (i=0; name[i]!='\0'; i++) {
```
onpon4, line 553: Whoops! Emmy's comment got lost when I navigated away. 😅

So we're not the biggest fan of this style, personally. strlen just kinda feels a bit more meaning-precise. But I'm ok with it if the others are ok with it. 🙂 Wanna wait for someone else's input since I've never seen this particular style before and don't fully know the pros and cons of each.

🕷️

UncombedCoconut: What makes me happy is to avoid calling strlen every loop iteration. Unless the compiler realizes the result isn't changing and hoists it, it makes the loop O(n^2).
Of course, this "n" is small enough that this hasn't bothered anyone. It's the principle of the thing. :)

onpon4: True! :slightly_smiling_face: Personally I would just store it in a variable, but not sure if that's my Python habits leeching in. :sweat_smile:

:spider:

UncombedCoconut: In other words, I'd be fine saving the length and then looping, though that is more verbose and not a style choice I'd make when left to my own devices. (The original loop style is pleasantly clear and concise; the only problem is that strlen() is a search rather than a lookup in C.)

UncombedCoconut:
And meanwhile the solution to button_updateHotkey may simply be to assign it a different job.
This calls for withdrawing and splitting the change.s

Issue #1375, opened 2020-11-14
Translation: rethink how we avoid writing "License: Large Civilian Vessel License" in land_shipyard.c and map_system.c


UncombedCoconut:
I refer to this land_shipyard.c code (and its twin in map_system.c), which looks dodgy from an i18n perspective:
```C
   /* Remove the word " License".  It's redundant and makes the text overflow
      into another text box */  
   license_text = ship->license;
   if (license_text) {
      len = strlen(ship->license);
      if (strcmp(" License", ship->license + len - 8) == 0) {
         license_text = malloc(len - 7);
         assert(license_text);
         memcpy(license_text, ship->license, len - 8);
         license_text[len - 8] = '\0';
      }
   }
```
In theory, it could work if the update-po script detected license names and emitted both translatable strings and we _(license_text) after this operation. IMO, that's absurd.
But what instead? Tweak the UI? Add a suitable extra text field?

onpon4:
Yes, that is very very dodgy. There used to be even more stuff like that, like for example `"%d ship%s"` where `"%s"` was replaced with `""` or `"s"`. Probably we should replace that with just a "short name" text field or something.

🕵️

bobbens:
I see a few options:
1. Add a short name field to the outfits (optionally of course). However, this crufts stuff up so it's not best.
2. Change it so that it does not become a problem. Something as simple as Large Civilian Vessel Permit would work, although it could be possible to change how it is displayed instead.

onpon4:
Accepting use of the word "License" twice could be fine too.

🕵️

bobbens:
> Accepting use of the word "License" twice could be fine too.

Could also add a license to be able to buy licenses: `License: License license` ;)

UncombedCoconut:
OK, so the comment's concerns about "overflow into another text box" are averted in the system map. In the shipyard, it looks like the below (en, de). It looks like we must seize more horizontal space, and then any non-silly wording will suffice.

![Schiffswerft-lizenz-lizenz](https://user-images.githubusercontent.com/1224111/99209492-48336100-2791-11eb-8327-1b035894a141.png)
![shipyard-license-license](https://user-images.githubusercontent.com/1224111/99209494-48336100-2791-11eb-8742-7b36f70e888a.png)

UncombedCoconut:
Honestly, I'm feeling pretty unbothered by "License: _ License" now that there's more space for the text. Should we just close this?

bobbens:
Agreed. Closing.

Issue #1280, opened 2020-10-09
Short space for some translations


Wuzzy2:
The space for translators is very short for these strings (there is either a forced line-break too early or the string is just cut off):
- "Stop Speedup at:"
- "Volume Levels"
- "Available Outfits:"
- "Overwrite GUI"
- "Automatically save changes"
- "Exit only"

fcb73daf7f9f7f51b8024359b6551a3f0d3c66d9

Wuzzy2:
It seems that everything is fixed now. In German it looks fine now.

UncombedCoconut:
Lovely, thanks!

PR #1379, opened 2020-11-16
Spacing

UncombedCoconut:
Helps #1280, mostly solves #1375.

### DIFF
--- a/src/dev_sysedit.c
+++ b/src/dev_sysedit.c
@@ -191,7 +191,7 @@ void sysedit_open( StarSystem *sys )
    i = 1;
 
    /* Autosave toggle. */
-   window_addCheckbox( wid, -150, 25, 250, 20,
+   window_addCheckbox( wid, -150, 25, SCREEN_W/2 - 150, 20,
          "chkEditAutoSave", _("Automatically save changes"), uniedit_autosave, conf.devautosave );
 
    /* Scale. */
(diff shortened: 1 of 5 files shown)


UncombedCoconut:
Sigh, I'm getting careless preparing the "clean" changesets. Force-pushing with forgotten file.

onpon4:
This seems to have a whole bunch of stuff that I think is unrelated to the goal of the PR. What's happening there?

:spider:

UncombedCoconut:
WOW! Yeah, that's not right.

UncombedCoconut:
It is better, but it makes sense to treat this with some skepticism til I get some rest. :) In particular, I should double-check the right space-grab was committed.

UncombedCoconut:
Famous last words, but I think it's where it should be now.
The hard part is the shipyard. It looks like this (except without the white outlines around textboxes, which are an uncommitted layout debugging aid).
The video mode here is OPENGL_WINDOW_MIN_WIDTH x OPENGL_WINDOW_MIN_HEIGHT windowed.
At larger sizes, the image grid shows more ships. (The stats are still narrow enough to require line-wrapping in German, but it looks bearable.)

shipyard-960x720-en
Schiffswerft-960x720-de
Schiffswerft-960x720-de-2

onpon4:
I'd actually be in favor of increasing the size even more both to add some extra space for the headings (I notice that the German translation for "speed" is rather tightly packed there) and to add a buffer for the descriptions, and perhaps just a little more horizontal space for stats. (The wrapping there isn't a big deal but making it a tad less tightly packed could still be nice.) It'd mean only two columns of ships at 1024x768, but I think it'd be worth the gains in text  space.

In any case, what you've done so far looks to me like an improvement. 👍

🕵️

onpon4:
Any updates on this one? It's still marked as a draft.

🕵️

UncombedCoconut:
I haven't gotten to your suggestions at all. May make sense to merge this
now and iterate later.

Issue #1189, opened 2020-07-31
ngettext usage


onpon4:
Naev is currently very weak on ngettext usage, so I figured it would be a good idea to have an issue dedicated to it.

Currently there's only a few places ngettext is used, which can (depending on the language) make accurate translation impossible. This needs to be worked on. Basically, whenever you insert a number via code, ngettext should be used to pick the form (with a singular and plural form defined for English). This is done via the `ngettext(singular, plural, n)` function in C, or via the `gettext.ngettext(singular, plural, n)` function in Lua.

🕵️

UncombedCoconut:
Here is an approach that I find tempting, and something we can apparently get away with doing most of the time (for en+de), but I'm concerned about whether it's wrong: https://github.com/UncombedCoconut/naev/commit/fe97b7dae53b0c1f9819e99eb2f519c86de8842b.
Being new at this, I'm wondering if shortcuts are ever possible when a sentence includes "[N] credits", or if it's important to run the entire sentence through ngettext every time (and worth the potential burden on people like @Wuzzy2 who are contributing translations).
It's even more tempting to try this when a sentence contains both a number of credits and another plural form. (Example: "Destroy some pirates! You have killed %d and have earned %s credits. If finished, return to %s.")

UncombedCoconut:
Here's my checklist of code to look at before totally closing this issue. Arranged in groups of similar cases, in roughly descending order of importance.

```
src/board.c:373:              player_message(_("\apYou looted %d %s(s)"), nadded, ammo->name);
src/equipment.c:1575:         "%d / \a%c%s%d\a0 tonnes\n"
src/equipment.c:1576:         "%d units (%d jumps)\n"
src/info.c:223:         "%d (%d Jumps)"),
src/info.c:454:         "%d\n"
src/info.c:467:         "%d / %d tonnes\n"
src/info.c:468:         "%d / %d units (%d jumps)\n"
src/info.c:777:         nsnprintf(buf[i],128, "%s%s %d",
src/land_outfits.c:447:         _("%d\n"
src/land_shipyard.c:258:         "%d\n"
src/land_shipyard.c:272:         "%d units\n"
src/land_shipyard.c:273:         "%d units\n"
src/land_trade.c:147:           "%d tonnes\n"
src/land_trade.c:173:              _( "%d tonnes\n"
src/land_trade.c:176:                 "%d tonnes\n"
src/map_find.c:461:               _("%s (%d jumps, %.0fk distance)"),
[...]

PR #1381, opened 2020-11-16
Replace formatting _("%s credits") with a call to creditstring(). 

UncombedCoconut:
This covers the absolute easiest case of #1189, in which the entire translatable string is a number of credits. (When we want to refer to a number of credits within a sentence, it's not clear whether we can get away with using creditstring() or we'll have to translate the entire sentence using ngettext().)


onpon4:
Awesome, thanks for doing this! :+1:

Regarding in-sentence contexts, using creditstring() in combination with `%s` (as opposed to `%s credits`) should be fine I think. That's what I've been doing at any rate. I actually have an idea to go further and replace the "credits" text with a symbol to represent credits, like "¢" (a good choice I think since it's already a currency symbol).

🕵️

PR #1386, opened 2020-11-18
fix crash in sh00

nenau-again:
This `0` was there for 5 years without any reason.
A recent change on `nlua_pilot.c` made it crash. A nice lua error (or warning) would have been better. According to git blame, it could be linked to the commit https://github.com/naev/naev/commit/bc78f71ab94e12eb547247248ac61e3d38dfd91e by @nloewen , but I didn't take the time to be sure.


bobbens:
Even so, it should be done so it doesn't crash... Thanks!

PR #1387, opened 2020-11-18
Strangely, _putenv_s can be declared but not getenv_s. The simplest fix is just using getenv here.

UncombedCoconut:
(env_detect() directly above needs it anyway.)

### DIFF
--- a/src/env.c
+++ b/src/env.c
@@ -48,16 +48,9 @@ int nsetenv( const char *name, const char *value, int overwrite )
    return setenv( name, value, overwrite );
 #else /* HAVE_DECL_SETENV */
    if (!overwrite) {
-#if HAVE_DECL__PUTENV_S
-      size_t envsize = 0;
-      int errcode = getenv_s( &envsize, NULL, 0, name );
-      if (errcode || envsize)
-         return errcode;
-#else /* HAVE_DECL__PUTENV_S */
       const char *envval = getenv( name );
       if (envval != NULL)
          return 0;
-#endif /* HAVE_DECL__PUTENV_S */
    }
 #if HAVE_DECL__PUTENV_S
    return _putenv_s(name, value);


bobbens:
Wouldn't it be better to put a check for `getenv_s`?

UncombedCoconut:
Depending on your priorities, it could be, and I can put that change together. (Up to and including the implementation of an "ngetenv", I guess.)
Though I'll note that this mode (!overwrite) is only used in case HAS_UNIX is set, to default SDL_VIDEO_X11_WMCLASS. So the main benefit to changing anything is avoiding an implicit getenv_s declaration warning.

bobbens:
> Depending on your priorities, it could be, and I can put that change together. (Up to and including the implementation of an "ngetenv", I guess.)
> Though I'll note that this mode (!overwrite) is only used in case HAS_UNIX is set, to default SDL_VIDEO_X11_WMCLASS. So the main benefit to changing anything is avoiding an implicit getenv_s declaration warning.

That's a good point, we should probably not overwrite `LANGUAGE` too if the user set it.

bobbens:
I've added a check for `getenv_s` instead, just to be safe! 50343ca270c1ce0dff577a7c72f5df82ded9ee7e

UncombedCoconut:
> > Depending on your priorities, it could be, and I can put that change together. (Up to and including the implementation of an "ngetenv", I guess.)
> > Though I'll note that this mode (!overwrite) is only used in case HAS_UNIX is set, to default SDL_VIDEO_X11_WMCLASS. So the main benefit to changing anything is avoiding an implicit getenv_s declaration warning.
> 
> That's a good point, we should probably not overwrite `LANGUAGE` too if the user set it.

Hmm, it's probably OK to do that, as long as the user has picked the language in the options dialog and restarted Naev.

PR #1389, opened 2020-11-18
Replaced "credits" notation with "¢".

onpon4:
This notation change has a couple of benefits:

1. It uses less space.
2. It gives credits a currency symbol.
3. It means we don't need to worry about plural forms.
4. It looks cool.

This PR also includes a change from "tonnes" to "t" most places, for similar benefits.

Some screenshots of how it looks:

Screenshot at 2020-11-18 17-35-40
Screenshot at 2020-11-18 17-35-02
Screenshot at 2020-11-18 17-33-04
Screenshot at 2020-11-18 17-16-21
Screenshot at 2020-11-18 17-32-16


bobbens:
Two things:
1. I'm not sure on changing ton to t. We don't gain much space, and many people may not be used to it.
2. I'm not entirely sold on ¢. What about using the currency sign? https://en.wikipedia.org/wiki/Currency_sign_(typography)

onpon4:
Valid! So I'll change ¢ to ¤ and revert the "t" usage.

🕷️

onpon4:
Okey dokey, I've done those changes!

Screenshot at 2020-11-18 21-52-42

How's that, @bobbens? ¤ as the symbol instead of ¢ and back to showing "tonnes" etc.

onpon4:
> Might be worth adding API for this in Lua.

Are you talking about an API like `creditstring()` in numstring.lua?

🕷️

bobbens:
I think it should work fine now, thanks!

PR #1394, opened 2020-11-19
Consistently use credit formatting functions on the C side, too. Sync…

UncombedCoconut:
…and test relevant messages.
This is intended as a direct follow-up to https://github.com/naev/naev/pull/1389/
The key benefit is better localized knowledge of how we write amounts of credits. (Side benefit: a few communications gain the automatic text coloring for too-high prices.)

### DIFF
--- a/dat/scripts/numstring.lua
+++ b/dat/scripts/numstring.lua
@@ -23,8 +23,7 @@ end
 --    @return A string taking the form of "X ¤".
 --]]
 function creditstring( credits )
-   return gettext.ngettext( "%s ¤", "%s ¤", credits ):format(
-         numstring(credits) )
+   return _("%s ¤"):format( numstring(credits) )
 end
 
 
--- a/dat/scripts/numstring.lua
+++ b/dat/scripts/numstring.lua
@@ -23,7 +23,8 @@ end
 --    @return A string taking the form of "X ¤".
 --]]
 function creditstring( credits )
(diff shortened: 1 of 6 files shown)


dat/scripts/numstring.lua
```diff
@@ -23,8 +23,7 @@ end
 --    @return A string taking the form of "X ¤".
 --]]
 function creditstring( credits )
-   return gettext.ngettext( "%s ¤", "%s ¤", credits ):format(
-         numstring(credits) )
+   return _("%s ¤"):format( numstring(credits) )
```
onpon4, line 26: Is this desirable? My intuition was that using ngettext could theoretically be useful and doesn't hurt anything, though I could be wrong.

🕵️

UncombedCoconut: All right, I can back that bit out.
Sometime soon, I think we should get this and credits2str doing the same thing, so this may look different soon either way.
(Otherwise we get weird effects like "Pirate: Yar, pay me 23,456 ¤. .... Pay 23456 ¤?" when both functions are styling part of a message.)

bobbens:
Thanks!

dat/scripts/numstring.lua
onpon4: They don't need to do the same thing, but credits2str's output should be made translatable.

🕵️

PR #1395, opened 2020-11-19
Make the bottom info line of the commodity screen visible -- was under the next text box.

UncombedCoconut:
(While here, add forgotten _() calls.)


bobbens:
Thanks!

PR #1400, opened 2020-11-20
Fix mingw-bundledlls step: ...

UncombedCoconut:
must calculate SOURCEROOT,BUILDPATH before using them, mustn't "export MINGW_BUNDLEDLLS_SEARCH_PATH=(value)" (it's not a Bash array), must name every subdir of subprojects/ (the search is non-recursive).
With that done, my Python script can die.


bobbens:
python genocide? Thanks!

PR #1401, opened 2020-11-20
Added a space in "s" and "p" notation (for consistency).

onpon4:
Doing this as a PR because I know @bobbens expressed uncertainty about doing it this way. It is the correct way to how SI abbreviations in English, though.

🕵️


bobbens:
I think it should be fine. Thanks!

PR #1404, opened 2020-11-21
Fix steam deployment script

ProjectSynchro:
Looks like I missed a single comment :(

A manual trigger is needed for steam to deploy again


bobbens:
Thanks!

PR #1407, opened 2020-11-21
Add star systems to the detected translatable strings.

UncombedCoconut:
(Tweak update-po.sh so I can generate POTFILES* that are otherwise in sync with the ones that are checked in. For example, @ProjectSynchro recently took shaders_c_gen out of POTFILES.in for a reason, so let's not undo that.)


bobbens:
Thanks!

PR #1408, opened 2020-11-21
Fix where 2fa.txt is saved.. Whoops!

ProjectSynchro:
Last one I swear..

The following build failure is caused by the location in which the 2fa file was being saved:
https://github.com/naev/naev/actions/runs/375608131

This rectifies that, by saving the 2fa file in the same path as the 2fa script, which is where SteamDeploy.sh expects it to be.


bobbens:
Thanks!

PR #1412, opened 2020-11-21
Some nerfs to the overpowered Quicksilver.

onpon4:
The Quicksilver had a bit of a problem where it was overpowered enough that it may have sometimes made other ships seem pointless. Most notably, it was reasonably competent in offensive combat, which it shouldn't be, and had barely less cargo capacity potential than the Koala. This fixes that by doing two things: first, it makes the Improved Stabilizer a "one per ship" outfit, and second, it removes one weapon slot and one structural slot. We did some extensive testing and found it to still excel at its purpose (rush cargo). With an Improved Stabilizer and Engine Reroute, it's able to outrun most enemies and confidently shoot down the few that can catch up (typically only the fastest Hyenas).

🕵️


bobbens:
Sounds good. Maybe some of the other modifications should be made one-per ship to be more consistent?

PR #1413, opened 2020-11-21
1281 translatability

UncombedCoconut:
#1281 never ends :)
Well, actually, it's pretty close, as long as the new tickets don't count.
Hopefully all changes are noncontroversial, else we can discuss. (A few involve boundary drawing between the internal and translated world.)


bobbens:
Thanks!

PR #1414, opened 2020-11-21
1281 image array captions

UncombedCoconut:
We need to translate the captions on our image arrays!
This means that, one way or another, we need to remember what each item is.
This changeset gets it all working.

As a whole, it's an untidy little diff (and actually not so little), in a way that suggests some refactoring might make sense.
But I don't want to change the way the toolkit works (or even just what extra data these screens attach to the window) without talking about it first.

Part of #1281.


bobbens:
Thanks!

PR #1415, opened 2020-11-22
Hopefully fix the last of the issues with SteamCMD

ProjectSynchro:
It appears that steamcmd may require an absolute path to be passed in when referring to app build scripts.

This additionally add some verbose output when files are being copied,

A manual trigger is needed to get this to run.


bobbens:
Thanks!

PR #1416, opened 2020-11-22
Translate the intro/credits text.

UncombedCoconut:
The solution here is a little hair-raising. Suggestions welcome.
The overall strategy is:

- Keep the intro and AUTHORS file in their plain-text format. (Especially AUTHORS.)
- Emit a pot file that collects the lines of text. I'm not aware of a tool that does this, so I shelled out to awk, which does everything.
- Have intro.c cut the source text at line boundaries while collecting its "intro_lines", so that each line can be an input string to gettext() when we should do that. The C string crunching gets pretty subtle, so comment as nicely as I can...

I can't tell you how tempting it is to replace these sequences with Lua scripts (which call _) and some sort of thin player on the C side. But that would likely cause a lot of problems, too.

### DIFF
--- a/po/POTFILES_COMBINED.in
+++ b/po/POTFILES_COMBINED.in
@@ -2549,3 +2549,4 @@ dat/ssys/zintar.xml
 dat/ssys/zylex.xml
 dat/start.xml
 dat/tech.xml
+po/credits.pot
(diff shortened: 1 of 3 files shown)


bobbens:
Might be easier to just translate everything as one block, but this also works. Thanks!

PR #1417, opened 2020-11-22
Back out a bad change in ad445a9a5194105eace52302f413625badaf965b.

UncombedCoconut:
Whoops, I broke npc naming instead of fixing it.


bobbens:
Thanks!

PR #1418, opened 2020-11-22
fix wrong usage of inRangePilot

nenau-again:
In function to target nearest enemy, we used to check inRange in the wrong order


bobbens:
Thanks!

PR #1422, opened 2020-11-23
diy-nerds: Refactor planet method-calls, add lots of gettext.

UncombedCoconut:
Updated script survived 30 seconds' testing, so I assume it completely works. :)


bobbens:
That's why we'll have a new beta ;). Thanks!

PR #1426, opened 2020-11-23
Cleanups

UncombedCoconut:
Test case for the input box changes (explained in commit message): Info, Ship Log, New Entry.
Type "Dear Diary, I continued you in Germ" -- at least with my font settings, it scrolls the "r" in "Dear" out of view but the final "m" still doesn't fit and isn't drawn. That feels very confusing.
Now start backspacing. Starting with the second deletion, it looks like letters are disappearing off both the front and the back. They aren't actually, but we're scrolling the leftmost letters out of view instead of into view.


bobbens:
Looks good, thanks!

PR #1427, opened 2020-11-23
Translate fleet pilot names (and "Bounty Hunter", which is special).

### LINKED ISSUE #1419: Translatability: the various forms of "pilot name"
Split off from #1281: There is a data flow along the lines of (xml/script -> FleetPilot name || human -> player name ) -> pilot name -> displayed text.
We need to make sure the non-player pilot names are translated at an appropriate stage, and the player's own input is never fed through gettext.

Solution should not impair  #775.
Maybe this is too simple for its own issue, but I don't know that yet.

UncombedCoconut:
Closes #1419. I think. It'd better; it's bedtime. :)


bobbens:
Thanks!

PR #1428, opened 2020-11-23
Speed up CI Runs

ProjectSynchro:
I haven't really been the happiest with how fast CI runs take.. so I improved things.
I have reduced CI runtimes by approx 10mins, reduced the overall amount of build artefacts, and have re-added meson testing to the mix.

Obviously some more things can be added in the future but this is a start.


bobbens:
Thanks!

PR #1429, opened 2020-11-23
an other targetting fix

nenau-again:
This time in case of omni-visible hostiles. Distance used not to be computed.


bobbens:
Thanks!

PR #1433, opened 2020-11-24
fuzzy pilots are not identified as enemies

nenau-again:
The AI used to identify pilots as enemies even if they were fuzzily detected.
I'm pretty sure it's not the desired behaviour (fuzzy pilots are not red neither green in order to hide many information including their faction)
That doesn't change very much in-game, except that it's slightly easier to navigate unnoticed with a stealth ship.


bobbens:
Nice catch! Would be nice to have AI pilots also lose track of the player and have to estimate where they might be and such, but that's out of scope of this PR. Thanks!

PR #1437, opened 2020-11-25
Add option to pass in $REPO as an arg in manuallyTriggerWorkflow.sh

ProjectSynchro:
This allows for an easy way for forks to pass in their repo names.

naev/naev is the default value for $REPO, and is not required to be passed in.


onpon4:
Thanks! I was thinking this would be useful but my sh knowledge is lacking.

🕵️

PR #1439, opened 2020-11-25
[sightseeing] When the objective is accomplished and it's time to collect your pay from the starting planet, update the map marker.

UncombedCoconut:


bobbens:
Thanks!

PR #1440, opened 2020-11-25
Translation cleanups (in principle)

UncombedCoconut:
This actually drops a working translation in order to use ngettext(), but should otherwise be a win. Assumption: it's worth the work and the translations need maintenance anyway.

### DIFF
--- a/dat/scripts/numstring.lua
+++ b/dat/scripts/numstring.lua
@@ -49,11 +49,11 @@ end
 -- @brief Like tonnestring, but for abbreviations.
 --
 --    @param tonnes Number of tonnes.
---    @return A short string like "22t" describing the given mass.
+--    @return A short string like "22 t" describing the given mass.
 --]]
 function tonnestring_short( tonnes )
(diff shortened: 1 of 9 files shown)


dat/scripts/numstring.lua
```diff
@@ -45,6 +45,18 @@ function tonnestring( tonnes )
...
 
+--[[
+-- @brief Like tonnestring, but for abbreviations.
+--
+--    @param tonnes Number of tonnes.
+--    @return A short string like "22t" describing the given mass.
+--]]
+function tonnestring_short( tonnes )
+   -- Translator note: this form represents an abbreviation of "_ tonnes".
+   return gettext.ngettext( "%dt", "%dt", tonnes ):format( tonnes )
```
bobbens, line 56: Shouldn't it be `"%d t"`?

bobbens (commented):
Other than the small space it looks good.

UncombedCoconut:
The only reason it isn't, is that it previously wasn't. Let's do it!

bobbens:
Thanks!

PR #1441, opened 2020-11-25
Centralize handling of various mission-description bits…

UncombedCoconut:
…so we can uplift ngettext usage (or whatever else) as needed.

### DIFF
--- a/dat/missions/empire/es_cargo.lua
+++ b/dat/missions/empire/es_cargo.lua
@@ -93,7 +93,7 @@ function create()
    misn.setTitle( string.format(
       _("ES: Cargo transport to %s in %s (%s)"), destplanet:name(), destsys:name(), tonnestring(amount) ) )
    misn.markerAdd(destsys, "computer")
-   cargo_setDesc( misn_desc, cargo, amount, destplanet, timelimit, piracyrisk );
+   cargo_setDesc( misn_desc:format( destplanet:name(), destsys:name() ), cargo, amount, destplanet, timelimit, piracyrisk );
    misn.setReward( creditstring(reward) )
 end
 
--- a/dat/missions/empire/es_cargo.lua
+++ b/dat/missions/empire/es_cargo.lua
@@ -23,14 +23,6 @@ require "numstring.lua"
 
 misn_desc  = _("Official Empire cargo transport to %s in the %s system.")
 
-misn_details = _([[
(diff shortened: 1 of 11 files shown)


dat/missions/empire/es_cargo.lua
```diff
@@ -23,14 +23,6 @@ require "numstring.lua"
 
 misn_desc  = _("Official Empire cargo transport to %s in the %s system.")
 
-misn_details = _([[
-Cargo: %s (%s)
-Jumps: %d
-Travel distance: %d
-%s
-Time limit: %s
-]])
```
bobbens, line 32: Couldn't this pattern be collapsed into a single function instead of multiples? Something like

```cargo_setDesc( cargo, amount, target_system, deadline, notes )```

Where notes could have piracy risk and something like that? Would allow changing order and everything up easier.

bobbens (commented):
Still can be simplified a bit I think.

dat/missions/empire/es_cargo.lua
UncombedCoconut: I'll give it a try.
These users are all unique and beautiful snowflakes (and, irritatingly, they have a reason for it), so more than half of the arguments will have to be nullable (or the function would have to take an arg table).
I hadn't thought about standardizing the order as a reason to do it anyway; that's a fair point.
And it does get nicer if our cargo_setDesc can recalc the routing info. (Caller also uses that to set the reward.)

UncombedCoconut: I'm also going to squash, because the first commit had at least one bug.

UncombedCoconut:
This does one fairly naughty thing to deal with the Sirius ferry mission: tweaks a format string so we can call format in 2 stages.
I'm nearly indifferent between that and having all the other callers do "misn_desc:format( target:name(), target:system():name() )" themselves.

bobbens:
> This does one fairly naughty thing to deal with the Sirius ferry mission: tweaks a format string so we can call format in 2 stages.
> I'm nearly indifferent between that and having all the other callers do "misn_desc:format( target:name(), target:system():name() )" themselves.

Well it's not necessary to do a one-fits-all template, just have it be flexible enough to cover 90% cases, and the current approach can be used for the remaining cases.

bobbens:
Thanks!

PR #1442, opened 2020-11-26
Fix accidental partial rename (destPlanet->destplanet) in script.

UncombedCoconut:
I went with the original name for continuity's sake, though other scripts universally prefer "destplanet".


bobbens:
Thanks!

PR #1445, opened 2020-11-26
Pass Lua scripts to the interpreter without transformation

UncombedCoconut:
(i.e., without skipping the commented XML block). This makes scripts easier to debug when the jettisoned waste hits the fan.


bobbens:
Looks good, thanks!

PR #1448, opened 2020-11-27
Draft: robust storage of Lua data in save files.

UncombedCoconut:
This draft PR extends the saved game format so that we can save any Lua string without crashing, and successfully load it back.

The Lua language spec [permits arbitrary binary data](https://www.lua.org/pil/2.4.html).
The XML spec [forbids "control characters"](https://en.wikipedia.org/wiki/Valid_characters_in_XML) (0x00..0x1F, with the exceptions of \t, \n, and \r), even entity-coded. (XML 1.1 is better, but libxml2 doesn't support it, and..) Save files being utf8-encoded imposes further restrictions. So, our version of persistent Lua data is a leaky abstraction. If a script's Lua data happens to be unsupported, Naev will crash and lose your game while trying to persist it.
It's effectively impossible to make this limitation clear and predictable -- at best, one can try to remove the reasons anyone would construct such a string (e.g., by going after the '\a' escapes in text). Rather than hope that's sufficient, it makes sense to fix this programmer feature altogether, ideally when we're about to bump our SemVer.

This patch sticks to the current way of doing things whenever possible, but switches to an encoded version in parallel data fields when necessary. The encoder/decoder are kept static and extremely small in terms of code footprint. (But if we get close to merging this, we might want to adopt Base64 because it's standard.)


bobbens:
Didn't we have a base64 implementation at some time that was used to encode stuff? Don't see it anymore though...

UncombedCoconut:
OK, at this point the PR could stop being a "draft", but the question is when it's wise to merge it and whether other changes should come along for the ride. (For example, we could easily support "commodity" objects here, with similar implications for save-file compatibility.)

onpon4:
Save file compatibility probably isn't much of a concern right now. If stuff like this breaks save compatibility then now's as good a time as any probably. (That's just my perspective though.)

🕵️

src/base64.c
```diff
@@ -0,0 +1,273 @@
...
+#include <stdlib.h>
+#include <stdint.h>
+
+#include "nstring.h"
+
+
+/* encode table - base 64 alphabet as defined by the rfc */
+static const char cb64[64]="ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+/* generate decode table at compile time */
+#define B64(_)                   \
```
onpon4, line 20: The fact that `_` is being used for the argument on this is my only criticism. Not causing a problem I guess, it just makes me feel possibly unjustifiably nervous. :wink:

🕵️

onpon4 (approved):
Looks good to me. I'll let @bobbens have the final say though. 🙂

🕵️

bobbens:
Looks good, but could you take a look at the merge conflict? It's fairly messy. Thanks!

UncombedCoconut:
Should be all rainbows and sunshine now, though let's wait for the CI.

bobbens:
Thanks!

PR #1450, opened 2020-11-27
Assorted cleanups.

UncombedCoconut:


bobbens:
Thanks!

src/faction.c
```diff
@@ -452,22 +452,22 @@ void faction_addEnemy( int f, int o)
...
    if (!faction_isFaction(o)) { /* o is invalid */
-      WARN(_("addEnemy: %d is an invalid faction"), o);
+      WARN(_("Faction id '%d' is invalid."), o);
       return;
    }
 
    /* player cannot be made an enemy this way */
    if (f==FACTION_PLAYER) {
-      WARN(_("addEnemy: %d is the player faction"), f);
+      WARN(_("Faction id '%d' is invalid."), f);
```
onpon4, line 466: This should still be warning that it's the player faction, shouldn't it?

🕵️

onpon4 (commented):
Just to check, are those indications of what function the warnings are from redundant? (I think so but just checking.)

UncombedCoconut:
> Just to check, are those indications of what function the warnings are from redundant? (I think so but just checking.)

Yes, ERR and WARN (only) include __func__, which expands to the function name. I'm guessing they weren't redundant when originally written, but they are now.

src/faction.c
UncombedCoconut: Oh no, good catch, thanks! Fixing now.

bobbens:
Thanks!

PR #1451, opened 2020-11-28
Trust the libc snprintf() in more cases.

UncombedCoconut:
Cleanup "Remove remnants of a 'pack' feature" hitched a ride. :)

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -619,7 +619,7 @@ static void menuKeybinds_genList( unsigned int wid )
    /* Create the list. */
    str = malloc( sizeof( char * ) * input_numbinds );
    for ( j = 0; j < input_numbinds; j++ ) {
-      l = 64;
+      l = 68;
       str[j] = malloc(l);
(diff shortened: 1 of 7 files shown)


src/options.c
```diff
@@ -619,7 +619,7 @@ static void menuKeybinds_genList( unsigned int wid )
    /* Create the list. */
    str = malloc( sizeof( char * ) * input_numbinds );
    for ( j = 0; j < input_numbinds; j++ ) {
-      l = 64;
+      l = 68;
```
bobbens, line 622: Why 68?

UncombedCoconut: It's somewhat arbitrary. GCC deduces 68 because we have a format string "%s <%s%c>" where "char mod_text[64]" is one of the "%s" args. (That plus brackets plus %c + null gets to 68.)
In practice, we wouldn't expect the message to get that long, and a truncation is only mildly annoying.
Could just double the 64 to 128.

bobbens:
Thanks!

PR #1459, opened 2020-11-29
Added use of some unused graphics for "previous generation" systems

onpon4:
These seem like good fits and they weren't being used anyway.

🕷️


bobbens:
Thanks!

PR #1461, opened 2020-11-29
Added Rikaina (Thurion military station).

onpon4:
This splits off some of the more advanced Thurion tech into a military station, mainly the bigger ships and the Nebular Shielding Prototype. The primary purpose of this is to make sure you can't make it to Proteron space  too soon.

(This also includes a whole bunch of stuff with virtual assets moved around, that's just the editor's doing.)

🕵️


bobbens:
Not getting to proteron space soon is a good idea. Thanks!

PR #1462, opened 2020-11-29
use raw goto when one has lost contact with the target

nenau-again:
Because there is no need to be very precise, neither to brake


bobbens:
Thanks!

Issue #1467, opened 2020-11-29
Opening the overlay sometimes doesn't work.


onpon4:
This is something I mentioned on the Discord server, but it's probably worthwhile opening it here, too. There are times when, usually only while address sanitizer is enabled, opening the overlay with the Tab key (or whatever is configured to open it) doesn't work. Instead the game pauses for a brief moment, and nothing happens.

git bisect showed the problem to be caused by 527559d4d88d6e683a82c6e13bc93cefcb2406cf, but a change in 85f06bbdb1e170a2941285e6077f8783af702b0d may be related as well. Given the pause, it's probably connected to the for loop introduced in 85f06bbdb1e170a2941285e6077f8783af702b0d. This was the git bisect output, in any case:

```
527559d4d88d6e683a82c6e13bc93cefcb2406cf is the first bad commit
commit 527559d4d88d6e683a82c6e13bc93cefcb2406cf
Author: Edgar Simo-Serra <ess@waseda.jp>
Date:   Sun Nov 29 20:51:58 2020 +0900

    Dynamically compute the radius of jump points and planets.

:040000 040000 96851a03c41c0f63f5f80eb88dbc48e022cedbde e8f93159be458d5eaff429d1c5d9b9c3078e14cb M    src
```

To reproduce, open this save file (after extracting it, compressed so that GitHub accepts it as an attachment):

[Natasha.ns.zip](https://github.com/naev/naev/files/5612597/Natasha.ns.zip)

Then compile Naev with address sanitizer enabled. Do this in meson with:

```
meson setup builddir .
cd builddir
meson configure -Db_sanitize=address
meson compile
```

Then run, open the save, take off, and press the overlay key. Rather than seeing the overlay coming to view, you will experience a brief pause and the overlay will not come into view.

onpon4:
Fixed by #1468. It was caused by the time a loop takes making the game thinking I wanted to only temporarily open the overlay.

PR #1468, opened 2020-11-29
HACK: Only set the ovr_opened time after doing the first round of positioning work.

UncombedCoconut:


onpon4 (approved):
Looks good on my end! I think this will fix #1467, will close that bug if so.

🕷️

PR #1470, opened 2020-11-29
We need locale.h all the time, not just for Solaris…

UncombedCoconut:
because we use LC_MESSAGES. It just so happens that libintl.h includes it for us, but that's just something it happens to do conditioned on __OPTIMIZE__, not a guaranteed behavior.


onpon4:
Thank you~!

🕷️

PR #1477, opened 2020-11-30
Uniformize the treatment of planets and jump points in ovr_refresh.

UncombedCoconut:
Operate on the list of discovered assets. Call ovr_refresh again after discovering stuff.


bobbens:
Looks good, thanks! Makes it easier to add other constraints.

PR #1478, opened 2020-11-30
Fix another size mismatch, conceptually separate NEBULA_ from NDATA_ name lengths.

UncombedCoconut:


bobbens:
On places like windows where paths get very long could it surpass 256? Do we even support unicode paths? :/

UncombedCoconut:
> On places like windows where paths get very long could it surpass 256? Do we even support unicode paths? :/

This length is only for the names formed from "nebu_bg_%dx%d_%02d.png". Wound NEBULA_FILENAME_MAX be clearer, maybe?

bobbens:
> > On places like windows where paths get very long could it surpass 256? Do we even support unicode paths? :/
> 
> This length is only for the names formed from "nebu_bg_%dx%d_%02d.png". Wound NEBULA_FILENAME_MAX be clearer, maybe?

Ah, that would be better.

UncombedCoconut:
> On places like windows where paths get very long could it surpass 256? Do we even support unicode paths? :/

**Outside** the context of nebula.c: the main case is ndata. We are quietly dead if relative ndata paths (the part after /usr/share/naev or wherever) get that long, because our dists are tar. GNU tar format has a path limit of 99 bytes. If that's a problem, there's a setting to switch to a format called "ustar"with a path limit of 255 bytes.

"Support unicode paths" is complicated if OS's make it complicated. We mainly use POSIX APIs -- file paths are byte strings, a '/' byte is a separator, if the user's programs treat them as UTF-8 then great. Windows provides "Unicode" APIs using wide chars (LPCWSTR, UCS2 encoding) and we don't care.

UncombedCoconut:
Yuck, the diff is complicated for no reason, fixing.

bobbens:
Thanks!

PR #1479, opened 2020-11-30
More compile modes, more warning fixes.

UncombedCoconut:
- Address sanitizer
- -Wstringop-truncation (does not notice we null-terminate right after strncpy, and it's kinda got a point.)


bobbens:
Thanks!

Issue #1490, opened 2020-12-01
Warning on cargo_common.lua


Mutos:
Hello all,

I got a recurring warning on cargo_common.lua whenever I run NAEV (commit 5ce7b3707, tested on Ubuntu 20.04, not tested yet on Windows 7/10).

Trace is the following : 

```
Warning: [misn_runFunc] Mission 'Cargo Rush' -> 'create': [string "cargo_common.lua"]:201: bad argument #1 to 'format' (string expected, got userdata)
stack traceback:
	[C]: in function 'format'
	[string "cargo_common.lua"]:201: in function 'cargo_setDesc'
	[string "missions/neutral/cargo_rush.lua"]:127: in function <[string "missions/neutral/cargo_rush.lua"]:77>
```

I think it might be quite simple, but I fear not to have time to search and solve it. I may try but no guarantee of success...

@+

Benoît 'Mutos' Robin

Mutos:
Proposed fix in PR #1491

bobbens:
This seems very critical. Only affects Lua and not LuaJIT apparently though. Might warrant another beta...

PR #1491, opened 2020-12-01
Fix cargo common

Mutos:
Solves warning in cargo_common.lua (Issue #1490).
It was simply a matter of adding a tostring() to the argument to format().


bobbens:
Thanks!

PR #1493, opened 2020-12-01
Go clean on warnings (more build configurations, and in map_overlay.c since it's settled down a bit).

UncombedCoconut:
Changes are non-functional outside nlua.c and shiplog.c. All are tested.
The update to meson.build should make it harder for crap to sneak in going forward. (I think Autotools builds are still a bit stricter; we'll have to synchronize at some point.)


bobbens:
Thanks!

PR #1494, opened 2020-12-02
little fixes

UncombedCoconut:
Fortunately, not finding real lua-not-JIT issues, but spotted some other stuff. :)
I'm guessing our big-endian support isn't entirely real anyway, but I have to try. (My first Linux machine was a PowerPC Mac and I guess I'm sentimental. My first experience with endianness bugs might have been jumpnbump blasting my ears out instead of playing the audio it meant to.)


bobbens:
At some point in time people were playing naev 0.3.5 on powerpc macs without an issue ;). Thanks!

PR #1495, opened 2020-12-02
Korean translation

paparad0x:
Hello, we are Korean university students who asked about contribution last time. We've translated 10 percent of whole sentences.


bobbens:
Thanks. We need a Korean font to display characters. Do you have any recommendation for an open source Korean font similar to what we use (Cabin-SemiBold)?

PR #1497, opened 2020-12-02
There are two hard things in computer science…

UncombedCoconut:
… cache invalidation, naming things, and off-by-one errors.
Assert the correct byte is null. When doing a memmove *forward*, shrink the range to exclude the final byte.
(Since we memmove the entire rest of the buffer, and do not zero out backspaced bytes, this is necessary.)

Also, prevent underflow when we've ctrl-backspaced and we're adjusting the view position. We could be erasing everything from position 0 to the cursor (which is ahead of "view", which is the left side of the text box).


bobbens:
Thanks! ;)

PR #1498, opened 2020-12-02
Removed SOUND_SPEED_PLAY_LIMIT.

onpon4:
This thing seems to be wholly unnecessary and makes the game seem buggy. I did some testing following Dvaered and FLF pilots in Frontier space, and hearing a bit of battle sounds at high speed wasn't a problem at all. (In fact, OpenAL makes sounds quieter as they get faster, so under OpenAL it ended up being silent at the real high time compression levels I was experiencing. But even under SDL_Mixer, where that wasn't happening, it just wasn't a problem.)

🕵️

### DIFF
--- a/src/sound.h
+++ b/src/sound.h
@@ -8,9 +8,6 @@
 #  define SOUND_H
 
 
-#define SOUND_SPEED_PLAY_LIMIT      4.0 /**< Speed modifier at which sounds do not play. */
-
-
 extern int sound_disabled;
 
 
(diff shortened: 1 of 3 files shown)


bobbens:
When I tried it a long time ago in a galaxy far far away, they would just be weird beeps at really high speeds. Is this not an issue with say autonav where you can get 50x compression? I would still set it to something like 4.0 so that when autonav kicks in the player can chill to just the music.

bobbens:
Looking at the code, OpenAL actually does handle the speed and does stuff, while SDL_mixer doesn't. I wouldn't feel bad dropping support for SDL_mixer which would both simplify code and make us be more consistent in general.

PR #1499, opened 2020-12-03
Korean font added.

paparad0x:
Added 2 Korean fonts, D2CodingBold is monospace font. Also updated FONT_LICENSE.yaml file.


bobbens:
Thanks, I have also pushed an updated `po/ko.po` that uses the new fonts. You should be able to load the game and see the text in Korean now.

PR #1501, opened 2020-12-04
More thorough rework of the input widget. Main benefit is that it can't position the cursor inside multibyte characters...

UncombedCoconut:
... even if some maniacal tester enters them in the dev-mode multi-line text boxes and uses the up/down arrow keys.
(But if char widths vary enough across lines, the new approach should look better.)
The logic should generally be a little more foolproof/hackable and a little less efficient when clamping and moving up/down lines.

I don't expect 0.8 to need this, because the game itself uses one-line inputs and those basically work fine already.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -497,6 +497,22 @@ static size_t font_limitSize( glFontStash *ft_font, int *width,
  *    @param width Width to match.
  *    @return Number of characters that fit.
  */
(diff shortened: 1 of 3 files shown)


bobbens:
Thanks! It's scary, but I don't want to maintain separate branches if possible ;)

PR #1502, opened 2020-12-04
Make sure a bitfield test results in 0/1 as documented.

### LINKED ISSUE #1500: Auto-assigning of secondary weapons doesn't work properly
This is something I noticed some time ago but forgot to write an issue about. So I decided to look into it, and I couldn't figure out at a glance what causes this, so opening this now (since it's very important) and will probably do a git bisect later.

For some reason, automatic assignment of secondary weapons is broken. You can see this in the weapons screen, and as @nenau-again pointed out on the Discord server, this has also led to all AI ships not being able to use any weapon considered "secondary" (such as homing missiles and ion cannons).

What happens on the weapons screen is, primary weapons get assigned automatically properly, but secondary weapons instead become in a weird limbo state where they're unassigned, but the game also seems to think of them as assigned when determining whether or not the weapon set is active. This means if you want to use secondary weapons, you must configure your weapons manually (manual configuration works just fine).

Since the bug also occurs with AI ships, that would suggest that it has something to do with `pilot_weapSetAdd()` in pilot_weapon.c. That seems to have not changed in years, though, so it could be something within that, or maybe it could be a bug with the "level" argument somewhere. A git bisect should reveal this; I'll do one at some point when I have the spoons for it if no one else does. 27ce89666addbec90526c5b5700dbb8d7b2470e3 should be a good commit to start from.

🕵️

UncombedCoconut:
Looks like it fixes #1500.

### DIFF
--- a/src/outfit.c
+++ b/src/outfit.c
@@ -536,7 +536,7 @@ int outfit_isGUI( const Outfit* o )
  */
 int outfit_isSecondary( const Outfit* o )
 {
-   return (o->properties & OUTFIT_PROP_WEAP_SECONDARY);
+   return (o->properties & OUTFIT_PROP_WEAP_SECONDARY) != 0;
 }
 
 


bobbens:
Thanks!

PR #1503, opened 2020-12-04
Limit SystemPresence initialization to one code path and init everything…

UncombedCoconut:
… (including .disabled).


bobbens:
Thanks!

PR #1505, opened 2020-12-04
"Fix" econsim. Though maybe it should go away instead.

UncombedCoconut:


utils/econsim/econsim.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
```
onpon4, line 1: I've seen another file do this but that shebang is wrong, "python" canonically always means the same as "python2" and "python3" should be used if it's a Python 3 script. (Any distro making "python" point to python3 is going against convention.)

🕷️

🕷️

UncombedCoconut: This shouldn't be true anymore except in outdated or enterprise / long-term-support distros. The end-of-life date for Python 2 was January 1, 2020.
It should be assumed that this is going to be awful PEP-394 doesn't help. :) Unless the user creates a venv, in which case it recommends the shebang I used here.
I could just not write f-strings or print(thing, another_thing) -- and then it doesn't matter. (The .items()/.iteritems() difference is not a big deal given the size of the map.)

onpon4: It's still true in Debian. That's what I'm using. 😉 And even if that was totally over eventually we'll have python4 and have to deal with this roller-coaster again, but worse.

🕷️

PR #1506, opened 2020-12-05
Have map_renderPath draw in varying thickness. Warning: needs review.

UncombedCoconut:
(because I'm not actually competent with OpenGL)

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -62,7 +62,6 @@ static int listMapModeVisible = 0; /**< Whether the map mode list widget is visi
 static double commod_av_gal_price = 0; /**< Average price across the galaxy. */
 /* VBO. */
 static gl_vbo *map_vbo = NULL; /**< Map VBO. */
-static gl_vbo *map_vbo_2 = NULL; /**< Map VBO. */
 static gl_vbo *marker_vbo = NULL;
 
 /*
@@ -116,8 +115,7 @@ int map_init (void)
    GLfloat vertex[6];
 
    /* Create the VBO. */
-   map_vbo = gl_vboCreateStream( sizeof(GLfloat) * 3*(2+4), NULL );
-   map_vbo_2 = gl_vboCreateStream( sizeof(GLfloat) * 6*(2+4), NULL );
+   map_vbo = gl_vboCreateStream( sizeof(GLfloat) * 6*(2+4), NULL );
 
    vertex[0] = 1;
    vertex[1] = 0;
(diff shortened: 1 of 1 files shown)


onpon4 (approved):
I like! 😄 (Can't comment on / review how good the code is though, I know next to nothing about OpenGL)

🕷️

src/map.c
```diff
@@ -1113,63 +1113,58 @@ void map_renderSystems( double bx, double by, double x, double y,
...
-         vertex[13] = a / 2. + .5;
-         vertex[14] = col->r;
-         vertex[15] = col->g;
-         vertex[16] = col->b;
-         vertex[17] = a / 4. + .25;
-         gl_vboSubData( map_vbo, 0, sizeof(GLfloat) * 3*(2+4), vertex );
+	 for (k=0; k<3*2; k++) {
+            h0 = 1 - .5*(k/2);  /* Fraction of the way toward (x0, y0) */
+            h1 = .5*(k/2);      /* Fraction of the way toward (x1, y1) */
+            sign = k%2 - 1;     /* Alternating +/- */
```
bobbens, line 1149: this is not actually a sign right? It's -1 o 0

src/map.c
```diff
@@ -1113,63 +1113,58 @@ void map_renderSystems( double bx, double by, double x, double y,
...
          else
             col = &cYellow;
+         x0 = x + sys0->pos.x * map_zoom;
+         y0 = y + sys0->pos.y * map_zoom;
+         x1 = x + sys1->pos.x * map_zoom;
+         y1 = y + sys1->pos.y * map_zoom;
+         w0 = w1 = map_zoom / hypot( x0-x1, y0-y1 );
+	 w0 *= jcur >= 1 ? 4 : 2;
          jcur--;
+	 w1 *= jcur >= 1 ? 4 : 2;
```
bobbens, line 1143: whitespace problem?

UncombedCoconut: vim strikes again! I am going to engage in some revisionist history, because several commits contain tabs that shouldn't be tabs.

UncombedCoconut:
Fixing the "sign = ..." makes this twice as thick as it was when first opened. This may make the hops with sufficient fuel too thick for one's tastes. I'm not really sure which I prefer; maybe the current way but clamped?

bobbens:
Divide by 2.? :D

UncombedCoconut:
> Divide by 2.? :D

After playing around, I think I like the current values but with line thickness scaled by MIN( map_zoom, 1.5 ) instead of map_zoom.
However, if you liked original / half of current better, that works too. (Did you mean that about the whole path or just the "fat" part? I really liked the "skinny" part being thicker than before the PR.)

bobbens:
Looks good, thanks!

PR #1507, opened 2020-12-05
Removed puke-green coloring of imagearray selections.

onpon4:
Looks much better, and seems redundant; the white outline is perfectly sufficient and allows the background color to still be seen.

Screenshot at 2020-12-04 19-56-05

🕷️


bobbens:
I don't really like this. Keeping the dark green makes it consistent with the lists and also much easier to tell than the simple outline.

PR #1508, opened 2020-12-05
Changed the dark blue to midnight blue.

onpon4:
Screenshot at 2020-12-04 21-33-44


bobbens:
Looks better than before, thanks!

PR #1510, opened 2020-12-05
Modified colors for indicating pilots, planets, etc.

onpon4:
This replaces the "friend", "hostile", "neutral", "inert", and "restricted" colors with colors from the muted colors set. It's not perfect, but combined with the symbols on planets and in other similar places, it seems to be an improvement. Colors were chosen to maximize contrast between neutral, hostile, and friendly (since distinguishing pilots' coloring can't be replaced by any sort of symbol).

🕷️


bobbens:
Looks good, thanks!

PR #1511, opened 2020-12-05
Attempt to solve my Autotools problems using violence.

UncombedCoconut:


utils/gettext.sh
```diff
@@ -1,12 +1,19 @@
 #!/bin/bash
 
+set -ex
```
bobbens, line 3: Is this necessary?

UncombedCoconut:
The "x" isn't. Is the "e" fine? Without it, distcheck was hiding a failure.

bobbens:
Thanks!

Issue #1512, opened 2020-12-05
OS shows as 'Unknown OS' in the Options menu when building with meson


ProjectSynchro:
This is due to the fact that the flag the code is looking for is not being defined.
In options.c (line 313-323) the following code checks for those flags. (or whatever term it's called)
```c
#if defined(LINUX)
         "Linux\n"
#elif defined(FREEBSD)
         "FreeBSD\n"
#elif defined(MACOS)
         "macOS\n"
#elif defined(WIN32)
         "Windows\n"
#else
         "Unknown OS\n"
#endif
```
In meson these will need to be defined.

bobbens:
This should be fixed ;)

PR #1513, opened 2020-12-05
Add checks for delaring which OS we compiled with in meson

ProjectSynchro:
Solves #1512

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -169,7 +169,7 @@ if buildExec.disabled() == false
    config_data.set10('HAVE_OV_READ_FILTER', cc.has_header_symbol('vorbis/vorbisfile.h', 'ov_read_filter'))
    config_data.set10('HAVE_STDALIGN_H', cc.has_header('stdalign.h'))
 
-   # OS declaration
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -169,6 +169,17 @@ if buildExec.disabled() == false
...
 
+   # OS declaration (Shows which OS we built this with in the options menu)
+   if host_machine.system() == 'linux'
+      config_data.set10('LINUX', true)
+   elif host_machine.system() == 'freebsd'
+      config_data.set10('FREEBSD', true)
+   elif host_machine.system() == 'darwin'
+      config_data.set10('MACOS', true)
+   elif host_machine.system() == 'windows'
+      config_data.set10('WIN32', true)
```
bobbens, line 180: Theoretically it's 64 bit right? ;)

bobbens:
Thanks!

meson.build
ProjectSynchro: very true ;P I was thinking of adding more detail to options.c (like arch or something)

PR #1516, opened 2020-12-05
Fix workflow run failure

ProjectSynchro:
It looks like we were getting lucky by having some other package install python in MSYS2.

This PR simply adds the python package to all release workflows.


bobbens:
Thanks!

PR #1520, opened 2020-12-05
Fix format-security warning on Windows code path.

UncombedCoconut:
(All those users whose install paths contain "%s" can breathe a sigh of relief.)


bobbens:
Thanks!

PR #1522, opened 2020-12-06
a try at making PRINTF_FORMAT less error-prone

UncombedCoconut:
The last PR left a weird situation where PRINTF_FORMAT( fmt_arg_index, va_list_index ) could go wrong if you included "attributes.h" before <stdio.h>.
I never found a perfect answer, but this seems OK: (1) put it with nsnprintf in in nstring.h, (2) nstring.h includes stdio.h, (3) we stop having Autotools builds pre-include "attributes.h" because I don't think it makes sense. (This was never set up for Meson.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -347,7 +347,7 @@ case "$host" in
 esac
 
 # global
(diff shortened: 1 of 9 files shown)


bobbens:
Thanks!

PR #1524, opened 2020-12-06
IDEA: Add unit tests run by initialized Naev binary.

UncombedCoconut:
This is a trivial example of what such a mode can do.
Less trivially we could consider running the extras/autotests.lua script and checking some XML stuff.
(Why XML stuff? I'd like to silence the warnings when using libxml2+Windows+PRIu64, but also to prove that's not a deadly mistake.)
Open question: should any of this stuff get cut out of release builds, like the other dev modes do? (Unit testing a release build sounds like a good idea.)
Open question: even if we do this, is this the structure we want or should tests be off to the side somewhere?


bobbens:
One thing I would consider would be to directly have a different binary for unit testing. Something like `./naevtest` which would use the same object files, just a completely different `main(...)` that would run tests, since I don't see any reason to actually have this functionality baked into the base game. I think the main reason for using it would likely to be adding more tests to CI. Furthermore, this would also allow testing release and debugging mode separately depending on what flags it is compiled with.

UncombedCoconut:
That would also let us figure out how fully we really care about initializing Naev, and bypass user configs. Sharing object files also sounds good -- no conditional compilation, just let the linker strip out unused `unittest_` functions from Naev if it's able.
So I'm sold.
I suppose this would also make the "Reached main menu" test a total side issue. Converting it to a unit test under a separate main() doesn't make sense. A separate change making it a CI step (to catch bad PRs but not bother distro pacakgers), or outright removing it, might.

UncombedCoconut:
I hope to come back to this after reassessing. In the meantime, this doesn't need to be in the queue.

PR #1527, opened 2020-12-06
Fix issues compiling using autotools in older GCC versions

ProjectSynchro:
Some older gcc versions will default to c99, and since there's some code that's gnu11 it will fail to build.

This allows builds with the steam runtime without passing any std flags in CFLAGS..


bobbens:
Thanks!

PR #1528, opened 2020-12-07
"Fix" libxml2/MinGW warnings with Dvaered engineering. (After testing, I don't think they're bugs, and they're noisy enough to hide actual bugs.)

UncombedCoconut:
The caveat: that "testing" was under Wine with the DLL coming from my system's mingw-w64-libxml2 package.
Shouldn't matter, but might?
If we add asprintf() later, we can replace the xmlTextWriterWriteFormat* calls with a smarter macro and sidestep all of this crap. That sounds like 0.9 work to me.

### DIFF
--- a/src/nxml.h
+++ b/src/nxml.h
@@ -9,6 +9,17 @@
 
 #include <errno.h>
 
+#ifdef __MINGW64_VERSION_MAJOR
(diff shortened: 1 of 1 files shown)


bobbens:
OH GAWD WHAT IS GOING ON?!?!?! (Thanks)

PR #1529, opened 2020-12-07
Dvaered engineering: we can solve a "TODO get rid of this", and…

UncombedCoconut:
…legit warnings on that dumb OS where sizeof(wchar_t)==2, just by deleting unused functions. Meanwhile, document what the upstream library is (and what the author recommends in its place nowadays).


bobbens:
Fairly scary. Hope the code gods shine on us.

PR #1530, opened 2020-12-07
Only handle imagearray double-clicks as such if the same image is clicked twice.

UncombedCoconut:
(Actually, as a convenient substitute just check if the second click is to the selected image.)
Fixes weirdness reported by Emmy in the bar.

### DIFF
--- a/src/tk/widget/imagearray.c
+++ b/src/tk/widget/imagearray.c
@@ -58,6 +58,8 @@ static void iar_cleanup( Widget* iar );
  *    @param caption Caption array to use (freed).
  *    @param nelem Elements in tex and caption.
  *    @param call Callback when modified.
+ *    @param rmcall Callback when right-clicked.
+ *    @param dblcall Callback when selection is double-clicked.
  */
 void window_addImageArray( const unsigned int wid,
                            const int x, const int y, /* position */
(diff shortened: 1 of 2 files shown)


src/tk/widget/imagearray.c
```diff
@@ -421,22 +423,27 @@ static int iar_mclick( Widget* iar, int button, int x, int y )
...
 {
    /* Handle different mouse clicks. */
+   iar_setAltTextPos( iar, x, y );
    switch (button) {
       case SDL_BUTTON_LEFT:
-         iar_focus( iar, x, y );
-         if (iar->dat.iar.dblptr != NULL)
+         if (iar->dat.iar.dblptr != NULL
+             && iar->dat.iar.selected >= 0
+             && iar->dat.iar.selected == iar_focusImage( iar, x, y )) {
```
bobbens, line 438: wouldn't it make more sense to also add a motion check in the toolkit itself so if the double click is done with two clicks far away from each other it isn't a double click?

bobbens:
Thanks!

UncombedCoconut:
Hmm, maybe.
The widget is still going to have to do some filtering (e.g., a
double-click on negative space shouldn't activate the selection).
My bias is toward the to being a dumb router, and the widget code that
actually knows the contents being the brains behind the decision.
This colors my intuition, e.g. that there's a correct way to handle rapid
clicks on opposite sides of a cell border and it's to treat them as
separate single-clicks. I could see an argument that this is over-speccing
literal edge-case behavior, though.

Issue #1532, opened 2020-12-07
Game crashes if you switch from a ship carrying cargo to a ship with stowed cargo.


UncombedCoconut:
Repro: load this little fellow, go to the equipment screen, and double-click the pirate Kestrel.

[True pirate.ns.gz](https://github.com/naev/naev/files/5650772/True.pirate.ns.gz)

The save data says
```
<ship name="P-KISTAL" model="Pirate Kestrel>
   ...
   <commodities>
      <commodity quantity="35">Food</commodity>
      <commodity quantity="23">Diamond</commodity>
   </commodities>
   ...
</ship>
```

If that's a legitimate situation, the pilot_cargoMove code needs to handle it correctly.
(If it's not, that code still needs to change.)
It tries to make space for new entries in the dest->commodities array, but clobbers the originals rather than appending.
But also, can the source and destination ships have overlaps in the commodity types? Do we have to dedup? Should we simply call a helper method to add each cargo to the destination?

bobbens:
It shouldn't crash, but the way the game is, you should not have cargo in any ship other than the active ship. This prevents "abuse" like basically storing a shitload of resources on stored ships, going to a planet in a fighting ship, and selling it all. To answer your questions:

0. I would print a warning, and try to merge (or just clobber). This could also be sanitized at load.
1. Source and destination ship can't overlap, since destination can't have cargo.
2. No need to dedup since it should be illegal.

UncombedCoconut:
Wild. I pirated this save file from @asyasb (who reported some separate
issue once and posted a save file which added useful diversity to what I
had). I'd goofed around since then, but the original probably had the same
loaded Kestrel and we can ask about it.
Either way, if the save format says it can happen, it can happen. :)

UncombedCoconut:
Just capturing what came up on Discord: the block on leaving cargo on a ship is on this path:
GUI -> equipment_changeShip -> land_errDialogue( shipname, "swapEquipment" ) -> can_swapEquipment
There's a similar block on the shipyard trade-in path. (The "swapEquipment" and "swap" checks are near duplicates and should maybe be merged?)

There's a gap: Lua scripts can call player.swapShip directly. Example: missions/pirate/pir_ship_stealing.lua. That script's logic could be revisited; the author was already questioning it.
But we basically need to decide
(1) Is untransferred cargo allowed then, or does it have to get jettisoned (with mission-cargo handling as needed)?
(2) Should a call to player.swapShip() be allowed to fail?

[...]

PR #1535, opened 2020-12-08
Stop the bleeding from #1532 (memory bug in pilot_cargoMove).

UncombedCoconut:
The macro problem of cargo staying with your old ship when the plot has you switching ships is still here, but we now correctly move/consolidate cargo if it fits and otherwise leave *all* of it in place.


bobbens:
Thanks!

PR #1540, opened 2020-12-09
When we build Luajit as a Meson subproject, do a native Meson build.

### LINKED ISSUE #1525: Finish luajit submodule port to meson
Currently this submodule wraps make to build things.. and as the comments say, we should probably finish writing enough so meson can build luajit natively..

UncombedCoconut:
Fixes #1525.
No effect on official builds, which use Autotools.


bobbens:
Thanks!

PR #1541, opened 2020-12-10
Fix the fix to radar coordinate translation. We need scaling but also…

UncombedCoconut:
… compensation for the GUI viewport offset.

This may actually still be wrong; the answers for Legacy (with its round radar) don't make sense and seem like they could. But things are less wrong.


bobbens:
Thanks!

PR #1544, opened 2020-12-12
Fix some totally trivial issues.

UncombedCoconut:
One -Wstringop-truncation (it's got a point), one SEGV report where Occam's razor says the user hit a weapset-fire hotkey at the main menu. That does indeed crash.


bobbens:
Thanks!

PR #1555, opened 2020-12-13
Button tweaks

djvs:


src/colour.c
```diff
@@ -31,7 +31,9 @@ const glColour cGrey80     = { .r=0.80, .g=0.80, .b=0.80, .a=1. }; /**< Grey 80%
 const glColour cGrey70     = { .r=0.70, .g=0.70, .b=0.70, .a=1. }; /**< Grey 70% */
 const glColour cGrey60     = { .r=0.60, .g=0.60, .b=0.60, .a=1. }; /**< Grey 60% */
 const glColour cGrey50     = { .r=0.50, .g=0.50, .b=0.50, .a=1. }; /**< Grey 50% */
+const glColour cGrey45     = { .r=0.45, .g=0.45, .b=0.45, .a=1. }; /**< Grey 45% */
```
bobbens, line 34: I'm not a big fan of adding so many colors. In fact, I think we should reduce them a lot more. Are these new colors absolutely necessary?

src/toolkit.c
```diff
@@ -1325,15 +1325,15 @@ void window_render( Window *w )
...
+   if (w->focus != -1){
       wgt = toolkit_getFocus( w );
       if ((wgt == NULL) || wgt_isFlag(wgt, WGT_FLAG_KILL))
          return;
       x  += wgt->x;
       y  += wgt->y;
       wid = wgt->w;
       hei = wgt->h;
-      toolkit_drawOutlineThick( x, y, wid, hei, 3, 4, &cGrey10, NULL );
+      toolkit_drawOutlineThick( x, y, wid, hei, 0, 2, (wgt->type == WIDGET_BUTTON ? &cGrey70 : &cGrey20), NULL );
```
bobbens, line 1336: I don't really like the idea of having the focus outline change depending on the widget type. It would be better to try to be consistent I think.

bobbens (changes requested):
Overall, I think it is a bit better. However, could you try to do it without adding new colours and having widget focus outline be consistent? Additionally, I think the outline when not focused might be a bit too dark.

src/colour.c
djvs: Yeah, there's a pretty big difference between a 5% gradation.

src/toolkit.c
djvs: This "widget" concept doesn't really carry over into the user's perception of the interface.  A button needs to have additional prominence over something like an image/text widget.

djvs: Most of this stuff doesn't really even need an outline, e.g., this list widget



I'd be just as happy removing that code for the general use case.

bobbens: If the list widget is focused, you can navigate through it, which isn't possible if it isn't though. So in that sense, it is useful though.

bobbens:
OK, I've tried it. I'm fairly happy with all the changes except the fact that you can't tell the focus on the list widget and such. If you try to navigate, for example, the load menu using the keyboard only (tab to change widgets), you can't tell when the list is selected, which I believe is important for the general case as the selected widget gets special input handling in many cases (input widgets, etc...)

bobbens:
Thanks!

PR #1557, opened 2020-12-13
Opinionated code cleanup, fixes editors' number inputs.

UncombedCoconut:
"Fixes" means "blocks 'x' as intended". Input filters may need a rethink in the Unicode world anyway; constants should make that easier.


bobbens:
Thanks!

Issue #999, opened 2020-05-21
Align format of <asteroid><pos> tag in SSys files


Mutos:
I'm currently playing with the new asteroids feature and discovered an inconsistency in the way their position is written in the XML files, relative to the way every other position is expressed in both asset.xml or ssys.xml files.

In all NAEV XML files, positions tags look that way :

```
<pos>
  <x>-1644.922880</x>
  <y>-220.059200</y>
 </pos>
```
Here is an asteroids anchors position tag :

```
   <pos x="0.000000" y="0.000000"/>
```
I suggest to align the format for asteroids anchors positions to the general format with subtags <x> and <y>.

nenau-again:
That comes from the asteroid system simplifications that happened in https://github.com/naev/naev/pull/917

@onpon4

Mutos:
Thanks nenau for the information !

The idea should be to "just" replace **xmlw_attr** by **xmlw_elem** in **system.c** when writing XML for the SSystem, then find the place where you read them (**dev_system.c** or **space.c** ?) and replace whatever reading equivalent function appears there.

Of course, I put quotes around "just" because there is never such a simple thing, this will probably have implications unthought of.

nenau-again:
Yep, it's in function `system_parseAsteroidField` in `space.c`
And we'll have to make corresponding changes in the systems that contain asteroids. As there are not too many, it's largely doable.
But first, let's wait for the opinion of onpon on this, and of the devs.

onpon4:
I actually made it like that because defining these X and Y positions as separate tags is bloated and makes the XML files needlessly harder to read and write. It's three extra lines per position, which can add up a lot for asteroid fields with a lot of sub-fields and/or exclusion zones. It also makes writing more tedious: more indentation levels, more text. I don't see how there's any advantage at all in completely ignoring XML attributes.

If anything I think all position tags should be changed to what I used for asteroids. But I didn't do that because it would have meant painstakingly going through all the files and manually replacing them. In those other applications the problem of spreading out positions into four lines doesn't build up as bad as it does for asteroids, so it didn't seem worthwhile to me to change them.

Mutos:
Hi onpon4,

Thanks for your opinion !

In the meantime, I had seen other objects using attributes as well. I first thought the assets were the reference, but it didn't seem so. And you're right with the bloating factor.

So don't be afraid, I won't do anything stupid ^-^

[...]

PR #1559, opened 2020-12-14
[For 0.9] Introduce typed versions of xmlr_attr. Use them. Everything hurts less now.

UncombedCoconut:
Why 0.9?
(1) I think I've tested with good coverage, but I haven't checked *everything*.
(2) free(NULL) is valid and this code will do it. I forget where. I'm still operating as if such refactors are for after 0.8 final.
So we can deal with this later; I just wanted to put it out in the open ASAP.

This is a part of #999 - it won't be a pain to implement with this in place.
Future improvements: better macros for reading an attribute with a specified default, the FIXME in system_parseJumpPointDiff, change supported int types to specific things like int64_t. (A serialized data format shouldn't depend on how big the compiler makes an int or long.)


bobbens:
Thanks!

bobbens:
Pushed to dev branch.

PR #1562, opened 2020-12-15
Oh no, not the overlay optimization again!

UncombedCoconut:
Some observations:
- The current code is calculating some overlay figures using the radar res. ???
- The current code sorta tries to shrink everyone proportionally (same factor each iteration), and sorta tries to do something about object collisions, but really doesn't. (It's not all bad: the total clickable area of the objects is maximized when we're as *unfair* as possible, since A∝r². Failure to prevent overlap also gives the user more clickable area.)
- If you reach an unlucky state, the "regularization" will push the text coordinates more than 200% of the way toward the base each iteration, creating a positive feedback loop which lasts until the coordinate becomes NaN. If that happens, the label will disappear.

Maybe this strategy could be nice?
(I do think its results are a little worse for Doranthex. Top to bottom, the 4 objects are Secundus, S. Alpha, S. Beta, S. Station. But IMO it was already bad. Overall I think they're a fair bit better, unless one really liked the object radii before these changes.)

Screenshots of Brimstone, Doranthex, Gamma Polaris, Goddard...

Before:
master @ Brimstone
master @ Doranthex
master @ Gamma Polaris
master @ Goddard

After:
branch @ Brimstone
branch @ Doranthex
branch @ Gamma Polaris
branch @ Goddard

### DIFF
--- a/src/map_overlay.c
+++ b/src/map_overlay.c
@@ -23,15 +23,23 @@
 
 
 /**
+ * Structure for map overlay size.
+ */
+typedef struct MapOverlayRadiusConstraint_ {
+   int i; /**< this radius... */
+   int j; /**< plus this radius... */
+   double dist; /**< ... is at most this big. */
+} MapOverlayRadiusConstraint;
+
+
+/**
  * Structure for map overlay optimization.
  */
 typedef struct MapOverlayPosOpt_ {
    /* Same as MapOverlayPos (double buffering). */
-   float radius; /**< Radius for display on the map overlay. */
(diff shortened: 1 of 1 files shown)


UncombedCoconut:
It might also be good to drop the "* 1.2" when we calculate ovr_res, and instead have the GUIs set overlay boundaries that don't overlap the UI chrome. We might be able to gain space that way.

bobbens:
Thanks! Does look nicer ;)

PR #1563, opened 2020-12-15
I'm a bonehead. (value of INPUT_FILTER_RESOLUTION)

UncombedCoconut:


src/tk/widget/input.h
```diff
@@ -42,7 +42,7 @@ void window_setInputCallback( const unsigned int wid, char* name, void (*fptr)(u
 
 /* Filter constants. */
 #define INPUT_FILTER_NUMBER     "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ[]{}()-=*/\\'\"~<>!@#$%^&|_`"
-#define INPUT_FILTER_RESOLUTION "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ[]{}()-=*/\\'\"~<>!@#$%^&|_`"
+#define INPUT_FILTER_RESOLUTION "abcdefghijklmnopqrstuvwyzABCDEFGHIJKLMNOPQRSTUVWXYZ[]{}()-=*/\\'\"~<>!@#$%^&|_`"
```
bobbens, line 45: I don't understand this, why remove `x`?

UncombedCoconut: This is for screen resolutions like "1280x720" -- also, it's what options.c had before I moved *and* broke that filter.

bobbens: Aha! Thanks

PR #1564, opened 2020-12-15
Fix #1547: tidy up map_overlay changes, switch the map cost/trade view from extreme RGB colours to blended font colours.

### LINKED ISSUE #1547: Map/Mode minor issues
Using naev-0.8.0-beta.4-linux-x86-64.AppImage on Ubuntu 18.04, there are a couple minor issues with the map/mode functions:

1) Geron planet in Goddard system - label missing from travel view at one point ... haven't checked again to see if it is back or not
2) for me, dark blue or indigo used for high value goods is hard to read on black in the economic info Mode
- less color range would be better, using a lighter or medium blue

UncombedCoconut:
- Fixes #1547 
- Reduces the number of times this American has to spell "colour" in the code ;)

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -1281,7 +1281,6 @@ static void map_renderMarkers( double x, double y, double r, double a )
    }
 }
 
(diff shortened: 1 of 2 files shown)


bobbens:
Thanks!

PR #1565, opened 2020-12-15
Improve jump-point layout on the overlay.

UncombedCoconut:
(Details in commit messages. The text_width calc is simplified because we have to redo the layout if the "known" status changes.)


bobbens:
Thanks!

PR #1571, opened 2020-12-17
RIP Autotools.

UncombedCoconut:
To do: .github beyond CI stage, .vscode?, finish macOS.


bobbens:
RIP Autotools, you were always loved... wait no... good riddance! Thanks!

PR #1573, opened 2020-12-17
Remove the function pointers and sound_priv header which supported multiple audio backends.

UncombedCoconut:
There's a more extreme version here which shoehorns the "sound_openal" and "music_openal" code into "sound" and "music", respectively: https://github.com/UncombedCoconut/naev/tree/sound-monolithic
However, that step might be counterproductive. (Depends on how this code is to be revamped.) It definitely doesn't result in well-organized code.
This PR does stuff that's more clearly helpful.


bobbens:
Thanks!

PR #1575, opened 2020-12-17
Update German translation

Wuzzy2:
Hello, I've updated the German translation for the new strings that were added recently.


bobbens:
Thanks!

PR #1581, opened 2020-12-18
MinGW found trouble in nlua_file.c

UncombedCoconut:
Fun facts I just learned: (1) rw->size is a function pointer. You can call it (passing the rw), but casting its address to long is not so good. (2) luaL_optlong uses a potentially shorter data type (long) than luaL_optinteger (ptrdiff_t).


bobbens:
That's crazy. Thanks!

Issue #1405, opened 2020-11-21
Translatability: image assets


UncombedCoconut:
Naev includes several images which are captioned and used as part of the UI.
This is a tracker to maybe do something about that.

Primary example: dat/gfx/map/ has map background images. They have plot-relevant captions burned in.
Secondary example: dat/gfx/loading has 3D ships with English labels.
Tertiary example: dat/gui/legacy/minimal.png has "Cargo", "Target", "2nd", "Nav" labels. Ugh.
Is dat/gfx/commodity/tractor_capability.png for real?
Same for dat/gfx/commodity/diesel.png? But whatever.
Finally, there's dat/gfx/portraits/news.png, but (to a lesser extent) whatever.

onpon4:
Yeah, I was majorly confused by that when we added the ones we added (which are most of them, excepting the Empire one). Ideally it should be some really large font-drawn text at 50% opacity. I'm not sure how to do that though (in such a way that it can be fully customized for best possible positioning).

🕵️

bobbens:
In general, when possible, I would want to stay away from one image per language. That will bloat the repo + add more work load to translaters.

> Secondary example: dat/gfx/loading has 3D ships with English labels.

This could be done with independent image + text. In fact, I think loading time is long enough that we should probably give loading "hints" or something like that. Would be a good way to teach player about more niche stuff.

> Tertiary example: dat/gui/legacy/minimal.png has "Cargo", "Target", "2nd", "Nav" labels. Ugh.

This would ideally be done all lua-side.

> Yeah, I was majorly confused by that when we added the ones we added (Nebula and Proteron). Ideally it should be some really large font-drawn text at 50% opacity. I'm not sure how to do that though (in such a way that it can be fully customized for best possible positioning).

It should be possible to specify the image + text as separate items, allowing them to be placed independently. What might be tricky is to implement different types of alignment (right/left/center) to give more control.

Wuzzy2:
> Primary example: dat/gfx/map/ has map background images. They have plot-relevant captions burned in.

This is the most important thing IMO because these images are highly visible in the starmap and have a large text.
I suggest to just remove the text from these images entirely. The text itself is not critical anyway, as the faction logos are clearly visible.

> Secondary example: dat/gfx/loading has 3D ships with English labels.

Again, here the text can be just removed from the images.

Maybe the text can be later added back in some other form in both cases (as rendered text), but I think it's not urgent in either case.

[...]

PR #1582, opened 2020-12-18
Remove baked-in text in loading ship images

Wuzzy2:
Partial fix of #1405.

The baked-in ship name text in the ship images of the loading screen is removed.

Untested.


bobbens:
Leaving open until someone implements an alternative to display text in place of the removed text.

bobbens:
Thanks for the PR! I have added your images to the `naev-artwork-production` repository and included them in the main repository to not increase the main repository size. I have not added the text back in yet, but I think I have an idea of how to do it for now.

PR #1583, opened 2020-12-18
Remove baked-in text in starmap background images

Wuzzy2:
Partial fix of #1405.

The baked-in text in the background region images in the starmap (Dvaered, Empire, Nebula, etc.) is removed.
This was trivial for most images because the text did thankfully not overlap.
This was not the case for The Nebula, where the text DID overlap. For this image, I cropped the upper part of the image, faded it out and scretched the remaining nebula vertically.

UNTESTED! Please verify. I don't know how good (or bad) the new Nebular image will look in-game. The nebula image would ideally be redrawn by an actual artist.


onpon4 (changes requested):
The Nebula image shouldn't be modified from the image; if you look on https://github.com/naev/naev-artwork, there's a layered XCF file which can be used to make the edit more cleanly.

(By the way, when fixing that, I recommend ensuring that you use a PNG optimizer like optipng or whatever is popular these days; I use Trimage to do the job myself. And I suggest you force-push an amendment to your existing commit to avoid the unnecessary clutter of having the image modified twice in sequence.)

🕵️

Wuzzy2:
Done. I use `zopflipng` btw.

onpon4 (approved):
Awesome, thanks for fixing that. (Leaving it to @bobbens to decide whether or not to merge this right now; it's definitely something we want but ultimately we want to add text to replace the image-text as well. Not sure how bobbens wants to do this ordering-wise.)

bobbens:
I'm leaving this open as I want to implement translatable text support as an alternative, and I know if I pull it, we'll forget about it and never do it ;)

bobbens:
Thanks for the PR! I have converted all the png to webp and put them in the `naev-artwork-production` repository to not bloat the main repo.

Issue #1476, opened 2020-11-30
Implement Outlines with a Shader


bobbens:
I think it should be fairly easy (for someone who knows what they are doing) to implement a blur shader to generate the outlines we are doing for text, the map overlay, etc. This would avoid 4 additional rendering passes and probably look better.

ids1024:
Not sure if there's a good approach to add outlines in the fragment shader, with just a texture for a letter.

Using signed distance fields for text would help with this and other things: https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf

Or switching to something like Pango for rendering, but that introduces a lot of dependencies and has it's own complications.

bobbens:
I agree that signed distance fields would be the way to go for text. That would also allow for nice outlines in one pass. For primitives (lines, triangles, etc.) I guess it would also be applicable although it might be possible to get by with an easier approach.

On Pango, while it would be possible, it does feel like it's not a really easy dependency to add, plus it would likely add a ton of revamping to all the code.

ids1024:
For polygons, it looks like there are two approaches: a second draw call with `GL_LINES`, and using vertices in barycentric coordinates.

Perhaps I'll looking into how these "barycentric coordinates" work. Since that approach is more flexible, and I think there can be compatibility issues getting variable line width and good antialiasing that could be easier if we can just implement it in shaders.

bobbens:
Yes, from my reading anti-aliasing is fairly hit or miss depending on implementation. Doing it in GLSL would make it much more consistent and do fancier stuff :)

bobbens:
This has partially been done for text (although not entirely happy with it atm) using distance fields.

bobbens:
Done and working very well imho.

PR #1589, opened 2020-12-18
Fuss with the cooldown OSD message; supplying better duration data helps Naev use better text alpha/outline settings.

UncombedCoconut:
If you have to ask, you don't want to know. (It involves  #1476 being unfixed.)


bobbens:
Thanks!

PR #1590, opened 2020-12-18
Fix targeting graphics (turned from a frame to 4 circles) and slim GUI direction indicator (don't wanna know). Retire npng_metadata.

UncombedCoconut:


onpon4 (approved):

bobbens:
Thanks!

PR #1592, opened 2020-12-19
This may help get the outlines drawn by shaders.

UncombedCoconut:
Refactors the main gl_print-family functions.
Leaves us with this remark in gl_fontRenderStart:
```
   (void) outlineR; /* TODO: At this point, we ideally stuff the radius directly into the shader. */
   /* TODO: If we keep the separate shaders for outlined and non-outlined, we need to set up state so
    * that gl_fontRenderGlyph and gl_fontRenderEnd work with the right one (or pass outlineR to them too). */

```
It looks to me like there can be one "font" shader taking parameters (and an outline radius of 0 can degenerate to no outline). But I need to sleep and I wasn't sure if there was a reason not to consolidate that way.

### DIFF
--- a/src/font.c
+++ b/src/font.c
@@ -112,7 +112,6 @@ typedef struct glFontStashFreetype_s {
 typedef struct glFontStash_s {
    /* Core values (determine font). */
    char *fname; /**< Font list name. */
-   unsigned int h; /**< Font height. */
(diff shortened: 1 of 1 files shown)


bobbens:
Thanks!

PR #1593, opened 2020-12-19
Update German translation

Wuzzy2:
So I noticed the update-po was broken and is now fixed again. This caused a bit of problems.

So, here's a new update for German to add more missing strings.


bobbens:
Thanks!

PR #1598, opened 2020-12-19
Clippy asks if we meant...

UncombedCoconut:
(The font_shader_vertex and font_shader_tex_coord globals are only meant to be used mid-render. If we need these steps, do we need them for all font shaders?)


bobbens:
Clippy is spot on.

PR #1599, opened 2020-12-19
Add a global for the current font shader's projection matrix, too.

UncombedCoconut:


bobbens:
Whoa, totally missed that, thanks!

PR #1608, opened 2020-12-22
Font shader update

UncombedCoconut:
I did some algebra to figure out what fraction of a square one would expect to fall within the glyph/outline thresholds of a distance field, based on the value and directional derivatives.
It's not what we had.

EDIT: Then @bobbens fixed a whole lot of stuff, and the need for this mostly evaporated.
It still seems to achieve a crispness/smoothness tradeoff, which is likely also achievable by replacing smoothstep() and less change than we have here.

Probably we close this without merging. BUT here are updated comparisons.

land-master
land-merged
load-master
load-merged
map-master
map-merged
menu-master
menu-merged

### DIFF
--- a/dat/glsl/font.frag
+++ b/dat/glsl/font.frag
@@ -4,6 +4,7 @@ uniform sampler2D sampler;
 in vec2 tex_coord_out;
 out vec4 color_out;
 
+// How thick to render glyphs; 0 is very thick, 1 is very thin
 const float glyph_center   = 0.5;
 
 void main(void) {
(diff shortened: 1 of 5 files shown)


UncombedCoconut:
The originally submitted patch had an error in the formula translation (effectively using double the derivatives it should have).
The differences are very subtle but, overall, doing it correctly was better than doing it incorrectly.

land-new-algo-2
map-new-algo-2
menu-new-algo-2

UncombedCoconut:
I can beat both of these. I hate every moment of this. Why am I torturing myself so?

PR #1610, opened 2020-12-23
Scalefactor default applies correctly. A few other defaults enabled.

MightBeHarmless:
So, Scale factor was changing to 3.0 when I pressed the Set Defaults button under Video.
It turns out that the reason, was the default value being 1.0, while the slider (fader) widget treating 0 through 1 as 1 through 3 in terms of actual applied value.

I tried a different solution, but this final version I believe is more accurate and more readable than my non-disclosed effort.
The problem was, that there has to be equal functionality for Directly setting the value, as with loading from conf,lua or resetting to Defaults. The other two modes are Mouse based selection, and Keyboard based selection.

Have tested that each other Fader still operates as they were intended to,

This commit also adds a rounding of the Scale factor to the nearest Tenth of a value, because something like 1.4879 seems to me to be completely missing the mark of perhaps allowing scaling factors such as 150%, 200%, etc from being selected by the user.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -142,7 +142,7 @@ void conf_setDefaults (void)
    /* GUI. */
    conf.mesg_visible = 5;
    conf.map_overlay_opacity = MAP_OVERLAY_OPACITY_DEFAULT;
-   conf.big_icons = 1;
+   conf.big_icons = BIG_ICONS_DEFAULT;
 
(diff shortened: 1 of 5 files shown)


UncombedCoconut:
Makes me wonder if existing fader code should really work that way. No matter -- things are clearly better than before. Thanks!

PR #1611, opened 2020-12-23
I hate fonts

UncombedCoconut:
Rather than include screenshots I'm going to let @onpon4 either tear this apart or not.
But if you liked the old rendering at 1:1, you should like this. (But it's still using signed distance fields.)
The most important change is to use the FT character size that would make sense at 1:1 and a FT transformation matrix to render the glyph at the appropriate size. This helps size-dependent features of glyph outlines Do The Right Thing.
I also tweaked the shaders. Rather than introduce complexity, I changed a [-fwidth(dist),fwidth(dist)] interval to [-fwidth(dist)/2,fwidth(dist)/2] so the smooth steps are across a ~1px margin instead of a ~2px margin. This is a major improvement at 1:1.

### DIFF
--- a/dat/glsl/font.frag
+++ b/dat/glsl/font.frag
@@ -8,17 +8,10 @@ out vec4 color_out;
 const float glyph_center   = 0.5;
 
 void main(void) {
-   /* Standard rendering. */
-   /*
-   color_out = color;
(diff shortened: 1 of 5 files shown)


UncombedCoconut:
I can beat both of these. I hate every moment of this. Why am I torturing myself so?

PR #1621, opened 2020-12-30
Verify widget name before operating upon it. Fixes a bug.

MightBeHarmless:
Protecting from a Segfault that was encountered while pressing Left Shift on the Outfits tab while focused on something that possibly shouldn't permit focus. Use Tab to set focus, then press Left Shift.

Here's my hand-written stack trace.
```toolkit.c line 353
if (strcmp(wgt->name, name)==0)
	wgt->name is btnCloseOutfits
	name* is 0x0

called by: window_setFocus( const unsigned int wid, const char* wgtname ) line 2334
	with unmodified wgtname

called by: outfits_regenList( unsigned int wid, char *str ) land_outfits.c line 231
	as window_setFocus( wid, focused );
	where focused is 0x0, set by focused = strdup(window_getFocus(wid));

### DIFF
--- a/src/toolkit.c
+++ b/src/toolkit.c
@@ -2348,6 +2348,10 @@ void window_setFocus( const unsigned int wid, const char* wgtname )
    Window *wdw;
    Widget *wgt;
 
+   /* If widget name is NULL, we should not proceed */
+   if (wgtname == NULL)
+      return;
+
    /* Get window. */
    wdw = window_wget(wid);
    if (wdw == NULL)


MightBeHarmless:
This PR relates to one from UncombedCoconut, which was pushed against 0.8.0, and features different changes.

https://github.com/naev/naev/commit/dad79e2c6c0b7708bb1d3a3770ff552fb0b385f3

UncombedCoconut:
After https://github.com/naev/naev/commit/792210dec47012b312f22c3887f43f4b10c8b4df, I don't think this is required anymore.
The benefit of not exiting so early (in theory) is the ability to restore whatever state window_getFocus() saves -- including the state of no focused widget.
As a matter of style, I'd be fine switching from my way (?: expression) to the explicit exit, if that exit happens after toolkit_focusClear( wdw ). Your way may well be clearer.
What do you think?

MightBeHarmless:
I didn't notice that you had pushed 2 or 3 earlier commits to 0.8.x as well.
Maybe I should mention when I have a fix in the works when communicating on discord.
Not preferential on what source gets merged, so long as the solution is appropriate to the issue, and it finds its way to master as well. Although this is giving me a workout and git practice on keeping my local checkouts valid.

UncombedCoconut:
OK, in that case I'll close this and we can revisit if anything is still amiss. Sorry for the miscommunication.

MightBeHarmless:
Sure thing. Closing it now.

I would have liked to do more testing with it. Observations where Focus is actually going now.
As I mentioned before, I only saw one of your commits, and it looked like it was still calling a function I was trying to avoid calling, etc.

PR #1623, opened 2020-12-31
Hilighting a missing Ship License with Color

MightBeHarmless:
Issue: For the Light, Medium and Heavy Combat Vehicle Licenses, though I did not own them, their text was not in Red.
At least the Button was set to not enabled. Or enabled, with a message explaining the license was needed.

The button prohibiting purchasing was working as intended. The Red text indicating a missing license is what appeared to be missing.

I also reduced the decimal precision on the Local Map, to 0. Which makes oversized button fonts render nicely, but might affect any forks that charge 3.45 for local maps.

This is how the application behaves now:
License_Color

### DIFF
--- a/src/land.c
+++ b/src/land.c
@@ -827,10 +827,10 @@ void land_updateMainTab (void)
    /* Else create it. */
    else {
       /* Refuel button. */
-      credits2str( cred, o->price, 2 );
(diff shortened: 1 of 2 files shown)


UncombedCoconut:
Thanks!

PR #1624, opened 2020-12-31
Add macOS to CI (again) // Cross-compile Windows // Update pre-release and release workflows

ProjectSynchro:
Within this PR:

- [x] Adds Darwin back to CI, it will fail in it's current state due to issues when building the luajit submodule.
- [x] Opt to use MXE cross-compile environment to build Windows binaries in CI and in releases.
- [x] Adds cross-file for meson in utils/build


UncombedCoconut:
Huh. Didn't know tar -xf without the "J" option would handle .xz archives.

Issue #1606, opened 2020-12-21
Carefully ensure all "Redistributions in binary form must reproduce the above copyright notice" requirements are honored


UncombedCoconut:
Post-0.8, we've been adding a fair amount of stuff. Reminder to double-check later that all binary distributions fully credit our library authors. (In some cases, I'm not sure if embedding the library triggers the clause, but we're nice people and we may as well...)

bobbens:
Does it have to be done every run or can it just be thrown into credits? Additionally, a licensing window or something could also be displayed in options or something.

UncombedCoconut:
I think even a stand-alone text file is fine. The main concern is having the message in .c/.h files which don't make it into binary distributions.
(Including them in credits makes sense to me, but maybe a thanks there and the exact notice off to the side.)

onpon4:
If you're referring to the 3-clause BSD license, what it specifically says is:

> Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.

That's clearly referring to the license file. So that's not a thing asking for an entry in the credits (unless you're planning to put multiple copies of the 3-clause BSD license in the credits), it's just telling you to include the LICENSE.txt or whatever. This should go hand-in-hand with properly documenting Naev's exact license for the entire repository. I believe I mentioned a couple years ago, we need to have a license documentation file which clearly states licensing intent for everything, rather than just having a copy of the GNU GPL sitting there. I can do the bulk of the work (IANAL but I have enough experience with licensing stuff that it's not hard for me to do) but I really need @bobbens involved in this, namely to give a clear answer on what the licensing policy is going to be stated as, and in particular what our default license is (because while I believe I know the answer to that, I don't have the legal authority to decide the license of his work for him; I really need an explicit direction here or anything I do will be entirely meaningless).

Related: #638

If we get that done and over with, any additional licenses will just be a matter of listing those there similar to what SuiteSparse does.

🕵️

bobbens:
[...]

PR #1636, opened 2021-01-06
Clarified licensing terms.

### LINKED ISSUE #638: Lua script licenses need to be cleared up
If they're not at least under the GNU GPL v3, they're infringing copyright. But I don't like this situation where none of the Lua files have any sort of license information attached to them. This needs to be addressed somehow. It needs to be made clear that _all_ Lua scripts are licensed under at least version 3 of the GNU GPL. Preverably, every Lua script should have some sort of notice on it.

I was trying to address this with my "copyright" branch, but it was never merged.

onpon4:
This moves LICENSE to "gpl.txt" and uses LICENSE to instead explain
the licensing terms in full.

Closes #638.

Related to #1606.


bobbens:
I think it looks good. Thanks!

PR #1638, opened 2021-01-08
Switch to cross-compiling macos in CI

ProjectSynchro:
The doctor will be visiting.. (Crap joke about an apple a day)


UncombedCoconut:
Ahhh, that's what it meant. :)

PR #1647, opened 2021-01-08
Update German translation

Wuzzy2:
I fixed some mistakes in the German translations and updated the small number of new strings that were added recently. It's 100% again.


UncombedCoconut:
Thanks!

PR #1649, opened 2021-01-09
Fix issues pushing Naev releases. (0.8.x)

ProjectSynchro:
Looks like reading does in fact pay off..

We cannot set the default branch as live using the setlive option in build scripts.
Setting builds as live, will need to manually be done within Steamworks.


UncombedCoconut:
Thanks!

PR #1650, opened 2021-01-09
Fix issues pushing Naev releases.

ProjectSynchro:
Looks like reading does in fact pay off..

ditto of #1649


UncombedCoconut:
Thanks!

PR #1655, opened 2021-01-10
Update Readme.md

kernhe:
For *buntu the dependencies should also list:
* libsuitesparse-dev 
* libphysfs-dev (>=3.0.2)

So it might be difficult to build naev on a system older than *buntu 20.04.

Also missing are some basic dependencies:
* doxygen 
* cmake 
* git
... but meson will inform about this in a clear manner.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -31,16 +31,18 @@ If you're cross-compiling for Windows, you must install this soft dependency:
 
 
 ### Ubuntu
-
-Install compile-time dependencies on Ubuntu 16.04 (and hopefully later) with:
+Install compile-time dependencies on Ubuntu 20.04 (and hopefully later) with:
 
 ```
 apt-get install build-essential libsdl2-dev libsdl2-image-dev \
 libgl1-mesa-dev libxml2-dev libfreetype6-dev libpng-dev libopenal-dev \
 libvorbis-dev binutils-dev libiberty-dev autopoint intltool libfontconfig-dev \
-python3-pip
+libphysfs-dev libsuitesparse-dev python3-pip
 pip3 install meson ninja
 ```
+
+We do not suggest to build on Ubuntu releases < 20.04, as they do not provide the up-to-date releases 
+of the dependencies. Sorry.
 
 ### macOS
 


UncombedCoconut:
I appreciate the thought, but I'm dubious on most of these changes.

- libsuitesparse-dev and libphysfs-dev are optional dependencies (though, yes, let's recommend installing them, as that's fastest and most consistent with the distro).
- We fully support Ubuntu 16.04, in order to provide [AppImage](https://appimage.org/) builds. This will be true until its end-of-life date in April.
- cmake isn't an intentional dependency anymore, but probably no one's testing builds without it. Does something break?
- git isn't either, if someone fetches a source tarball from our releases or uses "apt-get source" to get the source.
- doxygen (and ldoc) are definitely optional dependencies. Meson should auto-detect whether it can enable the "docs_c" and "docs_lua" options, respectively. If it doesn't, that's a bug.

kernhe:
@UncombedCoconut , thanks for your fast response!


(1) libphysfs-dev is checked when you run "meson build" and I have on a Ubuntu 18.04 libphysfs-dev (3.0.1-1) installed. This is what meson is telling me: 
```
Run-time dependency physfs found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency physfs
```
If I use a newer libphysfs-dev (3.0.2-4) everything runs smoothly. But this version is only available from Focal Fossa onward. That leads to (2)

(2) Ok, I'm not an expert with meson ... I love make. So maybe I missed some tweaks with meson

Question: how do you tell meson that libphysfs is not needed?


(3) you're right, cmake is been checked and if it is not available "meson build" will continue

(4) git was required for the git subproject. But I checked it with the tarball and everything was fine.

(5) you're right with doxygen / ldoc.

UncombedCoconut:
(2) The Naev dev team provided Meson with a fallback directive for physfs: `dependency('physfs', fallback: ['physfs', 'physfs_dep'], required: true)` in meson.build. That's what this "fallback subproject" business is about: it's a way for physfs to be downloaded / compiled / linked in as a static library if it's not already present on the system.
It's still a good idea to add it to our "apt-get install" line in our instructions. That's the preferred setup, and the subproject is a fallback.
The same goes for luajit and libsuitesparse-dev. It's better to install them than not, but Naev is supposed to build if the system doesn't have them. (IMO, luajit is the key optional dependency: depending on your CPU architecture, it either speeds up Naev for free or is totally unavailable.)

kernhe:
With the old version of physfs meson will stop with 
`meson.build:100:3: ERROR: Subproject exists but has no meson.build file`

see log file
[meson-log.txt](https://github.com/naev/naev/files/5807722/meson-log.txt)

kernhe:
pull request withdrawn, because it is not 100% accurate.

PR #1660, opened 2021-01-15
Weapons rebalancing

onpon4:
This is a rebalancing effort I've been working on. I'm opening this draft PR to enable better discussion and understanding of what I'm doing.

I'm basing much of my work on this spreadsheet, which shows updated core system stats:

[cores_adjusted.ods.zip](https://github.com/naev/naev/files/5817727/cores_adjusted.ods.zip)

As of this writing I've not done much with weapons themselves yet, but basic cores and some utilities/structurals have been rebalanced.

For weapons, we're in general trying to have a set of weapon "tiers" enforced by two things: the penetrate/absorb values, and CPU requirements. Each serves a different purpose:

* Absorb values serve as a mitigation of the effectiveness of small weapons against large ships. In general, a weapon that's designed to effectively take on a ship will have enough penetration to cancel out absorption. What this will mean is, while lighter versions of weapons will only be marginally weaker than stronger versions against their intended targets (lighter fighters), they will be very ineffective against ships of the higher "tier". In general, it is meant for each ship class to be able to take on ships that are one "tier" above their own: ultralight fighters can take on regular fighters, regular fighters can take on corvettes, corvettes can take on destroyers, and so on. However, it will generally be very difficult for an ultralight fighter to take on a corvette, and nearly impossible for an ultralight fighter to take on a destroyer. A regular fighter, however, will be able to take on a corvette and may be able to take on a destroyer with difficulty.
* CPU requirements serve as an effective limiting factor for what weapon "tiers" a ship is reasonably able to equip. In general, the best CPU in a "tier" might allow equipping one weapon which is a tier above, but doing so will lead to there being next to no CPU left for other weapons, limiting effectiveness of this. CPU will not be used to divide every single tier, but of particular note, it will be used for making the ultralight fighter tier, regular fighter tier, and corvette tier use completely different weapons. This will also be used to effectively restrict feasible access to fighter bays to only the heaviest of ships, a niche which carriers are designed to fill.

It's worth noting that energy will *not* be used as a limiting factor for weapons, since my tests showed that this is not an effective means of dividing ships into tiers. Instead, energy will be used as a limiting factor for utilities, primarily. Still, stronger weapons will in general require more energy than weaker weapons.

In doing this rebalancing work, I'm referencing a small selection of "model ships":

* Yacht: Llama
* Courier: Koala
* Ultralight Fighter: Hyena, Shark
* Fighter: Lancelot
* Bomber: Ancestor
* Corvette: Admonisher
* Freighter: Mule
* Destroyer: Pacifier
* Light Cruiser: Kestrel
* Cruiser: Hawking
* Carrier: Empire Peacemaker

This PR is *not* going to take care of ship balance issues. That will be done afterwards, separately. This PR is solely focused on the outfits themselves.

This PR will be renaming or has already renamed several outfits as well:

* Unicorp core systems: named after CPU capacity (PT-18, PT-80, PT-280, PT-400, PT-750, and PT-3400)
* Unicorp platings: named after absorb value (D-2, B-2, D-4, B-4, D-12, B-12, D-24, B-24, D-48, B-48, D-72, B-72)
* Blasma Blaster MK3: new name undecided
* Razor MK3: new name undecided
* Heavy Laser: Heavy Laser Turret

Additionally, some outfits will be removed as they are redundant:

* All Laser PD outfits (to be replaced by Laser Turret MK1, which will be made a small outfit).
* Laser Cannon MK3 (to be replaced by Ripper Cannon).
* All but one of the Mass Driver weapons, which will be named simply "Mass Driver" and will be on the same tier as the Heavy Ripper Cannon.
* All but one of the Plasma Cluster Cannon weapons, which will be named simply "Plasma Cluster Cannon" and will b e on the same tier as the Heavy Ripper Cannon.
* Laser Turret MK3 (to be replaced by Laser Turret MK2 and Heavy Ripper Turret)
* Razor Turret MK3 (to be replaced by Razor Turret MK2 and Heavy Razor Turret)

Most likely I will also want to also repurpose the Turbolaser as an Empire outfit, do the same with Repeating Railgun and House Dvaered, and possibly add a new outfit to fit a similar niche for the Razor weapons that will be exclusive to House Sirius. I haven't finalized this yet as I'm starting with the lowest tiers.

Next on my agenda will be to finalize utilities and structurals balancing, and then begin with the lowest weapon tier.

The lowest weapon tier will consist of the Laser Cannon MK1, Plasma Blaster MK1, Gauss Gun, and Razor MK1. This tier will be used by ultralight fighters (Hyena, Shark) and use about 5 CPU.

Just above that will be the tier which consists of the Laser Cannon MK2, Plasma Blaster MK2, Vulcan Gun, and Razor MK2. This tier will be used by regular fighters (Lancelot) and use about 24 CPU.

The lowest turreted tier will be based on second-tier costs, but first-tier strength. This means that e.g. the Laser Turret MK1 will have strength similar to the Laser Cannon MK1, but a CPU cost similar to the Laser Cannon MK2. In this way, low-tier turrets will be effective as a defense against fighters which may pose a threat to a courier, but will not be an acceptable substitute for forward-facing weapons in a light combat ship. This will be true for all tiers up to and including the corvette tier. MK2 turrets, which will be on the same tier as the Heavy Ripper Cannon, Mass Driver, and Plasma Cluster Cannon, will become realistically available only to freighters, destroyers, and above.

🕵️


nenau-again:
Congratulations for all the work done! (and good luck for the work to go)
Is the branch testable right now, or do you want to add yet other important changes?
Looking forward to testing (and criticizing) what you've done.
Not looking forward to rebalancing all the missions after that.
I don't understand who is the light fighter. ultralight fighters are hyena and shark, and fighter is the lancelot.

onpon4:
It's not testable yet, no. It'll start being testable once I get the ultralight weapons put together.

Missions shouldn't need much rebalancing I don't think. I'm aiming for the same power dynamic, just enforced better. The one example of rebalancing I expect to be needed is to add some Pacifiers to the FLF patrol missions that feature a Goddard (and this is probably a good idea anyhow).

> I don't understand who is the light fighter. ultralight fighters are hyena and shark, and fighter is the lancelot.

Sorry, my terminology is a bit unclear. Let me explain what I mean:

* Ultralight Fighter: any fighter which seeks to stay under the 150t light engine mass limit (e.g. Hyena, Shark).
* Fighter / Light Fighter: any fighter which is meant to require engines that support a 300t mass limit (e.g. Lancelot, Vendetta).
* Light Cruiser: any Cruiser which can be light enough to fit under the 3800t mass limit of the Krain Remige Engine (currently this group only includes the Kestrel and Pirate Kestrel).

The "Ultralight Fighter" and "Light Fighter" terminology I tend to use in my head comes from the names of the S&K engines which correspond to these groups.

🕵️

onpon4:
I've just done a bit more of this, and the Laser Cannon MK1/MK2 and Ripper Cannon are now testable. I've also figured out what values I'm going to use for weapons at the various sizes:

* Ultralight: 5 CPU, 0 penetrate, 5 EPS, 21 DPS
* Light: 24 CPU, 12 penetrate, 10 EPS, 23 DPS
* Medium (Corvette): 100 CPU, 24 penetrate, 15 EPS, 32 DPS
* Medium-Heavy: 50 CPU, 48 penetrate, 20 EPS, 34 DPS
* Heavy (light Cruiser): 75 CPU, 72 penetrate, 25 EPS, 45 DPS
* Superheavy (capital class): 350 CPU, 100 penetrate, 45 EPS, 70 DPS
* Fighter bays: 450 CPU

The reason CPU usage drops for medium-heavy is because outfit size changes. It's considered legitimate to e.g. use a Lancelot's medium slot for weapons designed for Destroyers; what discourages that is engine mass limit. The same goes for heavy (but not capital-class) weaponry and medium ships that have heavy weapon slots, hence the modest CPU usage increase. Corvette outfits are then effectively a "limit break" with greater damage at a lower mass, lower energy usage, and in a small slot size, at the cost of increasing CPU usage (which is worthwhile for a Corvette given the fact that it needs to use the medium slots for missiles).

🕵️

onpon4:
Ok, I've done some rebalancing specifically with CPU costs. The primary mechanism I have used to achieve this is the improved TeraCom launchers, which I think works well since it's generally advisable for every ship to have at least one launcher and it can be useful to have the distinction between a 2 CPU Unicorp Banshee launcher and a 5 CPU TeraCom Banshee launcher. I may want to adjust the launchers themselves to ensure that the CPU cost differences are actually worthwhile, but so far I've focused specifically on the theory of how to put players in a "tough spot" with regard to choosing outfits (an issue that I think was brought up by @nenau-again if my memory doesn't fail me).

I've tested various configurations for the Hyena, Shark, Lancelot, Vendetta, and Ancestor and found that, at the shop level, these trade-offs that I've built in now seem to hold up pretty well. I did have to make one adjustment to the Ancestor: since it uses missiles, it tends to have excess CPU, so I've added a -20% CPU penalty for it. This seems to be a good threshold for creating a "hard choices" situation for the Ancestor, at least assuming the missile balancing that I'd like to have. As mentioned in the commit, I based the number on a threshold of allowing missiles that cost 10 CPU. I might adjust this later if I decide that I want to make the threshold something like 8 CPU instead, which may happen. (If I do change the threshold to that, then I'll probably change it to a -25% CPU penalty to compensate. We'll see after I do the balance work on these missile launchers themselves.)

🕵️

onpon4:
I think most of the hardest parts of this work are out of the way now. Most remaining weapons are turreted versions of other weapons; the way I'm handling those is to simply copy stats of other weapons. The general trend is to copy stats from one tier below, but copy the CPU and energy usage of the tier the turret is in. So for example, the Laser Turret MK1 copies most stats from the Laser Cannon MK1, but costs the same CPU and energy as the Laser Cannon MK2. MK2 turrets, I just found out, are a bit more complicated, because they're actually two tiers above the Laser Cannon MK2; as such, I used a mix of stats from the Laser Cannon MK2 and Ripper Cannon for the Laser Turret MK2. (Laser Turret MK3 will be removed as it is redundant.

So most of the rest of the work should mostly just involve applying these same rules elsewhere. Just a matter of getting the spoons to concentrate on this. :sweat_smile:

🕵️

bobbens:
> I've just done a bit more of this, and the Laser Cannon MK1/MK2 and Ripper Cannon are now testable. I've also figured out what values I'm going to use for weapons at the various sizes:
> 
>     * Ultralight: 5 CPU, 0 penetrate, 5 EPS, 21 DPS
>     * Light: 24 CPU, 12 penetrate, 10 EPS, 23 DPS
>     * Medium (Corvette): 100 CPU, 24 penetrate, 15 EPS, 32 DPS
>     * Medium-Heavy: 50 CPU, 48 penetrate, 20 EPS, 34 DPS
>     * Heavy (light Cruiser): 75 CPU, 72 penetrate, 25 EPS, 45 DPS
>     * Superheavy (capital class): 350 CPU, 100 penetrate, 45 EPS, 70 DPS
>     * Fighter bays: 450 CPU

I've been trying it and I like the light slots with high CPU usage that are similar. It seems to work very well and overall I feel like the weapon slots and core slots for small/medium ships feel pretty well polished. Some weapons may be a bit too similar still, but I do think it tends to work out pretty great.

Although this is still very much WIP, I would like to give some comments/feedback on some things I've seen:

* Masses don't seem rebalanced yet. I believe this just hasn't been started yet.
* I feel like are too many core outfits with too little diversity                               
   - "Milspec Orion 9901" and "Milspec Aegis 9801" are almost the same.         
   - "S&K Medium-Heavy Combat Plating" and "Unicorp B-24 Medium Plating" have pretty small differences, etc.
   - No light "Heavy Plating", "Patchwork Heavy Plating" is actually lightest   
* No medium/heavy weapons (affects kestrel, which can't really make good use of medium slots)
* Giving fighter bay 450 CPU usage might be too low? Maybe balance mainly on mass (they are fairly heavy now)?
   - Lancelot bay is 2x lancelot = 2x med, 6x sml weapons ~= 150 to 200 DPS with 10~20 penetrate
   - Za'lek bays are even more ridiculous
*  Low end "heavy CPU" have really bad CPU
   - Something between "Unicorp PT-3400" (3400 CPU) and "Milspec Orion 8601" (800 CPU) would seem fairly useful
* Fairly minor, but plasma Turret MK2 graphic

For weapons, other than stuff that hasn't been done yet, the main thing I feel would be fairly useful would be something like "Medium (Light Cruiser)" type weapons. Something like:

* Medium (light Cruiser): 350 CPU, 80 penetrate, 33 EPS, 55 DPS

For core outfits it does seem like it might be beneficial to both reduce the number so that they are fewer similar ones, and increase the diversity for niches that are missing as mentioned above.

onpon4:
> Masses don't seem rebalanced yet. I believe this just hasn't been started yet.

Yes, I've pretty much been ignoring masses at the moment and leaving them as-is mostly because I don't feel like they need the mass overhaul that everything else does. I figure I'll adjust as needed for this particular stat.

> "Milspec Orion 9901" and "Milspec Aegis 9801" are almost the same.

Aegis outfits have lower stats in general but better shielding (the same as how Prometheus specializes in battery and Hermes specializes in CPU). There's an argument to be had for redundancy (Orion is almost always the ideal choice), but the purpose of this to my mind is to offer more fine-tuned control over the stats of the ship (to pick what gives you the best bang for your buck, so to speak). I may look over the differential between specialized and Orion, though.

> "S&K Medium-Heavy Combat Plating" and "Unicorp B-24 Medium Plating"

Same as above, the different hull offerings are for fine-tuned customization. In the case of Unicorp outfits, they generally sacrifice capability for lower mass. Unicorp B- hulls also tend to have more cargo space than S&K combat plating. (Side note, I don't think I've changed hulls much, main thing I've adjusted hull-wise is the "medium" (Corvette) sized hulls. But I'm open to adjusting them to make the Unicorp B- hulls more desirable if you have an idea for what particular attribute would distinguish them. I personally don't really mind S&K hulls just being the best on the market for the most part and I also don't mind there being two Unicorp outfit versions. I've even found Unicorp B- hulls useful on occasion because of the lower mass. Perhaps that mass differentiation could be widened to increase that mass incentive, I'd be open to that.

> No light "Heavy Plating", "Patchwork Heavy Plating" is actually lightest

I'm not sure what you mean by this. Patchwork Heavy Plating is only slightly lighter than the Unicorp D-48 Heavy Plating, and the Unicorp B-48 Heavy Plating provides a middle-ground between D-48 and S&K Heavy.

> No medium/heavy weapons (affects kestrel, which can't really make good use of medium slots)

What do you mean? Kestrel is specifically targeted by the Railgun. Or if you really want to use turrets on it, the general turret weapons designated for cruisers are the Heavy Ripper Turret and Heavy Ion Turret. Grave Beam will also be added to that list when I get to it. For medium slots, there's the Laser Turret MK2, Razor Turret MK2, Orion Beam, Turreted Vulcan Gun, and Enygma Systems Turreted Fury Launcher. All of these outfits mentioned are designed to be compatible both with cruisers and with medium-heavy ships like armored transports and destroyers.

> Giving fighter bay 450 CPU usage might be too low? Maybe balance mainly on mass (they are fairly heavy now)?

Well that CPU cost makes using fighter bays on anything with less than the heaviest CPUs impractical (and especially impractical for medium vessels). Of course it's easy for a cruiser to equip heavy stuff to afford not only fighter bays but also the more powerful capital-class weaponry, but that doesn't really make much sense to do as carrier-class ships have better armor and shield stats as a general rule (and if it's a player's only option, I don't see that as a problem).

It's also a bit lower than I'm going to set more specialized bays to.

Regarding your calculations of DPS, I'm not necessarily sure why you would consider that a problem in general. Carriers are supposed to be the heaviest and most powerful ships in the galaxy. The outfit setup I mean to encourage is two weapons for emergency defense (of the highest class: Heavy Laser Turret, Turbolaser, Ragnarok Beam, Railgun Turret, Heavy Razor Turret), and the rest of the slots filled with fighter bays so you can launch and maintain a large fleet.

> Low end "heavy CPU" have really bad CPU

This is intentional. Cruisers are effectively upgraded Destroyers and use the same weapons, but mostly forego medium weapons in favor on a larger number of heavy weapons. (That said, you say "really bad", but heavy CPU values are almost double the comparable medium-heavy CPU values.)

That is, of course, except for particularly heavy Cruisers like the Goddard, which is designed to use superheavy cores. There's nothing stopping a player from doing this with a Hawking or even a Kestrel, but for the role of light cruisers, I want the player to have to trade away power for that extra speed. In return, the weapons they use are also much less draining on the battery than most of the strongest weapons are, and they heat up more slowly as well.

> Fairly minor, but plasma Turret MK2 graphic

I forgot to mention this, but Plasma Turrets aren't balanced yet. They'll have the same stats as the Laser Turrets but inflict impact damage.

As for the shop graphics, that's outside the scope of this PR and predates it. :wink:

> Medium (light Cruiser): 350 CPU, 80 penetrate, 33 EPS, 55 DPS

Why so much CPU? I've already got weapons for medium-heavy ships and for light cruisers as I mentioned above. They use only 75 CPU. Perhaps I don't understand what "Mediuim (light Cruiser)" means?

bobbens:
> > "Milspec Orion 9901" and "Milspec Aegis 9801" are almost the same.
> 
> Aegis outfits have lower stats in general but better shielding (the same as how Prometheus specializes in battery and Hermes specializes in CPU). There's an argument to be had for redundancy (Orion is almost always the ideal choice), but the purpose of this to my mind is to offer more fine-tuned control over the stats of the ship (to pick what gives you the best bang for your buck, so to speak). I may look over the differential between specialized and Orion, though.
> 
> > "S&K Medium-Heavy Combat Plating" and "Unicorp B-24 Medium Plating"
> 
> Same as above, the different hull offerings are for fine-tuned customization. In the case of Unicorp outfits, they generally sacrifice capability for lower mass. Unicorp B- hulls also tend to have more cargo space than S&K combat plating. (Side note, I don't think I've changed hulls much, main thing I've adjusted hull-wise is the "medium" (Corvette) sized hulls. But I'm open to adjusting them to make the Unicorp B- hulls more desirable if you have an idea for what particular attribute would distinguish them. I personally don't really mind S&K hulls just being the best on the market for the most part and I also don't mind there being two Unicorp outfit versions. I've even found Unicorp B- hulls useful on occasion because of the lower mass. Perhaps that mass differentiation could be widened to increase that mass incentive, I'd be open to that.

I do feel like it would be better to have less fine-tuned control over the stats to force the player to make more choices. In particular, you have this high level of granularity with smaller ships, when there are, for example, no core outfits with 2000ish CPU, you have to chose between 800 and 3400, which is a bit weird. I really don't think 9 different core slots for each type/size are necessary. Looking at a fairly glaring example:

* Unicorp B-4 Light Plating (90 armour, 4 absorption, 10 cargo, 60 tons)
* Unicorp D-4 Light Plating (75 armour, 4 absorption, 9 cargo, 55 tons)
* S&K Light Stealth Plating (90 armour, 5 absorption, 6 cargo, 60 tons, 30% cloaking)

You literally choose between "15 armour, 1 cargo" for 5 tons, and "1 absorption, 30% cloaking, -4 cargo". I don't think such fine-grained control seems to do too much other than increase complexity somewhat artificially. In fact, I don't really see any reason to not choose the "S&K Light Stealth Plating" out of these three other than when you really need a bit more cargo space for some reason, but even then, there seem to be better alternatives for that.

> > No light "Heavy Plating", "Patchwork Heavy Plating" is actually lightest
> 
> I'm not sure what you mean by this. Patchwork Heavy Plating is only slightly lighter than the Unicorp D-48 Heavy Plating, and the Unicorp B-48 Heavy Plating provides a middle-ground between D-48 and S&K Heavy.
> 
> > No medium/heavy weapons (affects kestrel, which can't really make good use of medium slots)
> 
> What do you mean? Kestrel is specifically targeted by the Railgun. Or if you really want to use turrets on it, the general turret weapons designated for cruisers are the Heavy Ripper Turret and Heavy Ion Turret. Grave Beam will also be added to that list when I get to it. For medium slots, there's the Laser Turret MK2, Razor Turret MK2, Orion Beam, Turreted Vulcan Gun, and Enygma Systems Turreted Fury Launcher. All of these outfits mentioned are designed to be compatible both with cruisers and with medium-heavy ships like armored transports and destroyers.
> 
> > Low end "heavy CPU" have really bad CPU
> 
> This is intentional. Cruisers are effectively upgraded Destroyers and use the same weapons, but mostly forego medium weapons in favor on a larger number of heavy weapons. (That said, you say "really bad", but heavy CPU values are almost double the comparable medium-heavy CPU values.)

Well I was trying to equip a light carrier Kestrel. Idea being trying to keep it as fast and as light as possible, while having a few fighter bays and long range weapons. Right now I don't really see how that is possible. You basically have to choose between a "heavy capship" build, or basically downgrade to a destroyer build which ends up being basically a waste of all your large slots. To summarize, what I found is:

* You are basically forced into the 3400 CPU "heavy core outfit", because 800 does not allow you to equip anything. However, this is too heavy and it ends up making you go really over your engine mass limit, which makes you end up being slow. I really don't see why there is not a 2000ish CPU core outfit between the 800 and 3400 ones. You see this also happens with the light CPU cores which jump from 20 to 80ish, with nothing around 50. However, medium CPU cores do seem to be much more granular between 280 and 400.
* Similar with plating cores, you seem to have really fine granularity at certain values, while there are some clearly missing types that would add more diversity. In particular, the lightest "heavy plating core" is 720 (beat up one), Instead of having granularity between "Unicorp D-48" (1200 armour, 48 absorb, 70 cargo, 800 mass) and "Unicorp B-48" (1400 armour, 48 absorb, 80 cargo, 900 mass), with mere difference of 200 armour, 10 cargo, and 100 mass, having an option of a lightweight armour something like "Unicorp D-40 Lightweight Plating" (1100 armour, 40 absorb, 50 cargo, 500 mass) would give something that would allow you to make more interesting choices like sacrificing armour for mass without having to use a medium slot which would make you significantly weaker.
* While your heavy slots will be equipped with turbolasers and other very good weapons, you will end up with a lot of CPU left over, and have to use destroyer weapons (50 CPU and much weaker than heavy slots). Why not do something like the "Medium (Corvette): 100 CPU, 24 penetrate, 15 EPS, 32 DPS" slots which use a lot of CPU and make better use of the slot when you have CPU to spare? Adding a stronger medium slot weapon that uses 300 CPU, while not being as strong/efficient as heavy slot, would allow the kestrel to make somewhat decent use of the medium slots. Furthermore, it could also be used in destroyers, although most likely the CPU requirement would not allow them.

> > Giving fighter bay 450 CPU usage might be too low? Maybe balance mainly on mass (they are fairly heavy now)?
> 
> Well that CPU cost makes using fighter bays on anything with less than the heaviest CPUs impractical (and especially impractical for medium vessels). Of course it's easy for a cruiser to equip heavy stuff to afford not only fighter bays but also the more powerful capital-class weaponry, but that doesn't really make much sense to do as carrier-class ships have better armor and shield stats as a general rule (and if it's a player's only option, I don't see that as a problem).
> 
> It's also a bit lower than I'm going to set more specialized bays to.
> 
> Regarding your calculations of DPS, I'm not necessarily sure why you would consider that a problem in general. Carriers are supposed to be the heaviest and most powerful ships in the galaxy. The outfit setup I mean to encourage is two weapons for emergency defense (of the highest class: Heavy Laser Turret, Turbolaser, Ragnarok Beam, Railgun Turret, Heavy Razor Turret), and the rest of the slots filled with fighter bays so you can launch and maintain a large fleet.

I guess it would need testing to make sure that it is not a problem. If you can choose between 90 DPS that has shorter range, uses energy and lots of CPU, and something that does 200ish DPS, has longer range, and doesn't use energy for a bit more CPU, it might be a bit of a simple choice, especially when there are no clear downsides.

> > Medium (light Cruiser): 350 CPU, 80 penetrate, 33 EPS, 55 DPS
> 
> Why so much CPU? I've already got weapons for medium-heavy ships and for light cruisers as I mentioned above. They use only 75 CPU. Perhaps I don't understand what "Medium (light Cruiser)" means?

As mentioned above, the idea would be to do something similar to the "light slot" corvette weapons which use a lot of CPU but are one tier over the standard "light slot" weapons, but for **medium** slots. 

In general, I just feel like you have all these outfits that are very close to each other clustered together, then you have a really big leap to the next tier. For CPU it goes ~20, ~80, 400~500, 800ish, 3500ish. There are no real options for something like 200, or around 1600 to 2000ish. However, you have 3 variants of 18 CPU that differ in trivial values.

onpon4:
> You literally choose between "15 armour, 1 cargo" for 5 tons, and "1 absorption, 30% cloaking, -4 cargo". I don't think such fine-grained control seems to do too much other than increase complexity somewhat artificially. In fact, I don't really see any reason to not choose the "S&K Light Stealth Plating" out of these three other than when you really need a bit more cargo space for some reason, but even then, there seem to be better alternatives for that.

I mean, technically speaking players would be fine with no choice at all. You could scrap everything but one hull and one core system for each size. But at that point, what's the purpose of having core systems at all? You almost might as well go back to the Naev 0.5 model at that point.

> Well I was trying to equip a light carrier Kestrel. Idea being trying to keep it as fast and as light as possible, while having a few fighter bays and long range weapons. Right now I don't really see how that is possible.

That's a thing I'm specifically trying to discourage. I don't understand, do you want the player to just be able to competently apply whatever stats they want onto a ship? That creates the opposite of balance. A fast ship that can launch fighters is a serious problem because, as you noted, fighters are extremely strong and versatile. So if you want to make a Kestrel into a carrier, you have to make it heavy, and the way to do that is to use a superheavy core system (as you noted). This is by design.

If you really want to equip a Pirate Kestrel as a carrier, this would be how you would do it:

Screenshot from 2021-05-02 09-58-39

Notice the larger engines. Yes, it's slower than the Krain Remige Engine, but that's the point. To get the versatility of fighter bays, you have to sacrifice speed.

> In particular, the lightest "heavy plating core" is 720 (beat up one), Instead of having granularity between "Unicorp D-48" (1200 armour, 48 absorb, 70 cargo, 800 mass) and "Unicorp B-48" (1400 armour, 48 absorb, 80 cargo, 900 mass), with mere difference of 200 armour, 10 cargo, and 100 mass, having an option of a lightweight armour something like "Unicorp D-40 Light Plating" (1100 armour, 40 absorb, 50 cargo, 500 mass) would give something that would allow you to make more interesting choices like sacrificing armour for mass without having to use a medium slot which would make you significantly weaker.

Again, this would go entirely against the point of balancing. Light ships are meant to be light, fast, and with little armor. If you propose sacrificing speed, then what is the purpose of using a light ship? At that point you use a Corvette or Destroyer or Cruiser. Unless you propose allowing any ship to have any stats, which would make combat needlessly confusing (since you wouldn't know at a glance whether or not you can take something on, especially if you pilot a light ship, let alone how to best fight it).

1100 armor is a lot. Lightweight weapons simply cannot inflict damage fast enough to drain thatt to zero in a reasonable time. At 21 DPS minus 40% absorb (meaning it would actually only be about 13 DPS), it would take 87 seconds of constant, perfect fire to take out that amount of armor, which is so long that your weapons would heat up tremendously.

> While your heavy slots will be equipped with turbolasers and other very good weapons, you will end up with a lot of CPU left over (when using 3400 CPU core), and have to use destroyer weapons (50 CPU and much weaker than heavy slots). Why not do something like the "Medium (Corvette): 100 CPU, 24 penetrate, 15 EPS, 32 DPS" slots which use a lot of CPU and make better use of the slot when you have CPU to spare? Adding a stronger medium slot weapon that uses 300 CPU, while not being as strong/efficient as heavy slot, would allow the kestrel to make somewhat decent use of the medium slots. Furthermore, it could also be used in destroyers, although most likely the CPU requirement would not allow them.

A light cruiser ought to have some sort of long-range missiles, and that's normally going to be the best thing to put in the Kestrel's medium slots. As for the CPU, that can go towards reactors and other utilities. Here's an example of how I might equip a Kestrel normally:

Screenshot from 2021-05-02 10-20-38

It has leftover CPU, but that's fine; it's pushing its engine capacity to the limit.

> I guess it would need testing to make sure that it is not a problem. If you can choose between 90 DPS that has shorter range, uses energy and lots of CPU, and something that does 200ish DPS, has longer range, and doesn't use energy for a bit more CPU, it might be a bit of a simple choice, especially when there are no clear downsides.

Well that's the intention. The 350 CPU weapons are only meant to be equipped onto carriers as a backup in case the carrier gets too close to an opponent, or possibly to assist the fighters against other carriers. My idea for an ideal carrier is almost entirely fighter bays, with a pair of capital-class weapons.

> As mentioned above, the idea would be to do something similar to the "light slot" corvette weapons which use a lot of CPU but are one tier over the standard "light slot" weapons, but for medium slots.

I'm not opposed to that idea, but I'm not presently convinced that it would have much utility. I suppose the Mule would have a use for that, and maybe the Kestrel kind of would have a slight use for that, but it seems to me it would be a rather niche outfit. Maybe just a new beam weapon in between the Orion Beam and Grave Beam?

> In general, I just feel like you have all these outfits that are very close to each other clustered together, then you have a really big leap to the next tier. For CPU it goes ~20, ~80, 400~500, 800ish, 3500ish. There are no real options for something like 200, or around 1600 to 2000ish. However, you have 3 variants of 18 CPU that differ in trivial values.

This is an important part of how my rebalancing works, and is very much intentional. Specifically, there are massive CPU spikes when going from ultralight to light, from light to medium, and from heavy to superheavy. All of these jumps have a specific purpose:

* Fighters and bombers have much more CPU than ultralight fighters to cut off access to MK2 weapons for ultralight fighters.
* Corvettes have much more CPU than fighters and bombers to cut off access to Ripper Cannon/Shredder/Slicer/Plasma Cannon for fighters and bombers.
* Carriers and heavy cruisers have much more CPU than light cruisers, destroyers, and armored transports to cut off access to Repeating Railgun/Turbolaser/Heavy Razor Turret/Railgun Turret for light cruisers, destroyers, and armored transports.

You'll notice that in all cases, the cost of the outfits the CPU hikes are supposed to cut off contain large CPU usage spikes as well. Laser Cannon MK2 costs 5 times as much as Laser Cannon MK1, Ripper Cannon costs 4 times as much as Laser Cannon MK2, and Repeating Railgun costs 5 times as much as Railgun. In fact in all cases, you can just barely manage to fit one of these outfits on a ship one tier below what it's supposed to go on, but in doing so you sacrifice flexibility as it consumes all the CPU you have available (and it also requires a Hermes core system in most cases).

If I made the transition between classes granular, it would reintroduce the problem that I originally sought to solve with this rebalancing work, where smaller weapons become useless as it is always ideal to equip the strongest weapons you can possibly use without sacrificing the primary advantage of the ship type you're flying.

🕵️

bobbens:
> > You literally choose between "15 armour, 1 cargo" for 5 tons, and "1 absorption, 30% cloaking, -4 cargo". I don't think such fine-grained control seems to do too much other than increase complexity somewhat artificially. In fact, I don't really see any reason to not choose the "S&K Light Stealth Plating" out of these three other than when you really need a bit more cargo space for some reason, but even then, there seem to be better alternatives for that.
> 
> I mean, technically speaking players would be fine with no choice at all. You could scrap everything but one hull and one core system for each size. But at that point, what's the purpose of having core systems at all? You almost might as well go back to the Naev 0.5 model at that point.

There is a pretty big difference between remove redundant/inconsequential choices and removing all choice. I don't understand how you jump to the conclusion of removing all choice from my comment where I say that rather than having almost indistinguishable outfits that they should have more diversity.

> > Well I was trying to equip a light carrier Kestrel. Idea being trying to keep it as fast and as light as possible, while having a few fighter bays and long range weapons. Right now I don't really see how that is possible.
> 
> That's a thing I'm specifically trying to discourage. I don't understand, do you want the player to just be able to competently apply whatever stats they want onto a ship? That creates the opposite of balance. A fast ship that can launch fighters is a serious problem because, as you noted, fighters are extremely strong and versatile. So if you want to make a Kestrel into a carrier, you have to make it heavy, and the way to do that is to use a superheavy core system (as you noted). This is by design.

I don't understand why you are doing a straw man argument. I am not arguing for allowing the player to do whatever they want, I am just arguing that you are removing flexibility by forcing the player to fit into some pre-defined roles. I really don't see how introducing a 1600 CPU core slot would break anything. In particular, using only 800 CPU and 3400 CPU tiers makes it so currently almost all the capships will go with a supercapship due to having too much inherent mass to be agile at all and thus make use of an 800 CPU core. The only ships that seem to have an option for light cruiser builds would be hawking and kestrel I guess, but even the hawking might be pushing it with 1800 mass. Having a 1600 CPU option would let you go somewhat inbetween, but in the case of the Hawking it would heavily limit your weapon set.

I can't think of a single case where a 1600 CPU core outfit would allow for any build that goes against what you are trying to do in your rebalance. A kestrel with 2 fighter bays (it wouldn't be able to fit more) using a 1600 CPU core outfit would end up having fairly low firepower and still being significantly slower than destroyers. However, it would allow for another playstyle which I see no reason to forbid.

> > In particular, the lightest "heavy plating core" is 720 (beat up one), Instead of having granularity between "Unicorp D-48" (1200 armour, 48 absorb, 70 cargo, 800 mass) and "Unicorp B-48" (1400 armour, 48 absorb, 80 cargo, 900 mass), with mere difference of 200 armour, 10 cargo, and 100 mass, having an option of a lightweight armour something like "Unicorp D-40 Light Plating" (1100 armour, 40 absorb, 50 cargo, 500 mass) would give something that would allow you to make more interesting choices like sacrificing armour for mass without having to use a medium slot which would make you significantly weaker.
> 
> Again, this would go entirely against the point of balancing. Light ships are meant to be light, fast, and with little armor. If you propose sacrificing speed, then what is the purpose of using a light ship? At that point you use a Corvette or Destroyer or Cruiser. Unless you propose allowing any ship to have any stats, which would make combat needlessly confusing (since you wouldn't know at a glance whether or not you can take something on, especially if you pilot a light ship, let alone how to best fight it).

Again, I feel you are just jumping to conclusions. Giving an option for a "lightweight" heavy armour does not mean a light cruiser will suddenly and magically fly like a corvette. In fact, unless you basically sacrifice everything, chances are you won't overcome the inherent 1400 mass (which is 1000 mass over a pacifier) of a kestrel. The lower absorption in combination with lower armour would also make the cruiser a much larger target for destroyers, which again, would be faster in general. Having alternatives like this would allow for more diverse builds _without_ going against the point of balancing.

> 1100 armor is a lot. Lightweight weapons simply cannot inflict damage fast enough to drain thatt to zero in a reasonable time. At 21 DPS minus 40% absorb (meaning it would actually only be about 13 DPS), it would take 87 seconds of constant, perfect fire to take out that amount of armor, which is so long that your weapons would heat up tremendously.

I don't see how that is a problem given that it would be a "heavy" slot. You are calculating how much time it would take a single ultra-light weapon to take down the armour of a cruiser. I don't see how that example is representative of anything. If you have 10 hyenas with 3 ultra-light weapons each that would be 3 seconds. Giving them non-ultra light weapons (which would be the fairly obvious thing to do if they were going to take down a cruiser) would make it even faster.

> This is an important part of how my rebalancing works, and is very much intentional. Specifically, there are massive CPU spikes when going from ultralight to light, from light to medium, and from heavy to superheavy. All of these jumps have a specific purpose:
>     * Fighters and bombers have much more CPU than ultralight fighters to cut off access to MK2 weapons for ultralight fighters.
>     * Corvettes have much more CPU than fighters and bombers to cut off access to Ripper Cannon/Shredder/Slicer/Plasma Cannon for fighters and bombers.
>     * Carriers and heavy cruisers have much more CPU than light cruisers, destroyers, and armored transports to cut off access to Repeating Railgun/Turbolaser/Heavy Razor Turret/Railgun Turret for light cruisers, destroyers, and armored transports.
> 
> You'll notice that in all cases, the cost of the outfits the CPU hikes are supposed to cut off contain large CPU usage spikes as well. Laser Cannon MK2 costs 5 times as much as Laser Cannon MK1, Ripper Cannon costs 4 times as much as Laser Cannon MK2, and Repeating Railgun costs 5 times as much as Railgun. In fact in all cases, you can just barely manage to fit one of these outfits on a ship one tier below what it's supposed to go on, but in doing so you sacrifice flexibility as it consumes all the CPU you have available (and it also requires a Hermes core system in most cases).

I find that your current system works exceptionally well for smaller ships because of how they can have light weapons with medium core slots. In particular, ships like the Admonisher feel like they are at a really fun and interesting point of outfitting. You can adapt it to quite a few roles, and at the same time it doesn't steal the role of other ship types, e.g., it will not outgun a destroyer nor outmaneuver a fighter. This is also likely due to the fact that medium slots have more diversity, especially with CPU being fairly granular. I would like this level of choice and diversity to be seen in all ships.

> If I made the transition between classes granular, it would reintroduce the problem that I originally sought to solve with this rebalancing work, where smaller weapons become useless as it is always ideal to equip the strongest weapons you can possibly use without sacrificing the primary advantage of the ship type you're flying.

Correct me if I'm wrong, but the only exception I can think of for light ships is I don't really see the role of "ultra-light fighters" given that in general you can just equip them to be "light fighters" without any downside. What would be the reason to equip the lower end light outfits? It seems like the speed gains are really marginal compared to the loss in armour and shooting power.

While I do think the system works great for ships like destroyers/corvettes, I do feel that the system sort of breaks down a bit with capital ships because they basically end up fairly homogeneous. You have lots of heavy slots and that is it. The only exception is the Kestrel which has fewer slots and still some medium slots that do make it more interesting, however, with only light cruiser and supercapship cores available, and the fact that the kestrel as a supercapship will get outclassed by pretty much any other supercapship, 

Again, I am not saying what you are doing is flawed, it works very well in many cases, I just believe it could be made better. I am just saying:
* I do not see the benefit in having so many very similar core outfits. I feel that it just leads to confusion and choice paralysis while at the same time not giving any real meaningful choice. This could be mitigated by making a much easier to compare and use interface, but I do feel like it's trying to kill a fly with a cannon.
* I don't see the necessity for some of the large jumps in tiers, and in particular, for "heavy" ships, it does seem like they are basically being limited to two playstyles. I don't see how adding some other inbetween stuff, say 1600 CPU core, would go against the balancing, while it would also allow doing other builds without being forced to be either a very light cruiser or a supercapship.

And to reiterate, I am not saying "one outfit per tier" nor am I saying "super fine granularity so that the player can do anything". I don't think it is necessary to go to one extreme or the other.

onpon4:
> There is a pretty big difference between remove redundant/inconsequential choices and removing all choice. I don't understand how you jump to the conclusion of removing all choice from my comment where I say that rather than having almost indistinguishable outfits that they should have more diversity.

Well what do you propose exactly?

> I am not arguing for allowing the player to do whatever they want, I am just arguing that you are removing flexibility by forcing the player to fit into some pre-defined roles. I really don't see how introducing a 1600 CPU core slot would break anything. In particular, using only 800 CPU and 3400 CPU tiers makes it so currently almost all the capships will go with a supercapship due to having too much inherent mass to be agile at all and thus make use of an 800 CPU core. The only ships that seem to have an option for light cruiser builds would be hawking and kestrel I guess, but even the hawking might be pushing it with 1800 mass. Having a 1600 CPU option would let you go somewhat inbetween, but in the case of the Hawking it would heavily limit your weapon set.

This is an example of what I consider a well-equipped Hawking:

Screenshot from 2021-05-02 12-40-56

The Hawking is largely what I base these stats on. Of course not every cruiser is going to want to be a light cruiser; the Goddard, in particular, is designed to be a heavy cruiser with equipment more like this:

Screenshot from 2021-05-02 12-49-36

As for ships other than the Hawking and Kestrel not supporting light cruiser builds, that's a problem with those ships, no? I'm using the Hawking and Kestrel as a base, and the plan is to adjust ships to meet these criteria afterward.

> Correct me if I'm wrong, but the only exception I can think of for light ships is I don't really see the role of "ultra-light fighters" given that in general you can just equip them to be "light fighters" without any downside. What would be the reason to equip the lower end light outfits?

Because they're faster.

Hyena and Shark are classed as "ultralight" under my system, as is the regular Collective Drone. Here's some examples of Hyena and Shark builds:

Screenshot from 2021-05-02 12-54-32
Screenshot from 2021-05-02 12-57-54

Yes, you can equip them as regular light too, but if you did that then you would have a slower ship without the benefit of the medium slot for missiles that the Lancelot has, or the many slots for heavy firepower like the Vendetta has.

> While I do think the system works great for ships like destroyers/corvettes, I do feel that the system sort of breaks down a bit with capital ships because they basically end up fairly homogeneous. You have lots of heavy slots and that is it. The only exception is the Kestrel which has fewer slots and still some medium slots that do make it more interesting, however, with only light cruiser and supercapship cores available, and the fact that the kestrel as a supercapship will get outclassed by pretty much any other supercapship,

I really have no clue what you're talking about here. Yes, the largest ships have only large slots, but the smallest ships also only have small slots, so what's the difference?

If you're referring to the lack of incentive to build a light cruiser, one possible solution to that is just to make superheavy engines slower. If making a speed vs. power differential like I'm trying to make is impractical, then I think that would suggest that attempting to have light cruisers is a mistake, but I don't think that's the case. Light cruisers can, in particular, occupy an important niche of existing to destroy carriers to stop them from continuing to launch fighters. To do that, they need to have significantly more speed than the carrier so they can catch up, but they also need heavy firepower, armor, and shields so they can contend with them. Heavy cruisers, on the other hand, serve roughly the same niche as carriers, just in a different way. Instead of launching fighters and trying to stay away from they action, they take a direct role in trying to provide heavy fire support.

> I do not see the benefit in having so many very similar core outfits. I feel that it just leads to confusion and choice paralysis while at the same time not giving any real meaningful choice. This could be mitigated by making a much easier to compare and use interface, but I do feel like it's trying to kill a fly with a cannon.

Well if the number per size you want is more than one, how many do you want, and how do you propose to differentiate them?

> I don't see the necessity for some of the large jumps in tiers, and in particular, for "heavy" ships, it does seem like they are basically being limited to two playstyles. I don't see how adding some other inbetween stuff, say 1600 CPU core, would go against the balancing, while it would also allow doing other builds without being forced to be either a very light cruiser or a supercapship.

So you're looking to be able to do something like this, but with sufficient CPU?

Screenshot from 2021-05-02 13-27-36

The reason this goes against the balancing is because light cruisers aren't supposed to really have access to the Heavy Laser Turret, let alone several Heavy Laser Turrets.

I'm open to increasing the CPU bonus you get for the Milspec Hermes system, but 1600 is pushing it a bit far. 950 or 1000 might be fine. Push it to 1050 and this would be possible:

Screenshot from 2021-05-02 13-33-56

I just tested that hypothetical, and it leads to a ship which can lead a much more devastating barrage, but only for a very short period of time before running out of battery, so I'm ok with this trade-off, I think. I think I'll actually make that further CPU increase official.

onpon4:
Ok, so having checked something, I do need to adjust masses (to prevent above-tier core systems from being really viable), so I've started doing so.

bobbens:
> > There is a pretty big difference between remove redundant/inconsequential choices and removing all choice. I don't understand how you jump to the conclusion of removing all choice from my comment where I say that rather than having almost indistinguishable outfits that they should have more diversity.
> 
> Well what do you propose exactly?

An example (light structural core slots) would be converting:

* Patchwork Light Plating          24 mass, 0 absorb, 35 armour, 3 cargo
* Unicorp D-2 Light Plating        27 mass, 2 absorb, 45 armour, 4 cargo
* Unicorp B-2 Light Plating        30 mass, 2 absorb, 50 armour, 5 cargo
* S&K Ultralight Stealth Plating   30 mass, 3 absorb, 55 armour, 3 cargo, 30% cloak
* S&K Light Stealth Plating        60 mass, 5 absorb, 95 armour, 6 cargo, 30% cloak
* S&K Ultralight Combat Plating    40 mass, 3 absorb, 65 armour, 4 cargo
* S&K Small Cargo Hull             40 mass, 2 absorb, 65 armour, 30 cargo
* Unicorp D-4 Light Plating        55 mass, 4 absorb, 75 armour, 9 cargo
* Unicorp B-4 Light Plating        60 mass, 4 absorb, 90 armour, 10 cargo
* S&K Light Combat Plating         80 mass, 6 absorb, 110 armour, 8 cargo

to

* Patchwork Light Plating          35 mass, 0 absorb, 35 armour, 5 cargo
* S&K Ultralight Stealth Plating   20 mass, 1 absorb, 40 armour, 3 cargo, 30% cloak
* Unicorp B-2 Light Plating        30 mass, 2 absorb, 50 armour, 10 cargo
* S&K Ultralight Combat Plating    40 mass, 3 absorb, 70 armour, 2 cargo, 5% heat dissipation
* S&K Small Cargo Hull             40 mass, 2 absorb, 60 armour, 30 cargo, 25% cargo inertia
* Unicorp B-4 Light Plating        60 mass, 4 absorb, 90 armour, 14 cargo
* S&K Light Combat Plating         80 mass, 6 absorb, 110 armour, 4 cargo, 5% heat dissipation

You have more clear roles and less overlap with the outfits, nothing really is a direct change for each other. I especially don't see a point in having Unicorp B and D series which are virtually the same.
 
> If you're referring to the lack of incentive to build a light cruiser, one possible solution to that is just to make superheavy engines slower. If making a speed vs. power differential like I'm trying to make is impractical, then I think that would suggest that attempting to have light cruisers is a mistake, but I don't think that's the case. Light cruisers can, in particular, occupy an important niche of existing to destroy carriers to stop them from continuing to launch fighters. To do that, they need to have significantly more speed than the carrier so they can catch up, but they also need heavy firepower, armor, and shields so they can contend with them. Heavy cruisers, on the other hand, serve roughly the same niche as carriers, just in a different way. Instead of launching fighters and trying to stay away from they action, they take a direct role in trying to provide heavy fire support.
> 
> > I do not see the benefit in having so many very similar core outfits. I feel that it just leads to confusion and choice paralysis while at the same time not giving any real meaningful choice. This could be mitigated by making a much easier to compare and use interface, but I do feel like it's trying to kill a fly with a cannon.
> 
> Well if the number per size you want is more than one, how many do you want, and how do you propose to differentiate them?

See above for an example (manually done), but some sort of "rule of thumb" would be to have them be over 40% different. Going with the following as an example:

* Unicorp D-2 Light Plating        27 mass, 2 absorb, 45 armour, 4 cargo
* Unicorp B-2 Light Plating        30 mass, 2 absorb, 50 armour, 5 cargo

If we make the values relative to the Unicorp D-2 Light Plating, we would get the following:

* Unicorp D-2 Light Plating        100% mass, 100% absorb, 100% armour, 100% cargo
* Unicorp B-2 Light Plating        111% mass, 100% absorb, 111% armour, 125% cargo

If you average this, you get that the Unicorp B-2 Light Plating is roughly "112%" of the D-2 one for a difference of `12%`. Note this would not be a hard rule, but a rule of thumb, so to make it more distinctive you would have to change stuff around. For reference comparing with `Unicorp B-4 Light Plating` would give a difference of "243%" or a difference of `143%` with the D-2 plating (and about the same with the B-2 one too).

> So you're looking to be able to do something like this, but with sufficient CPU?
> 
> The reason this goes against the balancing is because light cruisers aren't supposed to really have access to the Heavy Laser Turret, let alone several Heavy Laser Turrets.

I don't see why you take this dogmatic stance here. With a 1600 CPU core slot you could have something like a medium cruiser, which would stand between a light cruiser and supercapship. It would neither have the health and firepower of a supercapship nor the speed of a light cruiser. It's not like if a hawking or kestrel can suddenly equip two or three turbolasers it is going to go one-on-one with a goddard. I really don't see why it has to be either one or the other.

> I'm open to increasing the CPU bonus you get for the Milspec Hermes system, but 1600 is pushing it a bit far. 950 or 1000 might be fine. Push it to 1050 and this would be possible:
> 
> I just tested that hypothetical, and it leads to a ship which can lead a much more devastating barrage, but only for a very short period of time before running out of battery, so I'm ok with this trade-off, I think. I think I'll actually make that further CPU increase official.

That's the point that I'm making. Adding another tier would not suddenly break your entire balance because even if you can equip more powerful weapons, you would still lack the energy and such.

Looking over the heavy core slots, it does seem like energy regeneration doesn't follow the trend you see in smaller slots. Although CPU usage goes up 4x from 800 to 3400 in one tier, energy regen goes from 130 to 160 or so. Is there any reason for this? Increasing energy usage of turbolaser and high-end weapons while increasing the energy regen to 200+ for supercap slots would allow for more control of the turbolasers and other superheavy weapons.

With the above energy regeneration modification, something like a 120 energy / 2000 CPU core outfit could be added. While it would allow you to equip much better weapons (although probably only half the slots), you would still be limited by energy usage. This could allow for niche builds like either light carrier or some sort of hit-and-run cruiser. It would also not bake the 800 CPU ones obsolete because they would be much lighter and would be essential in light cruiser builds.

onpon4:
> Looking over the heavy core slots, it does seem like energy regeneration doesn't follow the trend you see in smaller slots. Although CPU usage goes up 4x from 800 to 3400 in one tier, energy regen goes from 130 to 160 or so. Is there any reason for this?

The values are based on growth somewhere between linear and exponential of what you might call "net energy regeneration", i.e. energy regeneration minus what you can expect engines to consume in general. Looking at my spreadsheet, I came up with the following base net energy generation values which the actual values reflect:

* Ultralight: 11-12
* Light: 22-24
* Medium: 38-40
* Medium-heavy: 64-66
* Heavy: 97-101
* Superheavy: 110-120

When you look at it that way, the biggest leap is actually between Medium-Heavy and Heavy. I believe the reason for that large leap is mainly because light Cruisers and Destroyers are meant to use the same weapons, but Cruisers need to be able to have more destructive power. That said, these numbers are largely arbitrary. Energy is a thing where I focused on designing the energy consumption of weapons around the energy generation values I picked, and the primary focus is on typical layouts generating about half as much surplus energy as the energy consumption of the weapons used.

🕵️

onpon4:
> I don't see why you take this dogmatic stance here. With a 1600 CPU core slot you could have something like a medium cruiser, which would stand between a light cruiser and supercapship. It would neither have the health and firepower of a supercapship nor the speed of a light cruiser.

If you're willing to sacrifice speed, you can do this:

Screenshot from 2021-05-02 23-48-44

I think calling this a "medium cruiser" would be perfectly fair. It's slightly faster than a heavy cruiser would be, but slightly less armored than it would be.

The important thing here is that you can get this "medium cruiser" effect by mixing and matching different core systems, half heavy and with just a superheavy core system. So if there were cores specifically designed for "medium cruisers", they would instead move the center toward "medium light cruisers", which likely would be pretty much identical to what I have here but without the sacrifice to speed. (Note that I actually had a very large amount of wiggle-room on mass; 200 tonnes could be cut easily just by switching to Unicorp B-48 plating at only a slight armor loss.) And if you can do that, that raises the question, what is the benefit of a heavy cruiser?

If anything, the amount of wiggle-room available with the Melendez Mammoth already suggests that either the weapons should be heavier or the mass limit gain of the Melendez Mammoth should be less.

🕵️

bobbens:
> When you look at it that way, the biggest leap is actually between Medium-Heavy and Heavy. I believe the reason for that large leap is mainly because light Cruisers and Destroyers are meant to use the same weapons, but Cruisers need to be able to have more destructive power. That said, these numbers are largely arbitrary. Energy is a thing where I focused on designing the energy consumption of weapons around the energy generation values I picked, and the primary focus is on typical layouts generating about half as much surplus energy as the energy consumption of the weapons used.

Would there be any downside to increasing energy usage of superheavy weapons and regeneration of superheavy cores? This would allow medium cruisers without them being able to replace heavy cruisers, which I believe is your main worry.

> I think calling this a "medium cruiser" would be perfectly fair. It's slightly faster than a heavy cruiser would be, but slightly less armored than it would be.

I disagree. It is just a "fastish" heavy cruiser. The entire weapon loadout is basically superheavy from the looks of it. You are only sacrificing a bit of armour for speed.

> The important thing here is that you can get this "medium cruiser" effect by mixing and matching different core systems, half heavy and with just a superheavy core system. So if there were cores specifically designed for "medium cruisers", they would instead move the center toward "medium light cruisers", which likely would be pretty much identical to what I have here but without the sacrifice to speed. (Note that I actually had a very large amount of wiggle-room on mass; 200 tonnes could be cut easily just by switching to Unicorp B-48 plating at only a slight armor loss.) And if you can do that, that raises the question, what is the benefit of a heavy cruiser?

As I said, you can't really make a medium cruiser when you have only 800 CPU and 3400 CPU slots. If you go 3400 CPU, you end up with all your weapon slots with superheavy weapons, while if you go 800 CPU, you have all your slots with heavy weapons. No cores forces you to mix both superheavy and heavy, and if you do mix superheavy and heavy, you are just wasting the potential of a core slot which winds up with overdimensioned CPU and energy. The closest you get to a medium cruiser is your example above, which is just a heavy cruiser that trades a bit of armour for speed.

If medium cruiser exist, they would not replace the heavy cruiser either because they would have both less DPS given the lower energy and less health overall, for speed. This is **literally** the same that there is with super-light and light fighters. If you don't have a problem with how that is done, I don't see why you have a problem with having medium cruisers.

onpon4:
> Would there be any downside to increasing energy usage of superheavy weapons and regeneration of superheavy cores? This would allow medium cruisers without them being able to replace heavy cruisers, which I believe is your main worry.

I do think it's likely that energy usage of superheavy turreted weapons may need to be increased further, though I don't think I want to increase energy generation in tandem. I'll have to do further testing to verify.

> If medium cruiser exist, they would not replace the heavy cruiser either because they would have both less DPS given the lower energy and less health overall, for speed. This is literally the same that there is with super-light and light fighters. If you don't have a problem with how that is done, I don't see why you have a problem with having medium cruisers.

Well no, it's not the same because for fighters there's two tiers, and for heavy there's two tiers, but you're proposing adding a third heavy tier. The closest thing to a "medium fighter" is maybe the Pirate Shark, since it blurs the line between fighter and ultralight fighter a bit. That said, the Pirate Shark is still and ultralight fighter with the usual ultralight fighter downsides; it just has an extra weapon slot, and that extra weapon slot makes it more plausible to use as something like a "medium fighter". This is more or less the same thing as what I pointed out is possible with a Hawking.

You say that doesn't count as a "medium cruiser", but I rather view that distinction as arbitrary. What I'm trying to point out is that "medium cruiser" isn't an intentional ship class, but it's perfectly possible for a determined player to twist things and get a ship that's not exactly a heavy cruiser and not exactly a light cruiser.

I don't think a "medium cruiser" tier is needed or desirable. If anything, I think that even having the "light/heavy cruiser" distinction is a bit much and if I were to change tiers, it would be to remove the "light cruiser" tier entirely. But that would entail throwing away the Kestrel as a useful ship and, honestly, making most cruisers simply a different name for the same kind of body a carrier is. So I'm keeping the heavy/superheavy distinction and trying to amplify that by giving light cruisers more of a purpose. Creating cores that go in between these two tiers would only get in the way of that and, I believe, most likely just make the light cruiser class a sub-optimal choice again.

🕵️

bobbens:
> > If medium cruiser exist, they would not replace the heavy cruiser either because they would have both less DPS given the lower energy and less health overall, for speed. This is literally the same that there is with super-light and light fighters. If you don't have a problem with how that is done, I don't see why you have a problem with having medium cruisers.
> 
> Well no, it's not the same because for fighters there's two tiers, and for heavy there's two tiers, but you're proposing adding a third heavy tier. The closest thing to a "medium fighter" is maybe the Pirate Shark, since it blurs the line between fighter and ultralight fighter a bit. That said, the Pirate Shark is still and ultralight fighter with the usual ultralight fighter downsides; it just has an extra weapon slot, and that extra weapon slot makes it more plausible to use as something like a "medium fighter". This is more or less the same thing as what I pointed out is possible with a Hawking.

For an all light slot fighter, you can only equip light or ultra-light weapons. The difference between both is fairly minimal (21→23 DPS and 0→12 penetrate). In the case of an all heavy slot cruiser, you have heavy and superheavy weapons. The difference is pretty large between them (45→70 DPS, 72→100 penetrate). So while a medium fighter would not really bring much to the table, a third tier would give a significant and new alternative.

Again, going to your Hawking example. If instead of your "medium cruiser" you do a "heavy cruiser" build, how much does it differ? I feel like it'll be a couple of slots at most. That is not giving any choice to the player.

> You say that doesn't count as a "medium cruiser", but I rather view that distinction as arbitrary. What I'm trying to point out is that "medium cruiser" isn't an intentional ship class, but it's perfectly possible for a determined player to twist things and get a ship that's not exactly a heavy cruiser and not exactly a light cruiser.

I do get that it's not an intentional ship class, because the only intention I can see is that it shouldn't be possible.

> I don't think a "medium cruiser" tier is needed or desirable. If anything, I think that even having the "light/heavy cruiser" distinction is a bit much and if I were to change tiers, it would be to remove the "light cruiser" tier entirely. But that would entail throwing away the Kestrel as a useful ship and, honestly, making most cruisers simply a different name for the same kind of body a carrier is. So I'm keeping the heavy/superheavy distinction and trying to amplify that by giving light cruisers more of a purpose. Creating cores that go in between these two tiers would only get in the way of that and, I believe, most likely just make the light cruiser class a sub-optimal choice again.

The difference between heavy and superheavy is already the largest of all outfits "tiers". I don't see how it needs to be amplified.

Anyway, since I don't think this is really going anyway, and I do feel like you are not really interested in my feedback, so I will look forward to and wait until you finish before testing and trying out new things.

onpon4:
> For an all light slot fighter, you can only equip light or ultra-light weapons. The difference between both is fairly minimal (21→23 DPS and 0→12 penetrate). In the case of an all heavy slot cruiser, you have heavy and superheavy weapons. The difference is pretty large between them (45→70 DPS, 72→100 penetrate). So while a medium fighter would not really bring much to the table, a third tier would give a significant and new alternative.

The reason damage increase is so huge between heavy and superheavy classes is because the expectation is for only a very small number of those weapons to be used. This is also why energy consumption is so much higher. But since I haven't thoroughly tested yet, it's very possible that I need to adjust things here and possibly even nerf the superheavy turrets. I'm already pretty sure I need to increase the energy consumption of superheavy turrets further and probably make them significantly heavier. I get the impression now that you must have noticed such a balancing problem.

> Anyway, since I don't think this is really going anyway, and I do feel like you are not really interested in my feedback, so I will look forward to and wait until you finish before testing and trying out new things.

I'm interested in feedback, but when all you give me is suggestions of what to do without explaining what problem those suggestions are meant to solve, it's hard to have a productive discussion.

That said, I understand now that my attempts to understand your reasoning by making assumptions were counterproductive, and I apologize for doing so. I will try to make a greater effort to learn the reasoning behind a suggestion next time.

🕵️

onpon4:
Ok, all the initial balance adjustments are complete and this is now fully testable.

I've merged this into Naikari and will do more thorough testing there. When I'm pretty sure with the results I'll mark this as ready for review. In the meantime, feel free to test.

🕵️

onpon4:
This branch has already been merged into https://github.com/naev/naev/tree/rebalance, which has substantially diverged from this since then, so I'm going to consider this PR obsolete and close it.

🕵️

PR #1665, opened 2021-01-16
use the PD controller for escort docking

nenau-again:
So that one can get one's escort to return to carrier while moving.


UncombedCoconut:
What an odd thing to want! Thanks. :)

PR #1667, opened 2021-01-17
consider hidden argument in getNextSystem

nenau-again:
getNextSystem didn't use this argument although it was defined


UncombedCoconut:
Cool -- looks like nothing was using it yet, so this is a pure correctness improvement.

PR #1670, opened 2021-01-18
Drop npng's vflip feature; instead, transform our texture matrices.

UncombedCoconut:
I think this is a good idea because it makes npng more replaceable. (For instance, it's almost possible to drop in SDL_image and gain support for whatever image format a mod or script author might want to use.)
I want to expose the change to feedback before committing; otherwise it might get confusing.


UncombedCoconut:
(accidentally presented a version a bad debug-print, now removed)

PR #1673, opened 2021-01-21
Fix Issues Launching Naev on Windows via Itch.io Client

ProjectSynchro:
Found an interesting issue when trying to play Naev through itch.io's desktop client. 

The default behaviour for the client is to launch the first executable available, and for previous releases this works fine, but in the latest nightlies (since blackjack and other love2d stuff was added to master) the itch client was actually trying to launch the love2d bundle directly.. (and to answer your question the bundle does launch via love2d assuming it's on PATH but fails to do anything exciting see: [here](https://cdn.discordapp.com/attachments/785497094366035989/801633532271132732/love_759VA4q1ee.png)) 

This PR adds an itch.io manifest to launch the correct binary for each respective supported platform (linux, windows, macos) by copying a master file with some placeholder text and uses sed to configure it correctly.
Documentation for manifests is [here](https://itch.io/docs/itch/integrating/manifest-actions.html)

Another plus is that we are now technically shipping '[Platinum tier](https://itch.io/docs/itch/integrating/compatibility-policy.html)' builds to itch.io everywhere

### DIFF
--- a/.github/workflows/naev_nightly.yml
+++ b/.github/workflows/naev_nightly.yml
@@ -28,10 +28,11 @@ jobs:
 
       - name: Collect Artifacts
         run: |
-          mkdir -p build/dist/steam
+          mkdir -p build/dist/{steam,itch}
           mv build/meson-dist/naev-*.tar.xz build/dist/source.tar.xz
           cp source/dat/VERSION build/dist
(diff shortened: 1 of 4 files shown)


bobbens:
Looks good, thanks!

PR #1682, opened 2021-01-24
[WIP] Tracks

nenau-again:
Ships leave trails.
Work in progress, with loads of commentarized shit all around, but it should be testable.
About half of the ships have their tracks that come roughly from the engine. The other half was not adressed yet.


src/pilot.c
```diff
@@ -2711,6 +2806,11 @@ void pilot_init( Pilot* pilot, Ship* ship, const char* name, int faction, const
    if (ai != NULL)
       ai_pinit( pilot, ai ); /* Must run before ai_create */
    pilot->shoot_indicator = 0;
+
+   /* Animated track. */
+   pilot->track = array_create( Vector2d );
```
bobbens, line 2811: Should be grouped into a single array with a struct.

src/pilot.c
```diff
@@ -2195,9 +2245,54 @@ void pilot_update( Pilot* pilot, const double dt )
    gl_getSpriteFromDir( &pilot->tsx, &pilot->tsy,
          pilot->ship->gfx_space, pilot->solid->dir );
 
-   /* See if there is commodities to gather */
+   /* See if there is commodities to gather. */
    gatherable_gather( pilot->id );
 
+   /* Update the track. */
+   if (!space_isSimulation()) {
+      now = ntime_get();
```
bobbens, line 2253: Should try to use `dt` as much as possible and not ntime_get().

src/pilot.c
```diff
@@ -1701,7 +1701,54 @@ void pilot_explode( double x, double y, double radius, const Damage *dmg, const
...
+      gl_drawTrack( x1, y1, x2, y2, now, p->times[ array_size(p->times)-2 ], now, &c1, &c2 );
+
+      if (array_size(p->times) > 2){
+         for ( i=array_size(p->times)-2; i>0; i--){
+            point = &p->track[i];
+            gl_gameToScreenCoords( &x1, &y1, point->x, point->y );
+            point = &p->track[i-1];
+            gl_gameToScreenCoords( &x2, &y2, point->x, point->y );
+            //gl_drawLine( x1, y1, x2, y2, &c );
+            gl_drawTrack( x1, y1, x2, y2, p->times[i], p->times[i-1], now, &p->colors[i], &p->colors[i-1] );
```
bobbens, line 1748: Tracks should be all batched into a single call.

src/space.c
```diff
@@ -1475,6 +1475,15 @@ void space_update( const double dt )
 }
 
 
+/**
+ * @brief returns wether we're just simulating.
+ */
+unsigned int space_isSimulation( void )
```
bobbens, line 1481: Don't think this is necessary if instead of using `ntime_t now;` you use the `dt` directly (which is only >0. when simulating).

src/pilot.c
```diff
@@ -1701,7 +1701,54 @@ void pilot_explode( double x, double y, double radius, const Damage *dmg, const
...
+      /* Lastly created control point is replaced by ship's position */
+      gl_gameToScreenCoords( &x1, &y1, p->solid->pos.x + dx, p->solid->pos.y + dy*M_SQRT1_2 + p->ship->h_engine*M_SQRT1_2 );
+      point = &p->track[ array_size(p->track)-2 ];
+      gl_gameToScreenCoords( &x2, &y2, point->x, point->y );
+      //gl_drawLine( x1, y1, x2, y2, &c );
+      gl_renderCross( x1, y1, 12., &cWhite );
+
+      /* Set the colour. TODO: function */
+      if (pilot_isFlag(p, PILOT_AFTERBURNER))
+         c1 = cGold;
```
bobbens, line 1730: probably red will be better (looks less like portal), but I guess it needs testing

src/ship.c
```diff
@@ -926,6 +931,12 @@ static int ship_parse( Ship *temp, xmlNodePtr parent )
       xmlr_strd(node,"fabricator",temp->fabricator);
       xmlr_strd(node,"description",temp->description);
       xmlr_int(node,"rarity",temp->rarity);
+      if (xml_isNode(node,"engine_offset")) {
```
bobbens, line 934: Maybe call it `engine_track` to make it more clear it's visual only?

bobbens (commented):
Just giving some feedback ;)

src/ship.c
nenau-again: I'll revamp this part : there will be a node `track_generators`, containing one or more `track_generator`
Each one will have its offset and colour set (engine on, engine off, afterburner, jumping). And maybe track's thickness and duration ?
I think there should also be `nebula_track_generator` that is only active in the nebula (and maybe deactivate the regular generators in the nebula ?)
By this mean, each of the engines of a ship will generate tracks.

This raises one question: should the trail colour sets be defined in a separate xml ?

bobbens: That does sound good, but I'm worried that it will make it more complicated to add and maintain ships. That said, if possible, as much of the common stuff should be moved into separate xml files. For example, only define the position and type of trail in each ship's xml, and have the color, thickness, duration, etc. get looked up would work. It would also allow defining things like "zalek_large_trail", which could be changed afterwards to include new graphics and such.

src/space.c
nenau-again: The problem is that in simulation, the tracks don't increment their age, so they don't vanish. When the player takes off, there are lines all over the space.
I think using `dt` won't change that.

bobbens: It should work, because that's how all the pilot timers and weapon timers work. Most of the timers are implemented by setting the time to live and decrementing. When the timer is `<0` it is over. Wouldn't something like that work?

nenau-again:
For info: I had been forced to include `pilot.h` in `spfx.h`, which led to ugly unnatural things. I've found how to get rid of this. I'll push it soon on this branch.
Along with `trail.xml`

nenau-again:
Let's sum up what is still to do :

* Debug mac compilation (???)
* Use `dt` instead of `ntime_get`, and see of ti's enough to get rid of the `space_isSimulation` function
* Modify opengl function to render a track in one call
* Adapt the colours (in the xml, so it's trivial to do for anyone who has a bit of artistic skills -- unlike me)
* Add trails to missiles. With the problem that missiles can only be in 1 situation (never engine off, neither afterburning nor juming)
* Somehow add nebula-specific trails
* Make the rendering more optimized, without passing the colours as uniforms, but via the VBO
* See how to make the trails de-activable, and maybe mutually exclusive with engine glow
* Maybe make the thickness variable (like thicker when afterburning)

UncombedCoconut:
There's a cat sitting on my non-Mac computer. I'll look at the Mac build.

UncombedCoconut:
This one is easy: do not declare "trail_col_stack" in spfx.h (except as "extern", with one .c file owning the definition). I suspect the choice of compiler and options (e.g., Clang with Link Time Optimization enabled) made the difference between successful and unsuccessful builds.

bobbens:
I've editted stuff and pushed to naev / tracks branch. Without trail.xml I wasn't able to test, but it seems to be OKish.

>     * Debug mac compilation (???)
Probably fixed.

>     * Use `dt` instead of `ntime_get`, and see of ti's enough to get rid of the `space_isSimulation` function
Done.

I also combined the three arrays into one.

PR #1690, opened 2021-02-03
Update ARTWORK_LICENSE.yaml with a new entry

Bisc88:
added the entry for agricultural.webp which has been pull requested to naev-artwork-production


bobbens:
Thanks, instead of updating the main repo ARTWORK_LICENSE.yaml, I have created a new one in the artwork-production repository. As we move more and more artwork over there, that one should become the main one. Thanks!

PR #1694, opened 2021-02-06
Update dv_diversion.lua

Cornflakes91:
disarming "Hawk" to make it fit mission description and make the attacking force able to destroy it


UncombedCoconut:
The idea is right, and thanks for spotting the issue, but I think we need a different execution.

The briefing says, "Warlord... building his newest flagship... and will be onboard... as it tests its hyperspace capabilities. Since its engines and weapons have not been fully installed yet, it will be substantially slower than normal and unable to defend itself."

So we want a ship that's incompletely armed and slowed down (while still able to fly to a jump point) compared to a warlord's personal battle cruiser. `hawk:rmOutfit( "all" )` by itself leaves the ship totally immobile and unarmed.

Also, the commit message should preferably describe the change, as opposed to which file(s) changed, so I think the best resolution is a brand-new commit that credits you.

PR #1711, opened 2021-02-26
Shadow Vigil changes

nenau-again:
I split it in 4 for easier cherry-picking if needed.
1 - The player chooses to accept the mission or not during the preliminary event, which mainly enables us to place an osd arrow
2 - small eye-candy and user-friendliness changes
3 - added the possibility to disable and board one of the escorts after they killed the diplomat
4 - added a special accessory as a reward if the player does so.

with commit 4, it is theoretically possible for players to abort the mission, and re-do it again to get several copies of the accessory. But I think it's not worth fixing as it doesn't bring anything to the player.


dat/missions/shadow/shadowvigil.lua
```diff
@@ -262,7 +262,11 @@ end
 -- Function hooked to jumpin. Handles most of the events in the various systems.
 function jumpin()
     sysclear = false -- We've just jumped in, so the ambushers, if any, are not dead.
-    
+
+    if pahook then -- Remove the hook on the player being attacked, if needed
+        hook.rm( pahook )
+    end    
+
```
UncombedCoconut, line 275: FYI, as of https://github.com/naev/naev/commit/caf4e168f2b14a72eb2c70f31cdc2f5ee5300dd0 it's safe to call hook.rm(pahook) without testing for nil.
Checking anyway isn't an issue, of course.

dat/missions/shadow/shadowvigil.lua
```diff
@@ -61,7 +61,13 @@ text[4] = _([[Captain Rebina angrily drums her fingers on her captain's chair as
...
+   "Good job out there. Hum. But, you see? You won't catch me alive. Oh, no. The pill of silence is making sure for that. Hmf. That looks painful, doesn't it? You know what? It is even more painful than it looks. I guarantee it to you. Gbf. Oh, it makes me throw up... I'm sorry about that."
+   You approach the man, and ask him why he killed the diplomats. "You have no idea what is going on, right? Uh. Let me tell you a thing: they are coming! And they won't show any mercy. Oooh! They're so close! They're so worst than your darkest nightmares! And one day they'll get you. That day, you will envy... You will envy my vomit-eating dead-painful agony! Sooooo much!"
+   The man stops speaking and moving. His grimacing face still turned towards you make you wonder if he is already dead, but in absence of any response, you leave the ship as soon as possible before it explodes.]])
+
```
UncombedCoconut, line 70: As a native speaker, I think all added text is perfect except for this string, which needs only tiny edits.
I need to step away for a bit, but I assume I'll be merging the PR very soon.

PR #1736, opened 2021-03-20
No fleet_form

nenau-again:
* remove dependency of dvaered warlords battle event to the fleet_form script
* remove the now useless fleet_form script
* lower the probability of the battle event (10->5) because it triggers a bit too often by my standards (when testing frontier war for example)
* A few corrections/improvements on the formation system


dat/events/dvaered/warlords_battle.lua
```diff
@@ -14,9 +14,9 @@
 --  A battle between two Dvaered warlords. The player can join one of them and get a reward
 
 require "fleethelper"
-require "fleet_form"
 require "proximity"
 require "numstring"
+local formation = require "scripts/formation"
```
bobbens, line 19: you shouldn't need to prefix `scripts/` as it should be in the module loading path

dat/events/dvaered/warlords_battle.lua
```diff
@@ -133,44 +133,56 @@ function attack ()
    attackers[2*n+7]:memory().shield_return = 99
 
    attackers = arrangeList(attackers)  --The heaviest ships will surround the leader
+   form = formation.random_key()
 
    for i, j in ipairs(attackers) do
       j:rename("Invader")
-      j:setFaction("Thugs")  --I use Thugs and Associates because they won't interact with anybody
+      j:setFaction("Thugs") -- I use Thugs and Associates because they won't interact with anybody
```
bobbens, line 140: Might be worth using the new dynamic faction stuff ( https://naev.org/api/modules/faction.html#dynAdd ). I use some in the minerva missions if you want examples.

src/ai.c
```diff
@@ -2132,6 +2134,45 @@ static int aiL_getnearestplanet( lua_State *L )
...
+{
+   Vector2d *pos;
+   double dist, d;
+   int i, j;
+   LuaPlanet planet;
+
+   pos = luaL_checkvector(L,1);
+
+   /* cycle through planets */
+   for (dist=HUGE_VAL, j=-1, i=0; i<array_size(cur_system->planets); i++) {
```
bobbens, line 2154: Do you mean `math.huge`?

nenau-again: I don't know. I simply copy/pasted the previous function :/

dat/events/dvaered/warlords_battle.lua
```diff
@@ -2,7 +2,7 @@
 <?xml version='1.0' encoding='utf8'?>
 <event name="Warlords battle">
   <trigger>enter</trigger>
-  <chance>5</chance>
+  <chance>100</chance>
```
bobbens, line 5: Probably left in the 100% chance when debugging

nenau-again: Pfff... I should never push early in the morning.

PR #1774, opened 2021-05-08
Prioritize planet clearing over jump point clearing.

onpon4:
This change causes the target clear action (C key) to first deselect planets before deselecting hyperspace navigation. This allows clearing planet selections while autonaving without causing interruptions.

🕵️


bobbens:
Looks good, thanks!

PR #1777, opened 2021-05-08
Fixed symbol in map find dialog.

onpon4:
Oops! I didn't realize that calling planet_updateLand was necessary. This fixes the bug where planets in the find results would get the symbol of the previous planet.

🕵️


bobbens:
Looks good, thanks.

PR #1778, opened 2021-05-08
Use smallFont for search input box, not defFont.

onpon4:
This puts that input box in line with other input boxes throughout Naev. (It was the only one using defFont instead of smallFont, the default for this widget type.)

🕵️

### DIFF
--- a/src/map_find.c
+++ b/src/map_find.c
@@ -1024,7 +1024,7 @@ void map_inputFind( unsigned int parent, char* str )
 
    /* Create input. */
    window_addInput( wid, 30, y, w - 60, 20,
-         "inpSearch", 32, 1, &gl_defFont );
+         "inpSearch", 32, 1, &gl_smallFont );
    y -= 40;
 
    /* Create buttons. */


bobbens:
From what I see the dialogue input widget also uses the default font. In fact, from a quick grep it seems that the input widget in the equipment and outfit windows are outliers and probably they should be changed to default font instead.

onpon4:
Does it? I see this line:

```c
window_addInput( input_dialogue.input_wid, 20, -50-h, 200, 20, "inpInput", max, 1, NULL );
```

Specifying the font as NULL, which on my end defaults to gl_smallFont. Was this something I changed in Naikari? I don't remember doing so, but I might have.

If the default input box font in Naev is actually gl_defFont, feel free to close this PR.

🕵️

bobbens:
It should default to defFont when it is set to NULL, hence the name default font. Closing.

bobbens:
Yes, it needs harmonization. Having it default to small is the opposite of how all the font functions in `font.c` behave. For the small pop-ups (tk.input, map search, etc.) it should probably use default, while I guess embedded widgets which lack space should probably use small.

PR #1780, opened 2021-05-08
Fixed warlords battle failing if some were already dead at start.

onpon4:
This fixes an implicit assumption that all of the pilots in the team you fight against are alive when you first attack, which just isn't necessarily the case. Added checks to make sure the pilots exist before trying to do something with them.


UncombedCoconut:
🐗

PR #1789, opened 2021-05-12
Fixed reversed meaning of "Trade" and "Cost", made more robust.

### LINKED ISSUE #1788: map cost/trade swapped
When I select map mode Food:Cost, I get map mode Food:Trade, and vice versa.  Likewise for the other commodities.

onpon4:
It was using absolute array index to figure out what trade mode it was in, which got screwed up when Discovery mode was added. This fixes it and also hardens it so that such a thing shouldn't happen again.

Fixes #1788


UncombedCoconut:
definitely better! (better enough to trash-talk the old code that much? well, maybe. :)) thanks.

PR #1791, opened 2021-05-12
Fixed firing missile causing a mass reduction.

onpon4:
Looks like it was cutting mass *twice*, I assume as a leftover of a previous way that mass was handled, in combination with the fact that it wasn't just using pilot_rmAmmo. This meant that every time you decreased your ammo, your ship's mass decreased until something triggered a recalculation of mass (like adding or removing an outfit). Fixed by replacing that custom code with pilot_rmAmmo.


UncombedCoconut:
😸

PR #1797, opened 2021-05-14
Removed unused stuff in generic.lua, added generic pilot name function.

onpon4:
This is a generic version of the pirate name generator which is based on a simpler, less silly sounding [noun] [number] rule for use in pilots that aren't pirates (and thus need to be taken at least somewhat seriously). Added in preparation for my escorts thing, but also can be used in bounty and seek and destroy missions.

The stuff that was previously in pilot/generic is removed because it wasn't being used anywhere and they don't seem to be useful functions to use.


onpon4:
I've added use of the generic pilot name function for the Seek And Destroy and assassination job missions.

🕵️

bobbens:
Looks good, thanks!

PR #1798, opened 2021-05-14
Tentative EMP torpedo commit

Lineth:
Found that there's a possible need for a proper heavy disabling weapon. Hopefully this will do.


bobbens:
Thanks! Recently, we have split most of the new graphics and data into the https://github.com/naev/naev-artwork-production . If possible, could you make a PR with the artwork on the https://github.com/naev/naev-artwork-production repo (if possible in webp instead of png) and remove the artwork from this PR (you'll have to probably force push)?

 I know it's a bit more trouble to add new things, but it should pay off in the long run.

bobbens:
If you find this tricky to do, I can do it for your, but I don't think I'll be able to conserve your commits (although I will credit you in the commits).

Lineth:
Go ahead, I don't mind. I am a little confused as to the request, as I don't see any outfit files on there. I unfortunately don't have any .webp files on me, as all I actually did was change the hue and saturation of the Caesar images in paint.net. 😛

Lineth:
Also, we could probably give this thing to pirate Kestrels as a rare weapon. Would make sense for them and encourage their whole 'disable and plunder' thing that pirates are supposed to do.

bobbens:
OK, I have set it up and pushed. No enemy pilots get it yet, but giving it to pirate kestrels does seem like something that should be done.

PR #1799, opened 2021-05-14
Event-based escort system

### LINKED ISSUE #832: Proposal: Player Escort Hiring System

onpon4:
I, uh, kinda went back on my word on getting sleep. 😅

But this escort system is tested and works. It randomly creates pilot NPCs at the bar which you can hire for a certain deposit (actually just half the cost of the ship they fly with some random variation) plus a certain percentage "royalty" of your payment from missions. Once you hire them, they show up in the bar under their names and you can choose to fire them.

This depends on #1797 as it uses that pilot name function to name the escorts, and also depends on #1800 as the new Lua function introduced there is used.

Closes #832 if merged.

### DIFF
--- a/dat/events/escorts.lua
+++ b/dat/events/escorts.lua
@@ -272,6 +272,7 @@ function enter ()
          edata.pilot:setTemp(temp, true)
          edata.pilot:setHealth(armor, shield, stress)
          edata.pilot:setEnergy(energy)
+         edata.pilot:setFuel(true)
 
          if f == nil or f:playerStanding() >= 0 then
             edata.pilot:setLeader(player.pilot())
--- a/dat/events/escorts.lua
+++ b/dat/events/escorts.lua
@@ -168,6 +168,10 @@ function land ()
       hook.rm(standing_hook)
       standing_hook = nil
    end
+   if hail_hook ~= nil then
+      hook.rm(hail_hook)
+      hail_hook = nil
+   end
 
    -- Clean up dead escorts so it doesn't build up, and create NPCs for
    -- existing escorts.
@@ -220,6 +224,9 @@ end
 function enter ()
    if standing_hook == nil then
       standing_hook = hook.standing("standing")
+   end
+   if hail_hook == nil then
+      hail_hook = hook.hail("hail")
    end
 
    local spawnpoint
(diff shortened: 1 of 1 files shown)


onpon4:
Just added something to prevent escorts from getting cleared by pilot.clear(), so this is now dependent on #1800.

🕵️

bobbens:
Looks good, thanks! I pushed some fixes to the escorts branch if you are interested.

PR #1800, opened 2021-05-14
Added capability to set a pilot as exempt from pilot.clear().

onpon4:
This is something that can be used by Lua code to ensure that special pilots aren't removed by missions as a matter of course. Intended for use by #1799.


bobbens:
Thanks. I'll merge for now, but I'm still not entirely sure if we want to implement this behavior. Adding exceptions to `pilot.clear()` will make it so that `pilot.clear()` will sometimes not clear all pilots as expected, unless all missions work under the assumption that pilots have non-clearable escorts. In the case of escorts `pilot.clear()` already "forcibly docks them", so this would only affect escorts when implemented in pure Lua.

PR #1803, opened 2021-05-14
Don't allow player to gain info from own escorts.

onpon4:
This fixes a problem where the Seek and Destroy mission became real easy if you had escorts.

🕵️


UncombedCoconut:
lol

PR #1805, opened 2021-05-14
Fixed a bunch of missions to use the new nohook argument.

onpon4:
This adds use of the nohook argument in cases where player.pay()
is being used to take money away from the player, since these are
payments *from* the player (the same as e.g. buying an outfit).

### DIFF
--- a/dat/missions/baron/baron_prince.lua
+++ b/dat/missions/baron/baron_prince.lua
@@ -298,7 +298,7 @@ function seller()
       if tk.choice(title[8], text[9], buy:format(creditstring(reward * 0.15)), nobuy) == 1 then
          if player.credits() >= reward * 0.15 then
             misn.npcRm(sellnpc)
-            player.pay(-15000)
+            player.pay(-15000, true)
             artifactA = misn.cargoAdd("Artifact? A", 0)
             misn.markerRm(markerA)
          else
@@ -309,7 +309,7 @@ function seller()
       if tk.choice(title[8], text[10], buy:format(creditstring(reward * 0.15)), nobuy) == 1 then
          if player.credits() >= reward * 0.15 then
             misn.npcRm(sellnpc)
-            player.pay(-15000)
+            player.pay(-15000, true)
             artifactB = misn.cargoAdd("Artifact? B", 0)
             misn.markerRm(markerB)
          else
@@ -320,7 +320,7 @@ function seller()
       if tk.choice(title[8], text[11], buy:format(creditstring(reward * 0.15)), nobuy) == 1 then
          if player.credits() >= reward * 0.15 then
             misn.npcRm(sellnpc)
-            player.pay(-15000)
+            player.pay(-15000, true)
             artifactC = misn.cargoAdd("Artifact? C", 0)
             misn.markerRm(markerC)
          else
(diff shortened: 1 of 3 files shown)


UncombedCoconut:
This all makes me wonder about the boundary between credits changes a script should observe and those a script shouldn't.
What really makes sense for an extensible and cleanly scriptable game engine? Is it the same as what can reasonably be called "paying"?
(I see the `bioship_upgrade` script as not really decisive -- it could gain an `amount < 0` check if need be.)

Looking at the engine code, the player credit changes break roughly into

- `player.pay(n)`
- `player.pilot.credits(n) -- unused? but looking taught me that this exists, and is essentially the same as a nohook pay.`
- looting boarded ships
- C-side uses of `player_modCredits()` as when buying/selling commodities/outfits/ships.

bobbens:
> (I see the `bioship_upgrade` script as not really decisive -- it could gain an `amount < 0` check if need be.)

I agree with @UncombedCoconut 's assessment. If we start sprinkling all these nohooks just for the bioship event, this could lead to missions expecting the `pilot.pay()` hook to trigger when it doesn't. The only case I can really think of where the nohook is necessary is when doing `player.pay()` inside of a payment hook, given that it will probably infinitely recurse.

In the case of negative payments, I would say it is up to the hook function itself to make sure it is positive, and just making negative payments not trigger the hook is weird.

> Looking at the engine code, the player credit changes break roughly into
>     * `player.pay(n)`
>     * `player.pilot.credits(n) -- unused? but looking taught me that this exists, and is essentially the same as a nohook pay.`
>     * looting boarded ships
>     * C-side uses of `player_modCredits()` as when buying/selling commodities/outfits/ships.

`player.pilot().credits()` I don't think is used, however, `pilot.credits()` is fairly useful and used for other pilots.

onpon4:
Well is there any particular reason a mission or event would expect to interact with the code in these missions that only exists to take money away from the player? These are framed in the game as amounts of money the player is deliberately handing over to someone and they're all very specific to the functioning of the respective mission. I'm not in favor of blanket-excluding all negative payments, but it seems worthwhile to me to keep the expectation that the pay hook is for when modification of the player's credits happens *to* the player instead of *by* the player. So I would be in favor of keeping pay hooks enabled for e.g. the player's money being stolen, but I'm not so sure the player handing over money because of a dialog choice is the best idea.

Consider, for example, an event that's designed to lessen the impact of some kind of theft event. I'm not interested in such a concept, but let's just consider it as a thought experiment. Such a hook being triggered by these particular pay() calls would mean that the player would get a discount on the player paying the NPC, despite the fact that you still have to have the originally required amount. It would be weird and confusing.

I don't have especially strong feelings about this, though. (And I'm not worried about the bioship event, it already has a guard against negative pay hooks.)

🕵️

bobbens:
Probably the best way would be to make `nohook` only be used when it is really a problem, and in other cases, allow `player.pay()` to have some sort of string representing the reason. Then the hook would get the string and be able to decide what to do with it. This makes it all transparent to the hook, while still having default behaviour like now, and is more flexible than just choosing cases where it doesn't hook and cases when it does. Something like:
```lua
player.pay() -- normal hook trigger
player.pay( false ) -- no hook trigger
player.pay( "bribe" ) -- money is related to "bribe" string which is passed to the hook
```

onpon4:
True, that would make sense.

Ok, I'll close this then in favor of that other solution.

🕵️

PR #1808, opened 2021-05-16
Made escort commands NMOD_ALL by default.

onpon4:


bobbens:
Thanks, I had already fixed this in 34b72af68585c2bddf6ac0d22268517ccc6558c2 which I just cherry-picked to main (with some other escort-related fixes).

PR #1816, opened 2021-05-21
Some more missile shenanigans

Lineth:
Just using the headhunter turret sprite and the torpedo sound.


bobbens:
Looks good, thanks!

PR #1818, opened 2021-05-24
Kestrel shenanigans

Lineth:
First-pass rebalance of the iconic Kestrel.

- Hulls are 300 tons lighter and have small bonuses to speed, turning and shields.
- Krain Remige engines' power usage was absurdly low. Fixed, with a consolation of 100 extra mass limit.
Now on the proper branch! Pay no attention to the idiot behind the curtain.


bobbens:
Looks good, thanks!

PR #1821, opened 2021-05-25
Use exclude_directories instead of exclude_files

heirecka:
The .git from the artwork submodule isn't a file but a directory and
thus it wasn't excluded but installed.


UncombedCoconut:
Thanks!

PR #1823, opened 2021-05-26
Lock-on utility outfit

Lineth:
Just a little something for medium-heavy warships that want to improve their lock-on times with heavier missiles. Balance probably isn't perfect, but I tried.


Lineth:
Remade this to change fire rate to lock-on times for the preexisting Targeting Conduit. I find that fire rate is rarely an issue compared to getting a lock for missiles.

bobbens:
Looks good, thanks!

PR #1825, opened 2021-05-27
First Sirius rebalance

Lineth:
Make Sirius ships beefier and add the energy-draining tesla cannon (energy drain not yet implemented).


bobbens:
Thanks! One thing that's a bit weird is that some ships get a movement malus, but others get a movement bonus (preacher). Maybe it would be better to have all the ships across the board get a movement malus, although smaller on tiny ships.

Lineth:
Forgot the preacher, sorry.

bobbens:
Looks good, thanks!

PR #1829, opened 2021-06-01
Electron conflict fix

Lineth:
Makes the Electron Burst Cannon exclusive to the Za'lek. For the Collective a similar pseudo-rocket with very, very light tracking and high armor-piercing was added to their arsenal for a similar purpose.


bobbens:
Looks good, thanks!

PR #1833, opened 2021-06-08
Changed Repeating Railgun, Ripper and Heavy Laser Turrets, and Collec…

Lineth:
…tive Shatterers.

Made repeating railguns into actual repeaters. Switched ripper turrets and heavy laser stats and usage. Shatterer launchers buffed.

### DIFF
--- a/dat/ships/drone_heavy.xml
+++ b/dat/ships/drone_heavy.xml
@@ -24,8 +24,8 @@
  <slots>
   <weapon size="medium" x="8" y="1" h="-5" />
   <weapon size="medium" x="8" y="-1" h="-5" />
-  <weapon size="medium" x="5" y="5" h="-5" />
-  <weapon size="medium" x="5" y="-5" h="-5" />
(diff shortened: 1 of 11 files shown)


dat/ships/drone_heavy.xml
```diff
@@ -24,8 +24,8 @@
  <slots>
   <weapon size="medium" x="8" y="1" h="-5" />
   <weapon size="medium" x="8" y="-1" h="-5" />
-  <weapon size="small" x="5" y="5" h="-5" />
-  <weapon size="small" x="5" y="-5" h="-5" />
+  <weapon size="medium" x="5" y="5" h="-5" />
```
bobbens, line 27: It's a bit weird that the heavy drone has all medium weapon slots while being a small ship. Furthermore, if I'm correct, this will cause it to only have the medium shatterer missiles and not have heavy neutron blasters. It might be better to leave this as is (2x and 2x small).

bobbens:
Looks good, thanks!

PR #1834, opened 2021-06-09
Update toy_drone.xml

Lineth:
Quick and dirty edit to the toy drone to give it something better than the placeholder image.


bobbens:
Looks good, thanks!

PR #1836, opened 2021-06-10
Added commodities to worlds

Lineth:
Should fix a dearth of cargo missions in large parts of Empire space.

Will likely need several passes to make sure everyone is included


bobbens:
Looks good, thanks!

PR #1838, opened 2021-06-11
Fix reference to  Unicorp PT-100 Core System

heirecka:
It was changed in 3abcb707217336556d9c94d8c5260723bca80422.


UncombedCoconut:
Thanks!

PR #1839, opened 2021-06-13
Tentative Missile Rebalance

Lineth:
Found that the heavy drone didn't actually have enough CPU for its loadout. This fixes that.

EDIT: Made some broader changes to the missiles:

- Za'lek hunters were pathetically weak. They now have MUCH lower mass and increased damage.

- Torpedoes are now stronger, fire slightly faster and lock-on faster to make them more attractive as a long-range weapon.

- Shatterer nerfed pretty much across the board.


bobbens:
Looks good, thanks! Might need more adjustments depending on testing.

Issue #1794, opened 2021-05-13
Making bounty missions more doable


bobbens:
A common request is to make pirate bounties and such be their own faction so that they aren't blasted to smithereens by patrolling Empire or whatnot forces. Something like that could be done, while compensating for the fact that without being hostile to other ships, they would literally not really do anything and probably fly away. Something like actively seeking out the player or spawning the player could work.

Lineth:
When I tried it on a home build by making the "Bounty" faction hostile to merchants, civilians, and the player it seemed to work. Bounties already seem to spawn near-ish to the player when they warp in. But a command to actively pursue the player would be a good fail-safe if they don't have any valid targets nearby.

bobbens:
The issue is that if the bounty attacks a merchant and this merchant sends a distress signal, it should cause Empire ships nearby to attack it. I don't think this was implemented to only affect the player. However, it would be less likely than now.

Lineth:
I was just thinking that when I made the homment, d'oh. 🤐

amylaar:
In terms of payment, it's not that bad for dead-or-alive missions, since, if the pirate doesn't run away, you can generally take of them quickly, and even if you can't, there's the 'lesser reward' mechanism.
What's annoying is that you don't get even a fraction of the mission reputation gain when that happens.

Also, if the authorities want a pirate alive, what are the grunts doing killing the pirate?

So, a solution might be to give a reputation share with the 'lesser reward', plus make the 'wanted alive pirates another faction,
e.g. "Supergrass" (Assuming they are wanted alive because they might divulge info leading to higher-value targets in return for lenient sentencing and witness protection.  Actually, it seems possible that the only way they can get out alive is to be captured by force (keeping up plausible deniability and all that) .)

nenau-again:
> Also, if the authorities want a pirate alive, what are the grunts doing killing the pirate?

Very few people in naev universe seem to fully understand the concept of "not killing". Some get quite close to it and equip disabling weapons... But then they still shoot at disabled ships until they explode :o/

Sidenote : I thought the alive bounty pirates were part of a special faction that was not attacked by others. Was I wrong ?

Lineth:
Alive bounty pirates are indeed part of the "Wanted Pirate" faction, but that faction is still hostile to pretty much everyone.

bobbens:
> Alive bounty pirates are indeed part of the "Wanted Pirate" faction, but that faction is still hostile to pretty much everyone.

[...]

PR #1844, opened 2021-06-24
small fixes

Lineth:
No way to change ship in the primary bounty grinding location in Za'lek space. Dvaered will no longer murder your wanted alive pirates. Miners use their icon.


bobbens:
Looks good, although we'll still have to do something like mentioned in #1794 and probably do away with the fixed "Wanted Pirate" faction. Thanks!

PR #1846, opened 2021-07-02
za'lek beams use new sounds!

Lineth:
also heavy drones have their firing positions fixed

needs https://github.com/naev/naev-artwork-production/pull/19 to be merged


bobbens:
Looks good, thanks!

PR #1851, opened 2021-07-05
added the possibility to draw some trails on top of the ship gfx

nenau-again:
When the ship is facing the top of the screen, and if the trail generator is marked as "back_trail", the trail is drawn after the ship, and no more during `spfx_render`.
Examples are proposed with the Thurion capships, but actually, this back_trail stuff might as well become the default (for example, the Hawking or the Rhino could benefit of it. I guess only a few zalek ships should not use it.

The main drawback is that the trail is also drawn after all ships that are drawn before the ship in question.


bobbens:
Some notes:
1. When the pilot dies, you'll have to set `ontop=0` or the trail will disappear. That's the main reason it was moved from `pilot.c` to `spfx.c`
2. If you want `back_trail` to be the default behavior and have za'lek ships or the likes not use it. It might be better to rename it to `always_under` and make it be 0 be default so that the default behaviour would be `back_trail` in the current implementation. 

Side note: if trails look good enough, it might be a good idea to get rid of engine glow and just have everything be trail-based.

nenau-again:
On point 1, I thought this line : https://github.com/naev/naev/blob/main/src/pilot.c#L3078 means that the trail is removed anyway when the pilot is suppressed. Did I miss something?

On point 2: I will do that. The hardest will be to identify the (rare) ships that need this option. I know there are some, but I don't remember which ones...

UncombedCoconut:
> On point 1, I thought this line : https://github.com/naev/naev/blob/main/src/pilot.c#L3078 means that the trail is removed anyway when the pilot is suppressed. Did I miss something?

Maybe the function should be renamed -- trails are reference counted, and `spfx_trail_remove` just unreferences the trail. The spfx code counts an additional reference as long as there are trail segments to display. `spfx_trail_free` actually deallocates stuff and stops the trail from being drawn.

UncombedCoconut:
Since point 2 is in progress, I'll leave this alone, but I've looked at this code and the strategy seems fine!
Since there's time to nit-pick: you can assign a boolean expression directly to p->trail[i] instead of using an if/else. :)

UncombedCoconut:
[ballet_test.zip](https://github.com/naev/naev/files/6772108/ballet_test.zip)
This test appears to show: (1) most trail locations do not seem problematic, (2) the major weak point of this idea is that stuff can look weird when your velocity and acceleration aren't lined up. To be clear, the motion of these ships is totally artificial, but you can produce similar effects yourself: just hold the turn/thrust keys and watch the entire trail change its mind about whether it's in the foreground or background.

bobbens:
> [ballet_test.zip](https://github.com/naev/naev/files/6772108/ballet_test.zip)
> This test appears to show: (1) most trail locations do not seem problematic, (2) the major weak point of this idea is that stuff can look weird when your velocity and acceleration aren't lined up. To be clear, the motion of these ships is totally artificial, but you can produce similar effects yourself: just hold the turn/thrust keys and watch the entire trail change its mind about whether it's in the foreground or background.

Maybe the "below/ontop" should be per vertex? Having vertices after a certain distance all automatically become "below" could also avoid some inconsistencies.

nenau-again:
The problem when velocity and direction are not aligned is the main reason why my corrections on trails takes so long. But I believe I can come up with something acceptable.

src/nlua_debug.c
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @brief Toggles the emitter marker.
+ *
+ * @usage dispEmitters() -- Trail emitters are marked with crosses.
+ * @usage dispEmitters(false) -- Remove the markers.
+ *
+ *    @luatparam[opt=true] boolean state Whether to set or unset markers.
+ * @luafunc dispEmitters
+ */
+static int debugL_dispEmitters( lua_State *L )
```
bobbens, line 55: why not just have the default be false and do
```
if (lua_toboolean(L,1))
   debug_setFlag(MARK_EMITTER);
else
   debug_rmFlag(MARK_EMITTER);
```?

File: src/nlua_debug.c [hunk 2]
```diff
@@ -0,0 +1,74 @@
...
+#include "nlua_debug.h"
+
+#include "debug.h"
+#include "nluadef.h"
+
+/* Debug metatable methods. */
+
+static int debugL_dispEmitters( lua_State *L );
+static const luaL_Reg debugL_methods[] = {
+   { "dispEmitters", debugL_dispEmitters },
```
bobbens, line 26: showEmitters might be a better name than dispEmitters

src/debug.h
```diff
@@ -12,5 +12,23 @@ void debug_sigClose (void);
 void debug_enableFPUExcept (void);
 
 
+#ifdef DEBUGGING
+#define debug_isFlag(f)     (debug_flags[f])     /**< Checks if flag f is set. */
+#define debug_setFlag(f)    (debug_flags[f] = 1) /**< Sets flag f. */
+#define debug_rmFlag(f)     (debug_flags[f] = 0) /**< Removes flag f. */
+
+enum {
+   MARK_EMITTER,        /**< Mark the trail emitters with a cross. */
```
bobbens, line 21: Probably better to prefix the flags like `DEBUG_MARK_EMITTER`

src/nlua_debug.c
```diff
@@ -0,0 +1,74 @@
...
+
+   NLUA_CHECKRW(L);
+
+   /* Get state. */
+   if (lua_gettop(L) > 0)
+      state = lua_toboolean(L, 1);
+   else
+      state = 1;
+
+   /* Set as hostile. */
```
bobbens, line 67: "as hostile"

bobbens (commented):
Side note, you are creating lots of conflicts by working on a pre-equipopt branch. When merging, I can fix them, but just wanted to point it out. In particular, I was trying with lowering the time_compression on most ships to try to make the combat less instantaneous, and that is causing all the conflicts.

src/nlua_debug.c
nenau-again: copy/paste error. thnx

src/nlua_debug.c
nenau-again: It doesn't feel right to me to be false by default. Most people will use it to activate.

bobbens: It's fine in this case. However, in general, I'm trying to have boolean parameters default to false, as it avoids the `lua_gettop(L)`-dependent behavior. I don't like having `my_func()` and `my_func(nil)` behave differently, as I think it can be a source of bugs in some cases.

UncombedCoconut:
It's against ldoc's nature to be happy. :)

UncombedCoconut:
```diff
diff --git a/src/nlua_debug.c b/src/nlua_debug.c
index 6e5e73fc1..e791c5547 100644
--- a/src/nlua_debug.c
+++ b/src/nlua_debug.c
@@ -8,6 +8,7 @@
  * @brief Handles the Lua debug bindings.
  *
  * These bindings control the debug flags.
+ * @luamod debug
  */
 
 /** @cond */
```

PR #1875, opened 2021-07-25
Fix name of Wayline's second moon

nathan-b:
I'm assuming this is a typo :)


UncombedCoconut:
Good catch! Unfortunately the rename has to be done in a few more places before this will fully work. I'll take care of it and credit you in the commit message.

nathan-b:
Phew, looks like it was a lot more work than I realized! Thanks for taking care of it!

UncombedCoconut:
The po/ step was automated (on my end) and not critical. :) The important extra step was fixing what was `dat/assets/wayling_m2.xml`. Thanks again.

PR #1884, opened 2021-08-01
lighter cruisers

nenau-again:


bobbens:
I've been trying it and I'm not really sure what to say. I mean it's good as it forces more meaningful choices between remige and typhoon, plus gives another armour alternative. I also don't see it causing any major limitations nor issues. That said, on paper, it does seem like the difference between remige and typhoon is really big, although I like the fact that it lets you use typhoon on a kestrel to have fighters for a decrease in speed. For me I think the changes are OK, and we can tweak as needed.

Issue #922, opened 2020-02-10
Harmonize the use of relative and absolute velocity on rockets


nenau-again:
As I mentioned in https://github.com/naev/naev/pull/921, I believe the computation of velocity for rockets is odd and confusing. After having got a look at the code and some experimentations, here is how it works :
* When the rocket is fired, its velocity is the same as the pilot's, excepted for rockets with 0 thrust, that start at their max vel+the pilot's vel.
* Rocket is updated with Runge Kutta, and absolute velocity is bounded by max velocity, excepted again for rockets with 0 thrust, for whom velocity is not bounded
* In the thinking step of seekers, absolute velocity is again bounded (but only for seekers)
* Right now, only the mace rocket has nonzero thrust

So here are examples of behaviour :
* When you fire a seeker while moving towards the target, the seeker starts at its max vel+your vel, and just after, during the update step, this velocity is reduced, so the seeker starts at its max vel
* When you fire a seeker while moving backwards (you try to escape your target), it starts at its max vel - your vel (which is slow and handicaps the seeker)
* When you fire a banshee, it works exactly as a (slow) bolt weapon
* When you fire a Mace rocket, it starts at your velocity and then accelerates until it reaches its max vel in the absolute referential (but it's so fast that it's hard to make the difference).

And the question is now : what is the desired behaviour ?
* I believe that dumb ammo should use relative velocity because they are used the same way as bolt weapons (am I right ?). This is more or less the case right now as the speed of the mace rocket is far too high for that to make any difference.
* Seekers should use absolute velocity in all cases, otherwise, it's odd imho.

If we want to get both those behaviour in a consistent physics, the best way could be to get all the missiles to start at relative velocity and then make a slow transition towards absolute velocity. As the dumb weapons are short range, they would mainly remain at relative velocity while seekers would end up at absolute velocity.
...Or we could have different physics for dumb and seekers, but simply make sure all seekers end up at absolute velocity.

What do you think ?

What's more, thrust of missiles seems like a good way to add a few fanciness in the ballistic, but I guess there is a reason why it is nearly unused. What is it ?

onpon4:
[...]

PR #1885, opened 2021-08-04
guided ammo behave like bolt weapons and guided ammo always use absolute velocity

nenau-again:
This proposal would solve https://github.com/naev/naev/issues/922

* All unguided rockets (mace and banshee) have 0 thrust and behave like bolt ammo
* All guided rockets have non-zero thrust. The result is that their final speed is always equal to their max_speed, and they sometimes have nice trajectories when launched.
* The way trails colours are chosen has been changed to be coherent with these changes.
* Mace rockets are thus much more effective. As a result, I divided by 2 their reload rate, and their quantity in launcher. This results in this weapon being very effective, but if the shots miss, the attack is a failure.

I know it's not really in par with what was discussed on discord, but it's the easiest solution, implementation-wise, and also probably player-wise. Otherwise we would have to teach the AI to play with this absolute--relative velocity question.
Later on, we can implement unguided ammo sometimes running amok with a low probability, and turning all over the place, mainly for the cool factor.

### DIFF
--- a/dat/outfits/launchers/teracom_mace_launcher.xml
+++ b/dat/outfits/launchers/teracom_mace_launcher.xml
@@ -15,8 +15,8 @@ TeraCom's Mace launcher uses an advanced reload mechanism that allows the launch
  <specific type="launcher">
   <ammo>Mace Rocket</ammo>
   <delay>0.3</delay>
-  <reload_time>6</reload_time>
-  <amount>50</amount>
+  <reload_time>12</reload_time>
+  <amount>25</amount>
   <swivel>22</swivel>
   <trackmin>0</trackmin>
   <trackmax>5000</trackmax>
--- a/dat/outfits/launchers/unicorp_mace_launcher.xml
+++ b/dat/outfits/launchers/unicorp_mace_launcher.xml
@@ -13,8 +13,8 @@
  <specific type="launcher">
   <ammo>Mace Rocket</ammo>
   <delay>0.36</delay>
-  <reload_time>6</reload_time>
-  <amount>50</amount>
+  <reload_time>12</reload_time>
+  <amount>25</amount>
   <swivel>22</swivel>
   <trackmin>0</trackmin>
   <trackmax>5000</trackmax>
--- a/dat/outfits/rockets/caesar_iv_torpedo.xml
+++ b/dat/outfits/rockets/caesar_iv_torpedo.xml
@@ -15,7 +15,7 @@
   <spfx_armour>ExpL</spfx_armour>
   <trail_generator x="-10">fire-missileL</trail_generator>
   <duration blowup="armour">33</duration>
-  <thrust>0</thrust>
+  <thrust>70</thrust>
   <turn>25</turn>
   <speed>200</speed>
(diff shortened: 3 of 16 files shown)


onpon4:
I should note that it would be perfectly possible and desirable to make it so that unguided rockets with thrust are aimed for correctly. The current aiming code assumes that all projectiles fired have a linear speed. Add some slightly more advanced math calculating the distance traveled during acceleration to max speed and the time it takes to get to max speed, and thrusting mace rockets would work a whole lot better.

🕵️

nenau-again:
You're totally right. However there are two subtleties:
* There is this absolute/relative velocity problem that makes that the trajectory of thrusting rockets is curvilinear
* The travel time in the case of constant velocity (currently implemented) requires to solve a polynomial equation of order 2. In the present case, with a constant thrust, I suppose it might require to solve an equation of order 3, which solution is analytical IIRC. But if the gods of maths are against us, this relative/absolute issue could require to solve an equation which solution is not analytical, thus requiring to use dichotomy or such, and leading to much more complexity of the code and a possible non-negligible CPU cost.

The present proposal is simply an attempt to get rid of the need of such a feature by removing thrusting unguided rockets. But if the majority of people think, contrary to me, that those rockets should continue existing, I might end up implementing this aiming feature in a short future.

onpon4:
Well currently thrusting rockets don't inherit ships' velocity IIRC, so I'm a little confused. Wouldn't simple kinematic formulas suffice, followed by the same linear logic that is currently used all throughout?

Specifically, the time it takes to accelerate to max speed can be calculated with `v[f] = v[i] + a*t`. `v[i]` is 0, so `t = v[f] / a` (where `v[f]` is the rocket's max speed). Displacement can then be calculated with `d = (v[i]+v[f])/2 * t` (again, `v[i]` is 0, and `v[f]` is the rocket's max speed). This gives us the distance and time traveled while the rocket accelerates to max speed. Then for any additional distance involved, linear logic can determine that.

The only major complication I can think of is if thrusting rockets were made to start with the same velocity as the launching pilot. To be fair, making rockets with thrust have inertia would certainly be a good idea.

🕵️

bobbens:
I pushed my thoughts to `rockets` branch. Notable changes:
1. Split `speed` attribute into `speed_max` and `speed`. `speed_max` refers to maximum speed of the ammo (relative with thrust, absolute otherwise), and `speed` refers to initial speed.
2. All seekers have no initial `speed` (start at parent speed) and have `speed_max` set to the previous `speed` value
3. Banshee and mace rockets have `speed` and `thrust`. They will act similar to bolt weapons, but they do accelerate, and the `speed_max` is relative in this case.
4. Changed `outfit_speed` to handle the new code.
5. Display speed and thrust in outfit description.

I tried it and it feels sort of a blend between the old system and the PR here. AI is much more accurate with the new rockets and they still have a behaviour somewhat different from bolts. It should be possible to tweak the speed and thrush values more to change the behaviour. Except for 56d9803d7fbfe0ab7d1080a9ce3faccafafdc977 the branch should behave exactly like the PR.

PR #1889, opened 2021-08-08
Less energy waste with beams

nenau-again:
* When used via instant mode, beams de-activate when the key is released
* When getting out of range, beams try to turn off asap

I'm not really sure this is the most elegant solution, which is why I open this PR

### DIFF
--- a/src/pilot.c
+++ b/src/pilot.c
@@ -1962,6 +1962,8 @@ void pilot_update( Pilot* pilot, double dt )
    int ammo_threshold;
    unsigned int l;
    Pilot *target;
+   AsteroidAnchor *field;
+   Asteroid *ast;
    double a, px,py, vx,vy;
    char buf[16];
    PilotOutfitSlot *o;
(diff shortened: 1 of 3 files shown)


src/pilot.c
```diff
@@ -2058,6 +2060,41 @@ void pilot_update( Pilot* pilot, double dt )
...
       }
 
+      /* Get the asteroid target. */
+      if (pilot->nav_asteroid != -1) {
+         field = &cur_system->asteroids[pilot->nav_anchor];
+         ast = &field->asteroids[pilot->nav_asteroid];
+      }
+
+      /* Special case of beam weapons with weapon set only_inrange: turn off if needed. */
+      if (outfit_isBeam(o->outfit)) {
```
bobbens, line 2070: Wouldn't this part make more sense in `think_beam` in `src/weapon.c`? You can do the same thing as when energy is out by setting `w->timer = -1;` to stop the beam.

nenau-again: Yes, it works better, and without the bug that suppresses all other weapons.
I'll re-write history (because it's really messy), test, and push in a separate branch for now.

bobbens:
This is unnecessary now, right?

nenau-again:
True.

PR #1890, opened 2021-08-11
Update electron_energy_cell.xml

Lineth:
Fixes bug with EBC having no range.


bobbens:
Thanks for the PR, I fixed it in master. The issue was that `speed_max` was being set instead of `speed` (EMP grenades had the same issue).

PR #1891, opened 2021-08-14
proposal of new tracking values for bolt weapons

nenau-again:
I did not consider launcher, nor beams. just bolt cannons and turrets.

Considered values of evasion:
* Scout: 2000
* Interceptor: 4000
* Fighter: 6000
* Corvette: 10000
* Destroyer: 13000
* Cruiser: 22000
* Battleship: 35000

So the proposed tracking values are: (trackmin--trackmax)
* Smallest cannon (laser mk1) is 100% accurate with everything but ultrastealth scouts: 0--2000
* Cannon mk2 has problems with scouts: 0--3000
* Corvette miniaturized cannon (ripper): idem 0--3000
* Corvette heavy cannon (hvy ripper) unable to track scouts, and has problems with interceptors: 1000-6000
* Superheavy cannon (railgun and tesla) hits destroyers or heavier: 4000--12000
* Small turret (laser mk1) hits everything but scouts: 0--3000
* Medium Turret (laser mk2) has problems with stealth interceptors: 1000--4000
* Heavy turret (heavy laser) has problems with fighters, and tracks corvettes: 2000--8000
* Heavy ripper turret: tracks corvettes and above: 3000--10000
* Superheavy turret (turbolaser): has problems with destroyers, and tracks perfectly cruisers: 5000-20000

(Of course I also did in-game tests and I found it to be satisfactory for me).

Remarks:
* One thing I think could be considered would be giving the heavy turret (heavy laser) the same values as heavy ripper turret, to make them unusable against fighters.
* Is the trackmin value really necessary? It could be fixed to zero, or to a fixed fraction of trackmax. It would reduce the number of parameters the player has to consider.
* (unrelated) It would be really better for maintaining if the soromid outfits were defined only by one file, that contains info from which the upper stages stats would be computed.

### DIFF
--- a/dat/outfits/bio_weapons/bioplasma_claw_stage_1.xml
+++ b/dat/outfits/bio_weapons/bioplasma_claw_stage_1.xml
@@ -22,8 +22,8 @@
   <gfx_end>plasma2-end.png</gfx_end>
   <energy>10</energy>
   <heatup>20</heatup>
-  <trackmin>1000</trackmin>
-  <trackmax>4000</trackmax>
+  <trackmin>0</trackmin>
+  <trackmax>3000</trackmax>
   <swivel>22</swivel>
   <damage>
    <type>impact</type>
--- a/dat/outfits/bio_weapons/bioplasma_claw_stage_2.xml
+++ b/dat/outfits/bio_weapons/bioplasma_claw_stage_2.xml
@@ -22,8 +22,8 @@
   <gfx_end>plasma2-end.png</gfx_end>
   <energy>10</energy>
   <heatup>23</heatup>
-  <trackmin>1000</trackmin>
-  <trackmax>4000</trackmax>
+  <trackmin>0</trackmin>
+  <trackmax>3000</trackmax>
   <swivel>22</swivel>
   <damage>
    <type>impact</type>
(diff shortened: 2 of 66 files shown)


bobbens:
Overall it looks fairly OK. It does seem like a fairly big nerf to evasion, but given how strong it got with the bug-fix, I think it is necessary. Would have to play-test a bit to get a better feeling.

>     * One thing I think could be considered would be giving the heavy turret (heavy laser) the same values as heavy ripper turret, to make them unusable against fighters.

Yes, I think that in general all "heavy slots" should have problems with light crafts, with the exception of point-defense stuff if implemented.

>     * Is the trackmin value really necessary? It could be fixed to zero, or to a fixed fraction of trackmax. It would reduce the number of parameters the player has to consider.

I added it for more control, but you're right, it could be removed. I find that it might have more use with launchers though. That said, if you set it all to 0, it should not appear and not be used in-game, so it would be functionally the same as not having trackmin at all.

>     * (unrelated) It would be really better for maintaining if the soromid outfits were defined only by one file, that contains info from which the upper stages stats would be computed.

Yes, every time I have to edit them I think of that. I'm not sure the best way of doing it. There are several options I can think of, but not sure which is best atm, and they all are a bit tricky to implement.

bobbens:
Tried it and seeing there's no other comments, I think it's good!

PR #1892, opened 2021-08-15
Radar zoom

nenau-again:
Two things that I have considered desirable for a long time now:
* radar resolution is saved in player data
* maximal radar resolution is much higher (300) which allows to see nearly any detected ship. Player will probably want to change between 100 and 300 depending on the situation.

Maybe later, we could add an option to automatically adapt the radar resolution depending on what object is targeted.

### DIFF
--- a/src/gui.c
+++ b/src/gui.c
@@ -169,7 +169,6 @@ typedef struct Radar_ {
 #define RADAR_RES_MAX      100. /**< Maximum radar resolution. */
 #define RADAR_RES_MIN      10. /**< Minimum radar resolution. */
 #define RADAR_RES_INTERVAL 10. /**< Steps used to increase/decrease resolution. */
-#define RADAR_RES_DEFAULT  50. /**< Default resolution. */
 static Radar gui_radar;
 
 /* needed to render properly */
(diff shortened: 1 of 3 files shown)


bobbens:
Sounds good. The question though is what do we want the radar to do. It's mainly legacy from EV. Personally, I feel that getting rid of it to gain more screen space is a good solution. However, if people like and and find it useful for, say, dogfights, maybe it should be turned into something that scales automatically and gives you important dogfighting information like weapon ranges and nearby hostiles?

nenau-again:
I tend to mainly use the radar when doing fight on the screen, for avoiding perturbation of the overlay.

As it used to be only short-range, in some situations I used to periodically put overlay on to have a better view of the situation (see where are allied and enemies capships mainly). With this modification, in these situations, I put the resolution to 300 instead.

It would require long testing to see all the situations where I can use this new long-range radar, and if it's indeed better than putting the overlay on, but as people are still free to use only short-range radar, I guess there is no downside having that.

bobbens:
Sounds good, I'll merge and we can think about how to make it better in the future. Thanks!

PR #1901, opened 2021-08-22
Typo and punctuation

not-pyroman:
corrected what was likely a typo and added some punctuation


UncombedCoconut:
not sure if "there always seem to be mainly repeated by modified entries" was supposed to mean exactly that, but things are better :) Thanks.

not-pyroman:
if im going to be honest not even im sure what exactly it means so im just assuming

PR #1904, opened 2021-08-28
Small contribution to french translation

aWildElephant:
A while back a translated a few strings in french. Today I realized that Justin Blanchard was already properly contributing to the french translation; this is my (quite old) po file merged into the current one using `msgcat --use-first po/fr.po <my save directory>/fr.po > po/fr.po`, hoping it will save him some time.

I ran `./utils/update-po.sh` but I still have some diff in the comments, my fiddling must have reverted them to an older version. I'd greatly appreciate if someone can explain how to avoid this.

Best regards

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -2153,6 +2153,7 @@ dat/ai/core/idle/generic.lua
 dat/ai/core/idle/miner.lua
 dat/ai/core/idle/pirate.lua
 dat/ai/core/idle/trader.lua
+dat/ai/core/idle/wait.lua
 dat/ai/core/misc/distress.lua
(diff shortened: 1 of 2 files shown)


UncombedCoconut:
Hi! We've integrated [Weblate](https://hosted.weblate.org/projects/naev/naev/) into our workflow.
I should be able to merge this as-is and watch Weblate integrate the updates. We can also generate a fresh naev.pot file to sync up the comments with the current source tree. (Weblate's magic depends on naev.pot being committed, which leads to some churn as routine source code edits lead to line numbers and comments changing.)
BTW, thanks for the shout-out, but the real translation work (other than workflow/tooling stuff) is being done by other authors.

PR #1906, opened 2021-09-10
Fixed a bug where there is a missing character at the beginning of the line.

oo13:
The text with no space, such as Japanese text, makes it wrap forcibly, and a character at the beginning of the next line is deleted in some texts.
This PR avoids to delete it.

I checked a news text and a mission objective in Japanese.


bobbens:
Looks good, thanks!

PR #1907, opened 2021-09-11
Fix a bug that causes invalid unicode characters.

oo13:
When a text is forcibly wrapped, it may cause an invalid unicode character.
This PR fixes it.

I checked the tutorial mission in Japanese and Naev showed no 'Warning: [font_makeChar]...' messages. (a82f1ca87eea15b96c1526a305c6a4f43fbeed34 printed "Warning: [font_makeChar] Font 'fonts/IBMPlexSansJP-Medium.otf' unicode character '0xfffff279' not found in font! Using missing glyph.")


bobbens:
Looks good, thanks!

PR #1913, opened 2021-09-22
IFC addition

Lineth:
Peliminary adding of iflockon values

Torpedoes lockon fast, but calibration takes a long time. Aim carefully!

missiles kinda split on. Idea: Specialist missiles could take a long time to lockon, but calibration is fast. Regular missiles are split down the middle, more or less.


bobbens:
Looks good, thanks!

PR #1914, opened 2021-09-24
In case ai_newtask() is supposed to make a subtask of nothing, try to raise a Lua error?

UncombedCoconut:
Instead of letting a NULL dereference happen?
This idea based on -fanalyze feedback. I'm not sure if it's a good one or not. (Are we less dead if we pass `L`around and try to propagate the error?)


bobbens:
Raising an error does seem best.

PR #1927, opened 2021-10-18
fix energy usage

not-pyroman:
removed the old energy usage that bobbens forgot to remove


not-pyroman:
I chose the newer one which was -50, and it seems that bobbens forgot to remove the old energy usage when they made the change

PR #1944, opened 2021-11-24
Add some tags and a typo in zpp02

not-pyroman:
the only reason there's over 10 commits for a small number of things is merging


bobbens:
Looks good, thanks!

PR #1959, opened 2021-12-06
Minerva missions proofread

aguynamedryan:
I've done some minor proof reading for the missions based on Minerva.  I tried to separate them into minor edits (typos), smaller edits (word choice) and larger edits (changes to wording/sentences).


bobbens:
Some comments (not affecting merging though):
1) Some of the capitalization of **Nebula** is on purpose to distinguish between just normal nebula and "The Nebula" which is the particular one centered on Sol. This is inconsistent all around and probably could use some unification, in this case the PR makes it all common nouns, but I don't think this is necessarily an issue.
2) I originally went with gender-neutral pronouns most of the time given the artwork wasn't drawn, and wasn't entirely sure what to go with. However, I don't see any issue with changing them.

Thanks a ton for the proofread! Looks like it was a lot of work :)

PR #1960, opened 2021-12-07
Get the system instead of the string in maikki2

sigus:
Without this, the condition `system.cur() == eccsys` does not work properly.


UncombedCoconut:
Thanks, I'll leave it up to bobbens whether he wants to do system.cur():nameRaw() or just use an object like all the non-Minerva scripts do. :)

PR #1961, opened 2021-12-07
artifact => artefact

aguynamedryan:
A mia culpa for not realizing "artefacts" is legit British spelling


UncombedCoconut:
IMO, changing `mem.*` Lua variables is probably going too far. These are preserved by name in save files (if the player saves mid-mission), e.g. `mem.artifactA` produces `<data type="commodity" name="artifactA">...</data>`.
So the change creates a minor/conditional save-compatibility break.
This isn't a deal-beaker since we're pre-1.0, but I'm used to getting a bit more in return. :)

bobbens:
Thanks for the patch. So we'll revert some of the mem changes but leave the rest.

PR #1965, opened 2021-12-09
Dvaered mission edits

aguynamedryan:
Very minor edits to some Dvaered missions


UncombedCoconut:
Ooh, " paradigm objective" is a good find. :) Thanks!

PR #1969, opened 2021-12-14
Change "monkey" to "jockey"

aguynamedryan:
I might be oversensitive, but monkey is possibly offensive


UncombedCoconut:
Yeah, it's supposed to be rude to the player character only. I don't know whether the phrase would have been hurtful to anyone, but this edit won't be!

PR #1974, opened 2021-12-18
Minor edits to neutral missions

aguynamedryan:
- Mostly word choice and punctuation
- Rearranged a couple of awkward sentences


UncombedCoconut:
Looks good! Thanks! I'm going to pare this back just a little where I think the context justifies it:
- One old-person colloquialism in oldwoman.lua (based on a Google-fight between phrases, not firsthand knowledge, of course 😛)
- The sightseeing message about a non-luxury ship: only shown when you actively bait-and-switch passengers by exchanging ships, so I kept "you" as the subject of the clause.

PR #1976, opened 2021-12-19
No free transponders

sigus:


not-pyroman:
There's probably a cleaner way to do it other than this hacky method, does the player.pay thing even like negative numbers?

UncombedCoconut:
> There's probably a cleaner way to do it other than this hacky method, does the player.pay thing even like negative numbers?

Yes, in fact `player.pay(-cost)`, `player.pay(-fine)`, etc., is a pretty standard thing to do.

This looks good to me, thanks!

not-pyroman:
Huh, neat to know

PR #1978, opened 2021-12-20
Color => Colour

aguynamedryan:
Colour me grateful for Naev


UncombedCoconut:
Just to give me an idea of how much follow-up work will be needed: are you planning many more of these US->GB English changes?
(My workflow is something like: look over the changes, merge, prepare something like https://github.com/naev/naev/commit/c6ba7a3c58dcb162587c22cedecea94fd8d44266, regenerate the .pot file, push, make sure Weblate is happy.)

UncombedCoconut:
And in this case it should be very easy. Thanks! Merging!

PR #1979, opened 2021-12-20
More british

aguynamedryan:
Probably my last bout of US -> GB changes for a while

The ize => ise commit is a bit risky, but the rest should be safe, though I'd recommend someone else look over these changes


UncombedCoconut:
Yikes, I'll give it a look tomorrow. :)

aguynamedryan:
Just to shake my confidence in my work here, I found a file I hadn't saved so I have force-pushed an update :-/

UncombedCoconut:
I've merged a fair bit of this (commit at a time).
What remains:

- Merging this PR would convert -ize/-izat.../-yze endings. I think dev team consensus is we'd rather not, but in fairness that's based on 2 of us having an opinion and a third contributor adding information.
- The correct middle ground may be "Oxford spelling" which may mean we do want the analyze->analyze bit, just not most of the changes remaining in this PR.
- The PR included changes to the word "siphon". TODO: check on those and revert https://github.com/naev/naev/commit/fe7201ebc781a8844e316fe9da2402061c6b2eb2 if we like them.

UncombedCoconut:
Thanks for this! This round of merge conflicts tells me: it's time to move on from the part we didn't merge.

PR #1983, opened 2021-12-26
Rename dat/licenses directory

a-random-lemurian:
```
I thought the code for the in-game licenses was in the licenses folder,
and it wasn't. Hopefully if we rename it to copyright, that confusion
will be solved.

Copyright licenses haven't been deleted, I assume git or github suck
at showing diffs. You might want to review my fork to check.
```


a-random-lemurian:
Well, I suck at git. We make mistakes sometimes.

bobbens:
I do think it's more confusing if we have a `LICENSE` and other `ARTWORK_LICENSE.yaml` and such files all over, and the folder gets renamed to copyright. An easier more consistent way of distinguishing them would be to just make the directory name uppercase `LICENSE/`, which would be consistent with a) us calling the files LICENSE all over, and b) using all uppercase to denote license files.

a-random-lemurian:
I've force pushed with lease.

bobbens:
Looks good, thanks!

PR #1987, opened 2021-12-27
Replace "asset" to "spob" in translation.its.

oo13:


bobbens:
I knew I was forgetting something. Thanks!

PR #1993, opened 2021-12-31
Minor edits to sirius missions

aguynamedryan:


UncombedCoconut:
Thanks. FYI, I'll tinker with this a bit more after merging.

PR #1997, opened 2021-12-31
land and hyperspace thrid argument is now 'noshoot' and not 'shoot'

nenau-again:
We revert the default behaviour for `land` and `hyperspace` concerning shooting with turrets.
I tried to adapt missions and events to that, but I'll let this PR here for some time just in case someone notices that one of the missions they know well hasn't been adapted well.


UncombedCoconut:
Looks pretty good to me. I checked as exhaustively as I could for issues and saw two minor ones.

`dat/events/sirius/preach.lua`: this comment is now obsolete: "--attack back as they move away?"

`dat/missions/shark/sh01_corvette.lua`: based on its comment, we probably want to specify noshoot:
```lua
   --making sure the shark doesn't continue attacking the player
   sharkboy:hyperspace(escapesys)
```

nenau-again:
For the first one, I thought it was a question asked by the contributor, but now, it looks more like a TODO, that has indeed been solved by this. I'll remove the comment
For the second one, I had missed it, thanks (although the chances for the enemy to have a turret are close to 0).

bobbens:
I think this is good to merge, thanks!

PR #2011, opened 2022-01-17
Minor spelling mistake(s)

not-pyroman:


UncombedCoconut:
Looks like there are stray letters ("Ic") on line 170 here. :(

not-pyroman:
oof

PR #2015, opened 2022-01-21
Add Autonav button to mission screen

jjustra:
Player can set autonav to mission's target system, before accepting mission.


UncombedCoconut:
Cool! Tempting to change the main map's autonav key now (first bound in Feb 2021) to enable a better future...

PR #2017, opened 2022-01-22
H key cycles missions in map window

jjustra:
Player can cycle+autonav its missions from map window.

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -908,9 +908,45 @@ static void input_key( int keynum, double value, double kabs, int repeat )
          } else
             player_land(1);
       }
-   } else if (KEY("thyperspace") && NOHYP() && NOLAND() && NODEAD()) {
-      if (value==KEY_PRESS)
-         player_targetHyperspace();
+   } else if (KEY("thyperspace") && NOHYP() && (NOLAND()||map_isOpen()) && NODEAD()) {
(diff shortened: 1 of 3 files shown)


bobbens:
Cool idea! However, I don't think the hyperspace target key should be used for this, since it can already be useful for selecting a nearby system. What about using another keybinding like `target_next` and `target_prev`?

jjustra:
I personally used `thyperspace` practically exclusively for mission going and it seemed to make more sense to me, because you do set hyperspace, but then again I have not that much Neav hours under my belt, and some keys are still outside my muscle memory, so your judgement will probably be more insightful. I will adjust it to `target_` keys.

jjustra:
Change made. I've added bind to `target_nearest` (do same as `target_next`) for convenience, if that is ok.

bobbens:
Looks good, thanks for the patch! I've merged with a few changes:
1. Made the C formatting fit Naev style a bit more
2. Disable autonav functionality since it's a bit non-intuitive (you can just hit `A` to have it trigger autonav and close the window)
3. Made it center the target system when you toggle through them

jjustra:
Thanks for fix; autonav was artifact that I forgot to remove - initial version identified currently selected system by it.

PR #2018, opened 2022-01-23
Fix two typos in minigames/chuckaluck.lua.

oo13:
There is a typo in the singular text of n_().


onpon4:
This is an incorrect correction. The number that varies is {n}. In generall, the ngettext function (which `n_()` is an alias of) contains a variable number which is programmatically inserted; in this case, {n} is replaced with the number of Minerva Tokens won, and the purpose of ngettext is to properly handle plural forms for various languages.

🕵️

bobbens:
Looks good, thanks!

PR #2020, opened 2022-01-25
A couple updated descriptions.

LJDude:
Emergency Stasis Inducer's description should be a bit more clear now.

Changed "fighter" to "combat ship" because the Admonisher is not a fighter.


UncombedCoconut:
Thanks!

PR #2022, opened 2022-01-26
DIalogue option leading to nothing

not-pyroman:
Im just going to comment it out maybe until someone writes it


bobbens:
Looks good, thanks!

PR #2026, opened 2022-01-26
CPU Modification Outfits

LJDude:
I've added some outfits that increase ships' CPU capacity since I saw that nothing but the Left Boot did that:

Efficiency Combat AI: Medium utility, requires heavy weapon license, only one combat ai per ship etc. same as the other Combat AIs. This provides a 5% increase in CPU capacity. If that proves to be too much, it could be nerfed down to 4% or 3%.
If "Efficiency" turns our to be an unsuitable name, we could use some variation of "Overclock" instead. 

Auxiliary Processing Units: Structure outfits that provide a set amount of CPU capacity at the cost of mass and energy. Comparable to Shield Capacitors and Batteries. If we wanted to make the graphics relatively simple, we could reuse the graphics from those with some different colors and call these something like "Advanced Processor Capacitors" instead.

Please don't mind the extra commits of me merging stuff into main.

### DIFF
--- a/dat/outfits/structure/auxiliary_processing_unit_i.xml
+++ b/dat/outfits/structure/auxiliary_processing_unit_i.xml
@@ -1,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<outfit name="Auxiliary Proccessing Unit I">
+ <general>
+  <typename>CPU Modifier</typename>
+  <slot>structure</slot>
+  <size>small</size>
+  <mass>15</mass>
+  <price>20000</price>
+  <description>An additional proccessing unit to supplement the main processor of the core system, giving small ships a little more processing power.</description>
+  <gfx_store>placeholder.webp</gfx_store>
+  <cpu>5</cpu>
+ </general>
+ <specific type="modification">
+  <energy_regen_malus>4</energy_regen_malus>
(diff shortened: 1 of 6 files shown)


bobbens:
For graphics, I would just use the old APU graphics, and replace the graphics for the outfits they are being used now. I'm sort of trying the outfits now, but it could be just that everything is already balanced around not having extra CPU that I have a hard time thinking of good usage cases. Would you happen to have any examples where APUs and such really shine?

LJDude:
Sure. The three that come to my mind are:
1. Mini bays and docks on medium ships: mini bays and docks take up a lot of CPU power and its often hard to use them on a medium ship properly without using a medium-heavy core and/or the Left Boot. I'm not sure if there are any meodum shops besides the za'lek's that have bay slots, but the issue is still there.
2. Low stage soromid ships: I remember my soromid reaver starting out with _three_ cpu. I don't think that's enough to run even an ultralight weapon. I'm guessing that other soromid ships are similar, but I'd have to check to be sure.
3. Say someone wants to be a madman and put a ripper cannon on a shark. Well, with an APU, the ECAI, and Left Boot all together, you could do that and probably still have some left over.

bobbens:
Looks good, thanks!

PR #2027, opened 2022-01-27
Za'lek Cores and Outfit Rework

bobbens:
In order to further differentiate ships and make things a little less uniform, I'm working on giving Za'lek their own cores and reworking outfits to work with them. I'm interested in getting feedback on the changes.

Basic ideas for Za'lek outfits and cores is as follow:
1. More CPU
2. Less energy and energy regen
3. Less mass
4. Competitive with existing top tier cores

This is applied to the following outfits:
1. New system and engine cores
2. All beam weapons

One other change that I've had to do to make destroyer systems less appealing to corvettes is to increase their mass. However, at the same time I have lowered the mass of the destroyers so, in general, there should be no change for all ships equipping their corresponding core. There is, however, one exception. The Starbridge already has very low mass so I haven't touched it, which sort of leads to it being nerfed. I'm not entirely sure how to fix that without making the patagium engine either too good, or making it lighter than a corvette.

### Cores
The idea is for the Za'lek to use their own engines and systems, and rely on the low-end unicorp armours. For this reason two new lines of engines and systems are introduced. Currently these only exist in medium and large formats given that the Za'lek drones have all the functionality directly built in and are not player-acquirable.

* **Za'lek Central Kernel (ZCK):** are the new systems. They use the naming system `ZCK-(M,H)-{MASS}`. They have roughly 50% of the energy, 200% CPU, and 100% shields of MilSpec Orion cores.
* **Za'lek Primary Drives (ZPD):** are the new engines. They use the naming system `ZPD-(M,H)-{MASS LIMIT}`. They are roughly like the equivalent Tricon engines, but with 60%ish mass limit, much lower energy consumption, have a bit more speed vs thrust/turn, and also use a tiny amount of CPU (significant if not using ZCK).

The cores are designed to complement each other well, giving much more CPU with lower mass at the expense of much less energy regen and mass limit. This makes them work really well with the new Za'lek weapons, but trickier to use with other weapons due to low energy.

### (Beam) Weapons
The Za'lek beam weapons have had their CPU usage double, but mass and energy usage halved. They are usable with the Za'lek cores, but outside of specific set-ups, they are hard to use with other cores due to high CPU usage. However, they can be used and complemented with launchers.

### Fighter Bays
The Za'lek fighter bays are significantly lighter now, but use significantly more CPU. In general, you won't be able to use more than one or two without the Za'lek cores on other ships.

### Testing
This is still a sort of preliminary stage, so I haven't tightly integrated anything else. I'm interested in getting feedback on how it feels. In particular, the following questions are important:
1. Can you equip Za'lek ships enjoyably?
2. Are there any broken combinations that appear by mixing Za'lek and non-Za'lek cores?
3. Are the Za'lek weapons/bays still somewhat usable with other ships?
4. Do things feel balanced?

You can give yourself all the cores with the following code (they are also available at Zied):
```
for k,o in ipairs{
   "ZPD-H-2500 Drive",
   "ZCK-H-250 Kernel",
   "ZPD-H-4000 Drive",
   "ZCK-H-600 Kernel",
   "ZPD-M-500 Drive",
   "ZCK-M-40 Kernel",
   "ZPD-M-900 Drive",
   "ZCK-M-160 Kernel",
} do
   player.outfitAdd( o, 5 )
end
```

### TODO

- [x] Create medium and large core outfits
- [x] Tweak all the Za'lek beams and fighter bays to fit the paradigm
- [ ] In depth testing
- [ ] Add the outfits for sale
- [ ] Add the outfits to the equip scripts
- [ ] Tweak the launchers
- [ ] (optional) Add light cores too
- [ ] (optional) Add some structural outfits that work with the low mass paradigm (maybe cost CPU or something else?)


nenau-again:
I am in doubt on the topic of faction-specific cores.
Basically that would mean we tend towards a situation where every single ship has its own specific cores, from which would result that the concept of interchangeable cores would be pointless, but hard to defend against exploits.
If we create faction-specific cores, we will have to decide if the modifiers on ships are still useful: for example zaleks have stronger shields and lower armor, but that can be achieved via modifiers on ships (as now) or via special cores (which would lead all ships to be nearly equivalent, except for the number of slots), or finally it can be done via a bit of modifiers and a bit of special cores, which is confusing.

I think we should clearly decide what should be given by the cores and what should come with the ship model (maybe that has been done back then when the core concept was introduced) Because I get the feeling that diversifying cores like that is making things overly complex for no real benefit for the regular player, and exposes us to exploits. Plus, it makes balancing really hard.

bobbens:
> I am in doubt on the topic of faction-specific cores. Basically that would mean we tend towards a situation where every single ship has its own specific cores, from which would result that the concept of interchangeable cores would be pointless, but hard to defend against exploits. If we create faction-specific cores, we will have to decide if the modifiers on ships are still useful: for example zaleks have stronger shields and lower armor, but that can be achieved via modifiers on ships (as now) or via special cores (which would lead all ships to be nearly equivalent, except for the number of slots), or finally it can be done via a bit of modifiers and a bit of special cores, which is confusing.

I think that's a valid criticism. As you mention, giving exclusive access to slots to factions does lower the slot sharing mechanic, and blurs the line between ship and core bonuses.

> I think we should clearly decide what should be given by the cores and what should come with the ship model (maybe that has been done back then when the core concept was introduced) Because I get the feeling that diversifying cores like that is making things overly complex for no real benefit for the regular player, and exposes us to exploits. Plus, it makes balancing really hard.

I don't think we necessarily need a perfect balance. There will always be some stronger combinations, and as a single-player game, I don't see that as being much of an issue. Maybe instead of focusing it as a "Za'lek faction slots" it might make more sense to think of it as "high tech" vs "low tech". Right now, all the outfits are balanced on a single tier, which while it is easiest to do, it makes it so everything is boring and bland. Basically most weapons are just minor flavours of the same thing: same EPS, DPS, etc. with only the damage type changing. That is bad. Some ideas to combat this are:
1. Diversify weapons (like my changes to bioplasma and laser, but this basically focuses on changing range/damage not EPS/DPS)
2. Do a differentiation between high / low tech. Such that high tech outfits require more CPU, but less mass/energy, while low tech outfits require more mass/energy, but less CPU. Although I was working on doing it faction-specific, it could be done on a more general level. I can imagine Za'lek+Thurion using high tech, Dvaered using low tech, Soromid/Empire/Proteron/Sirius using a mix or mid tech. Currently things like mace rockets and launchers do act as low tech (although no energy usage), while other weapons would be mid-tech type

tl;dr I'm not set on doing faction cores and outfit trees, but we need to think about how to differentiate things a bit more than what we have

PhoenixRiver:
To add to the high/mid/low tech idea to make it a little more durable against "hacking" or at least balanced is maybe to have pirates (and maybe other factions via them or directly) steal ship organisation ideas from the player for themselves -> I feel like the pirates are a natural fit for this but government agents might be interested in doing the same if the player ship is clearly superior! (e.g. EPS/DPS calculation).
Maybe the faction wanting a copy of the has to "engage" in a mission against the player... so the player might get tailed by a faction ship and/or get scanning messages keep showing up on their flight feed and/or have suspicious people turn up at the bar interested in talking to the player about their ships and/or get attacked and disabled with the boarding party wanting to steal the ships' plans. The player might get the option of killing or trying to stop this or else trying to deceive them or just selling them the plans.
Granted all of this might lead to other "attacks" against the game like making pirates use high DPS/EPS ships that have terrible armour or stealth etc. but at that point maybe the player deserves to have an edge given the amount of effort that might take! ;)
It might also be fun for the player to face off against a mirror of their own ship piloted by the AI :D

PhoenixRiver:
Sudden thought... one thing that worried me when I was playing is that there are an awful lot of things, particularly core outfits on the outfit pages of stations and planets... however, most of the time, they are little to no use to me because they are not the correct size for my current vessel or even an vessel I own.
What seemed like a good idea then would be for the pages that have only specific items ("weapons", "cores" etc.) defaulting to only showing those that would fit your current ship and have a toggle to show everything else.
This way the player does not lose visibility of how much is offered in total on the first page of the outfit shop but does have their choice reduced when they choose specific areas to what is most useful to them at the present moment (with an option to change that if they want)...
I might also suggest that a page be dedicated to outfits the player already owns but does not have fitted that they can sell at this location (i.e. they are also sold here)... as, due to the large number of outfits, it can be hard to locate the outfit I want to sell at places where I am most likely to sell it... places with lots of interesting outfits! : D

PhoenixRiver:
> You'll be glad to hear that we have something like that already :) You can choose only stuff equipable by your ship or stuff equipable by any of your ships too. There is no selling interface (as in only show stuff you own), but that could be added. Might make more sense to just allow the player to sell anything from the equipment window though.

Oh! I hadn't realised... I guess it is not in the Steam version (0.9.2) yet or I'm doing something wrong?
Selling from the equipment screen would make sense though maybe an alert would have to pop up for items that can't be sold at that outfitters explaining why not.

I was also going to suggest that the little letters in the corners of the outfit icons match the ones for the menu tab in the outfitters but then I realised there is a mismatch due to Cores all being put together and being removed from the ones they would naturally be in. I don't know if that'd helpthough, the letters are pretty small (at least at my res. and font size.

bobbens:
Closing as this is quite dead.

PR #2033, opened 2022-01-29
Starmap mark system

jjustra:
Players can mark systems on map and write notes to them.
New button 'Mark system' is added with 'k' bind.
All is saved to save file (non-breaking change).

### DIFF
--- a/src/map.c
+++ b/src/map.c
@@ -43,7 +43,7 @@
 
 #define MAP_MARKER_CYCLE  750 /**< Time of a mission marker's animation cycle in milliseconds. */
 
-#define MAP_MOVE_THRESHOLD 10. /**< Mouse movement threshold */
+#define MAP_MOVE_THRESHOLD 20. /**< Mouse movement distance threshold */
 
 /**
  * @brief Faction presence container to be used for the map information stuff.
@@ -956,7 +956,7 @@ static void map_render( double bx, double by, double w, double h, void *data )
    /* Render system markers and notes. */
    if (cst->alpha_markers > 0.) {
       map_renderMarkers( x, y, z, r, cst->alpha_markers );
-      //map_renderNote( bx, by, x, y, z, w, h, 0, cst->alpha_markers );
+      map_renderNote( bx, by, x, y, z, w, h, 0, cst->alpha_markers );
    }
 
    /* Render commodity info. */
@@ -1365,7 +1365,7 @@ static void map_renderPath( double x, double y, double zoom, double radius, doub
 void map_renderNote( double bx, double by, double x, double y,
       double zoom, double w, double h, int editor, double alpha )
 {
-   double tx,ty, mx,my;
+   double tx,ty;
    glColour col;
    glFont *font;
    StarSystem *sys;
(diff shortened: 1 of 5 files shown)


src/map.c
```diff
@@ -1385,6 +1389,12 @@ void map_renderNames( double bx, double by, double x, double y,
       col.a = alpha;
       gl_printRaw( font, tx, ty, &col, -1, _(sys->name) );
 
+      /* Render note */
```
bobbens, line 1392: It might be better to render the note on mouse over like in the universe editor.

src/map.c
```diff
@@ -2023,6 +2033,33 @@ static void map_setMinimal( unsigned int wid, int value )
...
+      if (sys->note != NULL) {
+         free(sys->note);
+         sys->note = NULL;
+      }
+
+      /* Switch marking */
+      if (sys_isFlag(sys, SYSTEM_PMARKED))
+         sys_rmFlag(sys, SYSTEM_PMARKED);
+      else {
+         sys->note=dialogue_input(_("System note"), 0, 60, _("You can write note to this system"));
```
bobbens, line 2057: How about setting text to something like `"Write a note about this system."`?

src/map.c
```diff
@@ -1428,23 +1438,23 @@ static void map_renderMarkers( double x, double y, double zoom, double r, double
...
+      n  = (sys_isFlag(sys, SYSTEM_CMARKED | SYSTEM_PMARKED)) ? 1 : 0;
       n += sys->markers_plot;
       n += sys->markers_high;
       n += sys->markers_low;
       n += sys->markers_computer;
 
       /* Draw the markers. */
       j = 0;
-      if (sys_isFlag(sys, SYSTEM_CMARKED)) {
+      if (sys_isFlag(sys, SYSTEM_CMARKED | SYSTEM_PMARKED)) {
```
bobbens, line 1457: It would probably be better to have the player notes markers de independent of the computer markers and probably have a different shape to be easily identifiable. I guess a truncated rhombus or something like that could work. I can help with the Signed Distance Function Shader if you want.

bobbens (changes requested):
Overall it looks really cool. Nice work! I've made some comments, but the one that is probably most important is to make it visually separate from the computer markers, although the other changes would be nice to have too.

src/map.c
jjustra: That sounds much better

src/map.c
jjustra: Help would be great .. I know practically nothing about this now.

jjustra: I've separeted things like you've said .. it calls `map_drawMarker( tx, ty, zoom, r, a, n, j, 5 );`

src/map.c
jjustra: Universe editor is for now Terra incognita for me - will look into it asap ... it may also be useful if pressed `shift` key temporarely reveal all notes ?

bobbens: screenshot028

You need to set `devmode = true` in the conf for the editor, but I've attached a screenshot.

src/map.c
bobbens: Don't you need to increment `n` like?
```C
n  = (sys_isFlag(sys, SYSTEM_CMARKED)) ? 1 : 0;
n  += (sys_isFlag(sys, SYSTEM_PMARKED)) ? 1 : 0;
````

src/map.c
jjustra: Mouse over is WIP as of now - I have trouble sorting coordinates calculation.

src/map.c
jjustra: Note : marker type 5 colour is temporarely set to &cMarkerNew in map.c:867 .

src/map.c
jjustra: Mouse-over is done; I've added toggling with 'overlay' (tab) key for convenience.

bobbens:
Thanks for the fixes. I've touched some things and merged it. One thing I haven't gotten around to is making it so it fades in/out instead of just appearing and disappearing. I'm also not entirely sure on the positioning, but have left it as you implemented it for now. I think it'll be fairly useful. If you want to do changes or have any ideas to improve the visuals, I'm all ears. Thanks again :)

bobbens:
Another note mentioned in discord is that it would be nice to be able to change the colours of the notes. This would require creating a custom dialogue that has an input + a colour selector (custom widget). It sounds like a nice thing to have, but a bit tricky to implement.

jjustra:
I am glad I can be helpful, this is really fine project :) I'll probably look into colors in future.

PR #2034, opened 2022-01-29
extended per-size filter to the outfit window

nenau-again:
This addresses https://github.com/naev/naev/discussions/2025
I'm not fully satisfied with my implementation as it introduces code duplication. however, this seems to be the standard way to do things between equipment and outfit screen : there was already a lot of duplicated code...

One question I'd like to raise is that `equipment_outfitMode` is saved in the player's data. I don't think it's right as there is a risk a player sets it to "medium" for example, forgets about that button, and then never finds the other outfits back. This is why my new variable `outfit_Mode` for the outfit window is not saved.
On the other hand, I can understand it's handy if the setting is saved.

Last thing, the different modes use actual integers. Using an `enum` would feel more right, but as I don't master that part, I don't know if it's easy to do (there would be an operation to add in `equipment_outfitPopdownSelect` I guess.

### DIFF
--- a/src/equipment.c
+++ b/src/equipment.c
@@ -1464,12 +1464,26 @@ static void equipment_genShipList( unsigned int wid )
 }
 
 static int equipment_filter( const Outfit *o ) {
-   Pilot *p = eq_wgt.selected;
+   Pilot *p;
+   const PlayerShip_t *ps;
+
    switch (equipment_outfitMode) {
       case 0:
          return 1;
 
-      case 1:
+      case 1: /* Fits any ship of the player. */
+         ps = player_getShipStack();
+         for (int j=0; j < array_size(ps); j++) {
+            p = ps[j].p;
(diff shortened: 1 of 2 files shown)


UncombedCoconut:
The amount of code duplication looks OK to me. I agree with your view about `equipment_outfitMode` (though I'm not familiar with the reason it's saved). Changing an int to an enum in C is easy and we can do it after resolving this PR.

I'd merge right away, but I'll let bobbens look first (so as not to defeat the purpose of raising a PR).

bobbens:
I like having the mode saved, but it could just be me and my testing save with all the outfits. As I change ships a lot and equip all the time, having to click the option every time you start a new game is a bit annoying. I would like to make it default but since it has the potential for confusion I haven't (ship AI could go all clippy on the player potentially though).

That said, this PR looks great and will be very useful, and since @UncombedCoconut also has no qualms with it, into main :)

PR #2052, opened 2022-02-03
Made Teracom Headhunter Launcher available outside of Bon Sebb

LJDude:
That launcher accidentally got removed along with the Unicorp launchers. The outfit was added back in, but tech.xml wasn't updated to match.

Please don't mind the extra, obsolete commits.


UncombedCoconut:
looks good, thanks

PR #2055, opened 2022-02-05
Changed how the Schroedinger does fuel a bit

LJDude:
This is based on an idea @nenau-again mentioned in Discord. Link to the conversation for those in the server: https://discord.com/channels/661743268505190400/772643725255245875/939522574050787358

The idea was to have the Schroedinger fuel use lowered to 50 per jump and have the base fuel bonus reduced as well. The reason I've reduced the bonus so much (800 to 150) is because the amount of fuel that the engines give also doubles. With a Tricon engine, you'd still get 9 jumps; only two less than before. If you use a fuel pod with a Tricon engine, then you get the same amount as before: 13 jumps.

The maximum amount you can get (achieved by using a Mendelez engine and a fuel pod) is 17 jumps, only two more than the previous maximum. The minimum amount is 7, three less than the previous minimum, but who's going to be using a beat up engine, anyways?

### DIFF
--- a/dat/ships/schroedinger.xml
+++ b/dat/ships/schroedinger.xml
@@ -21,8 +21,8 @@
  <characteristics>
   <crew>2</crew>
   <mass>20</mass>
-  <fuel>800</fuel>
-  <fuel_consumption>100</fuel_consumption>
+  <fuel>150</fuel>
+  <fuel_consumption>50</fuel_consumption>
   <cargo>5</cargo>
  </characteristics>
  <slots>


nenau-again:
I obviously agree with that PR, but I guess we'll wait to see if anyone is against it.
How much did you test this out ? (after all, it's the first time we try fuel consumption lower than 100 on a player ship)
I'm not aware enough on how refuelling works to know what will happen if you try to buy for 50 fuel to a NPC, will they charge you for full 100 fuel ?

UncombedCoconut:
Yes, dat/comm.lua forces fuel transfers to be multiples of 100. However, I'm not against fixing it and doing this. (It's already possible for things like Stellar Winds system to make the current rule a bit iffy.)

LJDude:
> How much did you test this out ?

I actually haven't tested it out yet. It just made sense to me and I'm not sure how it would mess up, other than buying fuel from ships.

bobbens:
Looks good, thanks! I'll merge seeing there are no issues. Comm should probably be improved to consider sub-100 units, but I don't think it should be breaking, just makes the player waste more money :)

PR #2060, opened 2022-02-07
Medium Armor-Strengthening Structure Outfit

LJDude:
I found it a bit odd that there were just small and large outfits to improve armor strength and no medium one, so I made one.

Short conversation about this for those in the Discord server: https://discord.com/channels/661743268505190400/661743693279133706/939360810344521781

### DIFF
--- a/dat/outfits/structure/microbond_plating.xml
+++ b/dat/outfits/structure/microbond_plating.xml
@@ -1,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<outfit name="Microbond Plating">
+ <general>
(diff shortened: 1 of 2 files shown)


onpon4:
In fact, Plasteel Plating was originally a medium outfit. Considering how little use light ships have for armor expansion I have no idea why Plasteel Plating was converted into a small outfit.

🕵️

bobbens:
This looks good, thanks! As a follow-up also on discord, instead of using relative modifiers, it probably makes sense to use absolute values instead of relative ones. If you are interested in doing that I would be very interested in a PR!

PR #2061, opened 2022-02-08
invalid slot size

not-pyroman:
as far as Naev and I are concerned, Heavy is not a real slot size
Bobbens made an oops, and did not check, or forgot to push a fix


bobbens:
Late night commits aren't very good sometimes. Thanks!

PR #2064, opened 2022-02-12
Nebula Research Campaign

V-ktor:
Updated the nebula research campaign to use VN and added a new mission.
Also, give the player an accessory that grants stress dissipation and heat dissipation.


dat/outfits/accessory/broken_nebular_shielding_prototype.xml
```diff
@@ -0,0 +1,18 @@
...
+  <typename>Accessory</typename>
+  <slot prop="accessory">utility</slot>
+  <size>small</size>
+  <priority>3</priority>
+  <mass>0</mass>
+  <price>800000</price>
+  <description>A broken prototype for a novel shielding device that is supposed to enable travels into the Sol nebula. However, due to its malfunction you almost died. It seams, however, to absorb harmful energy and emit heat radiation into space.</description>
+  <gfx_store>misc02.webp</gfx_store>
+ <rarity>6</rarity></general>
+ <specific type="modification">
```
bobbens, line 14: 10% stress + heat might be too weak. Might be better to give some other property too or just increase the values to probably around 25%.

dat/missions/zalek/neburesearch/neburesearch_03.lua
```diff
@@ -82,124 +94,218 @@ end
...
+    vn.menu( {
+        { _("This is a good question."), "avoid_question" },
+        { _("It's an open question."), "avoid_question" },
+        { _("Run!"), "run" },
+    } )
+
+    vn.label( "avoid_question" )
+    vn.na(_([[You try to avoid the question. The scientist seems to be upset. There are no further questions. Apparently, there is another talk that starts very soon so people hurry off. You manage to leave without gaining much attention and return to your ship. Enough science for today!]]))
+    vn.func( function ()
         faction.modPlayerSingle("Za'lek", 1)
```
bobbens, line 299: Would be probably better to move the standing modifications into the common script to make it easier to edit if necessary.

dat/missions/zalek/neburesearch/neburesearch_04.lua
```diff
@@ -26,103 +27,148 @@
...
-         zlk.addNebuResearchLog(_([[You helped Dr. Mensing to acquire funding for a shielding prototype that will enable to explore the Sol nebula.]]))
-         misn.finish(true)
-      end
-   end
+    mem.landed = spob.cur()
+    if mem.landed == mem.dest_planet then
+        vn.clear()
+        vn.scene()
+        local mensing = vn.newCharacter( nebu_research.vn_mensing() )
+        vn.transition("fade")
```
bobbens, line 107: `vn.transition()` should be equal to `vn.transition("fade")` (doesn't necessarily need a change but might be more future-proof)

dat/missions/zalek/neburesearch/neburesearch_05.lua
```diff
@@ -0,0 +1,377 @@
...
+        ghost: moveto(player.pos(), false, false)
+        hook.timer(5.0, "moveto")
+    end
+end
+
+function endScan()
+    local osd_msg = {}
+    local ps = player.pilot()
+    mem.stage = 5
+    osd_msg[1] = _("Survive.")
```
bobbens, line 251: I don't think most of the OSD use punctuation at the end (no `.`, but sometimes `!`)

dat/missions/zalek/neburesearch/neburesearch_05.lua
```diff
@@ -0,0 +1,377 @@
...
+ <notes>
+  <campaign>Nebula Research</campaign>
+ </notes>
+</mission>
+--]]
+--[[
+   
+   Mission: Expert Nebula Research
+   
+   Description: The nebula shielding prototype is ready and the player has to test it. The first mem.stages go well, but in the final mem.stage the prototype fails and the player is about to die in the Sol nebula. He gets rescued by a Thurion carrier instead and is bought into Thurion space.
```
bobbens, line 23: Isn't this description wrong? Isn't it an Empire Peacemaker not a Thurion ship?

bobbens (commented):
Looks good, thanks! Added some minor nitpicking comments. It might also be an opportunity to add a `readme.md` (like the [black hole one](https://github.com/naev/naev/blob/main/dat/missions/zalek/blackhole/readme.md)).

One thing that you should do is set up `pre-commit` if possible (as explained on the main repor readme), because it does seem like luacheck isn't too happy with some stuff and is causing the github actions to fail.

bobbens:
Almost forgot, if the campaign is completed, and you want to give the player a faction standing boost at the end, something like
```xml
 <tags>
 <tag>zlk_cap_ch01_med</tag>
 </tags>
```
Would give 3 points to the player max standing limit (for now, value can be changed later). You can do `_lrg` for 5 instead.

UncombedCoconut:
Nice. Merging, and I'll fix Luacheck's nitpicks.

PR #2067, opened 2022-02-17
Ship description proofreading

PhoenixRiver:
Proofread all the ships; I'll note that the Kauweke and Taitamariki Soromid ships have descriptions marked "todo" - I changed them to highlight them ;) - as well as three of the Thurion ships (Apprehension, Certitude and Ingenuity) having "..." which seemed less urgent or possibly unnecessary to do. I'D do their descriptions but I know nothing about these ships yet :/

### DIFF
--- a/dat/ships/hyena.xml
+++ b/dat/ships/hyena.xml
@@ -13,7 +13,7 @@
  <trail_generator x="-15" y="-6" h="0">fire-thin</trail_generator>
  <trail_generator x="-15" y="6" h="0">fire-thin</trail_generator>
  <fabricator>Aerosys</fabricator>
- <description>The star product of Aerosys, the Hyena quickly has also become a favourite of guerrilla and paramilitary groups far and wide. Their top-notch thrust combined with fast reaction time allow them exhaust enemy energy supplies while wearing them down. While Aerosys claim that their ships are only sold to reputable pilots, it's suspected that they sell to anyone willing to pay, as it seems there are countless Hyenas among the ranks of the more notorious and unruly pilots.</description>
+ <description>The star product of Aerosys, the Hyena quickly has also become a favourite of guerrilla and paramilitary groups far and wide. Their top-notch thrust combined with fast reaction time allow them to exhaust enemy energy supplies while wearing them down. While Aerosys claim that their ships are only sold to reputable pilots, it's suspected that they sell to anyone willing to pay, as it seems there are countless Hyenas amoung the ranks of the more notorious and unruly pilots.</description>
  <health>
   <armour>25</armour>
   <armour_regen>0</armour_regen>
--- a/dat/ships/hyena.xml
+++ b/dat/ships/hyena.xml
@@ -13,7 +13,7 @@
  <trail_generator x="-15" y="-6" h="0">fire-thin</trail_generator>
  <trail_generator x="-15" y="6" h="0">fire-thin</trail_generator>
  <fabricator>Aerosys</fabricator>
- <description>The star product of Aerosys, the Hyena quickly has also become a favourite of guerrilla and paramilitary groups far and wide. Their top-notch thrust combined with fast reaction time allow them to exhaust enemy energy supplies while wearing them down. While Aerosys claim that their ships are only sold to reputable pilots, it's suspected that they sell to anyone willing to pay, as it seems there are countless Hyenas amoung the ranks of the more notorious and unruly pilots.</description>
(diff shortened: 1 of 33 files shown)


dat/ships/hyena.xml
```diff
@@ -13,7 +13,7 @@
  <trail_generator x="-15" y="-6" h="0">fire-thin</trail_generator>
  <trail_generator x="-15" y="6" h="0">fire-thin</trail_generator>
  <fabricator>Aerosys</fabricator>
- <description>The star product of Aerosys, the Hyena quickly has also become a favourite of guerrilla and paramilitary groups far and wide. Their top-notch thrust combined with fast reaction time allow them exhaust enemy energy supplies while wearing them down. While Aerosys claim that their ships are only sold to reputable pilots, it's suspected that they sell to anyone willing to pay, as it seems there are countless Hyenas among the ranks of the more notorious and unruly pilots.</description>
+ <description>The star product of Aerosys, the Hyena quickly has also become a favourite of guerrilla and paramilitary groups far and wide. Their top-notch thrust combined with fast reaction time allow them to exhaust enemy energy supplies while wearing them down. While Aerosys claim that their ships are only sold to reputable pilots, it's suspected that they sell to anyone willing to pay, as it seems there are countless Hyenas amoung the ranks of the more notorious and unruly pilots.</description>
```
bobbens, line 16: Is this `amoung` correct? I don't thing the spelling `amoung` is used at all these days, even in British English.

dat/ships/mule.xml
```diff
@@ -14,7 +14,7 @@
  <trail_generator x="-33" y="0" h="0">fire</trail_generator>
  <fabricator>Melendez Corp.</fabricator>
  <license>Large Civilian Vessel</license>
- <description>A heavy freighter specialized in transporting cargo throughout the galaxy. Also has sufficient accommodations for passengers depending on the model, allowing for enjoyable cruises. The Mule is a favourite target of pirates, so avoiding dangerous systems is strongly advised.</description>
+ <description>A heavy freighter specialised in transporting cargo throughout the galaxy. Also has sufficient accommodation for passengers, depending on the model, allowing for enjoyable cruises. The Mule is a favourite target of pirates, so avoiding dangerous systems is strongly advised.</description>
```
bobbens, line 17: Although we do use British English, I think we pretty much stick to [Oxford Spelling](https://en.wikipedia.org/wiki/Oxford_spelling) which means using endings in `-ize` instead of `-ise`.

dat/ships/pirate_admonisher.xml
```diff
@@ -16,7 +16,7 @@
  <trail_generator x="-20" y="5" h="-5" always_under="1">pirate</trail_generator>
  <trail_generator x="-20" y="-5" h="-5" always_under="1">pirate</trail_generator>
  <fabricator>Skull and Bones</fabricator>
- <description>The stock Admonisher is a powerful vessel, especially for its size. Through the usual Skull and Bones refit the Pirate Admonisher features improved manoeuvrability, more weapons space, and slightly improved defenses, at a cost of cargo capacity and armour regeneration rates.</description>
+ <description>The stock Admonisher is a powerful vessel, especially for its size. Through the usual Skull and Bones refit the Pirate Admonisher features improved manoeuvrability, more weapons space and slightly improved defences, at a cost of cargo capacity and armour regeneration rates.</description>
```
bobbens, line 19: Grepping the code, it looks like we use `defense` all over instead of `defence`. However, it does seem like we should be using `defence` all over. (this is not a change request, just an idle thought)

dat/ships/hyena.xml
PhoenixRiver: I stand corrected... apologies, I've been under the impression it was "amoung" and "amongst" - it seems that even the, largely hidden, British dictionaries are against me on this! (OED can go jump however, contrarian scholars that they are!)

dat/ships/mule.xml
PhoenixRiver: I'll remember to check for it that way in future... I have a niggling that I might have seen one or two in the ship descriptions... *sigh* No time now, maybe later! :)
I had noticed the use of British English spelling but I'm a little dubious about the punctuation... I only "corrected" it if there were other changes I definitely wanted to make!

dat/ships/pirate_admonisher.xml
PhoenixRiver: *heh* I'll make a little note in case I find time :) - though I'm note sure I have a way to find them easily not having grep here on Windows... gah, I need to get a Linux box again!

Should this be in the code as well as in the text is the big question... I guess for consistancy...

bobbens:
Looks good, thanks!

PR #2070, opened 2022-02-18
Added three new ship variants

PhoenixRiver:
Added the Alpaca, Guanaco and Vicuna ... little acknowledged but common variants of the Llama.

The intention is to add Llama mk IIs, mk IIIs and mk IVs with the same spite (at least for now) as varients of the Llama to add a little varienty into this venerable ship that must have, similar to the AK, spawned many varients that are all but the same but with their own tweeks depending on the reagion of space you find them... I figureed the Dvaered could be a little tougher, the Sirian have a little better electronics and the Frontier have a little more protection and leasve the original Llamas as the predominent in the Empire.

None of the above would replace each other in each space, they would just split the frequency between them... thoughts?

Do we already have a way to do this more simply than adding new ships?

P.S. The three files here are currently just identical copies of the Llama and are here to create the pull request as much as anything! :)

### DIFF
--- a/dat/ships/alpaca.xml
+++ b/dat/ships/alpaca.xml
@@ -1,0 +1,53 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<ship name="Llama">
+ <points>20</points>
+ <base_type>Llama</base_type>
+ <GFX size="47">llama</GFX>
+ <GUI>brushed</GUI>
+ <sound>engine</sound>
+ <class>Yacht</class>
+ <price>120000</price>
+ <time_mod>1</time_mod>
+ <trail_generator x="-12" y="-16" h="-2">nebula</trail_generator>
+ <trail_generator x="-12" y="16" h="-2">nebula</trail_generator>
+ <trail_generator x="-12" y="-6" h="0">fire-thin</trail_generator>
+ <trail_generator x="-12" y="0" h="0">fire-thin</trail_generator>
+ <trail_generator x="-12" y="6" h="0">fire-thin</trail_generator>
+ <fabricator>Melendez Corp.</fabricator>
+ <description>One of the most widely used ships in the galaxy. Renowned for its stability and stubbornness. The design hasn't been modified much since its creation many, many cycles ago. It was one of the first civilian use spacecrafts, first used by aristocracy and now used by everyone who cannot afford better.</description>
+ <characteristics>
+  <crew>2</crew>
+  <mass>80</mass>
+  <fuel_consumption>100</fuel_consumption>
+  <cargo>15</cargo>
+ </characteristics>
+ <health>
+  <armour>25</armour>
+  <armour_regen>0</armour_regen>
+ </health>
+ <slots>
+  <weapon size="small" x="7" y="0" h="1" />
+  <weapon size="small" x="-3" y="0" h="2" />
+  <utility size="small" prop="systems">Unicorp PT-16 Core System</utility>
+  <utility size="small" prop="accessory" />
+  <utility size="small" />
+  <utility size="small" />
+  <structure size="small" prop="engines">Nexus Dart 150 Engine</structure>
+  <structure size="small" prop="hull">Unicorp D-2 Light Plating</structure>
+  <structure size="small" />
+  <structure size="small" />
+ </slots>
+ <stats>
+  <turn_mod>-20</turn_mod>
+  <speed_mod>-10</speed_mod>
+  <turn_mod>-10</turn_mod>
+  <cargo_mod>20</cargo_mod>
+  <armour_mod>10</armour_mod>
+  <cargo_inertia>-20</cargo_inertia>
+  <ew_hide>-10</ew_hide>
+ </stats>
+ <tags>
+  <tag>standard</tag>
+  <tag>transport</tag>
+ </tags>
+</ship>
--- a/dat/ships/guanaco.xml
+++ b/dat/ships/guanaco.xml
@@ -1,0 +1,53 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<ship name="Llama">
+ <points>20</points>
+ <base_type>Llama</base_type>
+ <GFX size="47">llama</GFX>
+ <GUI>brushed</GUI>
+ <sound>engine</sound>
+ <class>Yacht</class>
+ <price>120000</price>
+ <time_mod>1</time_mod>
+ <trail_generator x="-12" y="-16" h="-2">nebula</trail_generator>
+ <trail_generator x="-12" y="16" h="-2">nebula</trail_generator>
+ <trail_generator x="-12" y="-6" h="0">fire-thin</trail_generator>
+ <trail_generator x="-12" y="0" h="0">fire-thin</trail_generator>
+ <trail_generator x="-12" y="6" h="0">fire-thin</trail_generator>
+ <fabricator>Melendez Corp.</fabricator>
+ <description>One of the most widely used ships in the galaxy. Renowned for its stability and stubbornness. The design hasn't been modified much since its creation many, many cycles ago. It was one of the first civilian use spacecrafts, first used by aristocracy and now used by everyone who cannot afford better.</description>
+ <characteristics>
+  <crew>2</crew>
+  <mass>80</mass>
+  <fuel_consumption>100</fuel_consumption>
+  <cargo>15</cargo>
+ </characteristics>
+ <health>
+  <armour>25</armour>
+  <armour_regen>0</armour_regen>
+ </health>
+ <slots>
+  <weapon size="small" x="7" y="0" h="1" />
+  <weapon size="small" x="-3" y="0" h="2" />
+  <utility size="small" prop="systems">Unicorp PT-16 Core System</utility>
+  <utility size="small" prop="accessory" />
+  <utility size="small" />
+  <utility size="small" />
+  <structure size="small" prop="engines">Nexus Dart 150 Engine</structure>
+  <structure size="small" prop="hull">Unicorp D-2 Light Plating</structure>
+  <structure size="small" />
+  <structure size="small" />
+ </slots>
+ <stats>
+  <turn_mod>-20</turn_mod>
+  <speed_mod>-10</speed_mod>
+  <turn_mod>-10</turn_mod>
+  <cargo_mod>20</cargo_mod>
+  <armour_mod>10</armour_mod>
+  <cargo_inertia>-20</cargo_inertia>
+  <ew_hide>-10</ew_hide>
+ </stats>
+ <tags>
+  <tag>standard</tag>
+  <tag>transport</tag>
+ </tags>
+</ship>
--- a/dat/ships/vicuna.xml
+++ b/dat/ships/vicuna.xml
@@ -1,0 +1,53 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<ship name="Llama">
+ <points>20</points>
+ <base_type>Llama</base_type>
+ <GFX size="47">llama</GFX>
+ <GUI>brushed</GUI>
+ <sound>engine</sound>
+ <class>Yacht</class>
+ <price>120000</price>
+ <time_mod>1</time_mod>
+ <trail_generator x="-12" y="-16" h="-2">nebula</trail_generator>
+ <trail_generator x="-12" y="16" h="-2">nebula</trail_generator>
+ <trail_generator x="-12" y="-6" h="0">fire-thin</trail_generator>
+ <trail_generator x="-12" y="0" h="0">fire-thin</trail_generator>
+ <trail_generator x="-12" y="6" h="0">fire-thin</trail_generator>
+ <fabricator>Melendez Corp.</fabricator>
+ <description>One of the most widely used ships in the galaxy. Renowned for its stability and stubbornness. The design hasn't been modified much since its creation many, many cycles ago. It was one of the first civilian use spacecrafts, first used by aristocracy and now used by everyone who cannot afford better.</description>
+ <characteristics>
+  <crew>2</crew>
+  <mass>80</mass>
+  <fuel_consumption>100</fuel_consumption>
+  <cargo>15</cargo>
+ </characteristics>
+ <health>
+  <armour>25</armour>
+  <armour_regen>0</armour_regen>
+ </health>
+ <slots>
+  <weapon size="small" x="7" y="0" h="1" />
+  <weapon size="small" x="-3" y="0" h="2" />
+  <utility size="small" prop="systems">Unicorp PT-16 Core System</utility>
+  <utility size="small" prop="accessory" />
+  <utility size="small" />
+  <utility size="small" />
+  <structure size="small" prop="engines">Nexus Dart 150 Engine</structure>
+  <structure size="small" prop="hull">Unicorp D-2 Light Plating</structure>
+  <structure size="small" />
+  <structure size="small" />
+ </slots>
+ <stats>
+  <turn_mod>-20</turn_mod>
+  <speed_mod>-10</speed_mod>
+  <turn_mod>-10</turn_mod>
+  <cargo_mod>20</cargo_mod>
+  <armour_mod>10</armour_mod>
+  <cargo_inertia>-20</cargo_inertia>
+  <ew_hide>-10</ew_hide>
+ </stats>
+ <tags>
+  <tag>standard</tag>
+  <tag>transport</tag>
+ </tags>
+</ship>


bobbens:
I've been thinking about ship variants and how to implement them. Anyway, some thoughts below:
1. First, are they necessary? Does having a lot of very similar but slightly nuanced ships add more to the game or just make it more confusing?
2. In they case they are necessary, how should they be implemented? Allow support for templates so that ships can inherit all properties except those that get overwritten? This does seem like a more elegant solution than just duplicating files all over. However, it does get a bit tricky if modifying slots.
3. An alternative implementation would be to keep the ship the same, and implement the variants through intrinsic outfits. Would need support for changing base ship names or something like that though.
4. If adding variants, how far should they go? How many should be added and how should they be spread? An easy solution would be to just have 3 types like flawed, normal, and superb for all the ships and just have them give a small boost. NPC would have random chosen (or normal by default) and it could be indicated with [F] or [S] suffix on the name for flawed or superb. However, this opens other issues like how should the player be able to get them? I believe EVN uses this approach and relies on allowing the player to capture ships. It would pollute interface a bit since the player, most likely, would ignore non-superb ones past a certain point in the game.

LJDude:
This reminds me of my idea for different yacht class ships where you could choose which one you start with:
1. A speed oriented one
2. A cargo oriented one (the Llama)
3. A combat oriented one

Another idea I had for a yacht was a sort of racing yacht that wouldn't have any weapons and would be faster than even a gawain (i.e. over 650 speed when maxed out).

PhoenixRiver:
Ship variants was just a off the cuff idea today for me so apologies that I have not thought very deeply about it... also note I have a query... the Llama specifies the "turn_mod" twice in stats, and each time with a different number "-20" and "-10" respectively... I figure this is a bug but I don't know which is correct, which might be used or even if that can be known - many languages work top to bottom and would overwirte the previous value but I hate to assume!

>     1. First, are they necessary? Does having a lot of very similar but slightly nuanced ships add more to the game or just make it more confusing?

I would imagine that if ships are only marginally different and with largely the same stats that it makes little difference from a confusion stand point however a player might be interested in one varient more than another because it costs the same (pretty much) but might allow them to squeeze that setup they really want to try into a ship they can actually afford... this likely matters more to a beginner than a player with a great ship since the latter can trawl the galaxy for the right outfits, even use hard to find one previously hoarded whereas to a beginner the ability to pick a ship even slightly more suited to their play style is likely to make a disproportionate difference.

So, the Llama is possibly the best case study for this since it is the starting ship; maybe a choice of which one could be given even, maybe that could even dictate which part of the Galaxy the player starts in? (My varients are going to be tailored to Empire, Dvaered, Sirius and Frontier ... I assume that the Soromid would have something differnt or a bio-Llama and I don't know enough about bioships to mess with them.)

Secondly the Llama's description would describe a VW beetle or Kalashnikov equally well but those latter two have dozens of varients that fall almost entirely under the same umbrella but have different particular trade offs while adhearing to the basic interchangability of basic parts e.g. wheels, engines, stocks, magazines, barrels so I felt like the Llama should have company in the form of very marginally different ships, more for flavour than for anything else... I'd be happy to name them Llama MK*s and have their alternate names in the description only if that would help reduce confusion.

Thirdly small ships would benefit varients more than large ships due to their tight limitations on space and size and so adding varients lets them breath a little more as the bigger ships are able (to some degree, the use of cores does restrict larger ships to a greater degree than used to be the case before cores existed.

>     2. In they case they are necessary, how should they be implemented? Allow support for templates so that ships can inherit all properties except those that get overwritten? This does seem like a more elegant solution than just duplicating files all over. However, it does get a bit tricky if modifying slots.

I had envisioned varients as not modifying slots as Rule Zero. To be honest I did not intend that they modify the sprite (except maybe colouring?) either and only made minor changes to the base stats... essentially I am looking at what slots exist on ships and discounting them from what the ship is... so... no engine, no central computer, no weapons, no specialist electrical gear and no bolt on armour... which leaves a chassis, bodywork and basic wiring and electronics. Then I tried to limit myself to what an enthusiast could do to a ship in his warehouse and have the authorities not notice that this was a "different" ship, i.e. no lopping bits off and sticking them on somewhere else but a little panel beating, shaving a little off the superstructure and rewiring it with gold contacts. Crutially, for every advantage there must be an opposite and even (numbered) disadvantage from the stock model; i.e. better turning (more nimble) might mean worse armour (because it is lighter).

If only the stats were considered (though maybe also the cargo space and base armour also?) then it might be possible to have a system where _every_ ship was potentially a varient just by choosing a random number of fixed modifiers to apply to every ship upon generation. In this case I would suggest that the pro/con pairs be a fixed list to be picked from and that vairants only ever get to modify a stat once - this way they never have a chance of becoming weirdly one-sided.

This system would also mean the current crop of ships would be the basis for the stock models and no new data files would need to be created. Various faction specific varients might also continue to be stock ships but the above system could also prefer certain pro/con pairs for ships of a specific faction.

>     3. An alternative implementation would be to keep the ship the same, and implement the variants through intrinsic outfits. Would need support for changing base ship names or something like that though.

A naming system similar to that for the pirates might be usable and this system would allow for greater variety and predictability from the players point of view, but it also means another set of things to keep track of for the player. Though if these intrinsic outfits were only obtainable at highly specific and obvious places then it might not be an issue... I'd suggest all outfits being available every place that offers these services but I'm still wedded to my very minor variant idea which would mean that the places offering these outfits would be kind of "tune-up" garages for spacecraft enthusiasts... maybe the spaceship enthusiast can direct players to the local garage? They get around a bit after all! ; )

>     4. If adding variants, how far should they go? How many should be added and how should they be spread? An easy solution would be to just have 3 types like flawed, normal, and superb for all the ships and just have them give a small boost. NPC would have random chosen (or normal by default) and it could be indicated with [F] or [S] suffix on the name for flawed or superb. However, this opens other issues like how should the player be able to get them? I believe EVN uses this approach and relies on allowing the player to capture ships. It would pollute interface a bit since the player, most likely, would ignore non-superb ones past a certain point in the game.

I specifically didn't want to make ships that are better or worse, just differently postured making them slightly better and worse at differnt roles... my four Llama variants are about to be the stock one, one with better manouverability, one with better armour (and shields?) and one with better cargo or maybe evasion? I am yet to catch up on all the stat modifiers!

The point in adding this is, for me, multifaceted; to add a bit more flavour to ships and the "world", to do so in a way that might help the player just starting out, to help traders and also to grow the traffic in the space lanes, which I feel is a little sparse but would also just become too repetative with more of the exact same...

Having reached this point I realise that I had not really considered fighters, maybe weapons that are largely interchangable could have a tag to group them which might allow ships to have more varied weaponry without having to enumerate every possibility in a file? Again, factions could have preferences... e.g. The Razor and Lazer turrets have the same weight, EPS and (almost) DPS and are so pretty much interchangable but ships with them mounted look different in combat due to their sound, weapon colour and the range they have to close to as well as the different impact they have on different ships due to the specifics of Razor and Lazer weapons fire on shield, hull etc.

And now I want to add another Llama variant, one that simply costs more but is classified as a luxury yatch... maybe it also has less cargo space because it has a pool or something! ;)

PhoenixRiver:
Right, here are the varients actually done to give you an idea of what I was envisioning... the luxury yacht is there too for a laugh! ;) I'm happy to cut it as it is the most out there one.
I should also not I have not added any of these ships anywhere else that they might need to be added to be spawned in the right places... any pointers would be appreciated if these are to go ahead in this form...

PhoenixRiver:
Oh, and I varied their "trails" so you can tell them apart if you know them well enough :) - I don't think it is necessary to do this but since I could... ;)

PhoenixRiver:
Any comments?

bobbens:
So my thoughts on the topic:
1. I would not give them different names when they're very close to the same ship. An adjective would probably suffice.
2. If variants were to be done, I think we would need a more high level overview instead of just adding some for the Llama. What ships get variants? How many? How are they distributed? I know it significantly increases the scale of the PR, but at least that sort of stuff should be decided before we commit to adding variants all over.
3. Implementation-wise, I really dislike having many copied files all over. I can look at it later, but I think it would be important to define some sort of base or standard template, and then define all the variants relative to that without duplicating code. Another option would be to use a python script, like how we do for bioships, and automatically generate all the variants using templates.

Overall my main concern is the high level plan for variants. What do they bring to gameplay? How do they make things more interesting? How to deal with GUI clutter? What ships get variants? How many do we need? Where are they sold or distributed? How are they acquired? etc...
Answering these questions should also naturally lead to how they should be implemented, like intrinsic slots vs ship definitions.

PhoenixRiver:
>     1. I would not give them different names when they're very close to the same ship. An adjective would probably suffice.

Fair point (I picked close relatives of the Llama for exactly this reason but I guess they are not well known enough and not so obviously connected and descriptive as an adjective could be :)

>     2. If variants were to be done, I think we would need a more high level overview instead of just adding some for the Llama. What ships get variants? How many? How are they distributed? I know it significantly increases the scale of the PR, but at least that sort of stuff should be decided before we commit to adding variants all over.

I absolutely agree that we need to do this with a view to how it would work for any ship, not just Llamas (I picked them for the reasons I mentioned above - i.e. easy and obvious starting point and as a demonstration of concept). I would suggest that every ship should be able to get variations in priciple but wether they do depends on if it makes any sensein each particular case.
How many and how they are distributed depend on what we want them for and that is maybe not yet sufficiently nailed down to allow further comment in that direction. so we definitely agree than a plan is necessary before any enlargement of the work done (probably in the wrong direction!)

>     3. Implementation-wise, I really dislike having many copied files all over. I can look at it later, but I think it would be important to define some sort of base or standard template, and then define all the variants relative to that without duplicating code. Another option would be to use a python script, like how we do for bioships, and automatically generate all the variants using templates.

I also really don't want to have to do the file route... I previously was dealing with the lack of asteroids by writing a script to produce asset files for each asteroid in my belt! So I understand the problem but also the solution that a script can provide (though it can still be messy) as with the bioships but I wonder if it might be possible to do this more elegantly depending on what exactly we want from varients... back to this problem.

> Overall my main concern is the high level plan for variants. What do they bring to gameplay? How do they make things more interesting? How to deal with GUI clutter? What ships get variants? How many do we need? Where are they sold or distributed? How are they acquired? etc... Answering these questions should also naturally lead to how they should be implemented, like intrinsic slots vs ship definitions.

I feel like, baring the first two questions, these points are a restatement of the above or much closer in questions we can't answer yet. So ... I'm going to try anyway to hopefully better explain my own rational (which I'm sure differes from that of others) as I'm feeling I've not done a good job so far!

GUI clutter - rename tabs that say "Ships" to "Hulls" and add a new tab called "Mods" or "Varients", this is a split page similar to the Equipment page with ships you own at the top and "Modifications" at the bottom, each time you choose a ship the Mods you can buy for that Hull from this planet/station appear (comments at the end ;) - hovering a mod will give you a tooltip with its exact modifiers and selecting it will show its effect on the ship stats in the top right of the page (also similar to the equipment page), clicking "Buy" will apply the mod to the Hull, clicking "Sell" will remove it.
One Mod per Hull, Mods can not be moved between Hulls, it is a part of the Hull, sell the Hull with the Mod applied on the Hulls page and you sell the Mod with it (Hull price = original Hull price + Mod price) even if this staion/planet doesn't sell that Mod or even Mods at all. The Hull can potentially be rebought until you take off but I'm not decided on that, it would be more consistant that way.
Hull varients are named with a letter or letters after the name to indicate which Mod e.g. "Llama S" - fast Mod applied, "Llama SX" - fast, maouverable Mod applied, "Llama DV" - Dvaered (armour) Mod applied, "Llama God" - Goddard (electronics) Mod applied etc.

Acquired from a few planets/stations that have not only Shipyards that can accomodate them but also skills to alter prepurchased Hulls i.e. the manufacturers of that Hull or Military shipyards or other specialist yards only working on a limited number of easier Hulls e.g. Llama, Koala, Mule - yes, these are all traders but combat Hulls are harder to modify without making them worse cf modifying a tank vs a truck - or on major changes to a Hulls original competance e.g. making a speed Hull a cargo Hull, making a carrier Hull a cargo Hull and other things the original designer would blanch at. (Maybe these could include slot changes? That's implementation though I think, if it's possible it might be nice but if not, no biggie)
We could implement buying Varient Hulls direct from the Hulls page but I wouldn't expect the variety and only faction specific ones from same faction shipyards... they would also be more costly of course.

I would suggest that each Major House and some other shipbuilders would be creating mods for those ships they would most likely use, as they already have as seperate files!
First those who have no Mods; The Empire, I feel, would use the originals, Soromid already have bioships so nothing from them(?) and FLF don't have the resources? Or maybe only for their most used ship... oh, and Collective is machine isn't it so its ships are perfectly fit for purpose so why would they ever Mod? : D
And the ones that do: Dvaered Mods for heavy / medium combat hulls, Sirius Mods for medium ships, Frontier for light ships, Za'lek for carriers and drones/fighters, Goddard general electronics Mod(s), Pirates general stealth Mod(s), general Mods for each of speed, manouver, armour, cargo, maybe one or two others and maybe an overhaul Mod or two to change a Hull's primary competance e.g. Carrier to Cargo, Destroyer to Explorer, Cargo to Combat - these would be a future potential extension of the system, might be a no go and would, anyway, likely not produce anything except middling results but allow a player to choose a Hull more on styling than ability as long as they weren't also expecting performance! :)
So all told a Heavy Combat Carrier might have as many as a Dvaered, Za'lek, Goddard, four general Mods and an overhaul Mod to convert it to cargo and another for making it a colony / exploration ship,
a Medium Combat ship a Dvaered, Sirius, Goddard plus the four general and overhaul mods to make it cargo or explorer or maybe a small carrier(?)
a Gawain might have a Frontier, Goddard, Pirate and speed and manouver general Mods (armour and cargo on a Gawain!?!) and an overhaul for exploration (lighter so lower fuel use(?))
a Shark might have the same as a Gawain but adding in a Za'lek Mod?

In the same way that a nearly good build could be made more viable with a Mod by the equipopt (I think that is what it is called?) the player can more closely ally themselves with their chosen faction by using their Mod as well as tune up their ship in precisely the direction they want to use it for the way they use it though I would expect for the player this is more a mid to late game mechanic except for maybe messing with it at the start on the Llama to get that early game edge to make the money for the proper ships! ;)
For the game it is introducing more variety between factions, making their space look more different (particularly if the House Mods also affect the ship trails consistantly e.g. Dvaered have bigger, thicker trails and more obvious engines) and so both looking less cookie cutter and identifying to the player when they move between different Houses' space or even showing more clearly that a system is shared space... potentially very useful if a player's faction standings are a little "inconsistant" ;) than they are for a new player.
It could also be that rather than using a fake transponder, or using it in addition, factions would be less likely to scan ships with their Mod applied, think of it as a basic IFF - probably less use in safe lanes but outside of them maybe they would protect, to some degree, scans - thereby acting as another degree of stealth for smuggling or backstabbing.

I readily admit that not all of these things can be done without work also being done on other parts of the game e.g. altering scanning to take account of Mods, maybe needing to add more engine trails to allow the necessary variety etc. but logs are currently in the same position, they are used but there is a lot of content that takes no account of them, ditto that the ship has an AI now, it is almost entirely absent from everything but the tutorial! So it will take time to fully take account of it within the game, just like any other potentially deep system.

So to implementation... the two things that come to mind are a file based solution (like outfits), one per Mod with each Mod containing a definition of it's effect and a list of ships that exact definition can be applied to, or, my preference at present, a file based solution (like outfits), one per Mod containing a definition of it's effect alongside changing each Hull definition to contain a list of Mods that can be applied to it... this would, I hope also allow us to remove all the faction specific Hulls we currently have in the ships directory where a original / generic Hull already exists while also making it easier to make future Hulls faction specific more consistantly and quickly i.e. in the definition of the Hull itself.
It might well also make choosing which ships to spawn easier since it could be limited to those with the potential to Mod to that faction as well as a category of "general" ships that everyone uses e.g. traders.
As for applying the changes I would guess that overriding, adding and multiplying the original stats would be position as well as adding new stats to the Hull (with negation also being included though by implication of being overridden to zero / default.


I think I have now covered all the points asked but I am sure that I have missed something and I am pretty sure that I don't have the skills right now to impliment this all in one go, certainly not the GUI and interaction with the AI equipment chooser, possibly not even the process of applying them but I can definitely do the Mod file definitions and Hull file lists... assuming this is anything like what might be chosen for implementation.

How does this proposal sound? Too long, I'd guess, and a bit rambling too. Apologies. However, comment away please :)

PhoenixRiver:
I'm starting to think this sounds more like an outfit that you can only have one of... the thing is that they couldn't be just added to any ship willy nilly unless they were really watered down from what they could be (certainly slots would be out)... it would also mean that hanging things like faction paint jobs or engine trails or faction stealth off them would be rather more of a challenge and dilute faction differentiation if they could be applied to ships that a faction wouldn't use... on top of that outfits are easy as chips to get ahold of for the most part but selling unusual ones is a pain in the neck so I guess that's why Mods are not Outfits?? Hmmm... will keep thinking on this.

bobbens:
Ship variants make sense in EV where you don't really have limits on outfits like in Naev. I think maybe the easiest way to implement would be with intrinsic outfits, which could allow the player to change their variant via an NPC or something, and potentially even change the visuals of the ship.

PhoenixRiver:
I think that that is the conclusion I am also coming to... I don't know how different skins for the ships would work but I guess we're on the right track now?

ThrosturX:
I am very interested in the idea of variants in this game and would like to express my agreement with the intrinsic outfit implementation. It is the most dynamic of the suggestions we have and removes the need to duplicate code (in this case, mostly data files and sprites and things). I also think that this will make writing generic yet interesting missions more accessible as missions could randomly select variants to create more diversity between instances.

With regards to the different skins, this is an easy future fix with shaders or by plugging minor modifications to the intrinsic outfits themselves. Shaders would be great and would allow both generic (code-based) and custom (image-based) customization but we could start out with something simple similar to the engine trail modification outlined quite near the top of this PR discussion. Maybe it makes sense that a variant that changes speed or turn has it's engines randomly displaced by a small margin?

bobbens:
> I am very interested in the idea of variants in this game and would like to express my agreement with the intrinsic outfit implementation. It is the most dynamic of the suggestions we have and removes the need to duplicate code (in this case, mostly data files and sprites and things). I also think that this will make writing generic yet interesting missions more accessible as missions could randomly select variants to create more diversity between instances.

Easiest implementation would be just to add a structural slot to all ships that could change things a bit, just like we have for the accessories. However, this slot could be made to be locked (the player can't directly modify it), and already have an outfit in it. That way it could only be changed through NPC / events or the likes. NPC could spawn with different variants. I do think this approach could be quite interesting and worth pursuing.

> With regards to the different skins, this is an easy future fix with shaders or by plugging minor modifications to the intrinsic outfits themselves. Shaders would be great and would allow both generic (code-based) and custom (image-based) customization but we could start out with something simple similar to the engine trail modification outlined quite near the top of this PR discussion. Maybe it makes sense that a variant that changes speed or turn has it's engines randomly displaced by a small margin?

Shaders are already possible with the new effect system (you can already see them in a certain cutscene in nightly for pure ship visuals). It should also be relatively easy to allow directly overriding the ship graphic via outfit, so that we could have different models rendered and changing the variant would change visuals slightly. We would need more models for that though.

PR #2078, opened 2022-02-19
Outfits proofread

PhoenixRiver:
A lot of the older outfits have been proofread, along with some of the rarer ones that seem to have more issues.
The weakness harmonizer ai file should probably be moved into the same directory as the other combat ais but I don't know if that can be safely done justby moving the file so I7ll not touch it, apologies.

### DIFF
--- a/dat/outfits/accessory/air_freshener.xml
+++ b/dat/outfits/accessory/air_freshener.xml
@@ -8,7 +8,7 @@
   <priority>3</priority>
   <mass>0</mass>
   <price>1000000</price>
-  <description>Apparently air fresheners have been traditionally used in human vehicles as an ornament for ages. They used to have scents, but since they have no effect with modern ships powerful filtration system, these days they are for cosmetic purposes only.</description>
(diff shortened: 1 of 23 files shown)


bobbens:
Looks good, thanks!

bobbens:
> A lot of the older outfits have been proofread, along with some of the rarer ones that seem to have more issues. The weakness harmonizer ai file should probably be moved into the same directory as the other combat ais but I don't know if that can be safely done justby moving the file so I7ll not touch it, apologies.

Strictly speaking the weakness harmonizer AI is not a combat AI (it is not exclusive with the other ones). Might make more sense to change the outfit name to make it less confusing though.

PhoenixRiver:
I had not realised it was not exclusive, or, indeed, that they were, having not had opportunity to use them and not having read their code (only descriptions!) - I'll get there eventually - but, while I agree the name is confusing, I'm at a loss as to how it might be improved unless more of the same category were created.

PR #2082, opened 2022-02-20
Added logging and shipai cameos to derelict events and missions

PhoenixRiver:
As far as I can tell I didn't fluff any of the coding so I hope that the adding of logs and shipai cameos to the derelict events and missions was okay... I also redirected the blackcat and spacefamily mission logs to be ones related to derelicts given that that's how they start. I expect this is not the best solution for neutral events but it makes it clearer for other people to add new ones to the bottom.

### DIFF
--- a/dat/events/neutral/derelict.lua
+++ b/dat/events/neutral/derelict.lua
@@ -91,7 +91,7 @@ function create ()
    hook.land("destroyevent")
 end
 
-local function derelict_msg( title, text )
+local function derelict_msg( title, text, log )
    vntk.msg( title, text, {
       pre = function ()
          vn.music( der.sfx.ambient )
@@ -101,6 +101,8 @@ local function derelict_msg( title, text )
          vn.sfx( der.sfx.unboard )
       end,
    } )
+
+   der.addMiscLog( log )
 end
 
 local function islucky ()
@@ -123,7 +125,7 @@ function board()
    local pp = player.pilot()
    local stats = pp:stats()
    if stats.fuel < stats.fuel_consumption and rnd.rnd() < 0.8 then
-      derelict_msg(_("Lucky find!"), _([[The derelict appears deserted, with most everything of value long gone. As explore the ship you suddenly pick up a back-up fuel tank hidden in the walls. The fuel is in good state and you siphon it off to fill your ships fuel tanks. Talk about good timing.]]) )
+      derelict_msg(_("Lucky find!"), _([[The derelict appears deserted, with almost everything of value long gone. As you explore the ship you suddenly pick up a back-up fuel tank hidden in the walls. The fuel is in a good state and you siphon it off to fill your ship's fuel tanks. Talk about good timing.]]), fmt.f(_([[Just as you were running out of fuel you found a derelict with some to spare in {sys}, what good fortune!]]), {sys=system.cur()}))
       pp:setFuel( true )
       destroyevent()
       return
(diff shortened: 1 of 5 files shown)


PhoenixRiver:
Umm...

bobbens:
One thing, it may be a good thing to make it clear that `{shipai}` is the shipai, because I think lots of players will give it random or bad names and be confused about that. We don't have any checks on the name, so it might be worth writing something like `"your ship AI {shipai}..."` instead of just `{shipai}`.

PhoenixRiver:
@UncombedCoconut Oooooohhh.... THANK YOU 👍  :)

Okay, I'll make it clearer that it is the AI talking though I'm hoping thi will become less necessary in the future as the AI is a more frequent NPC... though I guess we should probably put in a check at some point to check if the player is happy with that.

PhoenixRiver:
Fixed a missing comma (doh!), two instances of now unused variables and one missing variable declaration missing from the head of the file and finally a weirdness where 'pnt, sys=spob.cur()' does not work within fmt.f()
i.e.
`fmt.f(_("text{pnt}{sys}"), {pnt, sys=spob.cur()})`

The above code returns:
`accessing undefined variable 'pnt'`

Is this because it is inside the function call?
because it is in the table? Ooohhh... that's why, there are now two elements in the table pnt and sys=spob.cur() - Doh!
I was expecting the two elements to be pnt=spob.cur()[1] and sys=spob.cur()[2] but for that the code would have to be:
`pnt, sys=spob.cur()
fmt.f(_("text{pnt}{sys}"), {pnt, sys})`

Is this right?

PhoenixRiver:
Right... brackets that didn't get caught last time fixed and clarification added to the {shipai} additions... hopefully they won't date too badly once the Ship AI is more integrated... will do a different pull request once I find the time to query the player as to whether they _want_ the Ship's AI to be a recurring character or not.

UncombedCoconut:
As an independent statement, "pnt, sys=spob.cur()" is valid Lua code (assignment to two variables), but there's no equivalent syntax for the stuff inside table expressions.

You probably want something like

local pnt, sys=spob.cur()
fmt.f(_("text{pnt}{sys}"), {pnt=pnt, sys=sys})

PhoenixRiver:
> As an independent statement, "pnt, sys=spob.cur()" is valid Lua code (assignment to two variables), but there's no equivalent syntax for the stuff inside table expressions.

Thank you for the confirmation... it took a while for my brain to click but it got there eventually! : P

> You probably want something like
> 
> `local pnt, sys=spob.cur()`
> `fmt.f(_("text{pnt}{sys}"), {pnt=pnt, sys=sys})`

I ended up doing

`fmt.f(_("text{pnt}{sys}"), {pnt=spob.cur(), sys=system.cur()})`

because it was used in other places and I wrote the fix before I was sure what I guessed at would work... it wouldn't have because I missed the "local" and it didn't matter anyway since I had a bunch of other errors that didn't get picked up in the same pass, so these are now getting dealt with cf the latest specifically relates to a missing "local" and a missing "include" equivalent... apparently today's multitasking brain overload is making me revert to my original habits in a globals-only where variables / functions are defined at time of use language... BASIC was a good start but it took me too long to transition to something more formal (though, to be fair, I didn't have access to a machine with a compiler for anything else for that same length of time! Borland C was my second and C let's you get away with so much! Then again maybe it could have been worse, I learned FORTRAN later... 
`1 = 0`
`PRINT 1+1`
output: 0
: D  : D  : D
I swear it's why == was developed so early on! : )

PhoenixRiver:
*YAY* Passed :)

PR #2108, opened 2022-03-02
Fix two missing translation marks in the minerva missions.

oo13:
Fix two missing translation marks in the minerva missions.


bobbens:
Looks good, thanks!

PR #2088, opened 2022-02-21
Two trader clan home systems

PhoenixRiver:
Two trader home systems; Holly and Ivy.
Holly is an extremely ore rich system with a trinary planet system and an asteroid ring outside that, rich in mining prospects. Now the home to a Trader Clan vertically integrated in Ore and Industrial Machinery trading - heavily robotisised
Ivy is a system with a number of terrestrial worlds that were, for a short time, extremely productive in agriculture... until it was discovered, too late, that one of the local flora - locally known as demon kudzu - would overrun everything that the colonists built within a single year. Unfortunately the demon plant had already transplanted itself to every world in the system (at least it is assumed the ice world suffered the same fate) and the system was quarantined until The Incident inspired some (very hungry) traders to find a way to eat the demon kudzu. These days, with very careful quarantine measures, food, water and medicines are traded very successfully - especially given the rock bottom prices the clan charges, especially its endless supply of low quality vegetable replacement packs made from kudzu.

The system files are currently done but only one spob has been started. These systems are currently imagined to go below Goddard and to create a buffer between the Empire and Dvaered.

PR #2110, opened 2022-03-04
A couple of, I think,  improvements to the anxious merchant.

PhoenixRiver:
If there is anything else you know has changed in coding style (etc.) or that you think needs improved, please comment.
Related: Would it not be worth moving (some of) the cargo missions into the trader mission directory?

### DIFF
--- a/dat/missions/trader/anxiousmerchant.lua
+++ b/dat/missions/trader/anxiousmerchant.lua
@@ -68,9 +68,9 @@ function create()
 end
 
 function accept()
-   if not tk.yesno(_("Spaceport Bar"), fmt.f(_([[As you sit down, the merchant looks up at you with a panicked expression, "Ahh! What do you want? Can't you see I've enough on my plate as it is?" You tell the merchant to calm down and offer a drink. "Jeez, that's nice of you... Ha, maybe I can get a break today!"
+   if not tk.yesno(_("Spaceport Bar"), fmt.f(_([[As you sit down, the merchant looks up at you with a panicked expression, "Ahh! What do you want? Can't you see I've enough on my plate as it is?" You tell the merchant to calm down and offer a drink. "Jeez, that's nice of you… Ha, maybe I can get a break today!"
     You grab a couple of drinks and hand one to the slightly more relaxed looking merchant as they start to talk. "So, I work for the Traders Guild. I transport stuff for them and they pay me. Only problem is, I kinda strained my engines running from pirates on the way to the pick-up and now I'm realising that my engines just don't have the speed to get me back to beat the deadline. And to top it all off, I'm late on my bills as is; I can't afford new engines now! It's like I'm in the Sol nebula without a shield generator."
-    You attempt to reassure the merchant by telling them that, surely, the company will cut them some slack. "Like hell they will! I've already been scolded by management for this exact same thing before! If I don't get this shipment of {tonnes} of {cargo} to {pnt}... I really need this job, you know? I don't know what to do..." The merchant pauses. "Unless... Say, you wouldn't be able to help me out here, would you? I'd just need you to take the cargo to {pnt} in the {sys} system. Could you? I'll give you the payment for the mission if you do it; it means a lot!"]]), {tonnes=fmt.tonnes(mem.cargo_size), cargo=_(mem.cargo), pnt=mem.dest_planet, sys=mem.dest_sys})) then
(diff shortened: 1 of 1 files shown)


dat/missions/trader/anxiousmerchant.lua
```diff
@@ -130,3 +131,8 @@ function tick()
    end
    misn.osdCreate(_("Help the Merchant"), osd_msg)
 end
+
+function abort ()
```
bobbens, line 135: I would not add a faction hit penalty for aborting the mission. Currently, the Trader Guild is also somewhat removed from the game (it is set as unknown to the player) pending a complete revamp and rethinking of them.

dat/missions/trader/anxiousmerchant.lua
```diff
@@ -111,8 +111,9 @@ function land()
...
+    You wait to one side as the cargo is hauled off your ship at breakneck speed and wonder if you should have just dumped the stuff in space. Just as the last of the cargo is taken off your ship, the inspector, who has clearly cooled off a bit, comes up to you and says "Look, I know you were trying to do us a favour but next time don't bother if you can't make it on time. I'm glad you didn't just dump it all into space like some people have done, but I can't pay you for this." He shakes his head and walks away. "That pilot is so fired…"]]))
       end
       misn.cargoRm(mem.cargo_ID)
```
bobbens, line 118: No need to remove the cargo as all mission cargos get removed when the associated mission "finishes" with `misn.finish()`.

bobbens (commented):
It looks OK, thanks! Added some small comments, I guess the main one is that we need to start rethinking about the trader's guild and its role in Naev (related to #2088). I do think they should be given a better name for starters and have some more well-defined lore.

dat/missions/trader/anxiousmerchant.lua
PhoenixRiver: Fair enough, I'll fix that... is it something that would be reasonable after or as part of the revamp?

I just feel like aborting the mission kinda dumps the merchant in the doodoo from an in-universe perspective and that this is made clear before accepting the mission... it's also a relatively small change (even if it is greater than the late arrival one - on purpose!)

dat/missions/trader/anxiousmerchant.lua
PhoenixRiver: Yeah, that's a left over... will be fixed in a moment :)

Can I ask if we have a mission template or is there one I can use as such so I can learn what the basic format of a mission should be, at least it would mean I'm a little more sure of of things like this and the "misn.finish(false) -> return" thing, though I have to admit I'm not entirely clear on that.... my current working hypothesis is that return should be used when you want to allow the player to talk to the mission giver again but misn.finish(false) removes this instance of the mission (which for the cargo missions I felt was just the way it worked rather than necessarily an issue - though it was a little annoying for cargo missions with a time limit since a ship switch might be possible but then the missions disappeared. So I guess I'm on the fence in that respect! :)

PhoenixRiver:
Will further comment on the other things you said in #2088 seeing as this small fix might be ready to go live...
I did find that the cargo is theoretically capable of being "nil" hence the extra edit here... it would only happen if spob.cur and destination.planet didn't have any commodities that overlapped though... I don't know how possible that is now but just in case it changes (again :P) in the future ;)

dat/missions/trader/anxiousmerchant.lua
bobbens: It would be reasonable, but since the player expects most missions to be abortable without consequence I think it would need something like:
1. Maybe a yes/no prompt to confirm with a message indicating faction hit
2. Probably not have the standing go below 0

dat/missions/trader/anxiousmerchant.lua
bobbens: There is a template at https://github.com/naev/naev/blob/main/docs/missions/mission_template.lua . However, looking at the code of any recent mission should also work.

> my current working hypothesis is that return should be used when you want to allow the player to talk to the mission giver again but misn.finish(false) removes this instance of the mission 

Yes, that's the case. However, in general we want missions to have the following properties:
1. Allow the player to abort or fail without consequence (mission can be retaken again)
2. Don't make the mission go away if the player doesn't accept or have enough cargo

bobbens:
> Will further comment on the other things you said in #2088 seeing as this small fix might be ready to go live... I did find that the cargo is theoretically capable of being "nil" hence the extra edit here... it would only happen if spob.cur and destination.planet didn't have any commodities that overlapped though... I don't know how possible that is now but just in case it changes (again :P) in the future ;)

Cool, thanks! It may be possible in the future to not have overlapping commodities, but I don't think that's possible at the moment.

dat/missions/trader/anxiousmerchant.lua
PhoenixRiver: Both good points I'll keep in mind :)

dat/missions/trader/anxiousmerchant.lua
PhoenixRiver: Understood :)

PR #2111, opened 2022-03-06
Fix an outdated path in the mission "srm_comingout5."

oo13:
This PR silences a warning message and lets us to play the mission "srm_comingout5."


bobbens:
Looks good, thanks!

PR #2113, opened 2022-03-08
spobs starting with Z

PhoenixRiver:
Corrected planet classes (mostly from A to something more accurate according to their descriptions or space image), proofreading, added a lot of tags (including a few new ones - new and old for worlds that have long or newly been settled, immigration as a new category for worlds with high immigration or currently being colonised and station for space stations - not something that is always obvious)


bobbens:
Looks good, but could you add the new tags and their description to `dat/spob/readme.md`? Thanks!

bobbens:
Looks good, thanks!

PR #2116, opened 2022-03-12
Made mission markers point to planets where appropriate - fixes #2115

### LINKED ISSUE #2115: Steam 0.9.2 - Commodity Delivery doesn't mark planet...
When doing a Commodity Delivery the destination planet is not shown in the overlay unlike Shipments, etc. (i.e. everything else) even though it is marked in the Mission OSD...

PhoenixRiver:
I was only able to check the first six pages of results for misn.markerAdd as git wasn't happy listing more than the sixth page?!?
Also heretic0.lua has a couple of comments - is contraband doable now? If so this mission is supposed to use it - also apparently the author would like to reduce Sirius reputation while increasing Nasin reputation... but the code isn't included for this (only something that probably need to be removed!) Should I go ahead and implement this or was it removed with good reason?

This pull request fixes #2115


bobbens:
Looks good, thanks!

PR #2117, opened 2022-03-12
Proofed spobs starting with W, X and Y

PhoenixRiver:
Tags, proofreading, planetary class / station class and exterior shots all jigged to better match the descriptions and make them better. Also added two more tags; refuel - for refueling stops since they have nothing else to recommend them despite being busy (for the location) but having near zero other use, wormhole - for wormholes.
I should note that I have used the station tag also for gas giant spaceports but the class is left to refer to the Gas Giant itself... this also happens for a couple of other spobs where a station is tied to its planet (including a pirate clan homeworld).
Oh, also changed a few of the populations that seemed a little odd - a major argiculture station with barely anyone onboard seems odd given the exponential need for support services once you have even a "small" group of specialists working in a place, for example :)

Also: Are there any special values for population? I was wondering if it might be more appropriate to list military spobs as having classified population (and maybe other things) but then again maybe it would be better to keep the spob.xml with the actual data and use the military tag to restrict this information unless you have a high enough standing with the owning faction?


bobbens:
Great work, thanks!

bobbens:
Oops, closed instead of letting it get marked as merged. Sorry

PhoenixRiver:
How does that make a difference?

bobbens:
> How does that make a difference?

AFAIK the only difference is it says "closed" instead of "merged".

Issue #2119, opened 2022-03-12
Bombers and missiles


nenau-again:
"you fly them [bombers] like a shitty corvette or a shitty fighter" -- supreme leader bobbens on 13 Jan 2022

Bombers appears to be noone's favourite class of ships. And this is more than understandable. I would BTW be really interested to read if someone finds them enjoyable to fly and why.

Bombers are not agile enough to dodge light ships attacks, and lack the firepower to obliterate them before being destroyed in a dogfight. This means that their role is rather to support their allies that do the actual dogfight. However, the relatively short range of missiles prevents them to be at a safe range from their targets.
The only actual usage of bombers is to counter heavy ships by using torpedoes, but even in this role, corvettes are usually better than bombers because they are nearly as manoeuvrable, usually have 3 medium slots and have a chance to survive attacks by interceptors.

Here is a list of proposals (some are mutually exclusive, others are not) that could give interest to bombers :

1- Change how seekers work
In this proposal, guided rockets would have a higher range (~5000), be less numerous, have lower fire and regeneration rate, lower agility but higher damage. By this way bombers could stay further away and be protected from direct dogfight. They would have more of a support role. Moreover, less agile and numerous projectiles would mean that targets will be able to dodge them if their attention is not cought elsewhere (needs fine tuning during the balancing phase). So, it would reduce the effeciency of kiting.
As a sidenote, close range seekers (~1500) could be added to a specific faction. They would fill the role of dogfight damage boosters, like Dvaered's Mace rockets.
PROS : Adds/reinforces the dimension of dodging rockets to the gameplay
CONS : Need to tune the AI to make it effecient in this new paradigm
OTHER : The reduced range difference between rockets and torpedoes makes turreted launchers a viable option for capships to counter bombers

2- Add a new core engine between fighter's and corvette's
And remove the Bomber's speed penalty. That would allow bombers to use heavier launchers. Maybe it would be necessary to lower a bit the manoeuvrability of corvettes and up to make room for this new engine type.
PROS : make sure that Bombers are actually more agile than corvettes
CONS : Hard balance-wise to predict if the new engine could be exploited some way or other

3- Change the bombers' characteristics (exclusive with 2)
[...]

Issue #2053, opened 2022-02-04
Quadtree-based Collisions


bobbens:
As we start thinking about adding more asteroids and more weapon types with more subparticles, improving collision detection gets more and more important. In particular, large amount of weapon particles with many asteroids leads to abysmal performance with the current brute-force implementation. A quadtree implementation has the potential to significantly speed this up. It would be most beneficial when it comes to large asteroid fields. Furthermore, it would allow us to also implement fast point defense weapons with weapon-weapon collisions. A potential base implementation is https://github.com/Antymon/quadtree .

bobbens:
Implemented in main.

PR #2121, opened 2022-03-16
Recreated point-defense turrets

PhoenixRiver:
In response to #2119 - I made the following to help out trader players - and maybe the AI once I work out where to fix that... oh, and I also fixed the description of the MK1 laser turrets ; )

I was guessing that the "resist" entry in the launchers was how destructible the projectile was and have thus added it to all the launchers which describe launching a physical object (i.e. I left the plasma launchers alone) making a number more of the missiles, rockets and grenades destructable, I find it odd they weren't already.
I also created two point defense turrets. I'll explain the reasoning behind their stats:
They have high CPU - tracking objects that small and fast is harder than targetting ships (and traders have few weapons and can spare the CPU cycles so these turrets are less use to combat ships which have armour and shields too).
They are cheap but make up for that in combat effectiveness... 1200 trackmax and 350 range. The amount of heat they produce is disproportionate to their damage energy and speed due to the computation and finely tuned wavelengths specifically aimed at taking out individual types of missiles, rockets and grenades and thus are heat inefficient but then again a trader isn't produce heat from their main guns (like as not) so it's not an issue for a trader (unlike a combat ship).
Their shots penetrate well and damage is higher than standard lasers due to the precision of their shots and specificity of their wavelengths matching their target.

So, these weapons are heat and CPU inefficient but highly capable close in to the ship, firing faster and hitting harder at targets that get close enough to be a danger... indicentally they might well be useful against fighters that do suicide runs or hover over the ship but such sitting ducks were always easy chaff! ; )

Edit - I forgot to add that I am perfectly happy for these to receive changes but I thought I'd start the ball rolling in what I think is the right direction... I'm happy for their range to reduce further as long as they have time to track missiles... I also forgot to mention the damage means they'll destroy small ordinence in one shot, medium ordinence in two to five shots and large ordinence probably never since they won't have time before it hits due to their range (but it would be nine shots I think).

### DIFF
--- a/dat/outfits/launchers/emp_grenade_launcher.xml
+++ b/dat/outfits/launchers/emp_grenade_launcher.xml
@@ -24,6 +24,7 @@
   <duration blowup="shield">1.3</duration>
   <thrust>0</thrust>
   <speed>500</speed>
+  <resist>10</resist>
   <damage>
    <type>emp</type>
    <penetrate>48</penetrate>
--- a/dat/outfits/launchers/emp_grenade_launcher.xml
+++ b/dat/outfits/launchers/emp_grenade_launcher.xml
@@ -24,7 +24,6 @@
   <duration blowup="shield">1.3</duration>
   <thrust>0</thrust>
   <speed>500</speed>
-  <resist>10</resist>
   <damage>
    <type>emp</type>
    <penetrate>48</penetrate>
--- a/dat/outfits/launchers/enygma_systems_huntsman_launcher.xml
+++ b/dat/outfits/launchers/enygma_systems_huntsman_launcher.xml
@@ -7,7 +7,7 @@
   <rarity>2</rarity>
   <mass>60</mass>
   <price>200000</price>
-  <description>Enygma Systems found that there was an unexploited niche in disabling weapons aimed at larger warships. Finding Medusa missiles and ion weaponry insufficient, they redisgned Unicorp's Caesar IV torpedo as a heavy EMP weapon, capable of rapidly frying the systems of even the largest warships.</description>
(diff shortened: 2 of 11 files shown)


not-pyroman:
actually, the resist stat controls how resistant the projectile is to jamming, for projectiles that do lock on that is

PhoenixRiver:
> actually, the resist stat controls how resistant the projectile is to jamming, for projectiles that do lock on that is take for example the Za'lek Hunter launcher, in game the stats say it has a 35% jamming resistance, and if you look at its XML file you will see that the resist stat there is also 35

Gah! Do you know which stats is used for a missile's armour? I'll have another dig but I'm new to the weapons game :(

PhoenixRiver:
Oh, just discovered that we don't have weapon on weapon collision at the moment... so PD weapons don't work... gah... I'll change back the resist stats and leave this here for now in case someone wants example PD weapons for testing solutions to this problem e.g. #2053

PhoenixRiver:
I guess traders will just have to wait for protection... *sigh*

bobbens:
Yes, point defenses are planned, but I want to finish and merge the `fleets` branch first before working on that. Although I may take a hiatus to work on missions a bit. I'll keep this open until point defense gets implemented and then try to work it into that.

PhoenixRiver:
Thank you :)

bobbens:
As of latest main, point defense weapons are now possible. I'll be adding some in similar vein, but quite different than this PR, so I'm closing. If you want work on other point defense weapons I can share my notes and we can work from there. Thanks!

bobbens:
For reference, the new point defense outfits can be found at https://github.com/naev/naev/tree/main/dat/outfits/pointdefense

PR #2122, opened 2022-03-16
Improved Cargo hull cores

PhoenixRiver:
This is a response to #2119 that doesn't involve improving collision detection but trys to improve and differentiate traders from combat (and stealth) ships... think turtles :)
It may well also work alongside #2121 but maybe both might have to be reduced in effectiveness if PD turrets happen, otherwise trader might become impenitrable... we want the pirates to have some chance before security turns up... if security turns up! ;)

I've added a secondary function to the cargo hulls to make them rather more resistant to damage at the expense of CPU and detection. Traders wanting the largest possible cargo holds are generally not going to worry about CPU (they're not trying to enter combat) and they are cargo ships... being detected quicker makes it easier for security to find and help them and they were unlikely to hide that well anyway. At least this way they can better stand a beating before popping!
I also feel that this sort of higher defense a the cost to potential offense (CPU usage makes more weapons harder) might be a good way to differentiate trader ships from simply being bad combat ships. It also affords them protection without sacrificing cargo.

### DIFF
--- a/dat/outfits/core_hull/large/sk_large_cargo_hull.xml
+++ b/dat/outfits/core_hull/large/sk_large_cargo_hull.xml
@@ -7,14 +7,18 @@
   <priority>7</priority>
   <mass>550</mass>
   <price>1100000</price>
-  <description>This cargo hull from Schafer &amp; Kane Industries makes use of state-of-the-art defensive alloys, which allows it to free up even more space for cargo without sacrificing too much durability.</description>
+  <description>This cargo hull from Schafer &amp; Kane Industries makes use of state-of-the-art defensive alloys, which allows it to free up even more space for cargo without sacrificing too much durability. Along with the increased space their (in)famous electromagnetic shockwve absorbtion generator comes built in, turning CPU power and a little more obviousness into missile james, tracking zero-points and reduced secondary splash damage, turtling has never been safer.</description>
   <gfx_store>cargo_hull_l.webp</gfx_store>
+  <cpu>-160</cpu>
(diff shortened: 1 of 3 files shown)


bobbens:
I like the idea, but feel that the jam chance is a bit too much. What about creating a new "defensive" hull type that would give large bonuses like the ones you added, but with less cargo capacity and cargo inertia? That way a trader could choose to either maximize cargo or maximize defenses?

PhoenixRiver:
I like it :) - maybe tomorrow's job though I feel!

PhoenixRiver:
Since the entire contents of this pull request are going to be reverted I'm just going to close this request and create a new one to make things cleaner, is that okay?

bobbens:
Yeah, that's fine. You could also force push instead though.

PhoenixRiver:
Hmmm... I'll look but at the moment github is not letting me even fetch & merge main or publish a new branch sooooo *shrug* (or even comment??)

PhoenixRiver:
git hub won't repond to any of my git commands and even my comments only work when I try repeatedly so I've got a branch coded up for this but I can't pass it on to you, maybe tomorrow it will be fixed, apologies. :(

Zireael07:
@PhoenixRiver: Github seems to be having issues this afternoon, I had a moment where I couldn't star any repos...

PhoenixRiver:
This pull request has been superceeded by #2126 so 'ill close it, apologies I could not force push - apparently git desktop doesn't have that functionality?

PR #2123, opened 2022-03-16
Fix a bug where num2str() could suppress '0' in the middle digits.

oo13:
info
"Damage taken" should be 1,*0*98,662 MJ.


bobbens:
Looks good, thanks!

PR #2124, opened 2022-03-16
Fix two duplicate labels in zbh07.

oo13:
They seem to be mistaken when copy-pasting.

1. "cont01b" -> "cont01a"
2. "cont01b" OK
3. "cont01c" OK
4. "cont01c" -> "cont01d"


bobbens:
Looks good, thanks!

PR #2125, opened 2022-03-17
new tiny gui 'slimplus'

jjustra:
I've made a gui. Aim is to provide more space.
It saves some pixels. I've hide some things that I did not used ... not sure how much of a good idea was that in general, but on my notebook it has good effect.


PhoenixRiver:
Did you move the Communications message to the left side? I've been wondering if that wouldn't help leave more of the main screen free on widescreen screens and there is more space since combat ships take up more of the bottom of the right side with weapon readouts...

jjustra:
Do you mean narrowing that log window? That seem like a good idea actually, if I've got it right.

bobbens:
Closing as out of date.

PR #2126, opened 2022-03-18
Red Star Cargo hulls

PhoenixRiver:
So I've reduce the missile jam a little, reduced the cargo space by a third, reduced the cargo mod and inertia benefits from 25 to 10 and made these hulls more expensive. I hope this is a reasonable tradeoff, especially as the CPU costs are just as high and detection is still as high as it was before.
I also added into lore that the Red Star organisation created hem for their humanitarian work since it seems appropriate that an aid agency would do this if necessary given the higher risks their work poses and that (in theory) they shouldn't be being targeted but stray laser shots don't recognise international symbols! ;)
I should note that there are also a couple of changes to tech.xml to allow for future use in the guild stuff but which should probably be there now anyway.

This is a replacement request for #2122 as it appears my git interface (git desktop) doesn't have the ability to force push :/ ??

### DIFF
--- a/dat/outfits/core_hull/medium/rs_medium_cargo_hull.xml
+++ b/dat/outfits/core_hull/medium/rs_medium_cargo_hull.xml
@@ -1,0 +1,24 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<outfit name="Red Star Medium Cargo Hull">
+ <general>
+  <typename>Core Systems (Hull)</typename>
+  <slot prop="hull">structure</slot>
+  <size>medium</size>
+  <priority>7</priority>
+  <mass>140</mass>
+  <price>360000</price>
+  <description>The Red Star Organisation have developed a hull modification that provides both ample cargo space and protection for their own work and have now started selling it to others. Deciding "in for a penny, in for a pound", an electromagnetic shockwave absorbtion generator was added, typically this makes the hull rather pointy and scanner obvious but since the hull is already rather lumpy due to the cargo pods... on the other hand, for the price of some CPU power, missile jams, tracking zero-points and reduced secondary splash damage are all achieved making this hull rather more resiliant.</description>
+  <gfx_store>cargo_hull_m.webp</gfx_store>
+  <cpu>-64</cpu>
+ </general>
+ <specific type="modification">
+  <cargo>80</cargo>
+  <absorb>20</absorb>
+  <armour>180</armour>
+  <cargo_inertia>-10</cargo_inertia>
+  <cargo_mod>10</cargo_mod>
+  <ew_hide>-20</ew_hide>
+  <ew_evade>10</ew_evade>
+  <jam_chance>42</jam_chance>
+ </specific>
+</outfit>
--- a/dat/tech.xml
+++ b/dat/tech.xml
@@ -208,17 +208,20 @@
   <item>S&amp;K Ultralight Combat Plating</item>
   <item>S&amp;K Light Combat Plating</item>
   <item>Nexus Light Stealth Plating</item>
+  <item>Red Star Small Cargo Hull</item>
   <item>S&amp;K Small Cargo Hull</item>
   <item>Patchwork Medium Plating</item>
   <item>Unicorp D-12 Medium Plating</item>
   <item>Unicorp D-24 Medium Plating</item>
   <item>S&amp;K Medium Combat Plating</item>
   <item>Nexus Medium Stealth Plating</item>
+  <item>Red Star Medium Cargo Hull</item>
   <item>S&amp;K Medium Cargo Hull</item>
   <item>S&amp;K Medium-Heavy Combat Plating</item>
   <item>Patchwork Heavy Plating</item>
   <item>Unicorp D-48 Heavy Plating</item>
   <item>Unicorp D-68 Heavy Plating</item>
+  <item>Red Star Large Cargo Hull</item>
   <item>S&amp;K Large Cargo Hull</item>
   <item>S&amp;K Heavy Combat Plating</item>
   <item>S&amp;K Superheavy Combat Plating</item>
@@ -402,6 +405,15 @@
   <item>Grave Lance</item>
   <item>Grave Beam</item>
   <item>Ragnarok Beam</item>
+ </tech>
+ <tech name="Mining 1">
+  <item>Mining Lance MK1</item>
+  <item>S&amp;K Plasma Drill</item>
+ </tech>
+ <tech name="Mining 2">
+  <item>Mining Lance MK2</item>
+  <item>Nexus Drill Lance</item>
+  <item>S&amp;K Heavy Plasma Drill</item>
  </tech>
 
  <!-- Generic Outfits -->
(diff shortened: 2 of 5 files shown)


dat/tech.xml
```diff
@@ -635,6 +647,11 @@
  <tech name="Criminal Outfits">
   <item>Camouflage Burster</item>
  </tech>
+ <tech name="Mining Outfits">
+  <item>Asteroid Scanner</item>
+  <item>Medium Weapon License</item>
+  <item>Milspec Impacto-Plastic Coating</item>
```
bobbens, line 653: This should be a Dvaered-only outfit.

bobbens (changes requested):
One small comment on a coating that should be Dvaered-only, but the rest seems fine, thanks! Other than that, I do think this PR makes the hulls not attainable because there are no planets that have the new techs like the `Hulls Cargo`. That should probably also be addressed or lots of outfits will become inaccessible outside of Zied.

Unrelated to this PR, I do think that we have to somewhat rethink from the ground up how we are doing techs and their distribution. Some stuff is pretty hard to find, while other stuff is just everywhere. Ideally we would want to have more interesting distributions of things while not making things too hard to find.

dat/outfits/core_hull/medium/rs_medium_cargo_hull.xml
```diff
@@ -0,0 +1,24 @@
...
+  <mass>140</mass>
+  <price>360000</price>
+  <description>The Red Star Organisation have developed a hull modification that provides both ample cargo space and protection for their own work and have now started selling it to others. Deciding "in for a penny, in for a pound", an electromagnetic shockwave absorbtion generator was added, typically this makes the hull rather pointy and scanner obvious but since the hull is already rather lumpy due to the cargo pods... on the other hand, for the price of some CPU power, missile jams, tracking zero-points and reduced secondary splash damage are all achieved making this hull rather more resiliant.</description>
+  <gfx_store>cargo_hull_m.webp</gfx_store>
+  <cpu>-64</cpu>
+ </general>
+ <specific type="modification">
+  <cargo>80</cargo>
+  <absorb>20</absorb>
+  <armour>180</armour>
```
not-pyroman, line 17: I feel like the armour stats are too similar to existing cargo hulls to make it worth the extra cost and less space, considering most pirates i have seen harass me use unguided missiles, beams, or some other weapons
This applies to the small cargo hull especially
Also i feel the current CPU values are quite off-putting for at least the small hull

not-pyroman (commented):
I feel like the armour stats are too similar to existing cargo hulls to make it worth the extra cost and less space, considering most pirates i have seen harass me use unguided missiles, beams, or some other weapons
This applies to the small cargo hull especially
Also i feel the current CPU values are quite off-putting for at least the small hull

dat/tech.xml
PhoenixRiver: I added Hulls Cargo into the same place that the SK cargo hulls already were, does this not work like I think it does? (i.e. categories can't be added into other categories to include their contents as well?)

not-pyroman: > I added Hulls Cargo into the same place that the SK cargo hulls already were, does this not work like I think it does? (i.e. categories can't be added into other categories to include their contents as well?)

Bobbens is referring to the impacto-plastic coating

dat/outfits/core_hull/medium/rs_medium_cargo_hull.xml
PhoenixRiver: It's worth bearing in mind that they have a cargo percentage bonus meaning the only hulls that have the final amount of cargo these do is the less defensive S&Ks, also that they are decidedly more defensive than the S&Ks and even than the more basic Hulls. They are designed as specialist hulls similar to the Nexus stealth hull and the S&K ultralight hull so they won't be cheap, especially as the engineering needed to make a hull that esesentially bridges defensive and cargo at the same time is always going to be more expensive than one that does either... specifically also the time to lock is longer and absorbtion higher than standard armour hulls and for a cargo hauler CPU is a useless dump stat except for turrets, and even then the engine is the limit on weight and so higher level turrets that use m,ore CPU aren't really possible anyway.
This hull is designed for a trader who fills to the briom with cargo and just ignores attacks and tried to make it to the jump gate / spob before the pirates notice or kill him... before them this was nigh on impossible, now it is just difficult!
Also note I also tried submitting better cargo hulls and they were too good and were rejected (with good reason if we can get proper point defence turrets working too :)

not-pyroman: On that point, in a cargo ship i usually put a defensive hull with decent cargo space, and also fill the thing to the crim with cargo pods and it usually gives me a larger chance of survival plus good cargo space

PhoenixRiver:
>  Other than that, I do think this PR makes the hulls not attainable because there are no planets that have the new techs like the `Hulls Cargo`. That should probably also be addressed or lots of outfits will become inaccessible outside of Zied.

I added Hulls Cargo into the same place that the SK cargo hulls already were, does this not work like I think it does? (i.e. categories can't be added into other categories to include their contents as well?)

PhoenixRiver:
> Unrelated to this PR, I do think that we have to somewhat rethink from the ground up how we are doing techs and their distribution. Some stuff is pretty hard to find, while other stuff is just everywhere. Ideally we would want to have more interesting distributions of things while not making things too hard to find.

I started a discussion on this and the wider issue of how spobs are defined, an issue that proofreading the spobs has made me acutely aware of... sooo many Class A Oceanic worlds! :D

dat/tech.xml
PhoenixRiver: fixed

dat/outfits/core_hull/medium/rs_medium_cargo_hull.xml
PhoenixRiver: Sorry, not easy. For various reasons *ow - my ears* - that included - Discord is not very workable for me, constant moving workstations doesn't help either :/

With the standard hulls you don't get a percentage bonus on you cargo space so they are much less efficient... as I mentioned somewhere else, you can get 600 tonnes on a Rhino, 900 on a Mule and 5000 on a Zebra with the cargo hulls, plus cargo pods to the max.

I've been working on making pure traders more survivable but there is a lot to do with trade generally since there is little consistancy between population figures, the economy, the desire to not let the player massively impact those economies, the trade lanes and how to calculate them, the number and sizes of trade ships and how much of that we actually represent and what sort of interspob and interfactional trade levels we are looking at, but there is a discussion topic and a better pull request that already discusses a lot of these things... please feel free to help out :D

PhoenixRiver:
Okay, I feel like I have fixed the necessary on this but that's because I'm not sure how the tech.xml file works... is it possible to put techs into the description of a tech (i.e. nested) in techs.xml, or is this not possible? If it is not then I need to fix a bunch of stuff in there I think!

Can I also ask how this works for the tefchs sold at a planet, it seems like you can put a category in the planets file to include every tech in that category, this is correct, yes?

Also can the tech in each category overlap? e.g. can the mining_drill be in the laser 1 category as well as its own category? What happens if a planet sells both laser 1 category and mining category?

Sorry for the confusion about how this works, apparently I'm just too good at hitting edge cases ;)

PhoenixRiver:
For now, while we sort out how to deal with tech distribution, I have put all the cargo hulls into the medium tech hulls tech as well (and added a yachts tech, maybe we should have one for each category of ship? but certainly this is useful for tourist spobs since the player is more likely to want to purchase a ship in these places to provide sightseeing missions (but then again I guess the sightseeing missions don't happen more frequently in spobs tagged that way?)

bobbens:
OK, looks good, thanks! We'll have to adapt the techs and maybe make modifications as needed. However, that can be done in the near future.

PR #2129, opened 2022-03-18
Fix some missing translation marks in dv_lostagent.

oo13:
Fix some missing translation marks in dv_lostagent.


bobbens:
Looks good, thanks!

PR #2131, opened 2022-03-22
World updates (Frontier and Soromid stations)

PhoenixRiver:
Tags, techs and exteriors updated.
The "Hulls Cargo" tech has also been added to relevant spobs; currently it duplicates some of the "Hulls Medium" hulls but this can be fixed once "Hulls Cargo" has been better propagated if this is wanted.
Also added a few lesser used tags that could be important; government (mostly for faction homeworlds), medicine (mostly Soromid and Red Star spobs), shipbuilding (for shipbuilding stations and the like), religious (for spobs of religious importance).

### DIFF
--- a/dat/spob/atalanta.xml
+++ b/dat/spob/atalanta.xml
@@ -36,4 +36,10 @@
   <item>Missiles 1</item>
   <item>Missiles 2</item>
  </tech>
+ <tags>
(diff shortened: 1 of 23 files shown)


bobbens:
Looks good, thanks!

PR #2133, opened 2022-03-26
spobs named V* (plus a few extras) tagged and checked

PhoenixRiver:
Spobs proofread and updated tags, popluation number format, exterior image and class - wiht apologies to the fans of A class planets... many fewer now I'm afraid, they probably only make up 50% of the galaxy at present! ;)


bobbens:
Looks good, thanks!

PR #2143, opened 2022-05-11
Goddard Battlecruiser → Battleship

LJDude:
As Messerbär on Discord said, calling it a battlecruiser would imply that it is to other battleships as the kestrel is to other cruisers.


bobbens:
Looks good, thanks!

PR #2145, opened 2022-05-23
Za'lek and Soromid news

theElerd:


bobbens:
Looks great! The only thing that is a bit problematic lore-wise is the phrase about "living ships". So I swapped it around to "intelligent pets" signing up and the manager lamenting about "unionizing their bioship next". Thanks!

PR #2147, opened 2022-06-08
Allow saving at the right time

sigus:


bobbens:
With my latest changes the code you presented on discord should work again (replicated below). So this PR is not necessary anymore. Thanks!
```lua
player.allowSave(false)
player.takeoff()
player.allowSave()
```

PR #2152, opened 2022-06-30
Hired Escorts reimplemented as Hireable Vagabonds with mini-personalities

ThrosturX:
This might be ready for merging, but I recommend perhaps revising my naming choices.

Summary:

"Hired Escorts" event forked & modified to make the escorts more like individuals, unlike the ships in your fleet who are more robotic, these guys have some chatter and a bunch of internal variables that make them feel more alive and playful. They talk less and less with time but every generated person has their own internal "chatter" variable determining their chance of idle chitchat. So the player can get fed up with a chatterbox and fire him, or whatever. It's interesting. Oh, and they cost a lot to maintain, so even though I let the player have 6 of them, they are a huge credit sink unless the player meticulously "trains" each one individually, or at least in small groups.

What they do:

- They mostly behave like regular escorts, but they buy their own ships based on a calculated budget
- They will board enemies and share the profits with the player (immediately on boarding for credits and on landing in case of commodities)
- They have their own dream ship, their ultimate goal, and they will try to buy that ship more often and be more generous when they are in one
- They like to say stuff (might need to be expanded upon, currently uses pirate AI for support)
- When they arrive, there is a chance of chatter, same when they are attacked, when they board something, when they have a lot of debt (they buy their own ships, remember? They need to take loans if they aren't good at making money). The chatter they choose is affected by their last sentiment, if they have one (plundered a large bounty, was defeated in battle, etc).
- When they have debt, some outfits might malfunction and they might complain about it or remain more silent than usual and it will take them longer to arrive in the system with the player
- Each escort will remember some "favorite ships", and there is a chance that they will buy a "favorite ship" any time they need a replacement

I should note here that if they board an "Alive" bounty, the mission will be failed because the player can't board the target again.

What you can do with them:
- You can make them attack and board things (though they need to pick up the attack mission before the target is disabled)
- Boarding and profiting increases their experience level
- You can give them tips, which increases their chances of buying a better ship next time they need a new ship
- You can upgrade them at a shipyard, the escort will buy a random ship from the existing shipyard, or the most expensive one within its budget if it is a "commander" (they can still buy a dream ship even if it isn't at the shipyard)
- Adding to the above, if you "force" your escort into a nice ship like a pirate starbridge (by upgrading at a pirate shipyard until you get the right ship) and then successfully board enough enemies to make it a favorite ship contender, there is a chance that your escort will buy that nice ship again

Okay, hold on, what's a commander?
A commander is the escort that can buy a large ship, like a kestrel or a hawking. All the other escorts can fly destroyers or the Zebra, and smaller. There can only be one commander for balance reasons, otherwise it's ridiculous for sure. Maybe in the future the commander is more interesting, but for now sometimes it will drop leadership and recruit the players other escorts and do whatever the native AI tells it to do. This adds a little bit of mystery and playfulness to the escorts. (I should add here that my pirate escort commander caused me a lot of trouble!)

Anything else?
Yes, I'm probably forgetting something, but one thing to mention is that I gave the "transport" escorts an intrinsic outfit that makes their defenses much better but their weapons weaker, basically increasing the chances of them being disabled but not destroyed and decreasing the chances of them destroying an enemy ship. This is to provide a more interesting gameplay experience when your escorts are flying transport ships for some reason, since these escorts don't share their cargo space with the player. A thing to note though is of course that the escorts with more cargo space can plunder more commodities to sell at the next station.


dat/scripts/equipopt/templates/escort.lua
```diff
@@ -0,0 +1,186 @@
...
+		 choose_one{"S&K Medium Combat Plating", "Unicorp D-24 Medium Plating", "S&K Medium-Heavy Combat Plating", "Patchwork Medium Plating" },
+   } end,
+}
+
+local escort_params_overwrite = {
+   weap = 1.5, -- Focus on weapons
+   disable = 2.4, -- prefer disablers
+   -- some nice preferable escort outfits
+  prefer = {
+		[ "TeraCom Medusa Launcher"] = 6 , ["TeraCom Medusa Launcher"] = 2,
```
bobbens, line 130: You're defining "TeraCom Medusa Launcher" twice.

dat/events/scavenger_event.lua
```diff
@@ -0,0 +1,2077 @@
...
+        mem.refuel = mem.refuel * 0.5
+    end
+    mem.refuel_msg = fmt.f(_([["For you, only {credits} for a jump of fuel."]]), {credits = fmt.credits(mem.refuel)})
+
+    -- Deal with bribeability
+    mem.bribe = mem.bribe_base
+    if mem.allowbribe or (mem.natural and mem.bribe_rng < 0.95) then
+        mem.bribe_prompt = fmt.f(bribe_prompt_list[rnd.rnd(1, #bribe_prompt_list)], {credits = fmt.credits(mem.bribe)})
+        mem.bribe_prompt_nearby = bribe_prompt_nearby_list[rnd.rnd(1, #bribe_prompt_nearby_list)]
+        mem.bribe_paid = bribe_paid_list[rnd.rnd(1, #bribe_paid_list)]
```
bobbens, line 2068: `bribe_prompt_list`, `bribe_prompt_nearby`, and `bribe_paid_list` is not actually defined anywhere...

dat/events/scavenger_event.lua
```diff
@@ -0,0 +1,2077 @@
...
+            -- get a new ship
+            _createReplacementShip(edata, limit_ships)
+            -- apply upgrade experience bonus
+            edata.wallet = edata.wallet + math.floor(upgrade_amount * edata.experience * 0.05)
+            edata.last_sentiment = "content"
+        end
+    elseif
+        n == 2 and tk.yesno("", fmt.f(_("Are you sure you want to get rid of {name}? This cannot be undone."), edata))
+     then
+        local k = #mem.persons
```
bobbens, line 1947: This `k` is defined, but not actually used and shadowed by the `k` in the for loop.

dat/scripts/pilotname/human.lua
```diff
@@ -0,0 +1,951 @@
+local fmt = require "format"
+
+-- For translation, just transliterate if necessary.
+local articles = {
```
bobbens, line 4: Seems to have lots of unused code (likely from copy pasting at the beginning). Should be made minimum.

dat/scripts/equipopt/wrapper.lua
```diff
@@ -13,4 +14,5 @@ equipopt.soromid  = require 'equipopt.templates.soromid'
 equipopt.pirate   = require 'equipopt.templates.pirate'
 equipopt.thurion  = require 'equipopt.templates.thurion'
 equipopt.proteron = require 'equipopt.templates.proteron'
+equipopt.scavenger   = require 'equipopt.templates.scavenger'
```
bobbens, line 17: I don't think you committed the scavenger template.

bobbens (commented):
Just a few quick comments here and there. I can't test without the templates.scavenger file atm.

dat/scripts/pilotname/human.lua
bobbens: I also pushed to the `escorts` branch on this repo some minor fixes. You might want to pull them. They are basically warnings from luacheck, although there's still a lot left.

dat/scripts/equipopt/wrapper.lua
ThrosturX: Shouldn't be needed, artifact from my other modding activity...

bobbens:
Thanks for the PR. Closing as per an old discussion on discord. While the concept is interesting, it is probably best as a plugin as we want to focus the story and everything around the player and not their hirable escorts. Furthermore, hiring escorts is not a thing in master, in lieu of player fleets.

PR #2155, opened 2022-07-15
separate some merged msgids into different contexts.

oo13:
These msgids mostly merged from a pirate's name and a planet's.


bobbens:
Looks good, thanks!

PR #2156, opened 2022-07-16
Fixes off-by-one error in mission id calculation

ThrosturX:
Pilot cargo logic assumes that anything with mission id 0 is a non-mission cargo item, but the first mission item that is added becomes a mission item with id 0. This is an off-by-one error where regular commodities and mission commodities can get mixed up, but you will for instance find that if you have only one mission of 10 diamonds to be moved and buy 30 diamonds of commodities (before accepting the mission), those 30 "normal" diamonds will be locked while you are in space, and you will not be able to jettison them.

This PR rectifies the error by setting the minimum mission id to 1 (by setting the variable `max_id` to 1).

There might be other ways to achieve the same effect, but IMHO, a `max_id` for mission ids can never be 0 because the minimum mission id is 1 for all valid mission ids.

### DIFF
--- a/src/pilot_cargo.c
+++ b/src/pilot_cargo.c
@@ -236,7 +236,7 @@ unsigned int pilot_addMissionCargo( Pilot* pilot, const Commodity* cargo, int qu
    id = ++mission_cargo_id;
 
    /* Check for collisions with pilot and set ID generator to the max. */
-   max_id = 0;
+   max_id = 1;
    for (int i=0; i<array_size(pilot->commodities); i++)
       if (pilot->commodities[i].id > max_id)
          max_id = pilot->commodities[i].id;


bobbens:
I don't see how this is necessary. `id` gets set to `++mission_cargo_id`, which is at least `1`, because `mission_cargo_id` is initialized to 1. Then only if `max_id >= id`, it sets `mission_cargo_id` to `max_id` and increments by `1`. So there is no way for a mission cargo to be initialized with an id of 0. Am I missing something?

ThrosturX:
> I don't see how this is necessary. [..] Am I missing something?

You might be right, but I think you're missing the potential for an overflow, but this isn't confirmed/tested and based solely on observation. I was scratching my head for a long time and suggested this modification based on the observed behavior vs what I expected. This might be fixed by the last nightly, I haven't checked yet, but the (at least previously) reproducible issue was as follows:

1. Have no missions
2. Find/select a mission to move a Commodity to a spob
3. Purchase some same Commodity
4. Accept mission
5. -> commodity and mission cargo is in the same stack (incorrect, observed)
6. repeat for second mission with a different Commodity
7. -> commodity and mission cargo in separate stacks (expected, observed)

Now, why this happens, or where the error is, I'm not sure because I didn't look at all of the code, there's a lot :) but looking in `pilot_cargo.c` I made the logical assumption that somehow the mission ID **must** be set to zero when the error occurs, as that seems to be (based on the functions that add cargo) the only way to stack any two things in the cargo list (or I guess it's called `commodities` here).

So, tracing back to the source, well, I couldn't actually find it, but that was the only place where I could imagine an input that triggers the incorrect behavior. If I go full haywire with my assumptions, I would expect that wherever the actual error is, it's somehow setting the mission ID to `uint_max` so that it overflows to zero. My assumption here is, when the player has no missions, the ID generator is probably incorrectly set to `uint_max` somehow (perhaps it is on loading?) resulting in the first accepted mission to incorrectly inherit an invalid mission ID.

Now, I'm not even 100% sure that I'm correct, but I believed at the time when I proposed this change that making this change would at the very least cover the overflow edge case, since it doesn't seem to be protected anywhere else. What happens if `mission_cargo_id` is `uint_max`? We know what happens, the question is why it was that high to begin with, but I think I'll defer that quest to someone who has worked more with the code.

bobbens:
This was a bug in shipCargo and should be fixed in 3b7985f56965ed3e07323d1faa0f1c8b8dec7d58 . Thanks!

PR #2157, opened 2022-07-17
fix boarding hook edge case

ThrosturX:
if you board a pilot that is removed in a boarding hook, this line prevents the error that would have occurred when trying to reference the invalid pilot.


bobbens:
Looks good, thanks!

PR #2161, opened 2022-07-22
allow the player to steal ships while boarding

ThrosturX:
This is untested, but should work in theory, unless I made some mistakes (which is likely as I did this in about half a cup of coffee's worth of time).

### DIFF
--- a/dat/ai/escort_guardian.lua
+++ b/dat/ai/escort_guardian.lua
@@ -1,0 +1,137 @@
+require 'ai.escort'
+local fmt = require 'format'
+local atk = require "ai.core.attack.util"
+
+-- Settings
+mem.aggressive = true
+mem.guardbrake    = 350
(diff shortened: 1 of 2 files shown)


ThrosturX:
I've marked this as a draft, because:

- it needs the part where we compute whether or not the player can actually capture the ship (aside from `swapship` requirements)
- it needs an adequate amount of testing w.r.t. the whole `swapship`ping in space, but I have been doing a lot of it with relatively good success so far.

bobbens:
Inspired a bit by this PR, I've written up a proposal on my thoughts of ship capturing at #2599 .

TheGiraffe3:
So a comment in #2599, bobbens mentioned this was implemented, though still a little rough around the edges. In that case should this PR be closed?

bobbens:
Should be closed, thanks!

PR #2167, opened 2022-07-31
Add scary war paint

ThrosturX:
Allows the player to paint the ship in scary war paint (kind of crimson colored) to earn respect from pirates, and hate from others.

### DIFF
--- a/dat/outfits/unique/scary_paint.lua
+++ b/dat/outfits/unique/scary_paint.lua
@@ -1,0 +1,29 @@
+--[[
+   Scary War Paint
+
+   Makes the player feared by pirates and bounty hunters.
+--]]
+local fmt = require "format"
+local pir
+
+function onload( _o )
+   pir = require "common.pirate"
+end
+
+function init( p, po )
+   p:effectAdd("Paint", math.huge)
+end
+
+function onremove( p, po )
+   p:effectRm("Paint")
+end
+
+function onscanned( p, po, scanner )
+   if pir.factionIsPirate(scanner:faction()) then
+      scanner:setFriendly(p)
+   elseif scanner:name() == _("Bounty Hunter") then
+	-- unfortunately, bounty hunters don't scan yet I think
+      scanner:setHostile(p)
+   end
+end
+
--- a/dat/outfits/unique/scary_paint.xml
+++ b/dat/outfits/unique/scary_paint.xml
@@ -1,0 +1,28 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<outfit name="Scary Paint">
+ <general>
+  <rarity>6</rarity>
+  <unique />
+  <typename>Scary War Paint</typename>
+  <slot>utility</slot>
+  <size>small</size>
+  <mass>1</mass>
+  <price>1000000</price>
+  <description>Scary war paint can be used to strike fear in those that lay their eyes upon the ship that wears it.</description>
+  <desc_extra>Paints the ship a scary colour, causing pirates to respect you and bounty hunters to hunt you.</desc_extra>
+  <gfx_store>black_cat_doll.webp</gfx_store>
+  <limit>fake_transponder</limit>
+  <illegalto>
+   <faction>Empire</faction>
+   <faction>Dvaered</faction>
+   <faction>Soromid</faction>
+   <faction>Sirius</faction>
+   <faction>Za'lek</faction>
+   <faction>Goddard</faction>
+   <faction>Frontier</faction>
+  </illegalto>
+ </general>
+ <specific type="modification">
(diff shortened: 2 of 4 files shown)


dat/outfits/unique/scary_paint.lua
```diff
@@ -0,0 +1,33 @@
...
+   p:effectRm("Paint")
+end
+
+-- TODONOTE: when independents start scanning, change the else branch!
+function onscanned( p, _po, scanner )
+   if pir.factionIsPirate(scanner:faction()) then
+      if pir.maxClanStanding() > 20 then
+         scanner:setFriendly(p)
+      else
+         scanner.setHostile(p, false)
```
bobbens, line 26: Shouldn't this be `scanner:setHostile`?

ThrosturX: thanks!

dat/outfits/unique/scary_paint.lua
```diff
@@ -0,0 +1,33 @@
...
+function onload( _o )
+   pir = require "common.pirate"
+end
+
+function init( p, _po )
+   p:effectAdd("Paint", math.huge)
+end
+
+function onremove( p, _po )
+   p:effectRm("Paint")
```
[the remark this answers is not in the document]
bobbens: Implemented it in main.

dat/outfits/unique/scary_paint.xml
```diff
@@ -12,15 +12,6 @@
   <desc_extra>Projects an unsettling visual appearance onto the ship's hull, causing a mild shock to those that lay their eyes upon it. Pirates tend to respect those that go the extra mile just for appearance's sake.</desc_extra>
   <gfx_store>black_cat_doll.webp</gfx_store>
   <limit>fake_transponder</limit>
-  <illegalto>
```
bobbens: Why did you remove illegality? I like the idea of it being a trade-off between pirate love and other faction love.

ThrosturX: I found after some testing that illegality is not required, since all the factions that check for illegality also do a scan, and anyone that scans you and isn't a pirate will set you as hostile. So this gives the player a choice -- shoot back and lose standings or ignore them (but don't lose standings). I could go either way, but I found it was a bit annoying to lose 2% faction standing at the very end of a journey right before landing.
But, it's up to you and all the other players, as I said, I could go either way, but I think illegality here is a little bit overkill because whoever scans you will start harassing you anyway, which either forces you to respond or run away/tank.

dat/outfits/unique/scary_paint.xml
```diff
@@ -0,0 +1,19 @@
...
+<outfit name="Crimson Holo-Disfigurator">
+ <general>
+  <rarity>6</rarity>
+  <unique />
+  <typename>Ship Paint</typename>
+  <slot>utility</slot>
+  <size>small</size>
+  <mass>1</mass>
+  <price>1000000</price>
+  <description>War paint has been used to strike fear in those that lay their eyes upon the ships that bear it for millenia. This holographic matrix allows the visual appearance of a ship to be altered to make it appear as if it had been painted in bloody crimson.</description>
```
TheGiraffe3, line 11: millenia is Oxford English? Or should it be
```suggestion
  <description>War paint has been used to strike fear in those that lay their eyes upon the ships that bear it for millennia. This holographic matrix allows the visual appearance of a ship to be altered to make it appear as if it had been painted in bloody crimson.</description>
```

TheGiraffe3:
You'll need to update the branch to fix the spellcheck.

TheGiraffe3 (approved):

ThrosturX:
> You'll need to update the branch to fix the spellcheck.

Thanks! This PR is pretty old, huh!

bobbens:
I've done some fixes and pushed to the `scarypaint` branch. While I like the idea, I'm not sure how useful it can be as is.
I've been trying to think about how to improve it. I've given it some signature bonus in the branch, but not sure if it's enough.

I am thinking about maybe making it be an upgrade to the "Fake Transponder" (or have an option to merge it with it), that would basically make it so you get a bonus to pirate standing (marauders + space pirates, stacks with pirate ships), so that they wouldn't attack you at all if you had a clan positive. It would just end up being something to boost the fake transponder, that you'll probably be using anyway if you're a pirate.

So potentially have something like:
1. You buy fake transponder.
2. You acquire holo disfigurator (maybe a small quest or something? misi?).
3. You get to chapter 2 and unlock your space station.
4. If you have holo disfigurator + fake transponder, you get an option to research / upgrade the fake transponder to incorporate holo disfigurator bonuses with no downsides other than those of the fake transponder itself.

PR #2168, opened 2022-07-31
emergency shield booster shader effect

ThrosturX:
I felt it was missing and made one for myself. Then I felt guilty for not sharing it with the world, enjoy!

### DIFF
--- a/dat/outfits/utility/emergency_shield_booster.lua
+++ b/dat/outfits/utility/emergency_shield_booster.lua
@@ -24,7 +24,7 @@ local function turnon( p, po )
 
    -- apply nice shader effect
    p:effectAdd("Shield Boost")
-   
+
    return true
 end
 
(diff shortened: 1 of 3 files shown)


ThrosturX:
image
Looks a bit like this, intensity decreases with time.

dat/outfits/utility/emergency_shield_booster.lua
```diff
@@ -34,6 +37,7 @@ local function turnoff( _p, po )
    po:clear() -- clear stat modifications
    mem.timer = cooldown
    mem.active = false
+   p:effectAdd("Shield Boost", 1)
```
bobbens, line 40: The 1 parameter isn't necessary as it is default.

bobbens:
I've been trying the effect, and I think it's good, but maybe a bit too abrupt changes. I'll push some fixes, but it probably still needs a bit of work. Thanks!

PR #2169, opened 2022-08-01
Add auxiliary ship bay & outfit scanner drone

ThrosturX:
This PR adds functionality that uses the new `cargo_shuttle` model by adding a new outfit, the auxiliary ship bay.
Since I had the crazy idea of a small sized fighter bay (remote controlled drone) that would fit well on this ship, I added that too. So there's two new outfits:

- Auxiliary Ship Bay
- Za'lek Scanning Drone Interface

The Aux. bay adds a button in the info window that lets you ride around in the cargo shuttle, you board your old ship again to return. The cargo shuttle is automatically fitted based on a very simple algorithm that tries to put "useful science stuff" and then picks a random accessory every time.
The scanning drone interface is a fighter bay with a single za'lek scout drone that doesn't have any weapons. If it's close enough to the player's target, it will broadcast a random scanned outfit every once in a while. Otherwise, it works like a regular fighter (but obviously it can't actually attack anything).

Neither outfit is acquirable other than via the Lua console with this PR. If you give yourself a scanning drone interface and equip an auxiliary bay, your cargo shuttle will have the scanning bay interface equipped when you use it. Same goes for the pulse scanner and the S&K Plasma Drill :) If you don't have a plasma drill, a fake transponder or blink drive is equipped instead, depending on whether or not you have them and whether you might need the transponder. As a last resort, you get stealth utilities (veil of penelope, nexus stealth coating).


ThrosturX:
image

ThrosturX:
This PR is now available as a plugin and can therefore probably be closed: [ThrosturX/naev-auxiliary-bay](https://github.com/ThrosturX/naev-auxiliary-bay)

PR #2172, opened 2022-08-11
Make formations.keys in-order.

oo13:
This PR makes formations.keys deterministic.
formations.keys should not be changed when restarting the game, because the saved variable "player_formation" is an index of formations.keys.

This PR also introduces formations.names as translated texts for the panel of formation.


bobbens:
Looks good, thanks!

PR #2173, opened 2022-08-11
Fixed some issues of translation.

oo13:
Fixed some issues of translation and some typos.

### DIFF
--- a/dat/events/neutral/travelling_merchant.lua
+++ b/dat/events/neutral/travelling_merchant.lua
@@ -240,8 +240,8 @@ They get uncomfortably close
 "Data Matrices!"]]))
          mm(_([["You wouldn't have happened to come about some Data Matrices?"]]))
          vn.menu{
-            {_([["No"]]), "matrices_yes"},
-            {_([["Yes"]]), "matrices_yes"},
+            {p_("not to come", [["No"]]), "matrices_no"},
+            {p_("not to come", [["Yes"]]), "matrices_yes"},
             {_([["…"]]), "matrices_silent"},
          }
 
@@ -294,7 +294,7 @@ They get uncomfortably close
       special_virtue    = outfit.get("Machiavellian Virtue"),
    }
    vn.label("special")
-   mm(fmt.f(_("So you're interested in my special services. Quite a bargain might I say. Each services costs only {cost}.")
+   mm(fmt.f(_("So you're interested in my special services. Quite a bargain might I say. Each services costs only {cost}."),
       {cost=poi.data_str(upgrade_cost)}))
    vn.menu( function ()
       local opts = {
--- a/dat/missions/zalek/neburesearch/neburesearch_02.lua
+++ b/dat/missions/zalek/neburesearch/neburesearch_02.lua
@@ -98,7 +98,7 @@ But before I forget, there's some issue..."]]))
     misn.setDesc(fmt.f(_("Take Dr. Mensing to {pnt} in the {sys} system as fast as possible!"), {pnt=homeworld, sys=homeworld_sys}))
     mem.misn_marker = misn.markerAdd(homeworld, "low")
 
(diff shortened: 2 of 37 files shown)


dat/events/neutral/travelling_merchant.lua
```diff
@@ -240,8 +240,8 @@ They get uncomfortably close
 "Data Matrices!"]]))
          mm(_([["You wouldn't have happened to come about some Data Matrices?"]]))
          vn.menu{
-            {_([["No"]]), "matrices_yes"},
-            {_([["Yes"]]), "matrices_yes"},
+            {p_("not to come", [["No"]]), "matrices_no"},
```
oo13, line 243: Typo: "matrices_yes" -> "matrices_no"

dat/events/neutral/travelling_merchant.lua
```diff
@@ -294,7 +294,7 @@ They get uncomfortably close
       special_virtue    = outfit.get("Machiavellian Virtue"),
    }
    vn.label("special")
-   mm(fmt.f(_("So you're interested in my special services. Quite a bargain might I say. Each services costs only {cost}.")
+   mm(fmt.f(_("So you're interested in my special services. Quite a bargain might I say. Each services costs only {cost}."),
```
oo13, line 297: Fixed a syntax error.

But travelling merchant will cause an error at vn.run() when you have some data matrices.

src/start.c
```diff
@@ -164,7 +164,7 @@ void start_cleanup (void)
  */
 const char* start_name (void)
 {
-   return start_data.name;
+   return _(start_data.name);
```
oo13, line 167: I'm not quite sure the function should return a translated text because it's saved the version.data in the save files.

src/input.c
```diff
@@ -478,7 +478,7 @@ void input_getKeybindDisplay( const char *keybind, char *buf, int len )
          if (key < 0x100 && isalpha(key))
             p += scnprintf( &buf[p], len-p, "%c", toupper(key) );
          else
-            p += scnprintf( &buf[p], len-p, "%s", _(SDL_GetKeyName(key)) );
```
oo13, line 481: "key" is not translatable, because the POT file has no SDL key names.

dat/missions/zalek/neburesearch/neburesearch_02.lua
```diff
@@ -98,7 +98,7 @@ But before I forget, there's some issue..."]]))
     misn.setDesc(fmt.f(_("Take Dr. Mensing to {pnt} in the {sys} system as fast as possible!"), {pnt=homeworld, sys=homeworld_sys}))
     mem.misn_marker = misn.markerAdd(homeworld, "low")
 
-    local c = commodity.new( N_("Dr. Mensing"), fmt.f(N_("You need to bring Dr. Mensing to {sys} but the Empire will assume you have kidnapped her if they scan you!"), {sys=homeworld_sys}) )
+    local c = commodity.new( N_("Dr. Mensing"), fmt.f(_("You need to bring Dr. Mensing to {sys} but the Empire will assume you have kidnapped her if they scan you!"), {sys=homeworld_sys}) )
```
bobbens, line 101: I guess this is an issue due to using a dynamic string. Would probably make sense to make it not dynamic.

src/start.c
bobbens: It should be not. It might be best to have all the `start_` functions be untranslated, and wrap them in `_()` as necessary.

bobbens:
Very thorough, thanks! I'll fix the minor things here and there.

dat/missions/zalek/neburesearch/neburesearch_02.lua
bobbens: Made it static in main and use N_() again.

src/input.c
bobbens: Included the key names and made them translatable in main.

PR #2176, opened 2022-08-12
Fixed some issues of translation. (key names and start_* functions)

oo13:
- Make the key names and the keybind panel translatable.
- start_* function returns an untranslated text now

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -686,6 +686,7 @@ static void menuKeybinds_genList( unsigned int wid )
    /* Create the list. */
    str = malloc( sizeof( char * ) * input_numbinds );
    for (int j = 0; j < input_numbinds; j++) {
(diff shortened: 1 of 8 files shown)


src/options.c
```diff
@@ -783,7 +784,7 @@ static void menuKeybinds_update( unsigned int wid, const char *name )
    /* Remove the excess. */
    keybind = keybind_info[selected][0];
    opt_selectedKeybind = keybind;
-   window_modifyText( wid, "txtName", keybind );
+   window_modifyText( wid, "txtName", _(keybind_info[selected][1]) );
```
oo13, line 787: The keybind panel shows a short description like "Target Next" instead of a mnemonic "target_next".

bobbens:
Looks good, thanks!

PR #2177, opened 2022-08-12
typo correction

ThrosturX:
Corrects `pres` to `pers` as (presumably) originally intended.


bobbens:
Actually, it's pres from "presence", not "pers". Thanks!

PR #2183, opened 2022-08-22
asteroids collide properly with polygons and do spin

nenau-again:
Asteroids have polygons and use them to detect collision with all kinds of weapons.
When an ammo is suspected to hit the asteroid (from a distance check), the polygon is rotated on the fly to match the asteroid's angle, and then collision is tested.
As a bonus, asteroids do spin (their angle changes with time)

I had mentioned the possibility to get them to change speed and spin when hit by bolt weapons, but for speed, it raises the rather complicated question of preventing impacts to push an asteroid out of the field (interaction with thrust potentially complicated)

Rmk: there is a merge conflict, probably due to artwork. Can we solve it there or I should rather solve it on my own branch?

### DIFF
--- a/src/asteroid.c
+++ b/src/asteroid.c
@@ -55,6 +55,7 @@ static int asteroid_creating = 0;
 /* Prototypes. */
 static int asttype_cmp( const void *p1, const void *p2 );
 static int asttype_parse( AsteroidType *at, const char *file );
+static int asteroid_loadPLG( AsteroidType *temp, const char *buf );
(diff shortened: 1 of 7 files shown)


nenau-again:
Ok I guess I have the answer to my remark question just above :P

bobbens:
Looks good, thanks! It would be nicer to use spritesheets instead of rotation to be able to use consistent lighting (or 3D models), but that is probably future work.

PR #2184, opened 2022-08-23
Fixes potential problem with remote controlled shield boosters

ThrosturX:
Adding this simple check will save me from having to bunch of heavy lifting in the multiplayer server code in every single frame and there isn't any harm in it.

### DIFF
--- a/dat/outfits/utility/emergency_shield_booster.lua
+++ b/dat/outfits/utility/emergency_shield_booster.lua
@@ -27,7 +27,9 @@ local function turnon( p, po )
    mem.timer = active
 
    -- apply nice shader effect
-   p:effectAdd("Shield Boost")
+   if p:exists() then
+      p:effectAdd("Shield Boost")
+   end
 
    return true
 end
@@ -41,7 +43,9 @@ local function turnoff( p, po )
    po:clear() -- clear stat modifications
    mem.timer = cooldown
    mem.active = false
-   p:effectAdd("Shield Boost", 1)
+   if p:exists() then
+      p:effectAdd("Shield Boost", 1)
+   end
    return true
 end
 


bobbens:
What issue is this trying to solve? pilot should always exist when doing turnon/turnoff no? I'm trying to think of a more general way of doing the exist check but need to know the exact issue.

ThrosturX:
You might be able to replicate the exact scenario by running a multiplayer server with an equipopts that gives you a shield booster and connecting a client with one and activating it (must have low health on server of course). What ends up happening somehow is that the pilot that receives the remote control command doesn't exist. I haven't looked deeply into it, but the server code already has an exist check where it's used (if I'm right about the entry point, which is not 100% certain since I didn't get that information):

  
     406                 if clplt and clplt:exists() then
     407                     local memo = clplt:memory()._o
     408                     if memo then
     409                         local outno = memo[outf]
     410                         if outno then
     411                             clplt:outfitToggle(outno, on)
     412                         end
     413                     end
     414                 end

bobbens:
Does the fix work? It's quite weird looking at the code you say triggers it, that already has an `:exists()` check, triggering the issue, but adding another `:exists()` check in the outfit fixes it. Makes me think maybe it's getting triggered somewhere else? I would like the outfiton/off stuff to be able to assume that the pilot exists, so if we need to add an exist check, it would make sense to add it somewhere in the C code, however, I'm not sure the best place atm.

ThrosturX:
I haven't been able to reproduce the issue locally :/

bobbens:
OK, my theory is it depends on the order of the pilot spawn, so that the Lua code runs on pilots that don't exist. This happens when you turn off the shield boosters right when the pilot dies. That said, hopefully my latest commits fix this. If not, we'll have to debug again. Closing, thanks!

PR #2190, opened 2022-08-26
reynir.lua: fix off-by-one

reynir:
An off by one error was introduced in 8153ed2a0d2d27bc0fd5d31aad73e83981c849c9. The comment further down says "make sure there are at least two planets" so presumably one is not enough.


bobbens:
Looks good, thanks!

PR #2226, opened 2022-10-23
New proteron ships

V-ktor:
Adds new Poteron ships and changes the existing ones. The Kahan was renamed to Gauss. The Derivative should probably be renamed as well.

List of all Proteron ships:
- Proteron Euler (scout)
- Proteron Derivative (interceptor)
- Proteron Gauss (destroyer)
- Proteron Hippocrates (destroyer/carrier)
- Proteron Pythagoras (cruiser)
- Proteron Watson (carrier)
- Proteron Archimedes (battleship)


bobbens:
I've pushed some minor fixes to [naev:proteron](/naev/naev/tree/proteron) (ship replacement works now). For the Derivative, Powerpeon on discord proposed calling it the Democritus or Dalton. I think the Dalton would be very fitting.

One important issue is that it seems like `artwork/` submodule is not up to date, and the graphics haven't been uploaded to `naev-artwork-production` yet so I couldn't update them and test myself. That would be important to have before merging (and ideally renaming the Derivative too).

bobbens:
Looks great, thanks!

PR #2232, opened 2022-11-08
placeholder now non-existent images

not-pyroman:
the Daltons, or as i like to refer to them as Danger boxes, they're meant to "replace" the derivatives, right? if so that needs an image, but to avoid headaches I'll placeholder these so naev isn't pissy


bobbens:
Looks good, thanks!

PR #2234, opened 2022-11-09
Fixed so that stealth is improved instead of damaged

ispublic:
Machiavellian Fortune has a presumed typo making the stealth timer shorter/faster instead of slower. Changed to make consistent with other Machiavellian intrinsics, which are unambiguously upgrades.


bobbens:
Oops. Looks good, thanks!

PR #2237, opened 2022-11-10
Fix jump detection to also be affected by ew_detect

ispublic:
Jump detection is currently only affected by the ew_jump_detect (or misc_hidden_jump_detect) stats.

Per dev remarks, general detection buffs such as those from sensor array should also improve jump detection. This should implement that.

I'm _reasonably_ certain this won't break hidden jump detection, but you might want to double-check that. I didn't look too hard at how that workss

### DIFF
--- a/src/pilot_ew.c
+++ b/src/pilot_ew.c
@@ -362,10 +362,10 @@ int pilot_inRangeJump( const Pilot *p, int i )
    if (hide==0.)
       return 1;
 
-   sense = EW_JUMPDETECT_DIST * p->stats.ew_jump_detect;
+   sense = EW_JUMPDETECT_DIST * p->stats.ew_jump_detect * p->stats.ew_detect;
    /* Handle hidden jumps separately, as they use a special range parameter. */
    if (jp_isFlag(jp, JP_HIDDEN))
-      sense *= p->stats.misc_hidden_jump_detect;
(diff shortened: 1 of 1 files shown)


src/pilot_ew.c
```diff
@@ -362,10 +362,10 @@ int pilot_inRangeJump( const Pilot *p, int i )
    if (hide==0.)
       return 1;
 
-   sense = EW_JUMPDETECT_DIST * p->stats.ew_jump_detect;
+   sense = EW_JUMPDETECT_DIST * p->stats.ew_jump_detect * p->stats.ew_detect;
    /* Handle hidden jumps separately, as they use a special range parameter. */
    if (jp_isFlag(jp, JP_HIDDEN))
-      sense *= p->stats.misc_hidden_jump_detect;
+      sense *= p->stats.misc_hidden_jump_detect * p->stats.ew_detect;
```
bobbens, line 368: Since it's multiplying, no need to multiply again by `p->stats.ew_detect` here, or you effectively make ew_detect affect hidden jumps twice.

bobbens:
Looks good, thanks! I've modified it as I mentioned in the inline comment to not have ew_detect affect hidden jumps twice. Thanks!

ispublic:
Good shout. I figured I'd probably missed something with the hidden jumps check, so good catch there.

PR #2241, opened 2022-11-13
Add map revealing jump point between Kretogg and NGC-14549 to Kretogg stores

ispublic:
What it says on the tin. Probably best to have this be a reward for building Black Lotus hypergate or somesuch when that's a thing, but should be fine for now.


bobbens:
I'm a bit torn on this. I know we have to do something about this jump, but I'm worried about a temporary solution (map) becoming more permanent. That said, maybe just having the map available at all pirate places would make sense and be easiest.

ispublic:
Fair. Is it worth building a hypergate construction thing for BL right now, then?

bobbens:
I think it is for a bit of parity. I've pulled it, but modified it so the map only appears after chapter 1. Would like to make it so the map is cheaper when bought in Kretogg vs other systems, but haven't bothered yet. I think this could be a good compromise for now.

PR #2245, opened 2022-11-22
Fix some translation issues.

oo13:
1. Distinguished a ship name "Scavenger" in Taiomi campaign from a faction and a character "Scavenger" in Minerva campaign.
2. Made translatable a system name in the note dialogue of the system map.
3. Used the faction name instead of the raw name in Taiomi campaign.


bobbens:
Looks good, thanks!

PR #2246, opened 2022-11-25
Fix some missing translation marks.

oo13:
I would make translatable "#%c%s>#0 %s" in player.c since "#%c%s>#0 %s" in nlua_pilot.c is translatable and it's useful to adjust the spacing.


bobbens:
Looks good, thanks!

PR #2247, opened 2022-11-25
Remove erroneous lua script call

ispublic:
The outfit currently is defined by the poi_data script, which is not relevant to its function. This tricks the game into thinking the outfit is activated by the script and causes it not to function.

Removes the <lua> line, fixing the outfit.

### DIFF
--- a/dat/outfits/unique/veil_of_penelope.xml
+++ b/dat/outfits/unique/veil_of_penelope.xml
@@ -12,7 +12,6 @@
   <cpu>0</cpu>
  </general>
  <specific type="modification">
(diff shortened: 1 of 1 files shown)


UncombedCoconut:
Yeah, that doesn't smell right. Thanks!

bobbens:
Thanks for the PR! We probably have to go about this another way. With the patch, the outfit does work well again, however, you can't buy it for encrypted data matrices anymore, only cold hard credits.

bobbens:
OK, solved this by adding a check for a variable `inactive` that allows a Lua outfit script to forcibly make an outfit inactive.

PR #2252, opened 2022-12-13
Correct capitalization of Vendetta.ogg

not-pyroman:
Filename and file listed here have different capitalization


not-pyroman:
I'm not sure about windows as I've noticed it be both picky and not picky about capitalization, but I know Unix and Linux based systems usually are more picky

PR #2253, opened 2022-12-15
Red Star descriptions and Particle Beam availability

Ianator:
- Rewrote descriptions for Red Star Small, Medium and Large Cargo Hulls.
- Added Particle Beam to Za'lek T1 tech pool, allowing them to be purchased outside Zied.


bobbens:
Tweaked it slightly as large red star hulls can't actually turn away torpedoes, only missiles atm. Thanks!

PR #2255, opened 2022-12-18
Docs/Manual/Ships update

Ianator:
Updated ships plugin guide with info on .XML fields.


bobbens:
Looks good, thanks! Tweaked and updated things a bit more!

PR #2256, opened 2022-12-21
Update pronoun in crimelord mission

warp-core:
The crimelord is mostly referred to using gender neutral pronouns, but one instance of `He` remained. Changed that to `They`.


bobbens:
Looks good, thanks!

PR #2260, opened 2022-12-28
Soften Autonav sound

hollunder:
I haven't really contributed via Github so far, so not sure I am doing this correctly. This branch merge commit is a bit irritating.


hollunder:
Bobbens discovered and fixed a bug, which rendered this obsolete.

PR #2273, opened 2023-01-01
Add missing choices in the search for Cytnhia

hollunder:


bobbens:
Looks good, thanks! I reworded it a bit as accept/decline seems a bit weird in this case.

PR #2275, opened 2023-01-03
Dev Manual: How to write star systems

Ianator:


bobbens:
Looks good, thanks! Added some minor tweaks and corrected the /ssys/ path.

PR #2280, opened 2023-01-08
Fix missing translation marks.

oo13:


bobbens:
Looks great, thanks again!

PR #2287, opened 2023-02-01
Fix a missing translation mark.

oo13:
Fix a missing translation mark in the mission shark07.

The branch 0.10.x has no problem.


bobbens:
Thanks for the patch! The issue is that the outfit is being stored as a string, and not an outfit. Instead of adding `_()`, I've changed it to be an outfit, so `tostring` now returns the translated version and thus `fmt.reward` does too. I've fixed it in d72444eeae6a255b3126b5094a594368a571cddd . Thanks!

PR #2290, opened 2023-02-06
Fix some errors in the mission Seek And Destroy.

oo13:
- The mission computer indicates the class of the target ship, instead of the ambush's.
- Fix a script error.


UncombedCoconut:
Thanks!

PR #2296, opened 2023-02-17
Update ships.md

Ianator:
Updated the `/slots/prop` section after discovering exactly how Engine, Hull and Core Systems stats are defined.


bobbens:
Looks good, thanks!

PR #2298, opened 2023-02-18
Fix two messages in the mission New Europe Refugees.

oo13:
1. Fix a translation problem. 
2. Fix a typo.


UncombedCoconut:
Thanks!

PR #2300, opened 2023-03-01
Spob proofreading

Ianator:
Updates to spob descriptions for Tau Ceti Station (Tau Ceti), P-3030 (NGC-10180), P-2317 (NGC-10907) and Vorca (Kretogg).


bobbens:
Looks good, thanks!

PR #2313, opened 2023-03-30
Add a space after the outfit class.

oo13:
I use a NL instead of a space because the system map has enough vertical space.


bobbens:
Looks good, thanks!

PR #2314, opened 2023-03-31
Fix a bug that the bioship can lose fuel, shield, and energy.

oo13:
The bioship can lose fuel, shield, and energy on landing. It causes a real problem only when the player lands on a planet where has no refuel facility, but the player can notice the shield and energy bar on the GUI and the remaining fuel on the system map.


bobbens:
Looks good, thanks!

PR #2324, opened 2023-04-15
Improve the smart missile algorithm.

oo13:
The current algorithm of AMMO_AI_SMART uses a coarse estimation of the impact time, so a missile can make mysterious U-turn in some conditions, such as a head-on between a low speed missile and a high speed ship.

This PR solves it with a sort of a proportional navigation.

We can correct the problem with a more precise estimation of the time, however, we can use a proportional navigation, which has no need to estimate the time. The navigation results almost same as the more precise version.

The proportional navigation tries to make the quadrature of the relative speed to zero. It works only in a situation that the missile follows the target and the missile speed is higher than the target's, so the PR handles the other situations as the special cases.

I tested it by Za'lek Hunter and Unicorp Caesar IV.

I also created a code to estimate the more precise time of the impact. For your reference:
```diff
          if (w->outfit->u.lau.ai == AMMO_AI_SMART) {
 
             /* Calculate time to reach target. */
+            /*
+              v = (Xp, Yp), p->solid->vel = (Vpx, Vpy), w->solid->vel = (Vwx, Vwy)
+               (Xp + t*Vpx)**2 + (Yp + t*Vpy)**2 = (Vwx*t)**2 + (Vwy*t)**2
+            */
             vec2_cset( &v, p->solid->pos.x - w->solid->pos.x,
                   p->solid->pos.y - w->solid->pos.y );
-            t = vec2_odist( &v ) / w->outfit->u.lau.speed_max;
-
+            const double a = vec2_odist2(&p->solid->vel) - vec2_odist2(&w->solid->vel);
+            const double b = vec2_dot(&p->solid->vel, &v);
+            const double c = vec2_odist2(&v);
+            t = w->timer;
+            if (a == 0.0) {
+                if (b != 0.0) {
+                    t = -c / (2 * b);
+                }
+            } else {
+               const double b2_ac = b * b - a * c;
+               if (b2_ac >= 0) {
+                   t = (-b - sqrt(b2_ac)) / a;
+                   /* Another solution is a negative value or greater than t */
+               }
+            }
+            if (t < 0) {
+               /* The missile is unreachable. */
+               /* t is a negative value when a > 0 and b > 0. */
+               t = w->timer;
+            }
             /* Calculate target's movement. */
-            vec2_cset( &v, v.x + t*(p->solid->vel.x - w->solid->vel.x),
-                  v.y + t*(p->solid->vel.y - w->solid->vel.y) );
-
+            vec2_cset( &v, v.x + t*p->solid->vel.x,
+                  v.y + t*p->solid->vel.y );
             /* Get the angle now. */
             diff = angle_diff(w->solid->dir, VANGLE(v) );
          }
```

### DIFF
--- a/src/weapon.c
+++ b/src/weapon.c
@@ -323,7 +323,7 @@ static void think_seeker( Weapon* w, const double dt )
    double diff;
    Pilot *p;
    vec2 v;
-   double t, turn_max, d, jc, speed_mod;
+   double turn_max, d, jc, speed_mod;
 
    if (w->target == w->parent)
       return; /* no self shooting */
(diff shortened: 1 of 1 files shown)


bobbens:
Seems to work great, thanks!

PR #2326, opened 2023-04-25
Add ja translations to the desktop files.

oo13:
"org.naev.Naev.desktop" might be an auto-generated file, but the change is harmless.


bobbens:
Looks good, thanks!

PR #2327, opened 2023-05-01
Fix a bug that may cause SIGSEGV in naev.versionTest().

oo13:
free() in semver_free() may raise SIGSEGV if naev.versionTest() is set a parameter that results semver_is_valid() returns false.


bobbens:
Looks good, thanks!

PR #2332, opened 2023-05-16
typo, missing right parentheses, `(opened with {infokey},` should be `(opened with {infokey}),`

flyxyz123:


bobbens:
Looks good, thanks!

PR #2341, opened 2023-07-10
Fix some issues around "sirius_awakening".

oo13:
- Fix some errors in Lua scripts.
- Fix the selecting logic for the special events of the derelicts.
- Add translation contexts.
- Make the desc_extra of the ship translatable and add a newline in equipment_shipStats().


bobbens:
Looks good, thanks!
BTW, no need for portrait.getFullPath, the return values from vnimage were being used the wrong way.

PR #2347, opened 2023-07-15
Fix a bug that causes an extra derelict event after a lucky find.

oo13:
You experience an extra derelict event soon after finding a full of fuel in a derelict. This PR fixes it.


bobbens:
Nice catch, thanks!

PR #2352, opened 2023-07-28
Fix a bug that a lua code ai.face(-jmp:angle()) might cause a reverse rotation.

oo13:
angle_cleanup() should return a value in the range [0, +2PI) instead of (-2PI, +2PI) if angle_diff() is supposed  to return (-PI, +PI).


bobbens:
Looks good, thanks!

PR #2353, opened 2023-07-29
Improve autonav in instant jump.

oo13:
The high speed ship with instant jump might pass through the target jump point when enabling autonav, because the ship has relative low turn rate and the current autonav try to adjust the head of the ship after it's in the jump area.

This PR makes the ship turn in the direction of the jump out in advance and practically fixes the problem. (Technically, you can create a plugin that provides an outfit with very low turn rate and a player equips it and a reverse thruster into their ship, it can fail to jump.)

How to reproduce the problem:
1. Load [this save game](https://github.com/naev/naev/files/12207209/autosave.gz).
2. Set autonav to jump the Qex system.
3. Take off.
4. You ship passes through the jump point unless you are very lucky.

### DIFF
--- a/dat/autonav.lua
+++ b/dat/autonav.lua
@@ -345,12 +345,54 @@ function autonav_jump_approach ()
    end
 end
 
+-- More approaching a jump point and turning in the direction of the jump out.
+local function autonav_instant_jump_final_approach ()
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good, thanks!

PR #2355, opened 2023-07-31
PoI submission

Hraelzyr:
A press guidelines PoI for the Proteron.


bobbens:
Looks great! I've added double quotes when the SAI is talking for consistency with other events/missions. Thanks!

PR #2358, opened 2023-08-01
Language fixes to PoIs

Hraelzyr:
Fixes to the proteron_press and a minor fix to the proteron_agent

### DIFF
--- a/dat/missions/neutral/poi/proteron_agent.lua
+++ b/dat/missions/neutral/poi/proteron_agent.lua
@@ -22,7 +22,7 @@ return function ( mem )
          if faction.known( "Proteron" ) then
             vn.na(_([[You enter the ship and make way to the bridge. The entire ship is oddly quiet as you pass through, with no signs of life. You reach the bridge and find the ship's systems have just enough energy left to power up, letting you jack in.]]))
          else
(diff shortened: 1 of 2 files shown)


bobbens:
It's more of a gameplay mechanic than anything else. The players are sort of trained to think POI == reward, so giving some sort of reward is important. Data matrices are literally a sort of "currency" from POI to be able to get special stuff from the traveling merchant, so giving them would be most natural, instead of having them test their chances at getting the POI they want.

Originally I meant to have each POI give a unique accessory or something like that, but then thought that as the number of POI increases and such, the probability of a player getting something useful to them or that they want decreases and such. By introducing data matrices as an intermediate currency, the player has more control about what they want and don't have to fight against probability. Now, there's some exceptions like the Guardian Unit outfit, although chances are that could also probably benefit from giving a data matrix instead, and making it available for sale.

I guess this would be a bit different if you, for example, have the drunkard event at doeston or whatever that can kick start the POI so that the player could have a reliable way of triggering it and get the reward, which would more easily be an accessory in this case.

Having it be a monetary reward would also work, but I do sort of feel those are probably the worse rewards from the POI.

Hraelzyr:
Alright, I'll make the edit, thought I'd prefer to do it in a separate PR.

PR #2359, opened 2023-08-01
Proteron campaign and lore updates

Hraelzyr:
Gliese and GM have been shuffled around a bit in preparation for the strandees mission.
Station naming convention shifted to Castellum/Castrum for SPA.


Hraelzyr:
That would work if the player can survive Arandon when they leave

Hraelzyr:
As for the other aspect, give me a minute, working on the data matrix thing.

bobbens:
> That would work if the player can survive Arandon when they leave

The main issue atm is Gliese, it has 27.2 volatility. Arandon only has 15.6, and Behar has 3.1. So the player is most likely to get killed in Gliese, and if they can survive that, they can survive Arandon.

Hraelzyr:
Oh no, that wasn't my intention! I'll change it so that the player won't die in Gliese

Hraelzyr:
The problem would be the player estimating the volatility before entering. If it's jump in from Behar, which isn't very volatile if at all, they may be taken unawares by the jump exit. And there'll be no escape either. Perhaps I should remove the volatility altogether, because Gliese Minor has no volatility.

Hraelzyr:
Say it's a pristine pocket in the Nebula, something like that.

bobbens:
> Say it's a pristine pocket in the Nebula, something like that.

I would probably say that for Gliese Minor, for Gliese itself, just make it similar to Behar, that should not be too much of a problem no?

Hraelzyr:
That's fine. Give me a bit of time, I'll make the edit

Hraelzyr:
Please merge this unless you have any issues to post. I'll create a separate branch and PR for the strandees mission so that it can go ahead on its own track.

bobbens:
Some comments:
1. Should the Proteron have presence in Gliese Minor? That means it will spawn ships, and they will be hostile to the player (most likely). I thought the system would only have the crashlanded capital ship and no functional ships (or maybe a small one or two, but not doing patrols and just kept for important things). I would think that they probably should not, so not giving any presence to the spob would make sense (or add a minus virtual spob for removing proteron presence to the system).
2. Related to the faction, and it's not much of a problem without ships, but it may make sense to use a different faction instead of the Proteron for the spob. Mainly, because otherwise, the player will be hostile by default and it should show up as hostile, and although the player advances this campaign and such, given that they have no contact with the "real" proteron, it's not like faction standing changes here should propagate over there. Creating a small static faction would probably make sense, and just "fake" the standing with the mission / event scripts would probably be a good option.

The rest looks good and shouldn't be an issue though.

Hraelzyr:
To respond to both matters:

There shouldn't be a Proteron presence. I thought I had committed the change with the "Proteron Unpresence" flag. If not, I'll commit and push it.

The planet should NOT be landable. Part of the mission will be a special landing and comms script that does not check faction standing. I don't mind changing the faction here, but them being hostile is intentional. If we can get rid of the purple warnings on the map for the planet, we should be good.

As for it not reflecting faction standing with the SPA, I wanted some small broader consequence for aiding/hurting them

bobbens:
Double checked. Looks good, thanks!
When working on the mission I think it may be a good idea to try to spruce up the description a bit more of Protera Nova as it is a bit bare.
Also, regarding the faction I still think it would probably be best to create a fake proteron faction that would be just for the spob, giving fine grained control of this system without affecting the "normal" proteron. It could use the same names and everything, just have independent standing. You can still give faction as a reward or whatever, but it'll probably be a bit messy otherwise with the spob most likely constantly be hostile to the player. However, you can see when working on the missions.

PR #2361, opened 2023-08-02
Adding to AUTHORS

Hraelzyr:
I've added myself and the specific work I've done so far to the file. (I'm aware this isn't a language fix, I just have only so many branches)


bobbens:
Looks good, thanks!

PR #2362, opened 2023-08-02
Proteron landing messages changed.

Hraelzyr:
Old messages used short form, seemed less formal. Changed it to be consistent with adoption of SPA and made it more formal.


bobbens:
Looks good, thanks!

PR #2365, opened 2023-08-08
Improve autonav when boarding a derelict.

oo13:
With autonav, it takes a long time for fast ships to board a derelict because autonav estimates too much value as the distance to stop the ships. This PR makes autonav estimate smaller and more relevant value.

How to estimate the distance to stop in this PR:
```
distance_to_stop = velocity**2 / (2 * ship_thrust)

because
mass * ship_thrust * distance_to_stop = 1/2 * mass * velocity**2
```

How to estimate the distance to stop in the current version:
```
10 * distance_to_stop = velocity * (10.84 * ship_speed / ship_thrust - 10.82)
```
The equation doesn't fully consider the slowing down of the ship for breaking.

I don't grasp the meanings of the coefficients 10.84 and -10.82, but it seems to work fine without the margins.

I tested it with some light, medium, and heavy ships. The light ships have room to reduce the time to stop yet for some reasons I don't know.

### DIFF
--- a/dat/autonav.lua
+++ b/dat/autonav.lua
@@ -306,15 +306,16 @@ local function autonav_approach_vel( pos, vel, radius )
    local pp = player.pilot()
    local turn = math.rad(pp:turn()) -- TODO probably change the code
 
-   local timeFactor = math.pi/turn + pp:speed() / pp:thrust()
+   local rvel = vel - pp:vel()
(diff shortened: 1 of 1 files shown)


bobbens:
IIRC, the coefficients were tuned by hand by @nenau-again for the PD controller and they found them to work best.
Thanks for the PR, merged!

PR #2366, opened 2023-08-11
Fix a bug that makes the GUI buttons unclickable.

oo13:
"Set formation" buttons in silm and brushed GUI were not clickable.


UncombedCoconut:
Thanks!

PR #2367, opened 2023-08-14
Fix some typos in the mission "Capricorn Pirates."

oo13:
In the current version, eliminating the pirates in "Capricorn" seemed to cause nothing. The PR fixes it.

You mean the pirates lock down the system "Capricorn" and removing "restricted" tag from the SPOB "Durea", don't you?


bobbens:
Sorry about that, a bit too much copy-paste coding. Thanks for the fixes!

PR #2368, opened 2023-08-14
Fix a bug where the trading panel doesn't show commodities of the escorts.

oo13:
The trading panel didn't show the escorts' commodities if the player ship didn't have them.


bobbens:
Nice catch, thanks!

PR #2372, opened 2023-08-17
Wrong reward message; said agent instead of press

Hraelzyr:
Proteron Press PoI had a semantic mistake in one line, fix is made.


bobbens:
Nice catch, thanks!

PR #2375, opened 2023-08-25
Just some sanity checks to the blockade code

Hraelzyr:
Will change it in earnest as the main campaign for Proteron happens, for now just standing < 0 = hostile


dat/events/proteron/blockade.lua
```diff
@@ -36,14 +36,15 @@ local function spawn_fleet( pos )
...
+   local stand, _str = faction.get("Proteron"):playerStanding()
    for k,s in ipairs(ships) do
       local leader = plts[1]
       local p = pilot.add( s, "Proteron", pos, nil, {ai="guard"} )
       local aimem = p:memory()
       aimem.enemyclose    = 10e3
       aimem.guarddodist   = 10e3
       aimem.guardreturndist = 15e3
-      p:setHostile(true) -- TODO change this when something can make them non-hostile
+      p:setHostile(stand<0)
```
bobbens, line 47: Strictly speaking, I don't think this is necessary, because this is the default behaviour.

Hraelzyr: Ah, I see, fair enough

dat/events/proteron/blockade.lua
```diff
@@ -71,7 +72,8 @@ end
 
 function heartbeat( proteron_blockade )
    local pp = player.pilot()
-   for k,p in ipairs(proteron_blockade) do
+   local stand, _str = faction.get("Proteron"):playerStanding()
+   for k,p in ipairs(proteron_blockade) and stand < 0 do
```
bobbens, line 76: I'm guessing you wanted an `if` statement here.

bobbens:
I've fixed it up and polished a bit and merged it. Thanks!

dat/events/proteron/blockade.lua
Hraelzyr: I was trying to use the short-circuit fancy stuff in lua

Issue #2009, opened 2022-01-14
Compute autonav jump routes considering travel time


bobbens:
Currently, the game only considers number of jumps when computing shortest path between systems, it should consider total travel time instead. This will involve having to compute the Djikstra graph on a jump point level instead of a system level.

bobbens:
Closed by #2376

PR #2376, opened 2023-08-26
Jump routes considered the distance in the systems.

oo13:
This PR makes the map panel create the jump routes considered the distance in the systems. It's similar #2009, but priorizes less jumps over travel time, since the jump consumes fuel.

This PR also fixes some bugs in A_rm() and map_findDistance().


bobbens:
Great work like always, thanks!

PR #2377, opened 2023-08-26
Add a context to the button label "Approach."

oo13:
This PR allows to translate the key function and the UI button "Approach" into different texts.


bobbens:
Looks good, thanks!

PR #2379, opened 2023-08-26
Proteron flavoured landing-messages

Hraelzyr:
3 are loyalty based, 1 is just standard stuff to give a respite from loyalty oaths to pilots.


bobbens:
Looks good, thanks!

PR #2380, opened 2023-08-30
Enums instead of strings

Hraelzyr:
Start of modifications to input.c to make it use enums instead of strings

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -159,6 +159,7 @@ Naev's translation is handled with gettext. (It's custom, but C and Lua code can
 
 When content like missions is updated, new translatable text must be made available to Weblate.
 The key manual step is to regenerate the `po/naev.pot` file (`meson compile naev-pot` in the build dir) and commit it.
+To avoid merge conflicts, it is recommended to not include updated `po/naev.pot` in a pull request that isn't exclusively about translation.
 
 Under the hood: `po/POTFILES.in` is a catalog of files that may have translatable text.
 We keep it synced using pre-commit hooks (or manually: `meson compile potfiles`).


src/input.c
```diff
@@ -40,93 +40,173 @@
...
+
+/* Description of each key semantic type */
+const char *keybind_info[KST_PASTE+1][2] = {
    /* Movement */
-   { "accel", N_("Accelerate"), N_("Makes your ship accelerate forward.") },
-   { "left", N_("Turn Left"), N_("Makes your ship turn left.") },
-   { "right", N_("Turn Right"), N_("Makes your ship turn right.") },
-   { "reverse", N_("Reverse"), N_("Makes your ship face the direction you're moving from. Useful for braking.") },
-   { "stealth", N_("Stealth"), N_("Tries to enter stealth mode.") },
+   { N_("Accelerate"), N_("Makes your ship accelerate forward.") },
```
bobbens, line 127: Would be nicer to probably use some #define magic and explicitly include the KST_ code instead of depending on the order of the enum, which will likely cause issues sometime down the line.

Hraelzyr: Yeah, will fix later on in the draft; prototype bind_desc(enum, short_string, long_string)

Hraelzyr: Bit of a time shortage, so started draft with partially complete code. Will first get this thing working as expected, and then make sanity changes.

bobbens: OK, sounds good. Thanks!

Zireael07:
Why close?

bobbens:
OK!

PR #2382, opened 2023-09-09
Fixed two bugs in the mission "Bounty."

oo13:
This PR fixes the following bugs:
- The reason isn't shown in Japanese language.
- The mission isn't finished if you kill the target.


bobbens:
Nice catches! Thanks!

PR #2383, opened 2023-09-09
Fixed a typo.

oo13:


bobbens:
Thanks! I've changed it to associate in the latest version which is a bit more fitting in the mafia-esque theme of the black lotus.

PR #2389, opened 2023-09-12
Edited "Sirius Awakening" event

bene-dictator:
There's some repetitive bits and minor errors in here that I've tried to fix. Also, the end is clunky.

~~I don't have to just review ES work : )~~


bobbens:
Great edits, thanks!

PR #2390, opened 2023-09-12
Fixed segfault that occurs when a snapshot gets deleted.

sigus:
Steps to reproduce segfault:
1. Run Naev
2. New Game (do not select Load Game)
3. Create more than one snapshot (autosave and backup are sufficient)
4. Delete a snapshot from the small menu


bobbens:
Nice catch, thanks!

PR #2391, opened 2023-09-12
Some strings handleling rework for easier translation.

nicolas-arnaud:
~~I changed first the conditional ships and outfit coditional string to merge them all in one translation string.~~
Then i reworked the ships and outfits statistics display to remove format things from translation and it looks also cleaner i think.


bobbens:
I'm not sure this is a good idea, at least in the current implementation.

First of all, the cond+condstr framework, allows for arbitrary conditionals, while you have changed it so it's hard-coded for reputation + faction, which significantly lowers the flexibility of the system. That said, if you really wanted to, you could, for example, have a custom xml string that doesn't replace cond+condstr, but is perpendicular. For example something like:
```
    <condrep name="Empire" standing="70" />
```
This would naturally be turned internally to be equivalent to:
```
    <cond>faction.playerStanding("Empire") &gt;= 70</cond>
   <condstr>You need at least 70 reputation with the Empire.</condstr>
```
The load script would have to check to make sure either "cond+condstr" or "condrep" (or none) are defined.

Secondly, I don't think I would include the colon (':') in the shipstats names. If anything, I would leave it as as and have the string `"%s#%s%+g%% %s#0"` or the likes be translatable and include the colon by default or the likes. I'm not against changing the order, and it could be a bit more intuitive, but I do think it should be done with the printf-style string instead of modifying the shipstats names.

nicolas-arnaud:
Yeah i see about the other way to do. I did it knowing that's not the best way to do it x) About the fact that it allow to custom each faction message, I don't think it's harder like this, The faction name is still taken from the translation and and as it's a meta think i dont thing a message different per ship is a good thing so rn it's just easier as long as the message is the same for each one.

nicolas-arnaud:
About the stats, they are just not included alone rn in the po file. Keeping printf format is just bug prone for failed translations and i don't see any interest to it i guess as spliting stats description and value just get ride of any localization issue for the format except the colon which is keept in translation.

src/outfit.c
```diff
@@ -2089,11 +2091,11 @@ static void outfit_parseSFighterBay( Outfit *temp, const xmlNodePtr parent )
    temp->summary_raw = malloc( OUTFIT_SHORTDESC_MAX );
    l = 0;
    SDESC_ADD( l, temp, "%s", _(outfit_getType(temp)) );
-   SDESC_COND_COLOUR( l, temp, _("\n%.0f CPU"), temp->cpu );
-   SDESC_COND( l, temp, _("\n%.0f Mass"), temp->mass );
-   SDESC_COND( l, temp, _("\n%.1f Seconds Per Launch"), temp->u.bay.delay );
+   //SDESC_COND_COLOUR( l, temp, _("\n%.0f CPU"), temp->cpu );
```
bobbens, line 2094: Why are these commented out?

src/outfit.c
```diff
@@ -1830,53 +1828,56 @@ static void outfit_parseSLauncher( Outfit* temp, const xmlNodePtr parent )
...
    else
-      SDESC_COND( l, temp, _("\n%.0f Speed"), temp->u.lau.speed );
+      SDESC_COND( l, temp, "\n  %s %.0f", _("Initial Speed:"), temp->u.lau.speed );
    if (!(temp->u.lau.thrust > 0. && temp->u.lau.speed > 0.))
-      SDESC_ADD(  l, temp, _("\n%.0f Maximum Speed"), temp->u.lau.speed_max );
-   SDESC_ADD(  l, temp, _("\n%.1f Seconds to Reload"), temp->u.lau.reload_time );
-   SDESC_COND( l, temp, _("\n%.1f EPS [%.0f Energy]"), temp->u.lau.delay * temp->u.lau.energy, temp->u.lau.energy );
-   SDESC_COND( l, temp, _("\n%.1f MJ Armour [%.0f%%]"), temp->u.lau.armour, temp->u.lau.dmg_absorb*100.);
-   SDESC_COND( l, temp, _("\n%.0f%% Jam Resistance"), temp->u.lau.resist * 100. );
+      SDESC_ADD(  l, temp, "\n  %s %.0f", _("Maximum Speed:"), temp->u.lau.speed_max );
```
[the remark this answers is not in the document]
nicolas-arnaud: I inversed how it was originaly done. 
Instead of space only the armour and shield damages bonus i keeped them aligned and spaces de others.
I guess there should be major and minor stats and handle order and spacing according. I'll dig into statistics to classify them.

src/outfit.c
nicolas-arnaud: That's a fail.

src/outfit.c
```diff
@@ -1450,31 +1450,31 @@ static void outfit_parseSBolt( Outfit* temp, const xmlNodePtr parent )
...
-   SDESC_ADD(  l, temp, _("\n%.0f Speed"), temp->u.blt.speed );
-   SDESC_COND( l, temp, _("\n%.1f second heat up"), temp->u.blt.heatup);
-   SDESC_COND( l, temp, _("\n%.1f Degree Dispersion"), temp->u.blt.dispersion*180./M_PI );
+   SDESC_COLOURT(  l, temp, "\n%s %.0f%%", _("Armour Damages:"), 100., darmour*100. );
+   SDESC_COLOURT(  l, temp, "\n%s %.0f%%", _("Shield Damages:"), 100., dshield*100. );
+   SDESC_COND( l, temp, "\n  %s %.0f%%", _("Knockback:"), dknockback*100. );
+   SDESC_COND_COLOUR( l, temp, "\n  %s %.0f", _("CPU:"), temp->cpu );
+   SDESC_COND( l, temp, "\n  %s %.0f", _("Mass:"), temp->mass );
+   SDESC_ADD(  l, temp, "\n  %s %.0f%%", _("Penetration:"), temp->u.blt.dmg.penetration*100. );
+   SDESC_COND( l, temp, "\n  %s %.2f [%.0f /sec]", _("Damages:"), temp->u.blt.dmg.damage * (double)temp->u.blt.shots,
```
oo13, line 1462: "sec", "deg", "/sec", "min:", and "duration" in the formats should be translatable.

oo13:
As a translator, I hope to have a free hand to translate the format strings... in fact, the current Japanese translation of Naev uses some layout space characters in the formats, instead of U+0020 SPACE. It's not necessary, but more legible.

nicolas-arnaud:
Rewrote the outfit part so, Separated the Stats name, units and format.
Any review is welcome.

src/outfit.c
nicolas-arnaud: :heavy_check_mark:

src/outfit.c
```diff
@@ -2023,15 +2094,24 @@ static void outfit_parseSAfterburner( Outfit* temp, const xmlNodePtr parent )
...
-   SDESC_ADD( l, temp, _("\n%.0f Maximum Effective Mass"), temp->u.afb.mass_limit );
-   SDESC_ADD( l, temp, _("\n%.0f%% Thrust"), temp->u.afb.thrust + 100. );
-   SDESC_ADD( l, temp, _("\n%.0f%% Maximum Speed"), temp->u.afb.speed + 100. );
-   SDESC_COND( l, temp, _("\n%.1f EPS"), temp->u.afb.energy );
-   SDESC_COND( l, temp, _("\n%.1f Rumble"), temp->u.afb.rumble );
+
+   SDESC_ADD( l, temp, "\n#r%s#0", _("Activated Outfit") );
+
+   l = os_printD( temp->summary_raw, l, temp->cpu,
+                  os_opts(_("CPU"), _("PFLOP"), 1, 0, 1, 0) );
```
bobbens, line 2101: It may be better to have some global static variables for these common os_opts that are shared with outfit types.

src/outfit.c
```diff
@@ -2957,3 +3041,83 @@ void outfit_free (void)
...
+
+#define ItoL(i) OUTFIT_SHORTDESC_MAX-i
+
+/**
+ * @brief Print an outfit double statistic line
+ *
+ */
+int os_printD( char *buffer, int i, double value, t_os_stat opts)
+{
+   if (opts.hide_zero && value == 0)
```
bobbens, line 3065: probably better to do `fabs(value) < 1e-5` or something like that instead of `value == 0`

src/outfit.c
```diff
@@ -2957,3 +3041,83 @@ void outfit_free (void)
...
+                     value > opts.color_threshold ? "#g" :
+                     value < opts.color_threshold ? "#r" : "");
+   i += scnprintf(buffer + i, ItoL(i), p_("outfitstats", "%s: %d%s"), opts.name, value, opts.unit);
+   i += scnprintf(buffer + i, ItoL(i), "#0");
+   return i;
+}
+
+int os_printD_range( char *buffer, int i, double minValue, double maxValue, t_os_stat opts)
+{
+   if (opts.hide_zero && maxValue == 0)
```
bobbens, line 3100: Again, probably better to use fabs with doubles than maxValue==0

src/shipstats.c
```diff
@@ -149,36 +154,36 @@ static const ShipStatsLookup ss_lookup[] = {
...
-   A__ELEM( SS_TYPE_A_DISABLE,            disable,             N_("MW Disable") ),
-
-   A__ELEM( SS_TYPE_A_CPU_MAX,            cpu_max,             N_("CPU Capacity") ),
-   A__ELEM( SS_TYPE_A_ENGINE_LIMIT,       engine_limit,        N_("Engine Mass Limit") ),
-   A__ELEM( SS_TYPE_A_FUEL_REGEN,         fuel_regen,          N_("Fuel Regeneration") ),
-   A__ELEM( SS_TYPE_A_ASTEROID_SCAN,      asteroid_scan,       N_("Asteroid Scanner Range") ),
-   A__ELEM( SS_TYPE_A_NEBULA_VISIBILITY,  nebu_visibility,     N_("Nebula Visibility") ),
+   A__ELEM( SS_TYPE_A_THRUST,             thrust,              N_("Thrust"),              N_("t") ),
+   A__ELEM( SS_TYPE_A_TURN,               turn,                N_("Turn Rate"),           N_("deg/sec") ),
+   A__ELEM( SS_TYPE_A_SPEED,              speed,               N_("Maximum Speed"),       N_("m/s") ),
```
bobbens, line 159: Might make sense to have the unit stuff be set up as const strings instead of repeating.

bobbens (commented):
What did you change in Artwork? It shows a lot of polygon changes. Did you just not pull the latest version of artwork?
Also, I would recommend not committing `po/naev.pot`, we can update that later and it'll just do merge conflicts.

nicolas-arnaud:
About the artworks, i well updated the submodule and git diff don't show me anything locally so i guess that because of the .gitattributes missing. About the naev-pot, i gonna restore it. Is it better to never commit it? (so the README should be edited)

bobbens:
> About the naev-pot, i gonna restore it. Is it better to never commit it? (so the README should be edited)

It's probably best to do it in a separate PR or use weblate. The main reason being that it will potentially cause a ton of merge conflicts unrelated to the rest of the PR. I've modified the readme to include that.

bobbens:
I've started doing some edits in the units branch (https://github.com/naev/naev/tree/units) including fixing a crash I was getting. Once I polish it a bit more I'll merge.

bobbens:
Been working on it a bit more, and it seems like it'll have to be extended more to also cover the equipment stuff (`src/equipment.c`) that uses hardcoded units atm

bobbens:
I've done quite a lot of changes, and thought it was ready to merge into main. Thanks!
Probably still rough on the edges (especially translation-wise), but it looks much better than before.

PR #2395, opened 2023-09-16
Fixed neburesearch_00

sigus:


bobbens:
Oops! Nice catch, thanks!

PR #2396, opened 2023-09-20
WIP Use GLES3 for rendering

ids1024:
Worth considering, for portability (mobile, web, and layering on other APIs using ANGLE).

This seems to be basically working. Various shaders and gl calls still need to be fixed.


bobbens:
It might be a good idea to split this into two. Stuff like remove GLSL uniform initializers and some GL_ constant renaming could be merged into main already. Things like changing the context to ES and stuff that needs workarounds (dust shader), could be done in a separate PR.

ids1024:
Yeah. A lot of the changes here could be merged without issue. With the OpenGL version currently in use.

After looking into it more, it seems desktop platforms do not necessarily have OpenGL ES available, so for full compatibility this needs to use desktop OpenGL or OpenGL ES depending on what is available. Which should be fine, though they won't end up being equally well tested. (But if it defaults to OpenGL, that's strictly an improvement over the current state where platforms with only OpenGL ES aren't supported.)

Potentially for the best compatibility and consistency, using GLES on ANGLE is ideal. Though I couldn't get that working with SDL when I tried it and it's a quite large dependency to build and bundle.

bobbens:
I would probably stay away from angle as a dependency, it's a bit too much I think. I guess the main options would be either (1) OpenGL + OpenGL ES (conf option + autodetect) or (2) Vulkan.
Although Vulkan is probably a bit more work, it's probably more future-proof. Not sure if it still gives us geometry shader issues though. The main issue is probably that support is still not that widespread so we can't jump to it for the next some years. People already complain that we don't support OpenGL 2.1.

bobbens:
Closing as out of date and will be replaced with Rust stuff and maybe wgpu anyway.

PR #2397, opened 2023-09-20
Add Red Rackham Mission

JackStorm448:
Add a simple 3-stage POI mission following the story of a legendary pirate's missing treasure, also fixes a typo in two other POI missions.


bobbens:
Thanks! I've converted it from 4 space tabs to 3 space tabs and fixed some warnings from pre-commit (just unused fmt.f parameters). It looks good but the ending feels a bit anti-climatic. I'll toss some ideas on discord. Thanks!

PR #2401, opened 2023-09-21
Improve the aiming of the guided missile turret.

oo13:
The aiming for the guided missile turret is the same as for the bolt turret. It's not quite good because the missile is accelerated the launched direction and it goes a roundabout way. For example:
Peek 2023-09-20 22-10

This PR makes the guided missile turret aim at the estimated direction of the target, it's not perfect but easier to understand.


bobbens:
Good idea, thanks!

PR #2404, opened 2023-09-30
Fix some translation issues of the ship's and outfit's status.

oo13:
I'm afraid that the fix for the physical unit might be too complex, it can be replaced by a simpler way that is just added _() if you want.


bobbens:
I see, it is quite complex. I'm assuming the motivation is to allow changing things on the fly. Wouldn't it be easier to do something like below? It'll be a tad bit slower than your approach, but should be negligible.
```C
/* .h file */
extern const char _UNIT_TIME[];
#define UNIT_TIME       _(_UNIT_TIME)

/* .c file. */
const char _UNIT_TIME[] = N_("sec");
```

oo13:
I agree that the performance is not a major issue here, my intention was just to avoid many repetition of the statements _(UNIT_...).

I will recreate the last commit as your suggestion and force-push it.

bobbens:
Looks good, thanks!

PR #2405, opened 2023-10-04
Fix problems that cause omission of some shipstats in Japanese language.

oo13:
This PR extends the buffers which are insufficient.

Actually 1.5 times of the current sizes are enough in Japanese language, so STRMAX may be too big.


bobbens:
Looks good, thanks!

PR #2406, opened 2023-10-05
Rackham's Razor

JackStorm448:
Added unique accessory 'Rackham's Razor' to the rewards of the third stage of the Red Rackham POI mission chain.

To-Do: add speed boost on board, and/or double effect when on a pirate ship to accessory effects


bobbens:
I'll try to look at this today, but one note: I did some changes after your previous PR, and it seems like this PR is based on your old version instead of the current Naev version. Please make sure to update Naev before doing a new PR (or starting to work on one) to avoid merge conflicts.

bobbens:
Fixed up and merged, thanks! I've also implemented the "double effect if pirate ship" since it was easier to do than the boarding effect for now (boarding would require new hooks).

Issue #1068, opened 2020-06-18
Allow to see all known outfits/ships/planets/systems


Wuzzy2:
I suggest to improve the search feature (for planets/systems/outfits/ships).

If you leave the search field empty, display ALL known items. Could also add a button “Show all” as convenience.

This will be useful to see which things you know so far.

bobbens:
Implemented by @oo13 in #2407 . Thanks!

PR #2407, opened 2023-10-10
Add "Show all" button to the find window.

oo13:
This PR implements the feature suggested by Wuzzy2 in #1068, and fixes the assertion error in map_findSearchOutfits().


bobbens:
Looks good, thanks! I've done the following two tweaks:
1. Display a different message when doing "show all"
2. Sort ships/outfits when displaying

PR #2416, opened 2023-10-22
Fix some translation issues in the system's map.

oo13:
This PR adds some msgctxts to distinguish the different contexts, and makes some texts translatable.

It also makes the commas in the lists of the tags untranslatable since the tags have no need to translate.

I don't come to mind a same translated word of two "Dense" in "Dense Volatile Nebula" and "Nebula: Dense Volatile"... so I would add the different contexts to it.


bobbens:
Looks good, thanks!

PR #2417, opened 2023-10-23
Converted the "Naev Needs You!" event to vn.

sigus:
I didn't reformat the text of the event though.

Also, it only partially solves the music issue with this event. Although `vn.music()` works better with scripts that run on land/load than `music.play()`, it still does not work well.

### DIFF
--- a/dat/events/naev_needs_YOU.lua
+++ b/dat/events/naev_needs_YOU.lua
@@ -18,6 +18,8 @@ DESCRIPTION:
 
 ]]--
 
+local vn = require "vn"
+local lg = require "love.graphics"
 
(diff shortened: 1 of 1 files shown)


bobbens:
What do you mean vn.music does not work well?

sigus:
> I'm worried with the current text it probably looks less than ideal, but we can improve it later. Thanks!

Is it ok to break large texts into smaller ones while converting events and missions to vn?

> What do you mean vn.music does not work well?

I mean it does not stop background music sometimes.

bobbens:
Yes, it's perfectly fine to cut the text up and/or do other minor changes as necessary.

I'll have to look into vn.music not stopping background music sometimes.

PR #2441, opened 2023-11-15
Fix some lua errors in factions/spawn/lib/common.lua.

oo13:
This PR also fixes Za'lek Mammon and Diablo spam in some systems, such as Goddard system.


bobbens:
Oops, looks good, thanks!

bobbens:
As a side note, made it now spit out a warning if `w` field is not defined (only happens if not the last element in the table). So hopefully this bug won't repeat itself.

PR #2455, opened 2023-11-25
Fixed a bug in the mission "Assault on Unicorn."

### LINKED ISSUE #2445: [Bug]: Shipps killed at Assault on Unicorn
### What happened?

Ships killed by my drones are not counted as killed by myself / compesated althouth these ships count in the info tab in the information window. I don't know if "killed by drones" is intended to be excluded from the count. In my opinion it may be more consistent to count these ships. I also have the impression that ships killed but attacked by other ships before, do not count either.

### Version

0.11.0 alpha (current)

### What operating system are you seeing the problem on?

Linux

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

oo13:
Probably, this PR fixes #2445. I killed 60 pirates and earned 3.5M credits with Za'lek Sting and Drones.

### DIFF
--- a/dat/missions/dvaered/assault_on_unicorn.lua
+++ b/dat/missions/dvaered/assault_on_unicorn.lua
@@ -92,7 +92,7 @@ function jumpin()
 end
 
 function death( pilot, killer )
-   if pir.factionIsPirate(pilot:faction()) and killer:withPlayer() then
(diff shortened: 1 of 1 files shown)


UncombedCoconut:
Nice.

PR #2464, opened 2023-11-29
Typo + fixed  an incorrect variable

not-pyroman:
there was a typo in 04_took and 02_met subtracted from Za'lek, when it seems like it should subtract from Dvaered


bobbens:
Looks good, thanks!

PR #2489, opened 2023-12-15
Update srs_ferry.lua

MathewKJ2048:
Bug fix for Sirius missions not updating marker for low-ranking players


bobbens:
Looks good, thanks! I've fixed it so dest_marker gets saved to the mission memory or it will give an error when reloading halfway through the mission.

PR #2493, opened 2023-12-20
Spell check

MathewKJ2048:
Changed "Ordinance" to "Ordnance" which I understand is the intended word (funny as the thought of pummeling fighters with ordinances may be)

### DIFF
--- a/dat/events/tutorial/tut_land.lua
+++ b/dat/events/tutorial/tut_land.lua
@@ -85,7 +85,7 @@ function outfit_buy( o )
       vn.scene()
       local sai = vn.newCharacter( tut.vn_shipai() )
       vn.transition( tut.shipai.transition )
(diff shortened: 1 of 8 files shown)


UncombedCoconut:
Good catch! As far as I can tell, "ordnance" is also an uncountable noun, meaning pretty much every use of "ord[i]nance's" should also become "ordnance".

MathewKJ2048:
> Good catch! As far as I can tell, "ordnance" is also an uncountable noun, meaning pretty much every use of "ord[i]nance's" should also become "ordnance".

Done
I also added an API call for faction.getAll(). The C function existed but the lua implementation was missing, so I assumed it could be added. I know different features are usually different PRs but I'm still figuring out my workflow

UncombedCoconut:
The faction.getAll implementation looks correct to me, but you should array_free(factions) before you return.
Having both features in the same PR might be OK -- the commits are separate, after all -- but I'm more certain about merging the spelling fixes unilaterally than the API update.
In any case, I'll check back on your PR after work.

bobbens:
Looks good, thanks!

PR #2495, opened 2023-12-21
Fix bug in shadowrun mission.

TeaToCode:


UncombedCoconut:
Yep, that looks better. Thanks!

PR #2496, opened 2023-12-23
Fix a bug that makes the mission "Seek And Destroy" impossible to complete.

oo13:
The narrator says that the player wasted the chance when the player jumps out after destroying the target. This PR fixes the bug, which seems to be introduced by c42999c10.

I would also replace the hook function "target_land" with "target_flee" since there isn't the function "target_land" in the mission "Seek And Destroy", though I couldn't test the situation to call the hook.


bobbens:
Oops, looks good, thanks!

PR #2497, opened 2023-12-26
Ensure "an" is used before "Ass" in Dvaered warlord insults

a-random-lemurian:
image

As it says in the title.


bobbens:
Thanks! I've modified it so the article is included in badword, which should be better for translations.

PR #2499, opened 2023-12-27
Mission to destroy a ship spamming local comms with advertisements ("Adblocker")

a-random-lemurian:
There's a ship spamming the local comms with annoying advertisements, and a captain's gone mad and is now frantically searching for someone who can destroy it!

Writing could probably use some work, but I feel sleepy. :zzz:


dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+local vntk = require "vntk"
+local portrait = require "portrait"
+
+local mission = {
+   name = _("Adblocker"),
+   description = "Destroy a ship spamming the {sys} system with excessive advertising transmissions",
+   reward = 300e3,
+   npc = {
+      name = "Desperate captain",
+      img = "neutral/male7.webp",
```
bobbens, line 40: Probably better to use the `vnimage` library to get a portrait instead of hardcoding a generic one. `vni.genericMale()` should work in this case. It returns a vn image + portrait image paths.

EDIT: You'll have to save the image / portrait in the `mem` table or it'll change every time the mission is loaded. Best to do it in the `create function.

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+
+local fmt = require "format"
+local neu = require 'common.neutral'
+local vn = require "vn"
+local vntk = require "vntk"
+local portrait = require "portrait"
+
+local mission = {
+   name = _("Adblocker"),
+   description = "Destroy a ship spamming the {sys} system with excessive advertising transmissions",
```
bobbens, line 36: You need to add translations for this string. Just wrap it in _()

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
+--[[
+<?xml version='1.0' encoding='utf8'?>
+<mission name="Adblocker">
+ <unique />
+ <chance>20</chance>
+ <location>Bar</location>
```
bobbens, line 6: You should add the following line in the header:
```xml
<cond>return require("misn_test").reweight_active()</cond>
```
It just makes it so the probability of the mission appearing goes down the more active missions the player has.

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+      img = "neutral/male7.webp",
+      description = _("You see a desperate looking captain.")
+   }
+}
+
+-- fp means "full path"
+mission.npc["img_fp"] = portrait.getFullPath(mission.npc.img)
+
+function create()
+   mem.current_system = system.cur()
```
bobbens, line 49: You should do an inclusive claim in the current system, or the pilot you create could get erased by another mission.

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+   if not accepted then return end
+
+   misn.accept()
+   misn.setTitle(mission.name)
+   misn.setDesc(fmt.f(_("A ship is currently spamming {sys} with tons of unwanted advertisements. A desperate captain has asked you to destroy, or disable it.")
+      , { sys = mem.current_system }))
+   misn.setReward(fmt.credits(mission.reward))
+   misn.osdCreate(mission.name, {
+      mission.name,
+      fmt.f(_(mission.description), { sys = mem.current_system })
```
bobbens, line 87: You should probably add the "Return to planet" line in the osd here, and then use `misn.osdActive(2)` to switch to it instead of recreating the osd.

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+   end)
+   vn.run()
+
+   if not accepted then return end
+
+   misn.accept()
+   misn.setTitle(mission.name)
+   misn.setDesc(fmt.f(_("A ship is currently spamming {sys} with tons of unwanted advertisements. A desperate captain has asked you to destroy, or disable it.")
+      , { sys = mem.current_system }))
+   misn.setReward(fmt.credits(mission.reward))
```
bobbens, line 84: No need to use fmt.credits here. You can directly use a number value.

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+   vn.transition()
+   man(fmt.f(_([["Look, I don't have much time! There's this annoying ship that's been spamming the local comms with tons of advertisements! I can't take it anymore! Please, you've got to stop it! I'll give you ${creds} if you stop it!"]])
+      , { creds = fmt.credits(mission.reward) }))
+   vn.menu {
+      { _([[Accept]]), "accept" },
+      { _([[Refuse]]), "refuse" },
+   }
+
+   vn.label("refuse")
+   vn.done()
```
bobbens, line 69: Would probably be nice to have a short refuse message, like "You walk away" or whatever.

dat/missions/neutral/adblocker.lua
```diff
@@ -0,0 +1,164 @@
...
+end
+
+function accept()
+   local accepted = false
+
+   vn.clear()
+   vn.scene()
+   local man = vn.newCharacter(mission.npc.name, { image = mission.npc.img_fp })
+   vn.transition()
+   man(fmt.f(_([["Look, I don't have much time! There's this annoying ship that's been spamming the local comms with tons of advertisements! I can't take it anymore! Please, you've got to stop it! I'll give you ${creds} if you stop it!"]])
```
bobbens, line 61: the format scripts just need {}, no need for the $ sign.

bobbens (changes requested):
Added some feedback. Thanks!

bobbens:
Looks good, thanks! Fixed some minor issues here and there, but seems to work fine.

PR #2508, opened 2023-12-30
Fix a wrong system name in the event "Kidnapped."

oo13:
The said system name was inconsistent with the mission "Kidnapped."


bobbens:
Looks good, thanks!

PR #2509, opened 2023-12-30
Fix a bug that makes accepting the mission "Sirian Bounty" impossible.

oo13:


bobbens:
Looks good, thanks again!

PR #2510, opened 2023-12-31
Fix bugs that cause lua errors in achack03.

oo13:
Some dialogues in achack03 were skipped due to the bugs.

I remove the shader in board() since you meet Harja in person.


bobbens:
Looks good. Thanks!

PR #2513, opened 2024-01-01
Fix some bugs in "Shadow" missions.

oo13:
In "Shadow Vigil", Seiryu didn't appear in Pas system.

By the way, in "Shadowrun" Rebina requests you to go to Durea in 20 periods, you need to know that you can use hypergates, wormholes, or Homebrew Processing Unit. Four Winds has a relative high standard.


bobbens:
Looks good, thanks! Really have to add vn and other libraries to the luacheck stuff.

bobbens:
I've also changed the system so that now you only have to go to Anubis (Scorpius system), and then go to Gammacron which should be much easier to complete.

Issue #2507, opened 2023-12-30
[Bug]: Targeting Conduit makes the range calculation wrong.


oo13:
### What happened?

With a Targeting Conduit, the ship starts to fire Mace Rockets and Za'lek Hunter Launchers even though she is out of range yet.

I guess the cause is weapon_createAmmo() in weapon.c:
```
   w->timer    = w->outfit->u.lau.duration * parent->stats.launch_range;
```
`launch_range` is directly multiplied the time dimension, but pilot_weapSetUpdateRange() in pilot_weapon.c multiplys `launch_range` and the distance.
```
         range *= p->stats.launch_range;
```
 It may be easier that `launch_range` handles the time.

### Version

cd4d0a3a1 (0.11.x, local build)

### What operating system are you seeing the problem on?

Linux

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bobbens:
Thanks for the feedback. The main issue seems to be the inconsistency in `outfit_range` that doesn't account for acceleration. I've modified the code to use `pilot_outfitRange` which does the correct calculations. This should be fixed. The remaining issue is that the ship stat is called "range" and not "duration", when it effectively is duration, although I'm not sure how pressing that is, since calling it "duration" is quite a bit more unintuitive imho. Thanks!

PR #2514, opened 2024-01-01
Fix an issue with launch_range modifiers.

oo13:
I'm sorry, #2507 didn't directly cause too early fires. The commit 7a62e1392 is not useless since this PR fixes the issue by time dimension.


bobbens:
Oops, missed that one. Thanks!

PR #2515, opened 2024-01-02
Fix translation issues in the mission "Adblocker."

oo13:
mission.description is a translated text.


bobbens:
Oops, missed those. Thanks!

PR #2518, opened 2024-01-05
Use `SDL_ConvertSurfaceFormat` to fix rendering of indexed color pngs

ids1024:
The other branch of this `if` statement already did conversion. It makes sense that it would be needed for indexed color in either case.

I guess conversion should only be done if the existing format cannot be used? Instead of always calling this.

This issue occurred on Fedora Rawhide, but not other Linux installs. I'm guessing that would be related to the version of `SDL2_Image` or `libpng`.

### DIFF
--- a/src/opengl_tex.c
+++ b/src/opengl_tex.c
@@ -368,9 +368,11 @@ static GLuint gl_loadSurface( SDL_Surface* surface, unsigned int flags, int free
    }
    else {
       *vmax = 1.;
-      glPixelStorei( GL_UNPACK_ALIGNMENT, MIN( surface->pitch&-surface->pitch, 8 ) );
+      SDL_Surface *s = SDL_ConvertSurfaceFormat( surface, SDL_PIXELFORMAT_ARGB8888, 0 );
+      glPixelStorei( GL_UNPACK_ALIGNMENT, MIN( s->pitch&-surface->pitch, 8 ) );
       glTexImage2D( GL_TEXTURE_2D, 0, GL_SRGB_ALPHA,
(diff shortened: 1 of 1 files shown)


src/opengl_tex.c
```diff
@@ -368,9 +368,11 @@ static GLuint gl_loadSurface( SDL_Surface* surface, unsigned int flags, int free
    }
    else {
       *vmax = 1.;
-      glPixelStorei( GL_UNPACK_ALIGNMENT, MIN( surface->pitch&-surface->pitch, 8 ) );
+      SDL_Surface *s = SDL_ConvertSurfaceFormat( surface, SDL_PIXELFORMAT_ARGB8888, 0 );
+      glPixelStorei( GL_UNPACK_ALIGNMENT, MIN( s->pitch&-surface->pitch, 8 ) );
```
bobbens, line 372: Shouldn't this be `s->pitch&-s->pitch`?

src/opengl_tex.c
```diff
@@ -368,9 +368,11 @@ static GLuint gl_loadSurface( SDL_Surface* surface, unsigned int flags, int free
    }
    else {
       *vmax = 1.;
-      glPixelStorei( GL_UNPACK_ALIGNMENT, MIN( surface->pitch&-surface->pitch, 8 ) );
+      SDL_Surface *s = SDL_ConvertSurfaceFormat( surface, SDL_PIXELFORMAT_ARGB8888, 0 );
```
bobbens, line 371: It seems like I get the colours inverted here, but it works with `SDL_PIXELFORMAT_ABGR8888`... You don't have an issue with this?

bobbens:
I've applied the two changes I've left in the comments as it seems to work here and should do less unnecessary conversions and such. Since I cherry-picked into 0.11.x, it hasn't detected the merge, so manually closing. Thanks!

PR #2520, opened 2024-01-07
Make energy regen constant

clintonmead:
(i.e. no longer affected by current charge level)


bobbens:
I've been thinking about this PR quite a bit. At the end, I think it's probably a good first step to having a linear system. We can merge it and test for now, and then modify all the outfits so that we don't need that counterintuitive *0.5 there. Thanks!

PR #2521, opened 2024-01-07
Add the 'Vacuum Cleaner?' to the list of mission accessories Misi offers

not-pyroman:


not-pyroman:
Let's see how many "Merge branch 'naev:main' into main" commits I can accumulate before this gets merged

PR #2522, opened 2024-01-13
tutorial.lua slight style and typo edits

exoterws:
Edits to the tutorial.lua text strings (primarily AI dialogue), since this is one of the first missions new players encounter. Edits are primarily for typos and style, e.g. repeated words etc. The only substantial content change was the removal of the phrase "ship ornament" from the AI's opening joke about a previous owner of the AI since, despite being a native English speaker, the phrase doesn't make sense to me and I don't get the joke.

Needless to say, maintainers are free to reject that edit; let me know if I have to redo anything.

### DIFF
--- a/dat/autonav.lua
+++ b/dat/autonav.lua
@@ -336,7 +336,7 @@ local function autonav_rampdown( count_brake )
    else
       d = d + btime * vel * tc_base
    end
-   if dtravel > d then -- Slight margin to brake early
+   if dtravel > d then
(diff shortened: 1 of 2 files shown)


bobbens:
It looks fine. I've been looking into the "ship ornament" thing, and it seems like "hood ornament" can also be used for airplanes, so I guess we can use that instead of space ornament to have it somewhat understood. Thanks!

PR #2554, opened 2024-03-13
[WIP] Round corners and modify tabed windows.

Nevrdid:
This is my in progress modifications to gui.
Remaining work :
- [ ] Recenter home window and Esc menu
- [ ] Fix unaligned things in widgets (scrollbar, background ...) or just use regular corner for them

Preview:
image


bobbens:
It looks great! Some comments:
1. I find the border contrast a bit too strong. Maybe weaken it a bit?
2. Would be cool to have it be optional. Maybe down the line it would be possible to have themes for the toolkit. This would probably make more sense most of the interfaces where moved to Lua so we didn't have to keep Lua and C synchronized.
3. The tabs at the top may be a bit hard to read when not focused. It seems like the text is almost covered up by the main window.
4. Does it work well with fullscreen windows?
5. Would probably be necessary to add to `dat/scripts/luatk` too, as things like the comms and some Lua-based interfaces use that.
6. Potentially could make sense to add to the VN too.

Nevrdid:
>     1. I find the border contrast a bit too strong. Maybe weaken it a bit?
I changed it to the same color as the one for active buttons to "unify" active tab like the ones in Outfit window and active buttons.
Do you think we can also change the color for active buttons or it is better to check if the window is flaged as HASTABS and weaken only there ( so the tabs in Outfit for example will not change).

>     2. Would be cool to have it be optional. Maybe down the line it would be possible to have themes for the toolkit. This would probably make more sense most of the interfaces where moved to Lua so we didn't have to keep Lua and C synchronized.
I'll look at this. I also planned to remove all the hard coded values to defines ofc.

>     3. The tabs at the top may be a bit hard to read when not focused. It seems like the text is almost covered up by the main window.
True. I think the best is just to move all the tabbar up a little as this this will also fix the strang looking of tabbar inside windows.

>     4. Does it work well with fullscreen windows?
Fullscreen windows like map should not change as i not apply it to windows which not have tabs. About land one, it just appear like the screenshot, it's no more a real fullscreen but i can still align side and bottom borders to the border of the game window.

>     5. Would probably be necessary to add to `dat/scripts/luatk` too, as things like the comms and some Lua-based interfaces use that.
>
>     6. Potentially could make sense to add to the VN too.

I'm currently looking at this, i first try to change the graphics.rectangle method to behave like the love2d 0.10+ one.
I'm not sure however what i may change once the method is ready as I not yet dig into the lua codebase.

bobbens:
OK, I took a quick look at it. Some notes.
1. circle.frag doesn't compile on linux/nvidia. I've pushed a fix to the round branch.
2. Buttons seem to have less border (or at least less clear than before)
3. White border seems to be stronger when not rounded vs rounded.
4. Main Menu isn't displayed on the main menu anymore (not a bad thing though, just a note. Haven't checked if it's more general)

I still have to play around a bit more. However, it seems to be quite nice.

Nevrdid:
>     1. circle.frag doesn't compile on linux/nvidia. I've pushed a fix to the round branch.

Hum, strange but ok x')

>     2. Buttons seem to have less border (or at least less clear than before)
> 
>     3. White border seems to be stronger when not rounded vs rounded.

I guess some doubles values may be floored through the functions arguments typecasting on some calls. I thought about it at a point but forgot to double check. Else i guess we must choose right values for line_width as for borders radius. For multiple elements, i choosen values "logicaly" without more testing than required i guess.

>     4. Main Menu isn't displayed on the main menu anymore (not a bad thing though, just a note. Haven't checked if it's more general)

It is intended, i globally removed windows title while i was working on windows size and i just seen it as useless x')
That's in the toolkit's window_renderBorder function. It's also where the borders were removed for fullscreen

dat/scripts/luatk/core.lua
```diff
@@ -363,12 +363,13 @@ function luatk.Window:draw()
 
    -- Draw background
    lg.setColour( luatk.colour.bg )
-   lg.rectangle( "fill", x, y, w, h )
```
bobbens, line 366: Would probably make sense to have the 30s be some sort of variable instead of hardcoded all over.

dat/scripts/luatk/map.lua
```diff
@@ -108,6 +108,7 @@ function Map:draw( bx, by )
       local px, py = p:get()
       if not (px < -r or px > w+r or py < -r or py > h+r) then
          lg.setColour( sys.c )
+         lg.setLineWidth(1);
```
bobbens, line 111: Might make sense to assume setLineWidth(1) and only change and change back when necessary.

src/nlua_gfx.c
```diff
@@ -467,29 +467,28 @@ static int gfxL_renderTexH( lua_State *L )
...
-   w     = luaL_checknumber( L, 3 );
-   h     = luaL_checknumber( L, 4 );
-   col   = luaL_checkcolour( L, 5 );
-   empty = lua_toboolean( L, 6 );
+   x   = luaL_checknumber( L, 1 );
+   y   = luaL_checknumber( L, 2 );
+   w   = luaL_checknumber( L, 3 );
+   h   = luaL_checknumber( L, 4 );
+   col = luaL_checkcolour( L, 5 );
+   lw  = lua_tonumber( L, 6 );
```
bobbens, line 486: If you want to make it optional, you have to use `luaL_optnumber( L, 6, 0. );` where `0.` would be the default in this case. This will raise errors if non-numbers are passed.

src/nlua_gfx.c
```diff
@@ -507,9 +509,13 @@ static int gfxL_renderRectH( lua_State *L )
    const mat4     *H     = luaL_checktransform( L, 1 );
    const glColour *col   = luaL_optcolour( L, 2, &cWhite );
    int             empty = lua_toboolean( L, 3 );
+   int             lw    = lua_tointeger( L, 4 );
```
bobbens, line 512: luaL_optinteger

src/nlua_gfx.c
```diff
@@ -521,24 +527,23 @@ static int gfxL_renderRectH( lua_State *L )
...
-   x     = luaL_checknumber( L, 1 );
-   y     = luaL_checknumber( L, 2 );
-   r     = luaL_checknumber( L, 3 );
-   col   = luaL_checkcolour( L, 4 );
-   empty = lua_toboolean( L, 5 );
+   x   = luaL_checknumber( L, 1 );
+   y   = luaL_checknumber( L, 2 );
+   r   = luaL_checknumber( L, 3 );
+   col = luaL_checkcolour( L, 4 );
+   lw  = lua_tointeger( L, 5 );
```
bobbens, line 543: luaL_optinteger

src/nlua_gfx.c
```diff
@@ -558,9 +564,11 @@ static int gfxL_renderCircleH( lua_State *L )
    const mat4     *H     = luaL_checktransform( L, 1 );
    const glColour *col   = luaL_optcolour( L, 2, &cWhite );
    int             empty = lua_toboolean( L, 3 );
+   int             lw    = lua_tointeger( L, 4 );
```
bobbens, line 567: luaL_optinteger

src/opengl_render.c
```diff
@@ -100,37 +105,48 @@ void gl_renderRect( double x, double y, double w, double h, const glColour *c )
...
-void gl_renderRectH( const mat4 *H, const glColour *c, int filled )
+void gl_renderRectH( const mat4 *H, const glColour *c, int line_width, int rx,
+                     int ry )
 {
-   gl_beginSolidProgram( *H, c );
-   if ( filled ) {
-      gl_vboActivateAttribOffset( gl_squareVBO, shaders.solid.vertex, 0, 2,
-                                  GL_FLOAT, 0 );
-      glDrawArrays( GL_TRIANGLE_STRIP, 0, 4 );
+   if ( conf.round_gui && rx && ry ) {
```
bobbens, line 135: I don't like the conditioning the behaviour here. If you, for example, try to draw a curved rectangle irregardless of the gui setting, it won't work. This is probably more important for the Lua API. Would probably be best to have this set when rendering, for example an if statement in renderPane for the C toolkit, and some equivalent wrapper function for luatk or whatever.

src/opengl_render.c
```diff
@@ -1141,6 +1157,18 @@ int gl_initRender( void )
...
+   vertex[1]  = 0.;
+   vertex[2]  = 1.;
+   vertex[3]  = 0.;
+   vertex[4]  = 1.;
+   vertex[5]  = 1.;
+   vertex[6]  = 0.;
+   vertex[7]  = 1.;
+   vertex[8]  = 0.;
+   vertex[9]  = 0.;
+   gl_paneVBO = gl_vboCreateStatic( sizeof( GLfloat ) * 10, vertex );
```
bobbens, line 1170: You can get rid of this if you use GL_LINE_LOOP instead of GL_LINE_STRIP.

src/options.c
```diff
@@ -143,7 +143,7 @@ void opt_menu( void )
 
    /* Dimensions. */
    w = 680;
-   h = 525;
+   h = 550;
```
bobbens, line 146: Slight overlap, may be better to increase.

src/tk/widget/tabwin.c
```diff
@@ -115,9 +115,16 @@ unsigned int *window_addTabbedWindow( unsigned int wid, const int x,
       /* Create windows with flags.
        * Parent window handles events for the children.
        */
-      wgt->dat.tab.windows[i] = window_createFlags(
-         tabnames[i], tabnames[i], wx, wy, ww, wh,
-         WINDOW_NOFOCUS | WINDOW_NORENDER | WINDOW_NOINPUT | WINDOW_NOBORDER );
+      int flags = WINDOW_NOFOCUS | WINDOW_NORENDER | WINDOW_NOINPUT;
+      /* For windows with upper tabs, keep borders to let them overwrite hiden
+       * tabs */
+      if ( strcmp( name, "tabLand" ) && strcmp( name, "tabInfo" ) &&
```
bobbens, line 121: I don't think hardcoding names is a good idea here. Would be better to pass some parameter instead or it'll likely bite us in the ass in the future.

src/tk/widget/tabwin.c
```diff
@@ -364,46 +371,67 @@ static int tab_key( Widget *tab, SDL_Event *event )
...
    }
-
-   /* Render the active window. */
-   window_render( wdw, wgt_isFlag( tab, WGT_FLAG_FOCUSED ) );
+   h = wdw->h < 800 ? 40 : wdw->h / 20;
+   w = tab_getBarWidth( tab );
 
    /* Render tabs ontop. */
-   x = bx + tab->x + 3.;
+   if ( !strcmp( tab->name, "tabLand" ) || !strcmp( tab->name, "tabInfo" ) ||
```
bobbens, line 386: Again, would be best not to hardcode specific names here if possible.

src/toolkit.c
```diff
@@ -688,6 +689,12 @@ unsigned int window_get( const char *wdwname )
 unsigned int window_create( const char *name, const char *displayname,
                             const int x, const int y, const int w, const int h )
 {
+   /* For windows with upper tabs, let tabedWindow handle the border to hide the
+    * unactive tabs */
+   if ( !strcmp( name, "wdwLand" ) || !strcmp( name, "wdwOptions" ) ||
```
bobbens, line 694: Would like to not hardcode window names here.

src/toolkit.c
```diff
@@ -1484,29 +1495,8 @@ static void window_renderBorder( const Window *w )
    /* Position */
    double x = w->x;
    double y = w->y;
-
-   /*
-    * Case fullscreen.
-    */
-   if ( window_isFlag( w, WINDOW_FULLSCREEN ) ) {
```
bobbens, line 1491: I think it would be nice to have a fullscreen fallback for the land window and such otherwise the tabs look a bit weird, like we had without the rounded tabs. Although I don't think this is the place.

dat/glsl/rounded_rect.frag
```diff
@@ -0,0 +1,24 @@
+#include "lib/sdf.glsl"
+uniform vec4 colour;
+uniform vec4 dimensions;
+uniform int parami;
```
bobbens, line 4: should probably be `uniform float paramf;` or the setting thickness part could give errors on some drivers.

dat/glsl/circle.frag
```diff
@@ -9,14 +9,16 @@ out vec4 colour_out;
...
       d = abs(d);
+   int line_width = parami > 0 ? parami : 1;
+
    /*
    float alpha = smoothstep(-m, 0.0, -d);
    float beta  = smoothstep(-2.0*m, -m, -d);
    colour_out   = colour * vec4( vec3(alpha), beta );
    */
-   float alpha = smoothstep(-1.0, 0.0, -d);
+   float alpha = smoothstep(-line_width, 0.0, -d);
```
bobbens, line 21: This gives errors on linux/nvidia. Please see the round branch for a solution (using paramf)

bobbens (changes requested):
Overall, it looks good, however there's some hardcoded stuff that should probably be fixed, and some stuff that should be made explicit. Some of the major points are:
1. Don't do the conditional rounding in renderPane
2. Don't hardcode window / widget names, use parameters instead
3. Would be nice to have a fullscreen fallback for when running at minimum resolution. As in, don't show the background when landed at minimum resolution at all. Please see screenshot of 4. for an example.
4. The rectangular tabs have some weird offsets as shown in the screenshot below for the selected tab
rounded
5. The buttons are less visible, but the border is stronger compared to the main branch
round_comp
6. There is also a difference in border brightness between the round and not round variant. Likely due to using the GL_LINE_LOOP. Probably would make more sense to just use a shader for all. You can use the discard; command to avoid rendering a pixel to be a bit faster.
round_comp2

bobbens:
Another thing I just noticed, since it doesn't full resolution on landed at minimum resolution (1280x720), you actually can see the bottom bar as shown in the first screenshot of the previous post. In the current main, it covers up everything. This means that it is likely that the effective height es being further reduced, likely making worse some of the issues we have with lack of scrolling text and too large descriptions.

Nevrdid:
> Overall, it looks good, however there's some hardcoded stuff that should probably be fixed, and some stuff that should be made explicit. Some of the major points are:
> 
>     1. Don't do the conditional rounding in renderPane
 
I'm still thinking about it.
 
>     2. Don't hardcode window / widget names, use parameters instead

Done using WINDOW_TABED for wdwXXX, and saving parent of tabed windows for tabXXX. 
Not sure if the parent save is the better option but it work.

>     3. Would be nice to have a fullscreen fallback for when running at minimum resolution. As in, don't show the background when landed at minimum resolution at all. Please see screenshot of 4. for an example.

Done. I basicly restored the fullscreen conditions and kept the tabs at top of window. But we lose the dividers style.

>     4. The rectangular tabs have some weird offsets as shown in the screenshot below for the selected tab

Fixed. The frags were not that well setup so changed them a little bit and it looks pretty well like that.

>     5. The buttons are less visible, but the border is stronger compared to the main branch

Done but yeah, i'll must check again all widget and fine-tune the ones which not like the shader smoothing.
I'll also have to correctly do the drawOutlinesThick effect  and verify that i don't round corners to elements which should not have them.

>     6. There is also a difference in border brightness between the round and not round variant. Likely due to using the GL_LINE_LOOP. Probably would make more sense to just use a shader for all. You can use the discard; command to avoid rendering a pixel to be a bit faster.

That was because of a different smoothing between them. Looks ok now. About discard, i'm not sure, I quickly googled it and seen that it's not such a good idea as it break gpu optimizations.

bobbens:
Thanks for the fixes looks better than before. Please note I am using `scalefactor = 1.5`, which is probably the cause of many of these issues. Some (nitpicking) comments:
1. There's probably too little spacing at the top. It's really tight with other languages like Japanese. Probably could use 1-2 pixels on the top and bottom of the text.
tootight

2. In fullscreen, it seems to be partially clipping on screen.
misalign

3. Still a difference in brightness in the borders. Pretty sure this is related to the scalefactor, but not sure.
comparison

4. Here it is zoomed in with `scalefactor = 1`. Things to notice:
4.1.Bbuttons are thinner and border is much thicker
4.2. Window border colour is much much brighter.
4.3. The window border seems to get "thinner" at the corner parts.
comp2

5. Since window names are removed. It would be nice to add the spob name to the landing main window. Maybe instead of "Landing Main", the first tab could be the spob name?
6. One thing that sort of bothers me a bit with the tabs is that they are now far away (in mouse distance) from the takeoff button. So on main I tend to click on the tabs from left to right, then click on takeoff. With the tabs on top, that's no longer possible. In that sense, maybe it would make sense to move the tabs to the bottom again?
7. Some of the dialogues would benefit from having the titles. I can think of things like `tk.msg` and friends (sort of being phased out). Maybe it would make sense for the windows to still have titles by default, and use a flag to disable it for things like landing window or main menu?

dat/glsl/rounded_rect.frag
```diff
@@ -0,0 +1,25 @@
...
+                dimensions.zw);
+    colour_out = colour;
+    if (paramf > 0.) {
+        d = abs(d);
+        colour_out.a = smoothstep(
+                -paramf,
+                -paramf / 2.,
+                -d);
+    } else
+        colour_out.a = smoothstep(0, 0.5, -d);
```
bobbens, line 24: This actually doesn't compile on linux/nvidia. In glsl, int/float difference is much stricter than standard C. For most implementations, `0` will not be cast to float. So it errors because it tries to find a function `smoothstep( int, float, float )`. It's best to write floats with `0.0` (not just `0.` since some drivers don't like that).

src/conf.c
```diff
@@ -1071,6 +1072,9 @@ int conf_saveConfig( const char *file )
    conf_saveComment(
       _( "Use bigger icons in the outfit, shipyard, and other lists." ) );
    conf_saveBool( "big_icons", conf.big_icons );
+   conf_saveComment( _( "Disable rounded GUI." ) );
```
bobbens, line 1075: Should this be "Enable rounded GUI" since `true` is enabled?

dat/glsl/rounded_rect.frag
Nevrdid: rip, i did it again.. my bad x'/

Nevrdid:
> Thanks for the fixes looks better than before. Please note I am using `scalefactor = 1.5`, which is probably the cause of many of these issues. Some (nitpicking) comments:
> 
>     1. There's probably too little spacing at the top. It's really tight with other languages like Japanese. Probably could use 1-2 pixels on the top and bottom of the text.

>     2. In fullscreen, it seems to be partially clipping on screen.

Yeah, i stoped yesterday once i restored fullscreen but i noticed that. Gonna do it. Also, i'll try all with different scalefactors.

 
>     3. Still a difference in brightness in the borders. Pretty sure this is related to the scalefactor, but not sure.

I didn't try to change scalefactor as i missed those differences but i'll investigate it.

>     4. Here it is zoomed in with `scalefactor = 1`. Things to notice:
>        4.1.Bbuttons are thinner and border is much thicker
>        4.2. Window border colour is much much brighter.
>        4.3. The window border seems to get "thinner" at the corner parts.

Honnestly, i didn't yet tuned borders values comparing to main branch and that is a step i'll must do. I changed multiple colors and sizes values so i hope just revert to original values will fix this. 

>     5. Since window names are removed. It would be nice to add the spob name to the landing main window. Maybe instead of "Landing Main", the first tab could be the spob name?

Just spob name as first tab not a good solution imo. Maybe `$spob_name Terminal` ? But yeah i understand that we should quickly see which spob we're on no matter which tab is actually active.

>     6. One thing that sort of bothers me a bit with the tabs is that they are now far away (in mouse distance) from the takeoff button. So on main I tend to click on the tabs from left to right, then click on takeoff. With the tabs on top, that's no longer possible. In that sense, maybe it would make sense to move the tabs to the bottom again?

I'm not sure about the render but It is totally possible.
Another solution may be to totally move some buttons out of the window and place them like tabs.
Possibly, i can move tabs to the right and take-off to the left. I may try to create a function to generate left side buttons to see how it render.

>     7. Some of the dialogues would benefit from having the titles. I can think of things like `tk.msg` and friends (sort of being phased out). Maybe it would make sense for the windows to still have titles by default, and use a flag to disable it for things like landing window or main menu?

:+1: I'll add a WINDOW_NOTITLE

PR #2559, opened 2024-03-26
Correct typo

not-pyroman:
corrected a typo in the require for love_shaders


UncombedCoconut:
Probably, and I can't see any harm in merging this. :)

PR #2564, opened 2024-04-22
Fix a bug that naev.py doesn't copy the po directories.

oo13:
neav.py created a directory "LC_MESSAGES", instead of the language directories.
This PR fixes it.

I tested it on Debian GNU/Linux testing, meson 1.2.3, and python3 3.11.8.


bobbens:
Looks good, thanks!

PR #2565, opened 2024-05-04
Changes to time presentation in race_qex mission

awest03:
- Fixed problem where display_time would display the minutes wrong for negative times (problem was `floor(-0.1)=-1`).
- Changed message when goal time is not met to be the time over the goal time (meaning the aforementioned fix isn't actually used but could still be handy in the future) i.e. instead of displaying `goal_time - time`, displays `-goal_time + time`.

### DIFF
--- a/dat/missions/neutral/race_qex.lua
+++ b/dat/missions/neutral/race_qex.lua
@@ -84,7 +84,7 @@ end
 
 local function display_time( time )
    return fmt.f(_("{1:02.0f}:{2:02.0f}.{3:.0f}"),{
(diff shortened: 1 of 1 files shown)


awest03:
There's another bug I just noticed, the actual time (as shown with the mission and used for comparison) is 0.1 seconds longer than the time which shows in large fancy text once you finish. I'm not really sure why, I've tested and it's not due to my current changes but I'll try to have a look at it

bobbens:
Looks good, thanks!

PR #2566, opened 2024-05-17
Fix bug causing inconsistent time display in the race_qex mission

awest03:
The large time display shown on completion was always less than the recorded time by 0.1s, this should fix it. Tested it running through the first track


bobbens:
Looks good, thanks!

PR #2569, opened 2024-05-26
News

theElerd:


bobbens:
Looks good, thanks! Some changes I've done:
1. Replace `ancient Sorom` with `Sorom` to be more consistent with the other wording.
2. Some other trivial corrections.

Issue #2571, opened 2024-06-13
[Bug]: `debug_logBacktrace` used in log.c when debugging is off


AntimatterReactor:
### What happened?

When `DEBUGGING` is not defined (which means `debug_logBacktrace` is not even defined) `log.c`'s `log_warn` attempts to use it anyway which results in the c compiler throwing an error, as ISO C99 disallows implicit declarations. #2572 should be the easiest solution to this

### Version

main-178a8a9

### What operating system are you seeing the problem on?

Linux

### Relevant log output

```shell
[2/131] Compiling C object naev.p/src_log.c.o
FAILED: naev.p/src_log.c.o 
cc -Inaev.p -I. -I.. -Isrc -I../src -I../src/tk -I../src/tk/widget -I../subprojects/enet-1.3.17/include -I../subprojects/nativefiledialog-extended-1.1.1/src/include -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/harfbuzz -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/sysprof-6 -I/usr/include/libxml2 -I/usr/include/SDL2 -I/usr/include/webp -I/usr/include/luajit-2.1 -I/usr/include/AL -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Wpedantic -std=gnu11 -O3 -Wno-pedantic -Wshadow -include config.h -DHWY_SHARED_DEFINE -DAVIF_DLL -D_REENTRANT -pthread -MD -MQ naev.p/src_log.c.o -MF naev.p/src_log.c.o.d -o naev.p/src_log.c.o -c ../src/log.c
../src/log.c: In function ‘log_warn’:
../src/log.c:309:4: error: implicit declaration of function ‘debug_logBacktrace’ [-Wimplicit-function-declaration]
  309 |    debug_logBacktrace();
      |    ^~~~~~~~~~~~~~~~~~
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bobbens:
Thanks for the PR. Should be fixed in main.

PR #2572, opened 2024-06-13
Added #if to debug_logBacktrace in log_warn to allow non-debug builds

AntimatterReactor:
Solves a compilation bug as defined by #2571


bobbens:
Oops. Looks good, thanks!
I've modified it so that debug_logBacktrace itself becomes a NOOP instead of using the `#if DEBUGGING` checks for simplicity.

PR #2578, opened 2024-07-02
Fix: Warning from input overlap triggering wrong

RisingLeaf:
## Description
Warnings from input overlaps only triggered when the overlap was the same keybind as the one that was currently set, fixed that by changing an == to an !=


RisingLeaf:
Might be worth it to add a notification if you left the keybind the same :shrug:

bobbens:
Looks good, thanks!

PR #2585, opened 2024-07-10
Create a PR template

TheGiraffe3:
Creates a PR template to make it easier to see what people are doing.
It is kind of copied from Endless Sky's though.


bobbens:
Looks good, thanks! I've tweaked it a tad as it may not have an associated issue.

PR #2593, opened 2024-08-23
Fix running from dev environment

BrknRobot:
Using `meson devenv` in the run helper fixes cases of missing shared libs from subprojects


ProjectSynchro:
Sweet! Thanks.

PR #2596, opened 2024-08-24
 Get dll search path for bundling from meson (fixed!)

BrknRobot:


ProjectSynchro:
Thanks!
Glad you were able to find a way to catch errors, not sure why it didn't work previously.

PR #2607, opened 2024-09-06
Avoid to show an empty message area at scrolling up.

oo13:
**Bug Fix**
**New Feature**

## Summary
When scrolling up:
- recover the alpha value of the fading messages.
- prevent the message area scrolling up if the lowest line is NULL.

and fixed two typos in a comment.

## Testing Done
I scrolled up the message area until it shows the first line after 10 jumping.


bobbens:
Looks good, thanks!

PR #2608, opened 2024-09-08
Resolve the overlap between the slider and the button in the VN option.

oo13:
**Bug Fix**

## Summary
Resolve the overlap between the slider and the button in the VN option.

## Testing Done
I changed the text speed and pushed the button in a VN option.


bobbens:
Looks good, thanks!

PR #2610, opened 2024-09-16
Fix a bug that conf.max_3d_tex_size was saved as a boolean.

oo13:
**Bug Fix**

## Summary
conf.max_3d_tex_size is saved as an integer number, same as the type when loading.

## Testing Done
I checked 'max_3d_tex_size = 256' in the config file after exiting Naev.


bobbens:
Nice catch, thanks!

PR #2611, opened 2024-09-16
Fix a bug that causes a buffer overflow when scrolling up the message.

oo13:
**Bug Fix**

## Summary
mesg_stack[new_point] might be out of the range because the modulo operator may result in a negative value.

mesg_viewpoint could also happen to be -1, so the message area was rewound the last line.

This PR fixes it.

## Testing Done
I scrolled up the messages after the lines of the message is more than 128.


bobbens:
Nice catch, thanks!

PR #2612, opened 2024-09-17
Add a spellcheck

TheGiraffe3:
**GitHub Action added**
**Typo fixes**

## Summary
Adds a GitHub action I made that checks spelling and throws errors when there are typos.

## Testing Done
None - typos don't really need to be tested and the spellcheck changes nothing in-game.


bobbens:
The concept is interesting, but ultimately I think it makes the most sense as a manual process and such during development. Maybe have pre-commit warn things (although not everyone uses it).

My particular strategy is to use `utils/gettextlanguagetool.py` with `extras/vimrc`, however, that is editor dependent and it only really handles C / Lua atm. It works quite well, but it has too many false positives to use it to raise CI errors. Thought about trying to use custom dictionaries and such, but it's just more things to maintain and such.

TheGiraffe3:
Just checking how many conflicts/errors there are.
24 thousand lines is a bit much, don't you think?

PR #2618, opened 2024-10-05
EBC is made playable.

Lineth:
Makes Electron Burst Cannons buyable by the player, but requires good standing with House Za'lek.


bobbens:
Looks good, thanks!

PR #2619, opened 2024-10-06
Fix errors in missions

oo13:
**Bug Fix**
## Summary
This PR fixes some errors that I have found in the last playthrough.

Fixed:
- Trivial errors
  - Overwrapping some portraits in "The Search for Cynthia"
  - Overwrapping a portrait in "Shadowrun"
  - No violation happened in "Dvaered Meeting"
  - No common bikers in "Dvaered Negotiation 2"
  - Never occurred the event "Way to Wild Space"
  - A log wasn't recorded in "Old Friends at Protera Husk"
- Errors by some unwanted input
  - Skipped some stages by asking Major Tam twice in "Dvaered Ballet"
  - Duplicate events occurred if you don't stop playing the chuckaluck after you received the note
- There is no nearby system "Effetey" in "Minerva Pirates 4"
  Dvaered and pirates' reinforcements might come from the same jump point and Dvaered could defeat the pirates easily. The pirates should come from the other side.
- Two errors in "The one with the Shopping"
  - There was no nearby system "Holly"
  - Patrol warned you when returning to the originating system through the system "Waterhole" without phosphine
- Landing permissions in Thurion
  - The Habitation Stations didn't have their permission, so you may land before you were known.
  - Shipyard and military restricted areas would never give you the landing permission even if you could increased the reputation. (Simply speaking, there was no return statement in the function can_land())
  - Some granted messages were nil.
- Naevpedia
  - A link error in shipstats.md
  - Not specified translation in classes.lua

## Testing Done
I played the fixed missions.

### DIFF
--- a/dat/events/wildspace/intro_npc.lua
+++ b/dat/events/wildspace/intro_npc.lua
@@ -4,7 +4,7 @@
  <location>land</location>
  <unique />
  <chance>10</chance>
- <chapter>^0</chapter><!-- TODO bump to 2 when finished. -->
+ <chapter>[^0]</chapter><!-- TODO bump to 2 when finished. -->
  <cond>
    local s = spob.cur()
    if not s:services()["bar"] then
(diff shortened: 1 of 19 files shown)


bobbens:
That's a lot of fixes! They all look great. Thanks!

PR #2639, opened 2024-10-27
Preliminary drone nerf

Lineth:
Heavy Drones now only have one Orion beam that does full damage. EBCs are a little better at damaging armor, but no longer do any disable damage.

### DIFF
--- a/dat/outfits/launchers/electron_burst_cannon.xml
+++ b/dat/outfits/launchers/electron_burst_cannon.xml
@@ -33,7 +33,7 @@
   <damage>
    <type>ion</type>
    <penetrate>24</penetrate>
-   <physical>12</physical>
+   <physical>14</physical>
   </damage>
   <ai>unguided</ai>
  </specific>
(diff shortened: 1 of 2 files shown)


bobbens:
That's a big ECB nerf, from 43.33 DPS (+10 disable/s) to 30 DPS. However, looking around Vulcan gun is 28 DPS and has lower range (1,100 vs 1,500), so maybe not bad? I would consider bumping damage a bit, but since they don't use energy, I guess it's fair game? I wouldn't be against giving some energy usage and and buffing damage though.

Side note, Mace rockets are 50!!! DPS and even use less CPU, so they would probably also need a nerfing.

Lineth:
I retried, ZBH10, and it was a lot easier. I agree that the EBC could probably get a damage increase. Maces I find are rarely _too_ bad. They do hit hard, but they run out of ammo easily and their aim sucks.

bobbens:
How about 14 damage for the ECB, so it has 35 DPS?

bobbens:
Looks good, thanks!

PR #2644, opened 2024-11-04
adds trails to zlk mph type v

Lineth:
odd that it didn't inherit it from the base mephisto, but heyho


bobbens:
Mmm, if it didn't inherit them, it's bug. The other variants didn't have to define them. Let me check.

bobbens:
Just tried it and it seems to be alright without the trail generator tag?
Furthermore, with 3D ships that tag shouldn't be used. The trails are baked into the gltf.

bobbens:
If you still have issues, the problem is probably elsewhere.

PR #2650, opened 2024-11-17
Dvaered Colonel Escort mission

TheGiraffe3:
**New Feature**

## Summary
Added a new mission where you escort a Dvaered colonel flying an Arsenal to a random Dvaered planet.
Another warlord lies in ambush, hoping to kill your escort, but Radner has tact… and leaves.

## Testing Done
I'm not sure how to.


TheGiraffe3:
Does anybody know what the problem with the Lua check is?

bobbens:
> Does anybody know what the problem with the Lua check is?

I would recommend you install `pre-commit` and use that as it faster to check, but if you download the CI artefact logs, you'll see the error is:

```
==================================== 1/1 =====================================
test:         naev / luacheck
start time:   05:09:09
duration:     19.12s
result:       exit status 2
command:      MALLOC_PERTURB_=29 LUA_PATH='/__w/naev/naev/source/?.lua;;' MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MESON_TEST_ITERATION=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 /__w/naev/naev/source/utils/nluacheck.py dat docs extras utils --no-color
----------------------------------- stdout -----------------------------------
Checking dat/missions/dvaered/dvaered-colonel-transport.lua 1 error

    dat/missions/dvaered/dvaered-colonel-transport.lua:28:4: expected 'end' (to close 'function' on line 26) near 'else'

Total: 0 warnings / 1 error in 1 file
==============================================================================
```

So it seems like it catches a Lua syntax error.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
+--[[
+<?xml version='1.0' encoding='utf8'?>
+<mission name="Escort a Dvaered colonel to a system">
```
bobbens, line 3: I would recommend a more simple name. Something like `Dvaered Colonel Transport`. You can then start the mission from the console with `naev.missionStart("Dvaered Colonel Transport")` and it should spawn the NPC.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
...
+end
+
+function accept ()
+   local text
+   else
+      text = fmt.f(_([[You approach the Dvaered, who seems to be a soldier, probably one of the colonel rank. Arriving at their table, you are greeted, "Hello! Could you escort my ship, an Arsenal, to {pnt} in the {sys} system? I'll give you {rwd} for it, but I can't tell you why. Well, what do you say?"]]),
+         {pnt=misplanet, sys=missys, rwd=reward_text})
+      mem.talked = true
+   end
+   if vntk.yesno( _("Escort Agreed"), text ) then
```
bobbens, line 33: I would recommend you use the `vn` library, and not `vntk`. In general, most modern missions use `vn` and display the NPC portrait. You can generate one automatically with the `vnimage` module, such as `vnimage.dvaeredMilitary()` for a random dvaered military character (it can be male or female looking).

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
...
+      vntk.msg( _("Escort Agreed"), _([["Perfect! I'll pay you as soon as we get there."]]) )
+      misn.accept()
```
bobbens, line 35: You should also set the mission reward / description here with `misn.setReward` and `misn.setDesc`.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
+--[[
+<?xml version='1.0' encoding='utf8'?>
+<mission name="Escort a Dvaered colonel to a system">
+ <unique />
+ <priority>1</priority>
+ <chance>24</chance>
+ <location>Bar</location>
+ <spob>Dvaer Prime</spob>
```
bobbens, line 8: In general, we try to avoid limiting missions to a single spob (with some exceptions where it is necessary). If possible, it should be a bit more flexible.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
+--[[
+<?xml version='1.0' encoding='utf8'?>
+<mission name="Escort a Dvaered colonel to a system">
+ <unique />
+ <priority>1</priority>
```
bobbens, line 5: Priority 1 should be reserved for more critical missions. In this case, either default or something like 4 would make sense.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
...
+--]]
+--[[
+
+   MISSION: ESCORT TO A PLANET
+   DESCRIPTION: SMALL MISSION WHERE YOU ESCORT AN ARSENAL TO A PLANET: WIP
+
+--]]
+
+local fmt = require "format"
+
```
bobbens, line 19: You need to include all the libraries you are using, such as `neu` or `escort`.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
...
+   else
+      text = fmt.f(_([[You approach the Dvaered, who seems to be a soldier, probably one of the colonel rank. Arriving at their table, you are greeted, "Hello! Could you escort my ship, an Arsenal, to {pnt} in the {sys} system? I'll give you {rwd} for it, but I can't tell you why. Well, what do you say?"]]),
+         {pnt=misplanet, sys=missys, rwd=reward_text})
+      mem.talked = true
+   end
+   if vntk.yesno( _("Escort Agreed"), text ) then
+      vntk.msg( _("Escort Agreed"), _([["Perfect! I'll pay you as soon as we get there."]]) )
+      misn.accept()
+      misn.osdCreate(_("Dvaered colonel escort"), {
+         fmt.f(_("Escort a Dvaered colonel flying an Arsenal to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment.")), {pnt=mem.destspob, sys=mem.destsys}),
```
bobbens, line 37: For the OSD, you generally want to keep it very short. I would just do something like `"Escort a Dvaered Colonel to {pnt} ({sys} system)."`. The full explanation can be put into `misn.setDesc`.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
...
+function create ()
+   mem.talked = false
+   misn.setNPC( _("A Dvaered colonel"), "devaered/dv_military_f6.webp", _("This soldier seems to be a colonel.") )
+end
+
+function accept ()
+   local text
+   else
+      text = fmt.f(_([[You approach the Dvaered, who seems to be a soldier, probably one of the colonel rank. Arriving at their table, you are greeted, "Hello! Could you escort my ship, an Arsenal, to {pnt} in the {sys} system? I'll give you {rwd} for it, but I can't tell you why. Well, what do you say?"]]),
+         {pnt=misplanet, sys=missys, rwd=reward_text})
```
bobbens, line 30: `misplanet` and `missys` are never defined.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,56 @@
...
+   if vntk.yesno( _("Escort Agreed"), text ) then
+      vntk.msg( _("Escort Agreed"), _([["Perfect! I'll pay you as soon as we get there."]]) )
+      misn.accept()
+      misn.osdCreate(_("Dvaered colonel escort"), {
+         fmt.f(_("Escort a Dvaered colonel flying an Arsenal to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment.")), {pnt=mem.destspob, sys=mem.destsys}),
+      }
+      misn.markerAdd( mem.destspob )
+      hook.land( "land" )
+   end
+   local colonel_ship
```
bobbens, line 42: I'm guessing you want:
```lua
local colonel_ship = ship.get("Dvaered Arsenal")
```

bobbens (changes requested):
Thanks for the mission. I like that it is an escort mission, we need more of those.

Some comments:
1. When Naev starts it'll test missions to see if they compile. If they give errors, you'll see them in the logs and the console. You should use that to fix Lua errors, because I think your code has quite a few that need fixing.
2. You should probably look at the [anxious merchant](https://github.com/naev/naev/blob/main/dat/missions/trader/anxiousmerchant.lua) mission (or any other similar one), to see how to use the `vn` and `vnimage` libraries to start the mission in a bit more modern Naev way.
3. If possible, could you write up what your intention is with the mission? Is it meant to be a one-off thing or a start of something larger?
4. The mission text is a bit bare bones. It would be nice to flesh it out a bit more.

TheGiraffe3:
Where do you suggest that I make a summary of the mission?
It's meant to be a one-off, so not as important.

dat/missions/dvaered/dvaered-colonel-transport.lua
TheGiraffe3: Ah, I had been under the impression that 1 made it least priority.

bobbens:
> Where do you suggest that I make a summary of the mission? It's meant to be a one-off, so not as important.

Just the header in the mission should be fine then.

As for the mission,  fleshing it out a bit more, like why does the colonel need the escorting? Potentially adding some danger or some twist along the way.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -1,47 +1,92 @@
...
+local vni = require "vnimage"
 
 
-function create ()
-   mem.talked = false
-   misn.setNPC( _("A Dvaered colonel"), "devaered/dv_military_f6.webp", _("This soldier seems to be a colonel.") )
+local npc_name = _("Dvaered Colonel")
+local npc_portrait, npc_image
+function create()
+   npc_image, npc_portrait = vni.dvaered
```
bobbens, line 26: You'll have to save the npc data with the mission or it will be lost on load. You can do this by storing it in the special `mem` table like:
```lua
mem.npc_image, mem.npc_portrait = vni.dvaeredMilitary()
```

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -1,47 +1,92 @@
...
+     {pnt=mem.dest_planet, sys=mem.dest_sys}
+   vn.jump("choice")
+
+   vn.label("what is your name")
+   m(_([[The Dvaered seems slightly taken aback. "Well... that may be classified information... call me Radver."]]))
+   vn.jump("choice")
+
+   vn.label("sure")
+   m(_([["Wonderful! I'll be on your ship when you leave."]]))
+   misn.accept()
```
bobbens, line 69: So, if you want to do things in the VN, you have to wrap it in a `vn.func()`, otherwise it won't be executed as part of the VN, but automatically. For accepting the mission, usually what is done is to do something like:
```lua
local accepted = false

-- vn stuff

vn.label("accepted")
vn.func( function ()
   accepted = true
end )
-- ...

vn.run()

--  Stops execution if accepted wasn't set
if not accepted then return end 

misn.accept()
-- ...
```

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -1,47 +1,92 @@
...
+   vn.menu{
+      {_([["Yep, I'd be glad to!"]]), "sure"},
+      {_([["Not going to happen, sorry."]]), "never"},
+   }
+
+   vn.label("never")
+   m(_([["That's sad. Oh well, I'll ask someone else." The Dvaered leaves.]]))
+   vn.done()
+   
+   misn.setReward(2500)
```
bobbens, line 82: 2500 credits is really low... unique missions usually pay much better than equivalent generic missions.

TheGiraffe3:
> As for the mission, fleshing it out a bit more, like why does the colonel need the escorting? Potentially adding some danger or some twist along the way.

Hmm, well, that's the thing. The colonel is being mysterious on purpose.
I have added a pirate spawn though, as part of the danger.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,128 @@
...
+ <priority>4</priority>
+ <chance>86</chance>
+ <location>Bar</location>
+ <faction>Dvaered</faction>
+</mission>
+--]]
+--[[
+
+   Mission: Escort a Dvaered colonel
+   Description: Small mission where you escort a Dvaered Arsenal.
```
TheGiraffe3, line 14: I've seen other files that list the author before giving a description, but the mission template didn't do that.
Should I be doing that/are there rules about it that I haven't learned yet?

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,128 @@
...
+   if not accepted then return end
+
+   misn.accept()
+   
+   misn.setReward(750000)
+   misn.setDesc(fmt.f(_("Escort a Dvaered colonel, who is flying an Arsenal, to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment."), {pnt=mem.dest_planet, sys=mem.dest_sys}))
+   misn.osdCreate(_("Dvaered colonel escort"), {
+      fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system.")), {pnt=mem.dest_planet, sys=mem.dest_sys},
+   }
+   misn.markerAdd( mem.destspob )
```
TheGiraffe3, line 103:
```suggestion
   misn.markerAdd( mem.dest_planet )
```

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,128 @@
...
+
+   if not accepted then return end
+
+   misn.accept()
+   
+   misn.setReward(750000)
+   misn.setDesc(fmt.f(_("Escort a Dvaered colonel, who is flying an Arsenal, to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment."), {pnt=mem.dest_planet, sys=mem.dest_sys}))
+   misn.osdCreate(_("Dvaered colonel escort"), {
+      fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system.")), {pnt=mem.dest_planet, sys=mem.dest_sys},
+   }
```
TheGiraffe3, line 102:
```suggestion
   })
```

dat/missions/dvaered/dvaered-colonel-transport.lua
bobbens: There isn't really a convention there. I don't really think the author is necessary given it is a collective work, and all the changes are stored in git anyway.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,127 @@
...
+   })
+   misn.markerAdd( mem.dest_planet )
+   hook.land( "land" )
+   hook.enter( "pirate_ambush" )
+   local colonel_ship = ship.get("Dvaered Arsenal")
+   escort.init ( colonel_ship, {
+   })
+end
+
+function pirate_ambush ()
```
bobbens, line 111: This is a bit technical, but you should make sure the system isn't claimed exclusively when spawning the pirates. The following line at the top of the enter wook should work.
```lua
if not naev.claimTest( system.cur(), true ) then return end
```
 Is there a reason it's pirates after them? It might make more sense to have it be another Dvaered warlord.

TheGiraffe3: Is there a way to do that?
I may not be understanding the whole situation, but it looks like there are no hostile Dvaered governments.
Or were you thinking have a hostile-to-the-Dvaered government flying a Dvaered ship?

TheGiraffe3: Maybe the Dreamer Clan?

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -102,10 +102,13 @@ function accept()
    })
    misn.markerAdd( mem.dest_planet )
    hook.land( "land" )
-   hook.enter( "pirate_ambush" )
    local colonel_ship = ship.get("Dvaered Arsenal")
    escort.init ( colonel_ship, {
    })
+   
+   if not naev.claimTest( system.cur(), true ) then return end
```
bobbens, line 109: Sorry, I meant it should be at the top of the `function pirate_ambush()` function.

dat/missions/dvaered/dvaered-colonel-transport.lua
bobbens: You can do that with dynamic factions. See the dvaered warlord battle event for an example:
https://github.com/naev/naev/blob/main/dat/events/dvaered/warlords_battle.lua

> Maybe the Dreamer Clan?

FLF would probably make more sense. However, it depends on what backstory you want to give the character. I think the most "neutral" would be another Dvaered Warlord. Just create two dynamic factions: one for the colonel, and one for the enemies. Have them be hostile to each other (Colonel should also be hostile to pirates), and it should work.

TheGiraffe3: I'm afraid I can't figure out how to set the ship's governments to the dynamic; could you help with that?

TheGiraffe3:
Do I need to update the .po files?

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,142 @@
...
+
+function ambush ()
+   if not naev.claimTest( system.cur(), true ) then return end
+   local dvaered_factions = faction.get("Dvaered")
+
+   fhostile = faction.dynAdd( dvaered_factions, "warlords_hostile", _("Warlords") )
+   ffriendly = faction.dynAdd( dvaered_factions, "warlords_friendly", _("Warlords") ) -- codespell:ignore ffriendly
+   faction.dynEnemy( ffriendly, fhostile ) -- codespell:ignore ffriendly
+
+   pilot.add( "Dvaered Phalanx", "fhostile", source_system, "Asheron Anomaly" )
```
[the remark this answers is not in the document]
TheGiraffe3: Isn't that field for the name of the ship?
The ship should be called "Asheron Anomaly". It doesn't mean anything.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,142 @@
...
+   misn.setNPC(npc_name, npc_portrait, _("A Dvaered, very professional-looking, is sitting with an excellant posture at the bar.") )
+end
+
+function accept()
+   local accepted = false
+   vn.clear()
+   vn.scene()
+   local m = vn.newCharacter( npc_name {image=npc_image} )
+   vn.transition()
+   m(fmt.f([[As you approach, the Dvaered soldier stands. "Hello, captain {playername}!" they say. "Nice to see you around here. How's it going?"]]))
```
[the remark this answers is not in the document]
TheGiraffe3: It's one of the mysteries. I can remove it, but the idea is this Dvaered has spies working for them, and has learned information from the flight specifications or something like that.

dat/missions/dvaered/dvaered-colonel-transport.lua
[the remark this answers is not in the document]
bobbens: OK, you should then add a translatable mark: `_("Asheron Anomaly")`. However, I think it may be a bit confusing unless it is mentioned by Radver. Maybe `_("Warlord Lackey")` might be more easy to understand.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,142 @@
...
+   hook.land( "land" )
+   local colonel_ship = ship.get("Dvaered Arsenal")
+   escort.init ( colonel_ship, {
+         pilot.add( "Dvaered Arsenal", "ffriendly", source_system, "Radver" ) -- The Dvaered colonel is telling you to call them the name of the Arsenal; their real name is classified information.
+   })
+
+   hook.enter( "ambush" )
+end
+
+function ambush ()
```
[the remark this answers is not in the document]
bobbens: No, `hook.timer` and `hook.safe` only execute once, the rest will execute infinitely unless removed (or if you land on the wrong planet the hook would get removed and such).

dat/missions/dvaered/dvaered-colonel-transport.lua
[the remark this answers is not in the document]
bobbens: Mmm, if you want something like that, I would probably condition it on their reputation with Dvaered or something. Otherwise, if the player just starts out and gets to Dvaered space, it will be a bit weird. Alternative, you can make it a bit more "spy-y". Maybe something like `"Why, if it is not Captain {playername}..."`. And let the player ask "How did you know my name!?" and then you can add some sly response.

TheGiraffe3:
I tried to make Radver a bit less polite, and added a note that the player is surprised at how friendly they are.
Let me know if the Dvaered is still too friendly.

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,160 @@
...
+   end } )
+
+   hook.enter( "ambush" )
+   hook.enter( "factions" )
+end
+
+function ambush ()
+   if not naev.claimTest( system.cur(), true ) then return end
+
+   rm.hook( "ambush" )
```
bobbens, line 133: API is `hook.rm` and you would need to pass the id obtained from `hook.enter`. An easier way of doing this is using a global variable. Just do something like:
```lua
if mem.ambushed then return end
mem.ambushed = true
```
Which is easier to do.

dat/missions/dvaered/dvaered-colonel-transport.lua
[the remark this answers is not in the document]
bobbens: Added comment below.

bobbens:
Have you tried starting the game with the mission? It probably gives you errors which can help you debug. They should appear in the logs and console.

TheGiraffe3:
I may not have checked throughly enough, but I haven't been able to find the logs on MacOS, and I don't have a Windows computer.
It doesn't help that the checks don't build an Apple version of the PR...

dat/missions/dvaered/dvaered-colonel-transport.lua
```diff
@@ -0,0 +1,163 @@
...
+local fmt = require "format"
+local neu = require "neutral"
+local vn = require "vn"
+local vni = require "vnimage"
+
+local reward = 0
+
+local source_system = nil
+
+lmisn.getSpobAtDistance([sys=system.cur(Dvaer)], 1, 6, fct=faction.get("Dvaered")[, samefct=true[, filter=nil[, data=nil[, hidden=false]]]])
```
TheGiraffe3, line 31:
```suggestion
lmisn.getSpobAtDistance([sys=Dvaer], 1, 6, fct=faction.get("Dvaered")[, samefct=true[, filter=nil[, data=nil[, hidden=false]]]])
```

bobbens:
> I may not have checked throughly enough, but I haven't been able to find the logs on MacOS, and I don't have a Windows computer. It doesn't help that the checks don't build an Apple version of the PR...

Since it is only Lua, you don't actually need to compile or anything. You can just dump the mission in your data, or even just place it as a plugin and the game should pick it up.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,163 @@
...
+local fmt = require "format"
+local neu = require "neutral"
+local vn = require "vn"
+local vni = require "vnimage"
+
+local reward = 0
+
+local source_system = nil
+
+lmisn.getSpobAtDistance(sys=Dvaer, 1, 6, fct=faction.get("Dvaered")[, samefct=true[, filter=nil[, data=nil[, hidden=false]]]])
```
bobbens, line 31: should probably be:
```lua
local spbs = lmisn.getSpobAtDistance( nil, 4, 6, "Dvaered" )
if #spbs <= 0 then misn.finish(false) end

mem.destspb = spbs[ rnd.rnd(1,#spbs) ]
```

Also, you should put it in the `create()` function block at the top.

bobbens: This will get all spobs that meet the criteria, fail if there are none, and choose a random one to put into `mem.destspb`, that you can use as the destination later.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -28,13 +28,17 @@ local reward = 0
...
 
 local npc_name = _("Dvaered Colonel")
 local npc_portrait = nil
 local npc_image = nil
+
+local m = vn.newCharacter( npc_name, {image=npc_image} )
+
 function create()
-   local spbs = lmisn.getSpobAtDistance( nil, 4, 6, "Dvaered" )
+   local spbs = misn.getSpobAtDistance( nil, 2, 6, "Dvaered" )
```
bobbens, line 41: No, this should be `lmisn`. If you haven't imported the library, you should add `local lmisn = require "lmisn"` at the top of the file (where the other requires are).

TheGiraffe3:
Sorry to put you to all this trouble reviewing.

TheGiraffe3:
Naev uses the triple-period character instead of three period characters in a row, correct

bobbens:
> Sorry to put you to all this trouble reviewing.

No problem.

> Naev uses the triple-period character instead of three period characters in a row, correct?

Honestly, it's a bit inconsistent. The only main difference I can think of is that it takes 3 times longer to do in the VN when animated, so having 3 periods may give a better feeling in that case.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+   mem.ambushed = true
+
+   local dvaered_factions = faction.get("Dvaered")
+
+   local ffriendly = faction.dynAdd( dvaered_factions, "radver", _("Dvaered") ) -- codespell:ignore ffriendly
+   local fhostile = faction.dynAdd( dvaered_factions, "radver_baddie", _("Dvaered Warlord") )
+   faction.dynEnemy( ffriendly, fhostile ) -- codespell:ignore ffriendly
+   return ffriendly, fhostile -- codespell:ignore ffriendly
+
+   pilot.add( "Dvaered Phalanx", "fhostile", source_system, _("Asheron Anomaly") )
```
TheGiraffe3, line 149: The error log says:
>    dat/missions/dvaered/colonel-transport.lua:149:4: expected 'end' (to close 'function' on line 135) near 'pilot'

So it seems the game doesn't like this line.
But I don't really want to remove the Dvaered Warlord entirely.
So does this need to be moved somewhere else, or..?

awest03, line 149: This might fix it
```suggestion
   pilot.add( "Dvaered Phalanx", "fhostile", source_system, _("Asheron Anomaly") )

   return ffriendly, fhostile -- codespell:ignore ffriendly
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+local npc_portrait = nil
+local npc_image = nil
+
+local m = vn.newCharacter( npc_name, {image=npc_image} )
+
+function create()
+   local spbs = lmisn.getSpobAtDistance( nil, 2, 6, "Dvaered" )
+   if #spbs <= 0 then misn.finish(false) end
+   mem.destspb = spbs[ rnd.rnd(1,#spbs) ]
+
```
bobbens, line 45: Since you never define mem.dest_sys, it'll error out later. Add the following here:
```lua
mem.dest_sys = mem.destspb:system()
```

bobbens:
Wanted to add some suggestions, but it seems like they got committed instead :/

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,165 @@
...
+   misn.setDesc(fmt.f(_("Escort a Dvaered colonel, who is flying an Arsenal, to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment."), {pnt=mem.destspb, sys=mem.dest_sys}))
+   misn.osdCreate(_("Dvaered colonel escort"), {
+      fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system.")), {pnt=mem.destspb, sys=mem.dest_sys},
+   })
+   misn.markerAdd( mem.dest_planet )
+   hook.land( "land" )
+   local colonel_ship = ship.get("Dvaered Arsenal")
+   escort.init( colonel_ship, {
+      func_pilot_create = function ( p )
+        local fct = ffriendly()
```
bobbens, line 126: You do not need this line.
```suggestion
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,165 @@
...
+
+function create()
+   local spbs = lmisn.getSpobAtDistance( nil, 2, 6, "Dvaered" )
+   if #spbs <= 0 then misn.finish(false) end
+   mem.destspb = spbs[ rnd.rnd(1,#spbs) ]
+
+   mem.npc_image, mem.npc_portrait = vni.dvaeredMilitary()
+   misn.setNPC(npc_name, npc_portrait, _("A Dvaered, very professional-looking, is sitting with an excellant posture at the bar.") )
+end
+
```
bobbens, line 49:
```suggestion
local function factions()
   local dvaered_factions = faction.get("Dvaered")

   local ffriendly = faction.dynAdd( dvaered_factions, "radver", _("Dvaered") ) -- codespell:ignore ffriendly
   local fhostile = faction.dynAdd( dvaered_factions, "radver_baddie", _("Dvaered Warlord") )
   faction.dynEnemy( ffriendly, fhostile ) -- codespell:ignore ffriendly
   return ffriendly, fhostile -- codespell:ignore ffriendly
end
   ```
   
   Define the faction generation function here.

File: dat/missions/dvaered/colonel-transport.lua (hunk 25 above)
Comment (bobbens) [reply dat/missions/dvaered/colonel-transport.lua:149]:
```suggestion
local _ffriendly, fhostile = factions()
pilot.add( "Dvaered Phalanx", fhostile, source_system, _("Asheron Anomaly") )
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,165 @@
...
+   mem.ambushed = true
+
+   local dvaered_factions = faction.get("Dvaered")
+
+   local ffriendly = faction.dynAdd( dvaered_factions, "radver", _("Dvaered") ) -- codespell:ignore ffriendly
+   local fhostile = faction.dynAdd( dvaered_factions, "radver_baddie", _("Dvaered Warlord") )
+   faction.dynEnemy( ffriendly, fhostile ) -- codespell:ignore ffriendly
+   return ffriendly, fhostile -- codespell:ignore ffriendly
+
```
bobbens, line 148:
```suggestion
```

dat/missions/dvaered/colonel-transport.lua
TheGiraffe3:
```suggestion
   mem.dest_sys = mem.destspb:system()
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+function ambush ()
+   if not naev.claimTest( system.cur(), true ) then return end
+
+   if mem.ambushed then return end
+
+   mem.ambushed = true
+
+   local _ffriendly, fhostile = factions()
+   pilot.add( "Dvaered Phalanx", fhostile, source_system, _("Asheron Anomaly") )
```
bobbens, line 148: You never define source_system. I'm guessing you want them to jump into the system after the player?

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+
+local m = vn.newCharacter( npc_name, {image=npc_image} )
+
+function create()
+   local spbs = lmisn.getSpobAtDistance( nil, 2, 6, "Dvaered" )
+   if #spbs <= 0 then misn.finish(false) end
+   mem.destspb = spbs[ rnd.rnd(1,#spbs) ]
+   mem.dest_sys = mem.destspb:system()
+   mem.npc_image, mem.npc_portrait = vni.dvaeredMilitary()
+   misn.setNPC(npc_name, npc_portrait, _("A Dvaered, very professional-looking, is sitting with an excellant posture at the bar.") )
```
bobbens, line 44: Using Dvaered as a noun is a tad weird, maybe a Dvaered Officer?

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+
+local reward = 750e3
+
+local source_system = nil
+
+local npc_name = _("Dvaered Colonel")
+local npc_portrait = nil
+local npc_image = nil
+
+local m = vn.newCharacter( npc_name, {image=npc_image} )
```
bobbens, line 36: You can't just define the character here, because they get added to each scene. You have to define them after a `vn.scene()`.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+   local ffriendly = faction.dynAdd( dvaered_factions, "radver", _("Dvaered") ) -- codespell:ignore ffriendly
+   local fhostile = faction.dynAdd( dvaered_factions, "radver_baddie", _("Dvaered Warlord") )
+   faction.dynEnemy( ffriendly, fhostile ) -- codespell:ignore ffriendly
+   return ffriendly, fhostile -- codespell:ignore ffriendly
+end
+
+function accept()
+   local accepted = false
+   vn.clear()
+   vn.scene()
```
bobbens, line 59: You should add the `vn.newCharacter` here after `vn.scene()`.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+   m(_([["Alright."]]))
+   vn.jump("mission description")
+
+   vn.label("fine")
+   m(_([["As long as there's no trouble…"]]))
+   vn.jump("mission description")
+
+   vn.label("my name")
+   m(_([["Uhh… tough luck, captain, I'm not allowed to tell you. Too bad.]]))
+   m(_([["Let's just say that my espionage services are of the best, and leave it at that… shall we?"]])) -- When/if possible, the word "shall" should be italicized.
```
bobbens, line 79: Italics are not possible now, but you could colourize it maybe. Like `#nshall#0` would make it be darker than the standard white text.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+   vn.jump("mission description")
+
+   vn.label("mission description")
+   m(fmt.f([["Well, to the point. I need somebody who's not crazy to escort me and my Arsenal to {pnt} in the {sys} system. Would you do that? I can't tell you why, that would be classified information leaked to an outsider. Very dangerous, especially when we can't trust you.]]))
+     {pnt=mem.destspb, sys=mem.dest_sys}
+   m(_([["One more thing: another warlord would be only to happy to blow any of the colonel rank such as myself up, so you may need to expect attacks. The ship will probably be called 'Asheron Anomaly', but it may be something else."]]))
```
bobbens, line 85: You should probably add a menu option to ask about the `Asheron Anomaly`, because I'm pretty sure it'll stand out to players.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+
+   vn.label("what is your name")
+   m(_([[The Dvaered seems slightly taken aback. "Well… that may also be classified information… call me Radver."]]))
+   vn.jump("choice")
+
+   vn.label("sure")
+   m(_([["Surprising but good! I'll be on your ship in a while, so don't leave too soon."]]))
+   vn.func( function ()
+      accepted = true
+   end )
```
bobbens, line 100: You should add a `vn.done()` after this line or it'll continue down to the next message.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,164 @@
...
+   mem.ambushed = true
+
+   local _ffriendly, fhostile = factions()
+   pilot.add( "Dvaered Phalanx", fhostile, source_system, _("Asheron Anomaly") )
+end
+
+function land ()
+   if spob.cur() == mem.destspb then
+      vn.clear()
+      vn.scene()
```
bobbens, line 154: Should define the character again here after scene.

bobbens (changes requested):
Without playtesting it, I've found a few mistakes, however, it looks ready for play testing once they are fixed.

dat/missions/dvaered/colonel-transport.lua
TheGiraffe3: Correct.

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,174 @@
...
+local vn = require "vn"
+local vni = require "vnimage"
+
+local reward = 750e3
+
+local source_system = nil
+
+local npc_name = _("Dvaered Colonel")
+local npc_portrait = nil
+local npc_image = ni
```
bobbens, line 34:
```suggestion
local npc_image = nil
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,174 @@
...
+   if not accepted then return end
+
+   misn.accept()
+
+   misn.setReward(reward)
+   misn.setDesc(fmt.f(_("Escort a Dvaered colonel, who is flying an Arsenal, to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment."), {pnt=mem.destspb, sys=mem.dest_sys}))
+   misn.osdCreate(_("Dvaered colonel escort"), {
+      fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system.")), {pnt=mem.destspb, sys=mem.dest_sys},
+   })
+   misn.markerAdd( mem.dest_planet )
```
bobbens, line 136:
```suggestion
   misn.markerAdd( mem.destspb )
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,174 @@
...
+   hook.land( "land" )
+   local colonel_ship = ship.get("Dvaered Arsenal")
+   escort.init( colonel_ship, {
+      func_pilot_create = function ( p )
+        p:rename("Radver")
+        local ffriendly = factions()
+        p:setFaction( ffriendly() )
```
bobbens, line 143:
```suggestion
        p:setFaction( ffriendly )
```

dat/missions/dvaered/colonel-transport.lua
```diff
@@ -0,0 +1,174 @@
...
+   end } )
+
+   hook.enter( "ambush" )
+end
+
```
bobbens, line 148:
```suggestion

local source_system = system.cur()
```

bobbens (changes requested):
With these last suggestions, I think it should be playable. Could you test and see if you get any errors and either fix them if you can or tell me what you get? I'll try to test later.

TheGiraffe3:
Alright, I don't think there are any errors related to this mission.
<details>
<summary>Full error log:</summary>
WARNING ../source/src/opengl.c:232 [gl_checkHandleError]: OpenGL error [gl_fboCreate:305]: GL invalid value
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
WARNING ../source/src/opengl_shader.c:223 [gl_program_link]: link_status==1: [[
WARNING: Output of vertex shader 'tex_scale' not read by fragment shader

]]
LAST WARNING PRINTED 10 TIMES, SKIPPING FROM NOW ON
</details>

bobbens:
Tried to play it and it still has a lot of mistakes. I found the following:
```diff
diff --git a/dat/missions/dvaered/colonel-transport.lua b/dat/missions/dvaered/colonel-transport.lua
index 561d7ad601..c0d5c30fdf 100644
--- a/dat/missions/dvaered/colonel-transport.lua
+++ b/dat/missions/dvaered/colonel-transport.lua
@@ -21,15 +21,13 @@
 local escort = require "escort"
 local fmt = require "format"
 local lmisn = require "lmisn"
-local neu = require "neutral"
+local neu = require "common.neutral"
 local vn = require "vn"
 local vni = require "vnimage"

 local reward = 750e3

 local npc_name = _("Dvaered Colonel")
-local npc_portrait = nil
-local npc_image = nil

 function create()
    local spbs = lmisn.getSpobAtDistance( nil, 2, 6, "Dvaered" )
@@ -37,7 +35,7 @@ function create()
    mem.destspb = spbs[ rnd.rnd(1,#spbs) ]
    mem.dest_sys = mem.destspb:system()
    mem.npc_image, mem.npc_portrait = vni.dvaeredMilitary()
-   misn.setNPC(npc_name, npc_portrait, _("A Dvaered soldier, very professional-looking, is sitting with an excellant posture at the bar.") )
+   misn.setNPC(npc_name, mem.npc_portrait, _("A Dvaered soldier, very professional-looking, is sitting with an excellant posture at the bar.") )
 end

 local function factions()
@@ -53,10 +51,10 @@ function accept()
    local accepted = false
    vn.clear()
    vn.scene()
-   local m = vn.newCharacter( npc_name, {image=npc_image} )
+   local m = vn.newCharacter( npc_name, {image=mem.npc_image} )
    vn.transition()
-   m(fmt.f([[As you approach, the Dvaered soldier stands. "Oh, is it Captain {playername}?" they say. "Nice to see you around here. How's it going?"]]))
-      {playername=player.name()}
+   m(fmt.f(_([[As you approach, the Dvaered soldier stands. "Oh, is it Captain {playername}?" they say. "Nice to see you around here. How's it going?"]]),
+      {playername=player.name()}))
    vn.menu{
       {_([["Quite well, thank you!"]]), "well"},
       {_([["I guess it's going alright."]]), "fine"},
@@ -77,9 +75,9 @@ function accept()
    vn.jump("mission description")

    vn.label("mission description")
-   m(fmt.f([["Well, to the point. I need somebody who's not crazy to escort me and my Arsenal to {pnt} in the {sys} system. Would you do that? I can't tell you why, that would be classified information leaked to an outsider. Very dangerous, especially when we can't trust you.]]))
-     {pnt=mem.destspb, sys=mem.dest_sys}
-   m(_([["One more thing: another warlord would be only to happy to blow any of the colonel rank such as myself up, so you may need to expect attacks. The ship will probably be called 'Asheron Anomaly', but it may be something else."]]))
+   m(fmt.f(_([["Well, to the point. I need somebody who's not crazy to escort me and my Arsenal to {pnt} in the {sys} system. Would you do that? I can't tell you why, that would be classified information leaked to an outsider. Very dangerous, especially when we can't trust you.]]),
+     {pnt=mem.destspb, sys=mem.dest_sys}))
+   m(_([["One more thing: another warlord would be only too happy to blow any of the colonel rank such as myself up, so you may need to expect attacks. The ship will probably be called 'Asheron Anomaly', but it may be something else."]]))
    vn.menu{
       {_([["I'd be happy to do that!"]]), "sure"},
       {_([["What is your name?"]]), "what is your name"},
@@ -129,7 +127,7 @@ function accept()
    misn.setReward(reward)
    misn.setDesc(fmt.f(_("Escort a Dvaered colonel, who is flying an Arsenal, to {pnt} in the {sys} system. You haven't been told why, but there may be a large payment."), {pnt=mem.destspb, sys=mem.dest_sys}))
    misn.osdCreate(_("Dvaered colonel escort"), {
-      fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system.")), {pnt=mem.destspb, sys=mem.dest_sys},
+      fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system."), {pnt=mem.destspb, sys=mem.dest_sys}),
    })
    misn.markerAdd( mem.destspb )
    hook.land( "land" )
@@ -163,14 +161,16 @@ function land ()
    if spob.cur() == mem.destspb then
       vn.clear()
       vn.scene()
-      local m = vn.newCharacter( npc_name, {image=npc_image} )
+      local m = vn.newCharacter( npc_name, {image=mem.npc_image} )
       vn.transition()
-      m(fmt.f(_([[As you land on {pnt} with the Arsenal close behind, you receive an intercom message. The Dvaered seems to have slightly changed appearance. "Good job bringing me here!" says the colonel. "Here is {reward}, as we agreed."]]), {pnt=mem.des
tspb, reward=fmt.credits(reward)}) )
+      m(fmt.f(_([[As you land on {pnt} with the Arsenal close behind, you receive an intercom message. The Dvaered seems to have slightly changed appearance. "Good job bringing me here!" says the colonel. "Here is {reward}, as we agreed."]]),
+         {pnt=mem.destspb, reward=fmt.credits(reward)}) )
       vn.func( function () player.pay(reward) end )
       vn.sfxVictory()
       vn.na( fmt.reward(reward) )
       vn.run()
-      neu.addMiscLog( fmt.f(_([[You escorted a Dvaered colonel who was flying an Arsenal to {pnt}. This colonel, who said their name could be Radver, was very polite to you, though they didn't tell you why they needed the escort.]]), {pnt=mem.destspb} ) )
+      neu.addMiscLog( fmt.f(_([[You escorted a Dvaered colonel who was flying an Arsenal to {pnt}. This colonel, who said their name could be Radver, was very polite to you, though they didn't tell you why they needed the escort.]]),
+         {pnt=mem.destspb} ) )
       misn.finish( true )
    end
 end
 ```
 
 Could you try to play the mission and find any other issues please?

File: dat/missions/dvaered/colonel-escort.lua [hunk 42]
```diff
@@ -0,0 +1,176 @@
...
+   m(_([["Alright."]]))
+   vn.jump("mission description")
+
+   vn.label("fine")
+   m(_([["As long as there's no trouble…"]]))
+   vn.jump("mission description")
+
+   vn.label("my name")
+   m(_([["Uhh… tough luck, captain, I'm not allowed to tell you. Too bad.]]))
+   m(_([["Let's just say that my espionage services are of the best, and leave it at that… #nshall#0 we?"]])) -- When possible, the word "shall" should be italicized and the color change removed.
```
TheGiraffe3, line 74:
```suggestion
   m(_([["Let's just say that my espionage services are of the best, and leave it at that… *shall* we?"]])) -- When possible, the word "shall" should be italicized and the color change removed.
```

dat/missions/dvaered/colonel-escort.lua
```diff
@@ -71,7 +71,7 @@ function accept()
 
    vn.label("my name")
    m(_([["Uhh… tough luck, captain, I'm not allowed to tell you. Too bad.]]))
-   m(_([["Let's just say that my espionage services are of the best, and leave it at that… #nshall#0 we?"]])) -- When possible, the word "shall" should be italicized and the color change removed.
+   m(_([["Let's just say that my espionage services are of the best, and leave it at that… *shall* we?"]]))
```
bobbens, line 74: Sadly, Markdown is not supported in the VN, this is only the naevpedia. We would need more work to support it well. Furthermore, even our luatk widget just does bold as `#n%s#0`...

TheGiraffe3:
```suggestion
   m(_([["Let's just say that my espionage services are of the best, and leave it at that… shall we?"]])) -- When possible, the word "shall" should be italicized.
```

bobbens:
Ran through the mission and did some fixes / changes:
1. Fix mission API and such.
2. Don't spawn ancestor right away, delay 10 seconds.

The diff is as belows. This is playable and works.

```diff
diff --git a/dat/missions/dvaered/colonel-escort.lua b/dat/missions/dvaered/colonel-escort.lua
index 467d8243a2..be2e9b36fb 100644
--- a/dat/missions/dvaered/colonel-escort.lua
+++ b/dat/missions/dvaered/colonel-escort.lua
@@ -130,34 +130,44 @@ function accept()
       fmt.f(_("Escort a Dvaered colonel to {pnt} in the {sys} system."), {pnt=mem.destspb, sys=mem.dest_sys}),
    })
    misn.markerAdd( mem.destspb )
-   hook.land( "land" )
    local colonel_ship = ship.get("Dvaered Arsenal")
-   escort.init( colonel_ship, {
-      func_pilot_create = function ( p )
-        p:rename("Radver")
-        local ffriendly = factions()
-        p:setFaction( ffriendly )
-   end } )
-
-   hook.jumpin( "ambush" )
+   escort.init( {colonel_ship}, {
+      func_pilot_create = "create_radver",
+   } )
+   escort.setDest( mem.destspb, "success" )
+
+   hook.jumpin( "jumpin" )
+end
+
+function create_radver ( p ) -- luacheck: globals create_radver
+   p:rename("Radver")
+   local ffriendly = factions()
+   p:setFaction( ffriendly )
 end

 local source_system = system.cur()
-function ambush ()
+function jumpin ()
    if not naev.claimTest( system.cur(), true ) then
       source_system = system.cur()
       return
-    end
+   end

-   if mem.ambushed then return end
+   hook.timer( 10, "ambush" )
+end

+function ambush ()
+   if not naev.claimTest( system.cur(), true ) then
+      return
+   end
+
+   if mem.ambushed then return end
    mem.ambushed = true

    local _ffriendly, fhostile = factions()
    pilot.add( "Dvaered Phalanx", fhostile, source_system, _("Asheron Anomaly") )
 end

-function land ()
+function success () -- luacheck: globals success
    if spob.cur() == mem.destspb then
       vn.clear()
       vn.scene()
```

Some comments:
* The arsenal can easily solo the ancestor, plus, since it happens quite after the jump, it is quite fast combat.
* I would consider making the spawn in happen later (~20 seconds?), also, probably add mode ancestors, and make Radver say something when he sees them, highlight them on the map, and make them hostile right away. That way it is a bit harder and the player actually has to bother to engage them.
* If combat becomes required, which it should, probably add some requirement to start the mission. Like at least chapter 1, or maybe even make the player have a mercenary license too.
* Probably lower change to 20% or less, since it's quite easy to find the mission in Dvaered space.

dat/missions/dvaered/colonel-escort.lua
```diff
@@ -0,0 +1,176 @@
...
+
+local npc_name = _("Dvaered Colonel")
+
+function create()
+   local spbs = lmisn.getSpobAtDistance( nil, 2, 6, "Dvaered" )
+   if #spbs <= 0 then misn.finish(false) end
+   mem.destspb = spbs[ rnd.rnd(1,#spbs) ]
+   mem.dest_sys = mem.destspb:system()
+   mem.npc_image, mem.npc_portrait = vni.dvaeredMilitary()
+   misn.setNPC(npc_name, mem.npc_portrait, _("A Dvaered soldier, very professional-looking, is sitting with an excellant posture at the bar.") )
```
bobbens, line 38:
```suggestion
   misn.setNPC(npc_name, mem.npc_portrait, _("A Dvaered soldier, very professional-looking, is sitting with an excellent posture at the bar.") )
```

TheGiraffe3:
> * I would consider making the spawn in happen later (~20 seconds?), also, probably add mode ancestors, and make Radver say something when he sees them, highlight them on the map, and make them hostile right away. That way it is a bit harder and the player actually has to bother to engage them.

Not 100% sure how to highlight them, is there a mission I could use as an example?

bobbens:
> > * I would consider making the spawn in happen later (~20 seconds?), also, probably add mode ancestors, and make Radver say something when he sees them, highlight them on the map, and make them hostile right away. That way it is a bit harder and the player actually has to bother to engage them.
> 
> Not 100% sure how to highlight them, is there a mission I could use as an example?

You can just do `p:setHilight(true)` where `p` is the pilot object.

TheGiraffe3:
I'd recommend that when you merge this you squash it - this many commits probably don't need to all be added.
Or I can open a new PR with the commits squashed once this one is deemed merge-ready.

dat/missions/dvaered/colonel-escort.lua
```diff
@@ -0,0 +1,190 @@
...
+   if not naev.claimTest( system.cur(), true ) then
+      source_system = system.cur()
+      return
+   end
+
+   if mem.ambushed then return end
+
+   mem.ambushed = true
+
+   local fhostile = factions()
```
bobbens, line 170:
```suggestion
   local _ffriendly, fhostile = factions()
```

dat/missions/dvaered/colonel-escort.lua
```diff
@@ -0,0 +1,190 @@
...
+   pilot.add( "Dvaered Vigilance", fhostile, pilot:setHilight(true), source_system, _("Asheron Anomaly") )
```
bobbens, line 171:
```suggestion
   local p = pilot.add( "Dvaered Vigilance", fhostile, source_system, _("Asheron Anomaly") )
   p:setHilight(true)
```

bobbens (changes requested):
Thanks! Looks more or less good with the remaining changes. I'll wait until post 0.12.0 to squash + merge, and probably do some changes:
1. Make Radver say something when enemies jump in or whatever.
2. Add a couple of enemies.
3. Probably tweak start condition so it doesn't appear too early and the player meets their doom.

bobbens:
Probably will give it a few more tweaks after testing, but I squashed and merged into main. Thanks!

PR #2651, opened 2024-11-17
Fix two bugs in the mission "Seek And Destroy."

oo13:
**Bug Fix**
## Summary
- OSD showed you the next system even if no one had mentioned it yet.
- Lua error at trying hook.pilot()

## Testing Done
- I played "Seek And Destroy" and check OSD.
- I threatened and shot a pirate until they gave up.
- I threatened a pirate and another one, and shot latter until they gave up.


TheGiraffe3 (approved):

bobbens:
Thanks for the fix! I'm not really happy with how the mission uses mission variables, but we can try to look at that post 0.12.0.

PR #2652, opened 2024-11-17
Fix a double space in the Dvaered Duel event

TheGiraffe3:
**Bug Fix**

## Summary
Fixed a double space.
I'm not sure whether this is a breaking bug or not; best not to find out wrong.


bobbens:
Duplicate whitespace should affect Lua, thanks!

PR #2655, opened 2024-11-22
Fix the overflowing model name in the equipment tab.

oo13:
**Bug Fix**
## Summary
Empire Lancelot Golden Efreeti showed the ship stat at a wrong line. This PR fixes it.

## Testing Done
I opened the equipment tab and selected an Empire Lancelot Golden Efreeti.


bobbens:
Looks good, thanks!

PR #2674, opened 2024-11-29
Make stringguess less annoying/hard

a-random-lemurian:
**Balance**
Make the string guessing game a lot easier.

- Add red X for when the letter is not used at all in the final string
- Correctly put the ? and ! symbols by using a positional array of matches, instead of a count of how many fuzzy or exact matches there were.

Because it is so much easier now, it might affect balance, and I don't know how to offset it.

## Testing Done
old:
image

new:
image

## Remaining stuff
- [x] Make the x in the explanation red

### DIFF
--- a/dat/scripts/minigames/stringguess/core.lua
+++ b/dat/scripts/minigames/stringguess/core.lua
@@ -5,6 +5,12 @@ local love_shaders   = require "love_shaders"
 local fmt = require "format"
 local mg = {}
 
+local match = {
+   EXACT = 1,
+   FUZZY = 2,
+   NEVER = 3
+}
+
 local colours = {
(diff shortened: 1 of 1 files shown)


HyperSoop:
why is this nessesary?

HyperSoop:
this puzzle does not need nerfs

bobbens:
Thanks for the PR!
This sort of defeats the difficulty of the minigame as having to infer without knowing the order is the main property, so it would only be acceptable as an easy mode. However, as mentioned on discord, this would be better as a plugin because we can't reasonable offer a difficulty setting for every minigame mode, and can only support the skip mechanic. I would encourage you to make a plugin with the changes and PR the info to https://github.com/naev/naev-plugins so people can find it. Thanks!

PR #2684, opened 2024-12-07
Make VN flow in judgment.lua more consistent and clear

not-pyroman:
**Bug Fix**

## Summary
Makes the labels in judgement.lua more consistent and fixes an incorrect label

## Testing Done
HyperSoop (@hypersoop) in the Discord has played through this version fine when I gave it to them to replace it with in their extracted AppImage weirdness (which is also what pointed out the erroneous label of "cont04_took" that was apparently in main (for a moment I thought *I* did that :P)


bobbens:
Looks good ,thanks!

PR #2699, opened 2024-12-27
Allow mods to adjust POI NPC priority

clintonmead:
**New Feature**

## Summary
Gets the POI NPC priority from a variable, defaulting to the current default of 9. This allows for mods to adjust this (I have a mod in the works that adjusts this value to 5 so it shows up as an important NPC that gives an alert). 

## Testing Done
Made sure when landing there weren't a bunch of Lua errors.


clintonmead:
I'm not sure why this is failing the Luacheck.

clintonmead:
Plugin that will use this is here: https://github.com/clintonmead/naev-additional-options

Issue #2748, opened 2025-01-30
[Bug]: CPU requirement of some light weapons


johann1764:
### What happened?

There are 3 levels of light weapons, that can be characterized by damage, consumption, weight or CPU requirements:

1. Ultra light that require around 5 PFLOP
2. light that require around 24 PFLOP
3. "Heavy light" that require ... 100 PFLOP
 
This post is about the third category.
The involved weapons include Ripper cannon, Shredder, Plasma cannon and Razor S3.
The damage/consumption/weight caracteristics of these weapons lie between those of the corresponding level 2 light weapon (eg Razor S2)
and the corresponding medium size weapon (eg Disruptor S1)
However, the corresponding medium size weapon "only" require 50 PFLOP.

Maybe in some cases you're willing to give up CPU for space (using a small slot instead of a medium).

I tried to find such a situation.
My biggest ship with small weapon slots is the Hippocrates, a destroyer.
Its maximum CPU capacity is therefore the one you get with a thalos 5402 : 420PFLOP.
Using a level 3 light weapon is therefore _possible_ ... but absurd in this particular case
(The interesting side of this ship being able to use a fighter bay that will use almost all CPU)

I could not find a case were using these weapons would make sense.

My suggestion is to reduce their requirement to 50PFLOP or less.
Even so, using such weapon would buy space (smaller weapon slot) against power 
(slightly less powerfull than equivalent medium-sized weapon)

I am not sure they would be used at all even in this setting.

### Version

0.12.2

### What operating system are you seeing the problem on?

Linux

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

CCSheepfarmer:
I just came here to report on this as well. The 100 cpu cost on the basic ripper seems entirely incongruous and makes no sense. Setting aside the fact that compute requirements for basic pewpews going up with power level makes no sense and is rather immersion breaking, the 100 for the basic ripper makes no sense next to the 50 of the heavy ripper. I NEVER use the light because there's simply no situation where it will be worth putting it on the vessels it's actually useful on.

Kharrus:
So I've kept up a parallel build for combat testing ever since this change went live years ago since I disagreed with it, and I have organized the tier 3 light guns in with a much more streamlined outfitter.

[...]

PR #2750, opened 2025-02-06
Balancing v4?

bobbens:
## Summary

This is changes by @Kharrus . Small summary:

1. Heavy Ripper, Plasma Cluster Cannon, Disruption artillery buffed and increased to large.
2. Tier 3 cannons (100 cpu small) turned into medium.
3. Mass of Disruptor Artillery S2, Disruptor Battery S2, and Mass driver adjusted with some buffs.
4. All corvettes get a small weapon slot to medium.
5. Small and medium-sized turrets have damage and armour penetration adjusted to max fixed parts excepting Gauss Cannon.

This should cover #2748 .


Kharrus:
-CPU values follow a more unified curve, going at 5,15,45,60,90,120,350, and 500 for different weapon sizes to standardize their weight class and usage. Adjustments and feedback requested; Disruptor Artillery S1, Razor Artillery S2, Heavy Ripper, and Turbolaser CPU values set to be adjusted (missed a few).

-Playtesting with favoring logical damage and armor penetration values between fixed and turreted guns of the same type results in an overall increase in the value of absorption as a stat in the game, making medium and smaller turrets more effective against small ships, roughly similar against medium ships, and less effective overall against capital ships.  This encourages medium ships to punch down a little harder while not threatening capitals as much, with capital ships threatening each other more. Disruptors as a baseline got their damage buffed a bit due to them underperforming, and do double their base disable damage, excepting turrets that had their values readjusted to match equivalent fixed guns.

-Systems cores are adjusted to all be good relative to one another, including Unicorp cores. Orion remains the midline option with the second best defense, Thalos is adjusted to have the second best energy output and capacity with the best CPU, Prometheus has the best overall power output with significantly more energy capacity than other cores but have the worst defense, Aegis cores specialize in only defense, with having the worst energy output and capacity and second to worst CPU. Unicorp cores are 4th best at defense and energy management, and the worst at CPU output.  All of the heaver cores in each class are adjusted to be 4x more massive than their smaller counterparts, limiting the potential for upsizing cores being the default option for corvettes and cruisers to ensure all sizes are used in context.

-Increased damage values are in line with a higher HP ceiling with addition of Aegis cores, which have significantly more raw HP than the next best core for protection. This paired with energy costs for turrets being adjusted to strictly double their fixed counterpart means that overall defense is favored more than long fights, with Prometheus cores generally being more favored on the higher end. 

-Ship roles, particularly with capital ships, were added to both solidify the niches certain ships fill, as well as better specializing their existing talents or adding new bonuses and penalties where appropriate based on the existing ship description (example, most Dvaered ships getting CPU penalties, Za'lek Demon getting energy regen bonus). 
For the flat class bonuses, Interceptors get a 50% bonus to acceleration (to help with dogfighting and to make your carrier fighters more responsive). This makes them adequately slippery, like the Hyena.
Carriers get a hefty 40% bonus to detection, mostly to help better coordinate their fighters across vast distances, and see battleships creeping up on them.
Cruisers get a minor 10-25% tracking speed bonus, mostly to help with shooting down pirates like corvettes and heavy fighters. This helps specialize their common role, especially for the NPCs, as customs ships that patrol systems and burn down every last pirate they can find, large or small.
Battleships get a 30-40% bonus to weapon range, and act in the classical sense of battleships with firing the biggest, farthest-firing guns you can reasonably fit on a ship to sink the other guy first. This is very fun to watch on a Goddard.

HyperSoop:
I'm somewhat concerned about potential loss of flexibility in ship designs with those changes. I love equipping my ships "wrong" and that should remain a feasible option, for fun's sake.

Could we perhaps get this branch built by CI so playtesting is accessible?

bobbens:
> Could we perhaps get this branch built by CI so playtesting is accessible?

It only changes data, so you can just run Naev from the directory with `naev -d dat/` and it should prioritize the new data. Alternatively, you could symlink or copy `dat/` into your plugins folder (it will complain about missing plugins.xml, but that's fine).

Kharrus:
> > CPU values for weapons follow a more unified curve, with all (except mining drills) adjusted to 5,15, 45, 60, 90, 120, 350, and 500 for different weapon sizes to standardize their weight class and usage
> 
> Is there any good reason to standardize this? I feel like CPU usages should differ within weapon classes as well, there could be lore justifications and also having everything laid out in a very rigid way is just boring

I'm interested in having projectile weapons use lower CPU since they're more primitive weapons, but it didn't make it into this release since I wanted to get people's feedback on it. As for how you want to equip your ships, you can still do that however you like, and the bonuses and alterations to weapons are meant to open fitting flexibility, rather than restrict it. As a good example for that, it brings up many more options for medium and large weapons to choose from, as well as opening up more fitting choices for corvette and cruiser classes.

johann1764:
I disagree with @Kharrus' proposal related to ship roles, here is why.

> -Ship roles, particularly with capital ships, were added to both solidify the niches certain ships fill, as well as better specializing their existing talents or adding new bonuses and penalties where appropriate based on the existing ship description (example, most Dvaered ships getting CPU penalties, Za'lek Demon getting energy regen bonus).
> For the flat class bonuses, Interceptors get a 50% bonus to acceleration to help with dogfighting and to make your carrier fighters more responsive. This makes them adequately slippery, like the Hyena.

Interceptors already have the best acceleration, due to their use of the lightest engine.
If we feel this accelaration is not enough, let us increase small engine acceleration further.
I feel this acceleration is already very good, and clearly much higher than the one of other ship.

> Carriers get a hefty 40% bonus to detection, mostly to help better coordinate their fighters across vast distances, and see battleships creeping up on them.

 - This bonus ruins any chance of sneak attacking a carrier, making Yuri's Kiss much less interesting. This also makes all the stealth skill tree of bioship unusable. Preventing all these interesting ways of playing is bad IMO.
 - Your fighters _want_ to be in contact with you, therefore the distance we can see them should rely on radio contact more than detection. For simplicity, **we could always see our subordinate ships, and they should share what they see**. However, this might be hard to program.
 - "battleships creeping up on them" Not sure what you are talking about. Are you often sneak-attacked by battleships ? :-)
 - **Hive combat AI should offer a detection bonus, say 10%** (to match with its other bonuses)
 - Carriers usually have a lot of utility slots that can be used for example for sensor arrays. We can even propose **new, larger sensor arrays giving better detection bonuses**.

> Cruisers get a minor 10-25% tracking speed bonus, mostly to help with shooting down pirates like corvettes and heavy fighters. This helps specialize their common role, especially for the NPCs, as customs ships that patrol systems and burn down every last pirate they can find, large or small.

 - A 25% tracking bonus cannot be qualified as "minor".
 - The detection AI is what you need for this purpose.
 - Choosing a selection of weapons that allow to deal with a good range of threats is one of the classic interesting naev dilemnas.

> Battleships get a 20-40% bonus to weapon range, and act in the classical sense of battleships with firing the biggest, farthest-firing guns you can reasonably fit on a ship to sink the other guy first. This is very fun to watch on a Goddard.

Only battleships (and a bit carriers) can afford to use the guns with the longuest range.
Maybe all we need is **increase the range difference between ultraheavy guns and heavy guns ?**


# My Conclusion on Roles

The presented roles seem to cover actual usage of ship class.

However, I disagree with your implementation of it with ship class modifiers.
I think these roles should be consequences of more primitive facts (which is already the case, and can be reinforced)
as I tried to show you on every case.
Forcing the roles with modifiers is both redundant and really feels arbitrary.
It is also very pleasant to find unplanned ways to use a ship, which is discouraged by such modifiers.

That said, I hope that my proposals above will find interest.

johann1764:
For those who want to test cores in game, you can add the following lines to dat/tech/all_outfits.xml:
```
 <item>Milspec Aegis 2201 Core System</item>
 <item>Milspec Aegis 3601 Core System</item>
 <item>Milspec Prometheus 2203 Core System</item>
 <item>Milspec Prometheus 3603 Core System</item>
 <item>Milspec Aegis 4701 Core System</item>
 <item>Milspec Aegis 5401 Core System</item>
 <item>Milspec Prometheus 4703 Core System</item>
 <item>Milspec Prometheus 5403 Core System</item>
 <item>Milspec Aegis 8501 Core System</item>
 <item>Milspec Aegis 9801 Core System</item>
 <item>Milspec Prometheus 8503 Core System</item>
 <item>Milspec Prometheus 9803 Core System</item>
```
They will become sold in Zied.

It looks like some commit were forgotten (I saw something about where they are sold somewhere),
they don't appear anywhere in current version.

Please remove this comment when the related problem is solved.

johann1764:
> > Could we perhaps get this branch built by CI so playtesting is accessible?
> 
> It only changes data, so you can just run Naev from the directory with naev -d dat/ and it should prioritize the new data. Alternatively, you could symlink or copy dat/ into your plugins folder (it will complain about missing plugins.xml, but that's fine).

I tried both, it kind of works.. except that ship are not displayed (neither as miniature in shipyard, nor in space) and
firing weapons just "works" once. Plus a few other broken things...

This is ok for examining outfits and testing ship configurations, but not for actual testing.
I have this on the log output:
`error: shader output explicit location requires GL_ARB_explicit_attrib_location extension or GLSL 3.30`

Doing this:
`> glxinfo | grep ARB_explicit`
gives that:
`    GL_ARB_explicit_attrib_location, GL_ARB_explicit_uniform_location,`
and this:
`> glxinfo | grep GLSL`
gives that:
`OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20`

(Using naev-0.12.3)

johann1764:
Originally, this pull request were supposed to fix #2748. It also added re-using the old cores, and perform lots of changes on ships and weapons. Some but not all of these changes are documented above.

This pull request also perform some reworks of the dvaered, empire and sirius faction, and I think these topics are interesting and should be discussed.

## Undocumented Changes
### The Dvaered Ships
Dvaered tactics are oriented toward damage intensity: doing a lot of damage quickly. This is illustrated by their taste for mace rockets.
Their combat practice being mainly duels (between themselves), they use ships that are adapted to that with improved armor, which is a better option to survive overwhelmingly intense fire. Therefore, they use weapons that are more efficient with armor.
Their focus on duel and their little taste for electronics make them prefer forward weapons.
All that is consistent with their finiteness philosophy.

#### The Dvaered Goddard changes
The Dv Goddard had, for forward weapons, a +15% energy consumption and +30% damage.
It meant that, with the same amount of energy, it will convert it to damage in around -15% time,
for a total damage that is around 15% greater. Hit fast, hit hard, so dvaered.

With the new changes, it lost the first part, therefore getting a huge +30% damage bonus with no counterpart.
This change therefore both breaks the Dvaered signature modifiers on its flagship **and** breaks the balance.

#### The Dvaered  Retribution changes
The Dvaered Retribution lost its energy consumption malus and it damage bonus.
In addition to that, it gained a tracking bonus (!) for turrets (!!).
In that case, not only the dvaered signature modifiers are broken, but also some modifiers absolutely opposite to the dvaered are introduced.

#### The CPU penalty proposition
This is a good proposition (with related CPU requirement decrease for impact weapons). Maybe a positive counterpart would be nice.

#### A Remark and a Proposition
Dvaered are not easy to make nice to play: by design, they make absolutely fearsome opponents but are ill-suited for players, who tend to favor their survival over damage done. I have no solution to that but at least I have a couple of improvements we can make:
 - increase a bit their armor bonus, as it that is not that high and is part of their signature modifiers. Going up to +30% would be acceptable IMO.
 - Increase the armour value of armour structure outfits. I find them barely greater than their equivalent shield capacitors, which is unfair (and physically absurd). Should be at least 2 times greater than shields capa.
 
### The Empire Ships
#### The Empire Peacemaker
I also feel the empire lacks a battleship. Is this a good reason to turn the Peacemaker into one ? Maybe, maybe not. The resulting ship is just too powerfull, being the addition of both. Reducing the fighter bay slots were partly compensated by a bonus to fighter bay capacity, while the freed slots can be used for turrets.
A solution would be to introduce an empire battleship, but downpowering this monster also works.

#### The Empire Admonisher
It was given a +20% launch_speed and +20% launch_turn.
The idea is original and consistent with the empire faction (and with the admonisher line).
This ship lacked a little something, now it has it.

#### The (Normal) Empire Lancelot
It got -30% energy cost for forward weapons. That is weird, and too much (-15% is already a lot)

#### Other Empire Ships
Several other ship got bonuses related to turrets. This is consistent, but let us keep in mind that turret don't lack attractiveness. The idea is good but the bonuses are too high (also true for the turret-improved neutral ships).

### Other Ships
#### The Sirius Ships
For short: the speed malus and the armour bonus are part of the Sirius signature ship modifiers. They should be left as is.
The new cooldown bonus is not very interesting.

#### The Za'lek Ships
They now have small energy regen bonuses. Why not.
Energy bonuses instead would be more logical, as their energy consumption is with beam weapons that you do not use continuously.

#### The Phalanx
The phalanx has lost its agility penalties, and gained speed. (varies with Phalanx variant)
This ship was a corvette suited for bomber role, and had the related properties. Now its keeps its advantages but loses its counterparts.

#### The Kestrel
It has 2 turret tracking bonuses. Mistake ?

## The Documented Changes
I already expressed my (deep!) disagrement with ship roles modifiers. I just forgot to mention that they are really **huge**.
By the way, I checked and size-6 weapons already have +30% range compared to size-5, which makes it useless to further improve battleship range.

On corvette weapon slots promoted to medium: I think corvettes were already over-powered before the change, and it makes it even worse.
I almost never use bomber or fighters, as a fast corvette will be as fast or faster with much better firepower and shield/aromor.
I almost never use destroyers as they are just terribly slow, barely improved corvettes.
Even if some destroyers have been improved, this does not make it for the big increase of firepower of the corvettes.
As an exemple, a pirate admonisher now has the same firepower as a pirate starbridge.

On some Thurion ships, though, it makes sense to promote some weapon slots.

### The Weapons

> -Playtesting with favoring logical damage and armor penetration values between fixed and turreted guns of the same type results in an overall increase in the value of absorption as a stat in the game

Notice that, the greater the difference between armor absorption (AA) and penetration (P), the greater the efficiency of 1 point of AA:
 - if AA-P=90%, an increase of +1AA will reduce your  received damage by 10% (from 10% to 9%)
 - if AA-P=50%, an increase of +1AA will reduce your  received damage by 2% (from 50% to 49%)
 - if AA-P=0%, an increase of +1AA will reduce your  received damage by 1% (from 100% to 99%)
 - if AA-P<0%, an increase of +1AA will change nothing
 
Therefore, if we want to increase importance of AA and P, for a given size, AA-P should be positive(the greater the better).
For example, we can set it to 10%, which would require to increase all AA of all armor by 10%.

I found several inconsistencies in the weapons parameters.

As a first example, the gauss gun, a level 1 weapon, has been set a penetration of 11% (like a level 2 weapon), unlike all other weapons of the same level that are around 0.

As a second exemple, I found the turbolaser was changed from 70 damage to 180 (which is unlike anything). 
Even more, since it keeps the same energy consumption, it breaks the usual rule that bigger weapons are less energy-efficient. It gets a damage/energy ratio between the ratio of size 1 weapons and the ration of size 2 weapons.
The same happens with some other weapons, but not so spectacularly.

### The Cores

I like reintroducing the cores that what removed long ago, because it offers more choice.

However, outfits and ships parameters have been repeatedly tweaked, and the balances today are not the same.
For example, the old cores masses are not aligned with nowadays cores masses, that are balanced (or at least: no one complains about them). The energy regen values have nothing to do (more than halfed) with what they were, and these new values are adapted to the change in energy regen mecanism (now linear).

I think we could have both the benefits of past rebalancing and the old core diversity, if we adapted the old cores to the new context by guessing for example how the new Aegis would be, based on the old orion, the new orion, and the old aegis.

johann1764:
@Kharrus: We are going to reintroduce Aegis. We have an updated version of this core, but it is currently only sold in Zied. Most likely you already have ideas on where we should sell it, so if you want to do it, or just offer advice, it would be nice!

By the way, I hope my long lists of disagreements is not responsible for your withdrawal. If that is the case, I am sorry about that. Methodically listing everything I find wrong was intended at pointing out what to address in order to improve your proposition, it was not meant for discouraging you.

There is a certain number of ideas in your proposal that I would like to see included in Naev. The best would be that you do it, but if for any reason you want to delegate that, maybe I can open a PR and include you as co-author of the commits. You tell me.

bobbens:
PR have been imported on codeberg and will be handled there. Closing. Thanks!
https://codeberg.org/naev/naev/pulls

Issue #2749, opened 2025-02-04
[Proposal]: Bioship rework, updates and polishing


johann1764:
Here I do not give full arguments to defend the propositions, I will develop it as needed.

### Updating Builtin Armor

Bioships require some updates, as its builtin outfits are based on outfits that evolved significantly.

Here I focus on one specific aspect that, in my opinion has the most significant effect on the game.
In my opinion, other effects of outdated builtin outfits roughtly cancel out each other, resulting in a negligeable impact.
If needed, we can open a separate discussion on builtin outfit updating.

Here are, for ship of increasing size (1 for interceptor, 6 for carrier/battleship), the damage absorption of fully upgraded builtin plating
versus corresponding S&K combat plating.

1. 3% Vs 5% (2%)
2. 6% Vs 15% (12%)
3. 15% Vs 30% (27%)
4. 30% Vs 46% (43%)
5. 56% Vs 66% (63%)
6. 80% Vs 80% (77%)

- [x] I propose setting damage absorption rate to (S&K value)-3%, which is given between parenthesis above.
Notice it decreases the value in case of the Arx and Brigand.

Concerning the armor value, they match, except in the case case of Brigand (65GJ vs 70) and Odium (300 GJ vs 320).
- [x] I suggest realigning these ones.

By the way, the fuel provided by the engine should follow the evolutions of engines.
It was not up-to-date here, let us 
- [x] align with the new 0.13.x default fuel values. Notice Copia and Ira use a copy of the Melendez Mammoth that does not exist anymore, let us make it like the Mammoth XL.

### Polishing Some Skills

_Natural Healing I_ and _Natural Healing II_ have the same effect of adding +5 armor regeneration.
In my opinion, the effect of the first one is too strong, but the effect of the two together is OK.
In order to make the second one more attractive and the first one less attractive, I think the
better is to separate the ability to regenerate itself and the ability to regenerate quickly.
My suggestion is the following:
- [x] Make _Natural Healing I_ +3 armor regeneration (like bioarmor structural outfit) and _Natural Healing II_ +7 armor regeneration.

Skills _Cannibal I_ and _II_ are a bit weird, as disabling ship is clearly not what bioship are good at (due to their builtin weapons not made for that)
I do not see an easy way to balance that, but here are two suggestions:

[...]

PR #2755, opened 2025-02-22
Bioship Updates, Polishing and Improvements

johann1764:
**Balance**

This PR addresses the issue described in #2749

Any question/remark/disagrement on any detail is welcome! Here I tried to sum up things.

Changes that were not discussed in issue 2749 are marked with "(NEW)".

## Summary
I clarified how bioships and related core outfits are built.

As much as possible, I tried to give an explicit rule of how they are built, while at the same time keeping as close as possible to the current values,with the notable exception of damage absorption values that were unfairly low for no reason and that I realigned with standard levels.

For example, cerebrum are a bit better than standard core, and I justified it and paid the price with heavier and more expensive cores. This cancelled out partially with armor updates that reduced their weight. This can be discussed, taking the cerebrum recipe as a basis, which is better than discussing raw numbers.

Also, it occured that some bioships had inadequate gene drives, these were switched to more adapted ones.

A few additional tweaks are proposed.

### Bioship Ship Parameters & Modifiers

Applied bioships individual proposed changes (see issue 2749).
Also given a normal weight to Revenant (was far too light), compensated by core updates (see below).
Documented soromid general & specific ship modifiers in dat/ships/soromid/README.md

Instead of doubling experience needed to level up for the Revenant, made any non-soromid bioship gain experience twice mode slowly (each experience gain is halved).

Decreased Revenant Bioweapon maxlevel to Talon II, losing damage and range, gaining tracking and power efficiency.
This is more adapted to the Revenant, and also gives it a clean upgrade scheme. (NEW)

Soromid ships have twice less utility slots than equivalent ships. The Arx should have 2.5 large slots, and has 2. I turn the remaining 0.5 large slot into a full medium slot. (NEW)

I filled one of the free weapon slot of the Reaver with another instance of its bioweapon (stinger). (NEW)

### Bioship Core Outfits & Weapons

Updated cortex and gene drives. In particular, repaired unfair low armour absorb values.
Left the "recipe" of these in comments in generate.py. Introduced new gene drives called Laeviter, Grandis and Magnus. Relations between core bio-outfits and core outfits can now be listed with
 > grep '=>' dat/outfits/bioship/generate.py

The Arx switches its engine to an equivalent of the Typhoon II, now matching its description. The Ira switches its engine to an equivalent of the Typhoon. The Reaver switches its engine to an equivalent of the Zephyr II, now matching its description.

Updated Cerebrum and bioweapons (with new related recipes).
There were serious issues with size 5 bioweapons. They are much less powerfull and energy-consuming now. Ira and Copia now have tentacle organ max level at II instead of III, had to change level up scheme a bit.
Globally penetration have been decreased for the large bioweapons. 

Updated costs of core outfits increase bioship prices significantly, which makes sense as you don't need to buy these outfits.

#### Sidenote
As a refinement of Issue 2739, added +1Jump to Zephyr II, Cyclone II and Typhoon II, and related gene drives. Resulting new rule:
 - Nexus/Unicorp engine: _size_+3 Jumps
 - Tricon: the same with -1 penalty
 - Melendez: the same with +1 bonus

### Skills
 - Armor 1&3 Natural Healing I&II planned changes
 - Bite 3 - Cannibal II planned modification.
 - Stealth 1 - slightly improved Compound Eyes from +15 detect to +20 (NEW)
 - Stealth 3 - Renamed "Ambush Hunter" as "Ambush Hunter I"
 - Stealth 4 - Silence added effect: -15% stealth range on stealth mode. This allows to get closer to ships while in stealth mode but does not help entering stealth mode. (NEW)
 - Stealth 5 - Ambush Hunter II adds +30% weapons range 10 secs after destealthing. This is supposed to mimic Yuri's Kiss style. Opened the Stealth skill tree to the Ira. (NEW)

#### Cannibalize new behaviour
 - either gluttony : eat all even if useless.
 - or cannibalize : eat no more than needed, leave at least 1AP.

Gluttony is proposed either when (1) the boarded ship belongs to a _hated_ faction (-30 or less) or when (2) the player has nothing to gain in cannibalizing. If player cannibalizes, the interface swaps to gluttony mode, as the condition (2) is now met. In case of cannibalize (resp. gluttony), the players now loses standing for _looting_ (resp.  _destroying_). The gluttony immediately closes the board window. At each step, the player is informed of the potential consequences.


johann1764:
Could you please **Delete** the broken version of this PR #2752 ?

bobbens:
> Could you please **Delete** the broken version of this PR #2752 ?

I don't think it can be deleted. It's closed, is that not good enough?

johann1764:
Testing with naev-0.12.3-350 looks ok.

Should I substitute :
```
dat/outfits/bioship/skills/ambush_hunter.xml`
```
with
```
dat/outfits/bioship/skills/ambush_hunter_i.xml
dat/outfits/bioship/skills/ambush_hunter_ii.xml
``` 
in every po file, or are these generated from a file I already updated ?

johann1764:
Ready for review ! (I think)

dat/board.lua
```diff
@@ -454,6 +454,36 @@ You will still have to escort the ship and land with it to perform the repairs a
       end )
 end
 
+local function board_fcthit_apply ( eat )
+   -- Piss off nearby ships
+   board_plt:distress( player.pilot() )
+
+   local fct = board_plt:faction()
+   local loss = math.floor(-fct:hit( -board_fcthit, system.cur(), "board" ))
```
bobbens, line 462: `math.floor` is being used with one parameter, so it is a NOOP. Probably should be removed.

dat/board.lua
```diff
@@ -454,6 +454,36 @@ You will still have to escort the ship and land with it to perform the repairs a
...
+         return
+      end
+      msg = fmt.f(_("You have lost {fcthit} reputation with {fct} for looting this ship!"),
+         {fcthit=fmt.number(loss),fct=fct})
+   else
+      if loss == 0 then
+         msg = fmt.f(_("You have cannibalized this ship, gaining {eat} armor. Another cannibalize would destroy it."),
+            {eat=fmt.number(eat)})
+      else
+         local loss_d = math.floor(-fct:hitTest( -board_fcthit, system.cur(), "destroy" ))
```
bobbens, line 477: Again, `math.floor` is probably unnecessary here.

dat/board.lua
```diff
@@ -454,6 +454,36 @@ You will still have to escort the ship and land with it to perform the repairs a
...
 
+local function board_fcthit_apply ( eat )
+   -- Piss off nearby ships
+   board_plt:distress( player.pilot() )
+
+   local fct = board_plt:faction()
+   local loss = math.floor(-fct:hit( -board_fcthit, system.cur(), "board" ))
+   local msg
+
+   if eat==0 then
```
bobbens, line 465: eat should probably be a bool

dat/board.lua
```diff
@@ -454,6 +454,36 @@ You will still have to escort the ship and land with it to perform the repairs a
...
+      board_fcthit = 0
+      if loss == 0 then
```
bobbens, line 467: Probably do `loss <= 0` in case something goes wrong and it creates a negative loss.

dat/board.lua
```diff
@@ -470,31 +500,86 @@ local function can_cannibalize ()
    return false
 end
 
-local function board_cannibalize ()
+local function can_cannibalize_usefully()
+   local pp = player.pilot()
+   local parmour, _pshield, _pstress = pp:health(true)
+   local ba=board_plt:stats().armour
+
+   if (pp:stats().armour == parmour) then
```
bobbens, line 508: I take it this is to check full health?
Something like `pp:armour() >= 100` would probably work better.

dat/board.lua
```diff
@@ -470,31 +500,86 @@ local function can_cannibalize ()
...
+      return false
+   elseif (ba<=2) or (ba==3 and player.shipvarPeek("cannibal2")) then
```
bobbens, line 510: I don't understand the check for `ba==3`. This is an armour value right? Having exactly 3 armour is probably not going to trigger ever.

dat/board.lua
```diff
@@ -470,31 +500,86 @@ local function can_cannibalize ()
...
+      player.msg(fmt.f(_("Your ship cannibalized {armour} armour from {plt}."),{armour=fmt.number(heal_armour), plt=board_plt}))
+      board_fcthit_apply( heal_armour )
+   end
+
+   local left=armour-dmg
+
+   if left<=0 then
+      local fact=board_plt:faction()
+      player.msg(fmt.f(_("{plt} was destroyed."),{lft=fmt.number(left), plt=board_plt}))
+      fact:hit( -(board_plt:points()), system.cur(), "destroy")
```
bobbens, line 567: no need for parenthesis around `board_plt:points()`

dat/board.lua
```diff
@@ -470,31 +500,86 @@ local function can_cannibalize ()
...
+      board_close()
+   else
+      player.msg(fmt.f(_("{plt} was left with {lft} armour. Next time it won't !"),{lft=fmt.number(left), plt=board_plt}))
```
bobbens, line 570: Missing a verb no?
`Next time it won't survive!`
or
`Next time it will be destroyed!`

dat/board.lua
```diff
@@ -470,31 +500,86 @@ local function can_cannibalize ()
...
+   --   -- Does not work! How to re-build the board window ?
+   --   board_close()
```
bobbens, line 572: You could do something like change `board` so it is:

```lua
-- internal boarding function
function _board( plt, reload )
   -- paste current board function here, but disactivate sound and other effects if reload==true
end

function board( plt )
   return _board( plt, false )
end
```

Then you could use `_board` to regenerate the window.

dat/board.lua
```diff
@@ -611,59 +695,93 @@ function board( plt )
...
          btn_capture:disable()
          btn_capture:setAlt( msg )
       end
    end
+
+   local eat_mode
+
    if can_cannibalize() then
-      luatk.newButton( wdw, x-50, h-20-30, 130, 30, _("Cannibalize"), board_cannibalize )
+      --player.msg(fmt.f(_("faction {fac} reputation {rep}."),{rep=fmt.number(board_plt:reputation()), fac=board_plt:faction():nameRaw()}))
```
bobbens, line 725: Probably should be deleted.

dat/board.lua
```diff
@@ -611,59 +695,93 @@ function board( plt )
...
+         luatk.newButton( wdw, x-30, h-20-30+yoff, 110, 30, _("Cannibalize"), board_cannibalize )
+      end
       --x = x-100
+   else
+      eat_mode=0
+   end
+
+   local or_eat=""
+   if eat_mode==1 then
+      or_eat="(or cannibalizing) "
```
bobbens, line 740: Needs to be translatable. `_("(or cannibalizing)")`

dat/board.lua
```diff
@@ -611,59 +695,93 @@ function board( plt )
...
+
    board_fcthit_txt = luatk.newText( wdw, 20, 40, w-40, 20, fctmsg )
-   local fh = board_fcthit_txt:height()
-   local dfh = luatk._deffont:getLineHeight()
-   if fh > dfh then
-      h = h + fh-(30-dfh)
-      wdw:resize( w, h )
-   end
+
+   --local fh = board_fcthit_txt:height()
```
bobbens, line 768: Probably should be removed this comment block.

dat/outfits/bioship/skills/silence.lua
```diff
@@ -0,0 +1,18 @@
+
+function onstealth( _p, po, stealthed )
+   if stealthed then
+      po:state( "on" )
+   else
+      po:state( "off" )
+   end
+end
+
+function update( p, po )
```
bobbens, line 10: Strictly speaking, it shouldn't need an update function, should it? Just catching the stealth event should be enough.

dat/outfits/bioship/skills/silence.lua
```diff
@@ -0,0 +1,18 @@
...
+   onstealth( p, po, p:flags("stealth") )
+end
+
+function init( _p, po )
+    --po:clear()
+    po:set( "ew_hide", -15 )
```
bobbens, line 16: Why set it here? This should be equivalent to defining it in the XML no? Because it is never turned off.

dat/outfits/bioship/skills/the_bite.lua
```diff
@@ -156,9 +173,8 @@ function update( p, po, dt )
             -- TODO better calculation of asteroid mass
             p:knockback( 1000, t:vel(), t:pos(), 0.5 )
             -- Do the healing
-            if mem.improved then
-               local heal = 0.25 * dmg
-               p:addHealth( heal )
+            if mem.regen~=0.0 then
```
bobbens, line 176:
```suggestion
            if mem.regen > 0 then
```

dat/outfits/bioship/skills/the_bite.lua
```diff
@@ -183,9 +199,8 @@ function update( p, po, dt )
             t:damage( dmg, 0, 100, "kinetic", p )
             t:knockback( p, 0.5 )
             -- Do the healing
-            if mem.improved then
-               local heal = 0.25 * (ta - t:health(true))
-               p:addHealth( heal )
+            if mem.regen~=0.0 then
```
bobbens, line 202:
```suggestion
            if mem.regen > 0 then
```

dat/scripts/bioship.lua
```diff
@@ -57,7 +58,7 @@ local function _getskills( p )
    if pss <= 3 then
       table.insert( skilllist, "move" )
    end
-   if pss <= 4 then
+   if pss <= 4 or ps:baseType()=='Ira' then
```
bobbens, line 61: Would probably be cleaner to do this with tags. Like add a
`bioship_skills_stealth` tag to the Ira and check to see if it has the tag. Similarly, instead of using sizes, tags could be used for all the skillsets which may be cleaner and more flexible.

dat/ships/pirate/pirate_revenant.xml
```diff
@@ -17,13 +17,17 @@
...
+ <description>As Skull and Bones does not have the technology to grow new bioships from scratch they salvage stolen Soromid ships. While the adaptation process can be seen as cruel from an anthropomorphic point of view, leading to some conflicts with anti-animal cruelty pirate organizations, advocates argue that the process does not cause pain nor significant stress to the ships. The modifications do not target internal organs, and focus on redoing the externals, and there are many stories of pirates that have developed strong bonds with their bioships.
+
+The total experience the Revenant can get is only half of what its soromid counterpart can, but it also gains twice less experience.
```
bobbens, line 22: soromid should be capitalized.

dat/ships/pirate/pirate_revenant.xml
```diff
@@ -17,13 +17,17 @@
...
+Its free medium weapon slot and the possibilty to use any armor makes its significantly more flexible.
```
bobbens, line 23:
```suggestion
An additional medium weapon slot and the possibility to use any armour makes it significantly more flexible.
```

bobbens (changes requested):
Overall great! I can tell a lot of work went into it. Left some comments and areas for improvement here and there. I haven't tested it yet, though, but will next iteration.

dat/board.lua
johann1764: It is also used as a value when non-zero.

dat/board.lua
johann1764: I'll try tomorrow.

dat/board.lua
johann1764: Integer-based reasonning. It was before I realized there were no integer types in lua (I don't speak lua).
I need to reconsider this part of the code.

dat/outfits/bioship/skills/silence.lua
johann1764: Apparently not.

dat/board.lua
johann1764: It (together with next condition) ensures that, in the following:
```
      if cannibal2 then
         dmg = math.min( (armour-1), 3*(ps.armour-parmour)/2 )
      else
         dmg = math.min( (armour-1), 2*(ps.armour-parmour) )
      end
```
dmg will be strictly greater than 0. As these function work together, 
I'd prefer to keep clear how they relate.

dat/scripts/bioship.lua
bobbens: I would prefer pretty much anything other than matching ship name... It just is too fragile.
I would probably just add something like `bioship_skills_stealth`, and check for that, only in this case.

dat/board.lua
bobbens: Newer Lua supports integers, but we're stuck at 5.1 due to LuaJIT.

That said, Lua does do magic so that it **should** work.

My point is that the code is doing (relevant parts only):

```lua
local ba=board_plt:stats().armour
if ... then
elseif (ba<=2) or (ba==3 and ...)
... -- This will pretty much never trigger
else
end
```

So `ba` is set to the player's maximum armour, which is usually going to be some big number and not `<=2` or `==3`.  Am I missing something?

johann1764: How to make sure something can be gained from cannibalizing while leaving at least 1 ?
The exact answer requires simulating an attempt.

Instead, let us consider that it is not worth eating less than 1, and therefore the condition is `ba<2`.

This works, as it guarantees that, after cannibalizing, `can_cannibalize_usefully()` will return `false`.

It is not optimal, as, say `ba<1.5` would also work in strictly more cases.

Let us say this is optimally simple.

dat/scripts/bioship.lua
johann1764: I'll do that.

dat/board.lua
bobbens: I see. Maybe make it more clear with a comment then? Thanks for the explanation.

dat/outfits/bioship/skills/silence.lua
bobbens: By default, `po:set()` should replace.

dat/board.lua
johann1764: How. Just saw your message. I made a mistake, as ba should be the remaining armor.
The condition should mean "If ship is already so damaged that that is nothing left to eat"

bobbens: Ah, then it should be `parmour < 2`

dat/board.lua
```diff
@@ -503,11 +503,11 @@ end
 local function can_cannibalize_usefully()
    local pp = player.pilot()
    local parmour, _pshield, _pstress = pp:health(true)
-   local ba=board_plt:stats().armour
+   local ba, _shield = board_plt:health(true)
```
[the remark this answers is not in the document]
johann1764: That's exactly what I thought _after_ pushing. Will be next commit.

dat/board.lua
johann1764: done

dat/scripts/bioship.lua
johann1764: done

bobbens:
I think it looks great, thanks! I've done some minor fixes, such as using `pre-commit` to fix some missing translation entries and minor formatting issues. Thanks a bunch!

Issue #2753, opened 2025-02-19
[Bugs]: ambush hunter


johann1764:
### Bug 1

When you obtain skill ambush hunter, nothing odd happens. But then, if you take off and land, your ship now has a +50 weap dam.

To make sure nothing were wrong with my savegame, I followed this procedure:
 - create a new game
 - buy a brigand
 - take off with it
 - get paid a lot
 - land and upgrade the stealth branch up to stealth 3 - ambush hunter.
 - take off
 - land

At that point your ship has permanent +50% weapon damage. 

### Bug 2

If ambush hunter is active (visible icon) when you land, 
the effects (and icon) will remain permanently when you take off, until you re-activate it by stealthing on and off.

### More General Questions
On this example, we see that certain effects are not automatically removed when landing.
Is that desired, and desirable ?

In the bugfix I propose (#2756), effectRm() removes effects but not icon, is that a bug ?

### Notice

This bug is the main obstacle to completion of #2755.

### Version

0.12.3

### What operating system are you seeing the problem on?

Linux

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
Thanks for the report. Yes, the GUI not updating was a bug that I fixed. Should be working better now with your merged PR. Thanks!

PR #2756, opened 2025-02-23
ambush hunter bugfix

johann1764:
**Bug Fix**

This PR addresses the bugs described in #2753

## Summary
Bug 1 was fixed by removing outfit mods, as it is the effect that is supposed to add modifiers, not the outfit.

Bug 2 was fixed by always removing the effect when landing.
I couldn't think of another event that would interrupt the effect.

Notice that removing the effect does not removes the icon, 
and here I assume it is another bug.

Actually, it is possible to "fix" that by adding a new effect at take off (init) that adds a new effect that overwrites this one with a duration of 0. But I think it is better to fix the bug than to hide it in a dirty way.

## Testing Done
Yes, I did.

### DIFF
--- a/dat/outfits/bioship/generate.py
+++ b/dat/outfits/bioship/generate.py
@@ -11,7 +11,7 @@ N_ = lambda text: text
 class Build:
     ''' By default, build everything in-tree. '''
 
-    def read_tempate(self, name):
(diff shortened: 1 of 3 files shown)


bobbens:
Looks good, thanks!

Issue #2747, opened 2025-01-28
[Bugs]: dvaered scenario fw02_escape


johann1764:
PR #2760  aims at fixing it.

### Most obvious problem

After freeing the VIP in Hargen, you have far enough time to land in minerva,
and go to the hospital there. (EDIT: Can even go as far as empire space.)

Fleing after that is trivial, as you are already out of the blockade,
and the final dialogue mentions Minerva as if it were a Za'lek planet.
It also mentions an encounter with captain HewHew that did not happen.

### Unmanaged ways of fleing.

There are several different ways to go back to Dvaered space, some of which are legitimate
in my opinion. 

- With the right ship and outits and intrinsic outits, you can (barely) go through the blocade.
- You can use the hypergate in Ruadan or Gamma Polaris. I think this should be prevented with a blocade around hypergates.
- Going to Kretogg the long way through empty systems in the "south".
- Using jumps the Empire and the Za'lek don't know about (Kretogg hidden jumps or Quorelia hidden jumps) or wormholes which is legitimate from my POV.

In some cases, you don't encounter captain HewHew and this should be managed in the final dialog.
(Something like "But How the hell did you manage to escape ? Seing you would rather not answer, he continues : Anyway,...")

### An actual bug
One of my tests made me land to meet either the pirate or the empire agent.
I met none, took off and could go to Goddard without any blocade on the way.

### Version

0.12.3

### What operating system are you seeing the problem on?

Linux

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2760, opened 2025-03-01
Dvaered FW-escape mission

johann1764:
**Bug Fix**

This PR addresses the bugs described in #2747.

Several problems seem to be caused by the map change years ago.
Unfortunately, the fix I propose requires a small map change (an edge flip to be exact).

It occurs that pirate6 is broken, in part because of recent map changes.

A few additional bugs were found and fixed.

## Summary

 - [x] End mission dialog : Now the case Captain Hewhew was not encountered is managed.
 - [x] Missing Blocades (included Waterhole)
 - [x] Blocade of Hypergates
 -  Fixing mission stage 2 : Complicated, see below

### Fixing mission stage 2

When you free the VIP, after one jump you are given 3 periods to land.
In order to make the scenario work, the player is expected to land in Za'lek space.

(The other altenative would be to change the interception system, but no suitable system was found for that.)

To make that happen,
- [X]  this delay had to be reduced to less than 4500s (minimum jump time), and **1/4 period** was chosen to feel at least a bit challenging. As a consequence, had to:
   - [x]  add a bar in Bastion Center. Current bar description is very basic, would require improvement by someone else.

Making the rest work requires:
- [x]  replacing the jump from Pultatis to Limbo by a jump from Pultatis to Sollav. 
   - [x] Fix minerva mission 1 accordingly.
   - [x] Fix minerva mission 2 accordingly.
   - [x] Fix minerva mission 6 accordingly.
- [x]  moving the blocades one system away from Pultatis.

#### kex3
I discovered that minerva mission kex3 had a bug: when taking off from Trincea, the thug spawn at a bad position, therefore it is easy to jump out without triggering their hook, provoking an error in the new system.

This is likely caused by the changes in urbanus by commit "Added House Yetmer and House O'rez".
 - [x] Avoid triggering error if player jumps out without triggering the hook.
 - [x] Adjust Thugs spawning position and Trincea position to make it work.

#### fw-escape
Some ship attacked Shaefer while he was bringing some fuel. The bug is hard to reproduce (usually don't happen)

I could not find out why. Pirates can be ruled out (clearPirates).

- [x]  Proposed a workaround.

## Testing Done
- [x] fw-escape (yet another time)
- [x] Minerva missions pi2 to pi6

### DIFF
--- a/dat/missions/dvaered/frontier_war/fw02_escape.lua
+++ b/dat/missions/dvaered/frontier_war/fw02_escape.lua
@@ -79,7 +79,7 @@ local zlk_list = { -- Systems with patrols
    system.get("Pultatis"),
    system.get("Stone Table"),
    system.get("Xavier"),
-   system.get("Straight Row")
+   system.get("Straight Row"),
 }
 local zlk_lisj = { -- Index refers to zlk_list
    {system.get("Provectus Nova"), system.get("Limbo")}, -- from Pultatis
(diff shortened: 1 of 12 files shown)


johann1764:
Can I proceed to the modifications of the star map that I mention ?

PR #2773, opened 2025-03-07
Outfits2md bugfix & improvement 

johann1764:
Outfits2md
 - now correctly manages minimal column width;
 - now correctly manages best/worst value using a set called _lower_is_better_;
 - now processes the commandline correctly and issues warnings;
 - now offers three output variants:
   - the default text-aligned valid md output,
   - a smaller md output giving the same result, but not text-readable,
   - a color terminal output with multi-line header.
  
screenshot


bobbens:
Looks good, thanks!

Issue #2774, opened 2025-03-08
[Proposal]: autonav improvements


johann1764:
### Suggestion

Pull Request #2775 have solved some issues, the other ones are listed with task lists.

At some point in the future, I would like to implement a few autonav improvements:

 - Similarly to _"Jump to follow"_ setting, I would like to implement a _"Land to follow"_ setting.
      - Is there any reason it could be a bad idea ?
      - Other question: should it be a separate setting, or should we merge the two options into one single _"Jump or land to follow"_ setting ? I can't think of a situation where one would want one but not the other.
 - Solve the afterburner bug:  see #2775.
 - If _Land to follow_ is off, and _Brake at approach_ is true, the player following a ship that lands should stop at the planet position.; the same goes with _jump to follow_ and a followed ship jumping.
 - Turn off afterburner before:
      -  a jump, except if the ship has _instant jump_; (#2781)
      -  landing; (_idem_)
      - [x]  cooling down (CTRL+S).
         - [x] Camera bug when CTRL+S.
 - [x] wait until all the pilots you are following jump/land before jumping/landing.

Feel free to agree or disagree with any point, and to propose other relevant improvements.
_Any opinion welcomed!_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
I think "Jump and land when following" would be a good option. The thing that should be improved is that it should wait until all the pilots you are following jump. That will be a bit tricky, but I guess something could be done to handle that.

Regarding turning off afterburner, doesn't that happen already? The cool down / landing code should force the player's outfits off. Not sure how the jump handles it though.

I like the idea of having land to follow + jump to follow with brake at approach to brake when they are gone. I guess it should start braking + ramping down when they are lost and then stop autonav when fully braked?

johann1764:
> I think "Jump and land when following" would be a good option. The thing that should be improved is that it should wait until all the pilots you are following jump. That will be a bit tricky, but I guess something could be done to handle that.

I used Jump to follow when playing 0.12.x in every mission where it is relevant to, and I never had a problem.
(in shadow vigil, I follow the four winds excort that follow the diplomat..)

> Regarding turning off afterburner, doesn't that happen already? The cool down / landing code should force the player's outfits off. Not sure how the jump handles it though.

I just re-tried to be sure, but even with:
`Naev v0.13.0-alpha.1+697.ga7f06b275a (linux-x86_64)`
the afterburner does not turn off in either case.

[...]

PR #2775, opened 2025-03-11
Autonav

johann1764:
**Bug Fix** / **New Feature**

This PR addresses _some_ of the bugs/features described in #2774.

## Summary
 - [x] If _Jump or Land to follow_ is off and If _Brake at approach_ is on, the player following a ship that is lost due to landing or jumping should stop at the  position where the target was lost.
    - [x] In order to do that, the player should ignore lanes (land).
    - [x] In order to do that, the player should ignore lanes (jump).
 - [x] If can't follow a pilot in a jump, stop autonav and issue relevant message:
    - [x] if unknown jump or exit_only;
    - [x] if out of fuel.
 - [x] Extend _"Jump to follow"_ into _"Jump or land to follow"_.
    - [x] make substitutions and update UI.
    - [x] land where the pilot landed.
 - [x]  Solve **the afterburner bug** : you get into a system with your autonav on that should redirect you to your next jump, but if you hit the afterburner too early, the ship just continues with the orientation it had while entering. This was caused by not taking afterburner accel bonus into account while testing whether current speed was unreachable by normal means.
 - [x] When we follow a pilot with jump_to_follow on an the pilot is lost, fast forward mode is stopped, as there is usually not enough time to ramp down before jumping or landing if the pilot was followed closely.
 - [x] When we follow a pilot with jump_to_follow off but brake_at_pos on, when the pilot is lost, we stop fast forward mode (ramping down does not work convincingly) and go to the last position the pilot was seen. If he has jumped (resp. landed) the jump (resp the spob) is selected.
 - [x] `autonav_abort()` with _Brake at approach_ on should stop the ship before stopping autonav.

## Testing Done
Each feature, one at a time. A validated feature has been tested.

### Don't forget
 - [x] Comment out debug messages when done.


dat/events/settings.lua
```diff
@@ -194,7 +194,7 @@ function settings ()
          var.push( "autonav_compr_speed", compr_speed )
          var.push( "autonav_compr_max", compr_max )
          var.push( "autonav_match_fleet", match_fleet )
-         var.push( "autonav_follow_jump", follow_jump )
+         var.push( "autonav_follow_land_jump", follow_land_jump )
```
bobbens, line 197: I would probably not rename the variable name, just keep it at `"autonav_follow_jump"` or we'll have to add logic to update the variable names and stuff. The local variable can be called `follow_land_jump` though.

johann1764: Do you mean update it in savegames ?

bobbens: Yes. It's just a bit more trouble that we can easily avoid :)

johann1764: Sure. Next commit !

bobbens:
It looks good. I would remove some of the commented debug prints and such, but I think it's ready to merge. I assume this is not a draft anymore?

bobbens (approved):

johann1764:
I could not find a way to do the last task of the list.

I will leave a comment in the code about it, and remove the remaining debug messages.

I will also update the CHANGES.

I will mark this PR as ready when done.

johann1764:
I didn't try doing a new function, as I though I might be redundant with an existing one.
But since no existing func works for that, I will try it !

bobbens:
> It worked!
> 
> Now I am having a doubt whether these `_("...")` I added are OK, since the function that receives it also applies `_()` on top of it. Should I replace my `_("...")` them with just `"..."` ?
> 
> If not, and if luacheck test succeeds, It should be ok.

The way gettext works is you want to apply it to the source strings, or they won't be captured in the translation software. So something like:
```lua
local foo = _("bar") -- Translated here
print( foo )  -- Translated version is used
```
Is what you want. In the case you have to use a variable for some reason or other to translate later, you should do something like:
```lua
local foo = N_("bar") -- This is actually a noop, but gets translated
print( _(foo) ) -- Now it will properly get converted
```

What would be wrong would be something like:
```lua
local foo = _("bar") -- Translated here
print( _(foo) ) -- Translated version is translated again, could become something weird...
```
The string is getting translated twice, which means that it could potentially get converted into something weird.

johann1764:
Ok, just this last correction and we are good.

johann1764:
Ok, fixed my mistake. 
Some instances of that mistake were not my doing, I also fixed them. (system name and pilot name could be translated twice)

Now _luacheck_ will tell whether we are ok.

dat/autonav.lua
```diff
@@ -633,8 +659,9 @@ end
 -- Going for the landing approach
 function autonav_spob_land_brake ()
    local ret = ai.brake()
+
    if player.tryLand(false)=="impossible" then
-      return autonav_abort()
+      return autonav_abort("cannot land")
```
bobbens, line 664:
```suggestion
      return autonav_abort(_("cannot land"))
```
Should be translated, as this is used in a format string which is translated before formatted.

dat/autonav.lua
```diff
@@ -656,24 +683,50 @@ function autonav_plt_follow ()
...
+            local fuel, consumption = player.fuel()
+            if jmp:known() and not jmp:exitonly() and fuel >=consumption then
+               pp:navJumpSet( jmp )
+               _autonav_system(false)
+               autonav_reset(0)
+            else
+               local why=nil
+               if fuel < consumption then
+                  if fuel == 0 then
+                     why="no fuel"
```
bobbens, line 699:
```suggestion
                     why=_("no fuel")
```

dat/autonav.lua
```diff
@@ -656,24 +683,50 @@ function autonav_plt_follow ()
...
+                  else
+                     why="not enough fuel"
```
bobbens, line 701:
```suggestion
                     why=_("not enough fuel")
```

dat/autonav.lua
```diff
@@ -732,7 +785,7 @@ function autonav_plt_board_approach ()
    if brd=="ok" then
       autonav_end()
    elseif brd~="retry" then
-      autonav_abort()
+      autonav_abort("cannot board")
```
bobbens, line 788:
```suggestion
      autonav_abort(_("cannot board"))
```

dat/autonav.lua
```diff
@@ -792,8 +845,7 @@ function autonav_enter ()
       -- Must have fuel to continue
       local fuel, consumption = player.fuel()
       if fuel < consumption then
-         player.msg("#r".._("Autonav: not enough fuel to continue").."#0")
-         autonav_end()
+         autonav_abort("not enough fuel to continue")
```
bobbens, line 848:
```suggestion
         autonav_abort(_("not enough fuel to continue"))
```

bobbens (changes requested):
OK, so I mentioned double translated, this doesn't happen with format strings. So something like:
```lua
local foo = _("bar")
print( fmt.f( _("hoge {foo}"), {foo=foo} )
```
Is valid. It translates the first string and saves that as a variable, then it translates `"hoge {foo}"` where the `{foo}` part should remain unchanged, and then the translated `{foo}` gets replaced by the translated local variable string.

johann1764:
I think there is a couple occurences left in "get_..._name", no time now, do it later.

johann1764:
Looks like it's ok.

bobbens:
Looks good, thanks!

Issue #2766, opened 2025-03-05
[Proposal]: merge core slots


bobbens:
### Suggestion

# Motivation

One of the annoying issues with developing and balancing in Naev is the core slots. There are 6 ship sizes (ultra-light, light, medium, medium-heavy, heavy, ultra-heavy), and with 3 core slots that means with 3 variants per core slot you immediately have to define and balance 54 outfits. Most outfits follow curves and recipes, but since they are handled manually it is a lot of work. This makes it a pain to handle spreadsheets and try to balance all the different combinations. It is also really annoying to give the different techs and make sure they are well distributed, and it bloats the different outfitter stores.

Furthermore, there is relatively little choice in the outfitting. Although a Fighter can fit ultra-light cores, you are usually forced to equip the light cores due to the high base mass, more slots, and higher mass of other outfits. Similarly, an interceptor with light cores will just be a worse fighter. There are only very few cases where you can force a ship to either upscale a core, or downscale a core to make the ship light enough. For example, my Za'lek Sting uses the light stealth armour with medium thalos + engine cores. However, that is more of an exception than a rule.

# Proposal

The proposal is fairly straight forward. Collapse both variants of same slot size into a single outfit, so, both small Thalos cores would become a single Thalos small core. This would effectively halve the number of cores needed, making it easier to maintain and develop, and also more transparent to the player. Now a small slot would use small cores, no need to consider ultra-light vs light.

The main downside is that this merging, by itself, would effectively make the smaller variant (ultra-light, medium or heavy) worse than the larger variant (light, medium-heavy, or ultra-heavy), because they would share slots, but the larger version would have bigger bonuses. This could be solved by giving ship stats to compensate. So now, a fighter would get something like +25% health / energy / cpu and -25% movement (numbers would have to be tweaked so they are similar to the differences between the ultra-light and light cores now).

## Advantages

* Makes it much easier to tweak outfit locations and balance outfits themselves from a development point of view.
* Much easier to understand for the player, no need to have to understand things like ultra-light and light distinctions, also much easier to visualize in the outfitter.
* Makes it so that unique outfits are usable by 1/3 of the ships (same size) instead of 1/6 of the ships.
* Makes core slot choices more meaningful and lets us add more core outfits with different properties without having to slog through adding a ton of different variants.
[...]

PR #2777, opened 2025-03-15
Multicore - core systems

johann1764:
**New Feature**

This PR addresses part of the feature described in #2766.
  (only the _current_ core systems in this PR).

## Summary
- Maybe resurrect Prometheus and Aegis core systems, update them and apply them the procedure ? It could be redundant with #2750, but should be done at some point. -> next PR with hull cores.
- [x] Automatically convert a pair of matching outfits of respective sizes 2k and 2k+1 into an extended outfit of size 2k.
- [x] Automatically convert an extended outfit into {xml,lua} files that work with the game. The prices were finally computed with the arithmetic average.
- [x] Automatically mark the used 2k+1 outfit name (that has been usurped by the new outfit) as deprecated.
- [x] Generate the list of pairs of matching outfits of respective sizes 2k and 2k+1 ...
- [x] ... and process them. (no need to know which one is the small one: multicore.py can manage when they are reversed)
- [x] Tried the mixed xml/lua approach. This mitigates greatly most problems (mass and broken neutral ships), but the description of a secondary core is now a copy of a primary plus a list of modifications, which is not pretty.
- [x] Manage the paths more cleanly in the scripts.
- [x] Implement something to find the class of a ship that inherits from some other ship.
- [x] Automatically add multicore slot to all ships of size 2k -> class-based, "bulk freighter" considered as size 6.
- [x] Update ships default outfits -> use the addslot script.
- [x] Remove deprecated 2k-sized core systems from tech.
- [x] Manage solo outfits that will become useless as secondaries
   - [x]  Multicores.py should manage the case it has one arg as arg as main, and 0 as secondary.
   - [x]  Couple outfit processing should also process solo outfits with multicores.py.
   - [x]  Manage the S/M/L _previous generation core systems_.
- [x] Convert savegames.
   - [x] Replace 2k cores with 2k-1 cores in core main slot, and
   - [x] Offer a free 2k-1 core for the core secondary slot.
- Some cruisers (Pythagoras, Hawking, Retribution) are initialized with size 6 core system. Maybe we can consider them as heavy cruiser (size 5.5) and give them 1 large utility core slot + 1 medium utility secondary core slot. Or maybe this is just a mistake. -> Actually, this concerns all cruisers but the kestrel. Let us do nothing.
   - [x] In any case, they should be initialized with a size 5 core system.
- [x] Discard all size 2k core systems (mark them as deprecated)
- [x] Update dat/scripts/equipopt/cores.lua accordingly.
- [x] Figure out why ships don't get their secondary core system slot equipped.
- [x] Revert mixed xml/lua.
- [x] Manage the _homebrew processing unit_ -> made it totally useless as secondary.
- [x] Update `/my_naev/dat/scripts/equipopt/templates`
- Fix `WARNING [...] Outfit '[...]' not found in stack.`. -> Should not happen once merged.
- [x] Remove last references to deprecated outfits in missions and events
- [x] Keep the Multi-Valued-Xml (.mvx, not in Wikipedia's ext. list) together with the generated xml/lua couples in dat/outfits/multicore/. These files are easy to edit, can generate xml/lua or reverse-generate old xml outfits.
- The entry in the CHANGELOG probably needs some rewriting.
- Finish extra description:
    - [x] Regroup multiple instances of helper funcs in a single file.
    - [x] Include permanent effects.
    - _Maybe_ shouldn't manage the transformations _property_name_ -> _unit_type_ and _property_name_ -> _In-game name_ manually by finding some lua hook to do it.
    - [x] Wait for next nightly release.
    - [x] Make a distinction between unequipped description and equipped in main slot description.
    - [x] re-factorize lua code into desc.lua.
    - [x] Follow Bobbens' helper func proposal.
    - [x] When done, don't forget to check every outfit was updated with new desc. function.
    - Now repair the `desc.lua` file to make description _and code_ look pretty.
        - [x] Colors, Units, Names.
        - [x] Add "+" prefix to positive numbers.
        - [x] Formatting and translations.

## Testing Done
One feature at a time.


dat/outfits/core_system/small/milspec_orion_2301_core_system.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version='1.0' encoding='UTF-8'?>
```
bobbens, line 1: You shouldn't remove the first weird XML encoding line, I believe it's required by the XML spec.

dat/slots/systems_secondary.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<slot name="systems_secondary">
+ <display>Core Systems</display>
```
bobbens, line 3: Not sure if it would make sense to give it a slightly different name. Like "Core Systems (Secondary)" or something.

dat/outfits/core_system/small/milspec_orion_2301_core_system.xml
johann1764: I didn't, elementTree did. I trust him, and so far I did not see any consequence.

dat/slots/systems_secondary.xml
johann1764: It may clarify the status of the slot, as it is white-bordered like the accessory slot and has the core slot logo. Stating explicitly it is a secondary slot might both clarify the slot status and allow the player to make the connection between the secondary values and the secondary slot.

johann1764: That looks good in the game, as scrolling over the slot turns "Core Sys" into "Core Sys (Sec)" while keeping the alignment.

However, the second line of the contextual description that was "Small Utility slot"
has become followed by a third line "or Utility slot"

bobbens: I've pushed some display fixes. It should be fixed and more consistent now.

johann1764: Cool !

johann1764:
bug:
```
Pilot 'Llama' failed safety check: !! Insufficient Energy Regeneration
   [0] Nexus Dart 160 Engine
   [1] Unicorp D-2 Light Plating
   [6] Unicorp PT-16 Core System
```
I bought a llama with these default outfits, and it pt-16 had no properties (except type etc.)
Therefore, it failed the safety check.
As soon as I select it as main ship, its properties are restored, and everything is ok.

The neutral pilots that are created suffer the same problem.

My understanding of that is that, when you create a ship, its default outfits are there but do not trigger the lua, that is called when we manually equip these, or when they are all requiped for refresh, as with the selection.

I do not know how I can solve that.

bobbens:
> bug:
> ```
> Pilot 'Llama' failed safety check: !! Insufficient Energy Regeneration
>    [0] Nexus Dart 160 Engine
>    [1] Unicorp D-2 Light Plating
>    [6] Unicorp PT-16 Core System
> ```
> I bought a llama with these default outfits, and its pt-16 had no properties (except type etc.) Therefore, it failed the safety check. As soon as I select it as main ship, its properties are restored, and everything is ok.
> 
> The neutral pilots that are created suffer the same problem.
> 
> My understanding of that is that, when you create a ship, its default outfits are there but do not trigger the lua, that is called when we manually equip these, or when they are all requiped for refresh, as with the selection.
> 
> More clearly: the init func is not called in these cases, hence the problem.
> 
> I do not know whether I can solve that, and how.

I pushed a potential fix that will initialize the outfits when initializing the pilots. However, it's going to be a bit tricky to solve the case that the player's non-deployed ships since they don't have valid IDs... Probably not going to be too easy to solve until moving it to rust...

johann1764:
Ok, I think I have an idea.

Why not using the main values in the xml file, and then overwrite/modify them in the lua file ?
This way, all ships that use size 2k+1 core systems will have no problem.

I think I will try this, to see if we can mitigate the problem this way.

johann1764:
Tomorrow (or later), I will need a quick briefing (or some pointers) for the point "**Convert savegames**",
as I have no idea on how it works and what files to look at, and did not find documentation on that.

johann1764:
xml.elementTree also removes comments.
Do we really want to keep ship comments ? (and leading trivial encoding information )

I would like an opinion before applying addslot to all ships (only the right ones will be modifed).

bobbens:
> Why not using the main values in the xml file, and then overwrite/modify them in the lua file ? This way, all ships that use size 2k+1 core systems will have no problem.

That is possible, but it'll have the issue of showing stats on separate lines. So it is fine for testing, but would not work for production, as it would just be horribly confusing.

> Tomorrow (or later), I will need a quick briefing (or some pointers) for the point "**Convert savegames**", as I have no idea on how it works and what files to look at, and did not find documentation on that.

It's all in `dat/save_updater.lua`. The way it works is it it'll receive a string and return a string. So, it is great for a one-to-one mapping, but if we want to do something different, we'll have to do something else. I guess it would be possible to make it work by recording stuff added, and then add a finalize function that can do other voodoo magic and checks. I'm guessing you want a one-to-two mapping and such?

> xml.elementTree also removes comments. Do we really want to keep ship comments ? (and leading trivial encoding information ?)

Checking there's only the following ship comments:
```
$ grep -- "<\!--" dat/ships/**/*.xml
dat/ships/feral_bioship/kauweke.xml:  <!--<tag>bioship</tag>-->
dat/ships/feral_bioship/nohinohi.xml:  <!--<tag>bioship</tag>-->
dat/ships/feral_bioship/taitamariki.xml:  <!--<tag>bioship</tag>-->
dat/ships/misc/psychic_orb.xml: <!--<base_type>Psychic Orb</base_type>-->
dat/ships/neutral/lancelot.xml:  <tag>goddard</tag><!--Used by the faction.-->
```
So I'm guessing removing it is fine.

johann1764:
> It's all in dat/save_updater.lua. The way it works is it it'll receive a string and return a string. So, it is great for a one-to-one mapping, but if we want to do something different, we'll have to do something else. I guess it would be possible to make it work by recording stuff added, and then add a finalize function that can do other voodoo magic and checks. I'm guessing you want a one-to-two mapping and such?

Yep. Looks like it will be complicated.

bobbens:
> > It's all in dat/save_updater.lua. The way it works is it it'll receive a string and return a string. So, it is great for a one-to-one mapping, but if we want to do something different, we'll have to do something else. I guess it would be possible to make it work by recording stuff added, and then add a finalize function that can do other voodoo magic and checks. I'm guessing you want a one-to-two mapping and such?
> 
> Yep. Looks like it will be complicated.

Implemented and pushed a "start" and "finish" API to the `dat/save_updater.lua`. Basically, now it'll record all the changes, and print the changes. It should be possible to use this now to give the player additional outfits, or potentially other changes. Allowing giving the player more outfits to compensate the core changes. However, given that the outfits are equipped based on slot ids, the player will still have to most likely reequip all their ships...

johann1764:
Ok ! Thanks for the solution !

This time I will let github sync my branch with naev instead of merging locally and then pushing.
I hope this won't mess up the PR. In the worst case, I will open another PR.

Then, I should be able to finish using your patch.

dat/save_updater.lua
```diff
@@ -34,24 +34,29 @@ function finish ()
...
-   print( "Save game updated!" )
+   --print( "Save game updated!" )
+   print(changes_done)
    for original,value in pairs(changes_done) do
       print( fmt.f("   {original} => {new} [{q}]", {original=original, new=value.new, q=value.q} ) )
-      if split[original] then
-         print( fmt.f(_("We offer you a free {outf} for your secondary slot. (TODO!)"),{outf=value.new}) )
-      end
+      --if split[original] then
+      --   player.outfitAdd(value.new)
```
bobbens, line 56:
```suggestion
      --   player.outfitAdd( value.new, value.q )
```

johann1764:
I don't get it: I added debug messages to `save_updater.lua`, but they are not displayed.
```lua
function finish ()
   if not save_updated then
      return
   end

   print( "Save game updated!" )
   for _original,_value in pairs(changes_done) do
      print("loop!")
   end
   print( "Save game update finished!" )
end

local function apply_change( original, new, q )
   print ("apply")
   local value = changes_done[original]
   if value then
      value.q = value.q + q
   else
      changes_done[original] = { new=new, q=q }
   end
   save_updated = true
   return new
end
```
`"appli"` is displayed, then `"Save game updated!"` is displayed, and then.. nothing else :
```
Reached main menu in 26.788 s
Charted safe lanes for 1492 objects in 4.518 s
apply
Player save was updated!
[      0x17a1fb] log_warn at ../../source/src/log.c:333 | ./naev-0.13.0-alpha.1+822.ga1bf721323-linux-x86-64.AppImage(log_warn+0x1f0)
...
[       0xc1119] _start at ??:0                         | ./naev-0.13.0-alpha.1+822.ga1bf721323-linux-x86-64.AppImage(_start+0x29)
WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Milspec Orion 9901 Core System' not found in stack.
Charted safe lanes for 1492 objects in 4.522 s
```
The error therefore occurs in :
```lua
   for _original,_value in pairs(changes_done) do
      print("loop!")
   end
   print( "Save game update finished!" )
```
I don't get it.
Notice, however, that the outfit has been successfully replaced, thanks to the lines:
```lua
local outfit_list = {
   ["Milspec Orion 9901 Core System"] = "Milspec Orion 8601 Core System",
   ...
}
```
I am missing something obvious. Any idea ?
(these correspond to the sources in the last commit)

bobbens:
> I don't get it: I added debug messages to `save_updater.lua`, but they are not displayed.
> 
> ```lua
> function finish ()
>    if not save_updated then
>       return
>    end
> 
>    print( "Save game updated!" )
>    for _original,_value in pairs(changes_done) do
>       print("loop!")
>    end
>    print( "Save game update finished!" )
> end
> 
> local function apply_change( original, new, q )
>    print ("apply")
>    local value = changes_done[original]
>    if value then
>       value.q = value.q + q
>    else
>       changes_done[original] = { new=new, q=q }
>    end
>    save_updated = true
>    return new
> end
> ```
> 
> `"appli"` is displayed, then `"Save game updated!"` is displayed, and then.. nothing else :
> 
> ```
> Reached main menu in 26.788 s
> Charted safe lanes for 1492 objects in 4.518 s
> apply
> Player save was updated!
> [      0x17a1fb] log_warn at ../../source/src/log.c:333 | ./naev-0.13.0-alpha.1+822.ga1bf721323-linux-x86-64.AppImage(log_warn+0x1f0)
> ...
> [       0xc1119] _start at ??:0                         | ./naev-0.13.0-alpha.1+822.ga1bf721323-linux-x86-64.AppImage(_start+0x29)
> WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Milspec Orion 9901 Core System' not found in stack.
> Charted safe lanes for 1492 objects in 4.522 s
> ```
> 
> The error therefore occurs in :
> 
> ```lua
>    for _original,_value in pairs(changes_done) do
>       print("loop!")
>    end
>    print( "Save game update finished!" )
> ```
> 
> I don't get it. Notice, however, that the outfit has been successfully replaced, thanks to the lines:
> 
> ```lua
> local outfit_list = {
>    ["Milspec Orion 9901 Core System"] = "Milspec Orion 8601 Core System",
>    ...
> }
> ```
> 
> I am missing something obvious. Any idea ? (these correspond to the sources in the last commit)

It needs functionality that was implemented recently. It should be in the next nightly in 2.5 hours. Sorry.

johann1764:
Did you forget to add "valgrind.h" ?

bobbens:
> Did you forget to add "valgrind.h" ?

Sorry about that, should be up!

johann1764:
On the same example, I now get:
```
Reached main menu in 26.875 s
Charted safe lanes for 1492 objects in 4.651 s
apply
Player save was updated!
Save game updated!
loop!
Save game update finished!
[      0x17a1fb] log_warn at ../../source/src/log.c:333        | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(log_warn+0x1f0)
[      0x1c47e7] outfit_get at ../../source/src/outfit.c:225   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(outfit_get+0x4f)
...
[       0xc1119] _start at ??:0                                | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(_start+0x29)
WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Milspec Orion 9901 Core System' not found in stack.
Charted safe lanes for 1492 objects in 4.552 s
```
which is much better, as I can work with that.
Still, that is odd that it issues a warning.

johann1764:
I still have the warnings in the form:
```
WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Unicorp PT-310 Core System' not found in stack.
```

They still appear after reloading an already converted savegame.
There is one for each deprecated outfit, regardless they are present in the savegame or not (I tried with a new, fresh savegame).

Maybe it is related to the knowledge of the player, as my newly created savegame first of all landed in Zied,
and therefore the player has knowledge of where to find them.

I will give another try to test this hypothesis.

bobbens:
> I still have the warnings in the form:
> 
> ```
> WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Unicorp PT-310 Core System' not found in stack.
> ```
> 
> They still appear after reloading an already converted savegame. There is one for each deprecated outfit, regardless they are present in the savegame or not (I tried with a new, fresh savegame).
> 
> Maybe it is related to the knowledge of the player, as my newly created savegame first of all landed in Zied, and therefore the player has knowledge of where to find them.
> 
> I will give another try to test this hypothesis.

What does the backtrace say? Could this be coming from the tech groups?

johann1764:
My hypothesis is false, I tried with a new savegame were the player only knew bolero, the errors still mentioned all deprecated outfits.

The techs have already been cleaned off the deprecated oufits:
```
~/my_naev/dat/tech$ grep -r PT-
unicorp_outfits.xml: <item>Unicorp PT-440 Core System</item>
unicorp_outfits.xml: <item>Unicorp PT-200 Core System</item>
unicorp_outfits.xml: <item>Unicorp PT-16 Core System</item>
systems_low.xml: <item>Unicorp PT-16 Core System</item>
systems_low.xml: <item>Unicorp PT-200 Core System</item>
systems_low.xml: <item>Unicorp PT-440 Core System</item>
basic_outfits2.xml: <item>Unicorp PT-16 Core System</item>
basic_outfits2.xml: <item>Unicorp PT-200 Core System</item>
basic_outfits2.xml: <item>Unicorp PT-440 Core System</item>
all_outfits.xml: <item>Unicorp PT-16 Core System</item>
all_outfits.xml: <item>Unicorp PT-200 Core System</item>
all_outfits.xml: <item>Unicorp PT-440 Core System</item>
basic_outfits1.xml: <item>Unicorp PT-16 Core System</item>
basic_outfits1.xml: <item>Unicorp PT-200 Core System</item>
basic_outfits1.xml: <item>Unicorp PT-440 Core System</item>
```

A full backtrace:
```
[      0x17a1fb] log_warn at ../../source/src/log.c:333                                           | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(log_warn+0x1f0)
[      0x1c47e7] outfit_get at ../../source/src/outfit.c:225                                      | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(outfit_get+0x4f)
[      0x27c9af] luaL_validoutfit at ../../source/src/nlua_outfit.c:156                           | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(luaL_validoutfit+0x6d)
[      0x27cb61] outfitL_get at ../../source/src/nlua_outfit.c:240                                | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(outfitL_get+0x1c)
[      0x508266] lj_BC_FUNCC at ??:0                                                              | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(lj_BC_FUNCC+0x33)
[      0x1afa2e] nlua_require at ../../source/src/nlua.c:867                                      | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(nlua_require+0x2f3)
[      0x508266] lj_BC_FUNCC at ??:0                                                              | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(lj_BC_FUNCC+0x33)
[      0x4fb17e] lua_pcall at ??:0                                                                | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(lua_pcall+0x5e)
[      0x1afe09] nlua_pcall at ../../source/src/nlua.c:1008                                       | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(nlua_pcall+0xba)
[      0x1ae3fe] nlua_dochunkenv at ../../source/src/nlua.c:340                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(nlua_dochunkenv+0x3c)
[      0x11f535] event_create at ../../source/src/event.c:235                                     | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(event_create+0x191)
[      0x11fb18] events_trigger at ../../source/src/event.c:420                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(events_trigger+0x331)
[      0x17958e] load_gameInternalHook at ../../source/src/load.c:1391                            | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(load_gameInternalHook+0x214)
[      0x17932e] load_game at ../../source/src/load.c:1301                                        | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(load_game+0xa4)
[      0x178a20] load_menu_load at ../../source/src/load.c:1050                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(load_menu_load+0xaa)
[      0x23d3f3] toolkit_mouseEventWidget at ../../source/src/toolkit.c:1965                      | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventWidget+0x391)
[      0x23cf3f] toolkit_mouseEventSingle at ../../source/src/toolkit.c:1823                      | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventSingle+0x114)
[      0x23cfd9] toolkit_mouseEventReverse at ../../source/src/toolkit.c:1835                     | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventReverse+0x8d)
[      0x23cfa0] toolkit_mouseEventReverse at ../../source/src/toolkit.c:1830                     | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventReverse+0x54)
[      0x23cfa0] toolkit_mouseEventReverse at ../../source/src/toolkit.c:1830                     | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventReverse+0x54)
[      0x23cfa0] toolkit_mouseEventReverse at ../../source/src/toolkit.c:1830                     | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventReverse+0x54)
[      0x23cfa0] toolkit_mouseEventReverse at ../../source/src/toolkit.c:1830                     | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEventReverse+0x54)
[      0x23d053] toolkit_mouseEvent at ../../source/src/toolkit.c:1852                            | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_mouseEvent+0x73)
[      0x23cc5b] toolkit_inputWindow at ../../source/src/toolkit.c:1739                           | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_inputWindow+0x126)
[      0x23cb32] toolkit_input at ../../source/src/toolkit.c:1698                                 | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(toolkit_input+0x3a)
[      0x1661f7] input_handle at ../../source/src/input.c:1871                                    | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(input_handle+0x14f)
[       0xc3db3] main at ../../source/src/naev.c:491                                              | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(main+0xbc5)
[       0x24082] __libc_start_main at ../csu/libc-start.c:308                                     | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf2)
[       0xc1119] _start at ??:0                                                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(_start+0x29)
WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Unicorp PT-68 Core System' not found in stack.
```
If I remove the core_systems folder, **this** problem disapears, **but** I now have tons of these at startup:
```
[      0x17a1fb] log_warn at ../../source/src/log.c:333                                           | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(log_warn+0x1f0)
[      0x21140e] ship_parseSlot at ../../source/src/ship.c:879                                    | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(ship_parseSlot+0x49f)
[      0x213fd5] ship_parse at ../../source/src/ship.c:1216                                       | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(ship_parse+0x2b6d)
[      0x2161f9] ships_load at ../../source/src/ship.c:1656                                       | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(ships_load+0x3b0)
[       0xc44cb] load_all at ../../source/src/naev.c:713                                          | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(load_all+0x1be)
[       0xc3b2d] main at ../../source/src/naev.c:399                                              | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(main+0x93f)
[       0x24082] __libc_start_main at ../csu/libc-start.c:308                                     | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf2)
[       0xc1119] _start at ??:0                                                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(_start+0x29)
WARNING ../../source/src/ship.c:879 [ship_parseSlot]: Ship 'Empire Lancelot Golden Efreeti' has default outfit 'Unicorp PT-16 Core System' which does not fit in the slot it is equipped by default.
```

johann1764:
In lancelot_golden_efreeti.xml:
```lua
  <utility prop="systems" size="small">Unicorp PT-16 Core System</utility>
  <utility prop="systems_secondary" size="small">Unicorp PT-16 Core System</utility><utility prop="accessory" size="small" />
```
In multicores/[...]/unicorp_pt16[...].xml:
```xml
<slot prop="systems" prop_extra="systems_secondary">utility</slot>
```

bobbens:
> In lancelot_golden_efreeti.xml:
> 
> ```lua
>   <utility prop="systems" size="small">Unicorp PT-16 Core System</utility>
>   <utility prop="systems_secondary" size="small">Unicorp PT-16 Core System</utility><utility prop="accessory" size="small" />
> ```
> 
> In multicores/[...]/unicorp_pt16[...].xml:
> 
> ```
> <slot prop="systems" prop_extra="systems_secondary">utility</slot>
> ```

I'll test the branch, give me one sec, I was in the middle of some toolkit refactoring.

johann1764:
It occurs I did not update:
`/my_naev/dat/scripts/equipopt/templates`
I am not sure this is the cause of this error, but it is a cause of error.
I will do it.

johann1764:
I think equiopt was not the cause of these errors.
Cleaning it had no effect in either case.

By the way, when no core_system dir, also had the errors:
```
WARNING ../../source/src/outfit.c:3138 [outfit_load]: Name collision! 2 outfits are named 'Unicorp PT-16 Core System'
[      0x17a1fb] log_warn at ../../source/src/log.c:333                                           | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(log_warn+0x1f0)
[      0x1d20cc] outfit_load at ../../source/src/outfit.c:3138                                    | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(outfit_load+0xd64)
[       0xc448e] load_all at ../../source/src/naev.c:710                                          | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(load_all+0x181)
[       0xc3b2d] main at ../../source/src/naev.c:399                                              | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(main+0x93f)
[       0x24082] __libc_start_main at ../csu/libc-start.c:308                                     | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf2)
[       0xc1119] _start at ??:0                                                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(_start+0x29)
```
I think this is because I use:
```
./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage -d ~/my_naev/dat/
```
And the image uses its own version of files that are missing in my dat/ folder.
Therefore, I must keep my deprecated versions where they are with the same name.

So, back to square 1 with that at loading a game :
```
WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Unicorp PT-68 Core System' not found in stack.
```

bobbens:
> I think equiopt was not the cause of these errors. Cleaning it had no effect in either case.
> 
> By the way, when no core_system dir, also had the errors:
> 
> ```
> WARNING ../../source/src/outfit.c:3138 [outfit_load]: Name collision! 2 outfits are named 'Unicorp PT-16 Core System'
> [      0x17a1fb] log_warn at ../../source/src/log.c:333                                           | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(log_warn+0x1f0)
> [      0x1d20cc] outfit_load at ../../source/src/outfit.c:3138                                    | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(outfit_load+0xd64)
> [       0xc448e] load_all at ../../source/src/naev.c:710                                          | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(load_all+0x181)
> [       0xc3b2d] main at ../../source/src/naev.c:399                                              | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(main+0x93f)
> [       0x24082] __libc_start_main at ../csu/libc-start.c:308                                     | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf2)
> [       0xc1119] _start at ??:0                                                                   | ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage(_start+0x29)
> ```
> 
> I think this is because I use:
> 
> ```
> ./naev-0.13.0-alpha.1+833.gbd23e89111-linux-x86-64.AppImage -d ~/my_naev/dat/
> ```
> 
> And the image uses its own version of files that are missing in my dat/ folder. Therefore, I must keep my deprecated versions where they are with the same name.
> 
> So, back to square 1 with that at loading a game :
> 
> ```
> WARNING ../../source/src/outfit.c:225 [outfit_get]: Outfit 'Unicorp PT-68 Core System' not found in stack.
> ```

Yeah, if you are adding your data on top, it'll fallback to missing files in the original. I tested the branch real quick and got no warnings after the first load. You could technically remove the deprecated outfits from the branch, and then toss them in another `dat/` file, and run Naev with both the new branch data + the deprecated outfit data to keep the branch ready to commit. However, just removing before merging is also fine.

bobbens:
By the way, if you don't use [pre-commit](https://pre-commit.com/), I would recommend you to do so. All you have to do is install it and run `pre-commit install` on the repo, and whenever you commit, it'll do some nice checks and busywork for you (like adding files for translation and running luacheck if available). You can always bypass the checks with `git commit -n` too.

johann1764:
Actually, I tried. I don't remember all the details, but it failed.
Something to do with having only very old versions of all libs.

johann1764:
I should get a more recent machine soon, though.
I hope I can  install everything needed to build naev with current stable debian.

johann1764:
### What I had in mind about the description
A preview of what I would like as description:

secondary
outfit

However, these have the following defects:
 - I used blue instead of grey because.. I didn't find how to do grey
 - The slashes '/' should also be greyed out
 - The list should include permanent effects (that are in the xml)
 - The builtin description is displayed in addition (and in reverse order)

johann1764:
I have a fundamental problem here, as the `descextra` function does not take an outfit instance as argument, like init does, but an outfit name. In other words, I don't have access to my instance data in it. This forbids to distinguish the three different cases:
 - the outfit is not equiped
 - the outfit is equiped in a primary slot
 - the outfit is equiped in a secondary slot

Using globals is no solution, as they are shared by different instances.
This allows testing how output looks like in diff cases, but that is all.

johann1764:
>  - [ ] No way to see the specific outfit it's associated to.

Another, more precise way to put it: `desc(p,po)` should receive an outfit instance second arg, as `init(p,po)` does, instead of an outfit name as it currently does.

> I can try to fix them both this week, but not exactly sure when I'll be able to yet. I'll also review the PR.

Ok !

bobbens:
> Ok, cool !
> 
> At next nightly release, I will be able to complete my task list quickly (very little left).

There's still the issue that when the outfit is equipped, the active stats will probably appear below. It should probably be possible to add a flag to disable that functionality if necessary though.

johann1764:
> There's still the issue that when the outfit is equipped, the active stats will probably appear below. It should probably be possible to add a flag to disable that functionality if necessary though.

Sure.

Also, that functionality should display stats in order instead of reversed order. But that's out of the scope of this PR.

johann1764:
Oh, I just noticed you removed the duplicate description. We're getting close ! However:
 - The mass is still appended at the end.
 - Other, unrelated outfits lose their description, included in stores. But not all. Some weapons, but not all. Afterburners do.

johann1764:
My mistake..

However, it is true that:
> Other, unrelated outfits lose their description, included in stores. But not all. Some weapons, but not all. Afterburners do.

johann1764:
> API is now:
> 
> ```lua
> -- Called to get the extra description of the outfit. This gets appended to the
> -- summary and description, and should be short. The function receives the
> -- pilot p if applicable, the current outfit o, and the pilot outfit slot po
> -- (or nil if not applicable). The function should return a translated string.
> function descextra( _p, _o, _po )
>    return _("This outfit is very cool.")
> end
> ```

I don't understand what object is this po argument. I tried several things, but it occurs that:
 - po never has tags (it is not even a table)
 - po does not have slot method

How do I access the slot tags ? Obviously, not like this:
```lua
function descextra( _p, _o, po )
   local desc = ""
   local nosec
   local nomain

   if po and po.tags and po.tags.core then
      if po:slots().tags.secondary then
         nosec=false
         nomain=true
      else
         nosec=true
         nomain=false
      end
   else
      nosec=false
      nomain=false
   end
   ...
```

bobbens:
> > API is now:
> > ```lua
> > -- Called to get the extra description of the outfit. This gets appended to the
> > -- summary and description, and should be short. The function receives the
> > -- pilot p if applicable, the current outfit o, and the pilot outfit slot po
> > -- (or nil if not applicable). The function should return a translated string.
> > function descextra( _p, _o, _po )
> >    return _("This outfit is very cool.")
> > end
> > ```

It is a `pilotoutfit` type object (or nil if not available).

> I don't understand what object is this po argument. I tried several things, but it occurs that:
>     * po never has tags (it is not even a table)
>     * po does not have slot method

It should have a `:slot` method when not nil.

> How do I access the slot tags ? Obviously, not like this:
> 
> ```lua
> function descextra( _p, _o, po )
>    local desc = ""
>    local nosec
>    local nomain
> 
>    if po and po.tags and po.tags.core then
>       if po.tags.secondary then
>          nosec=false
>          nomain=true
>       else
>          nosec=true
>          nomain=false
>       end
>    else
>       nosec=false
>       nomain=false
>    end
>    ...
> ```

Something like the following should work:
```lua
function descextra( _p, _o, po )
   local desc = ""
   local nosec
   local nomain

   if po then
      if po:slot().tags.secondary then
         nosec=false
         nomain=true
      else
         nosec=true
         nomain=false
      end
   else
      nosec=false
      nomain=false
   end
   ...
```

johann1764:
Ok, it works.

Some quick code factorization and it will be it.

bobbens:
Pushed `naev.shipstats()`. It returns a table with the following fields:
* **name**: Gives the name of the ship state (cpu_max).
* **display**: Untranslated name of the ship stat (CPU Capacity).
* **unit**: Untranslated unit (TFLOP).
* **inverted**: Whether or not negative is better.

That should make it much easier without need to create the table manually in Lua.

johann1764:
> Pushed `naev.shipstats()`. It returns a table with the following fields:
> 
> * **name**: Gives the name of the ship state (cpu_max).
> * **display**: Untranslated name of the ship stat (CPU Capacity).
> * **unit**: Untranslated unit (TFLOP).
> * **inverted**: Whether or not negative is better.
> 
> That should make it much easier without need to create the table manually in Lua.

That's the TODO of your proposed helper library. So when the new nightly is released, I can substitute
```lua
local u_power = naev.unit("power")
local u_mass  = naev.unit("mass")
local u_cpu   = naev.unit("cpu")

-- TODO probably better to expose / export the C engine data to Lua than duplicate
local shipstat = {
   mass        = { unit = u_mass,   name = _("Ship Mass"), reverse=true },
   cpu_max     = { unit = u_cpu,    name = _("CPU Capacity") },
   energy_regen= { unit = u_power,  name = _("Energy Regeneration") },
}
```
with
``` lua
local shipstat = naev.shipstats()
```
which is indeed simpler.

bobbens:
> > Pushed `naev.shipstats()`. It returns a table with the following fields:
> > 
> > * **name**: Gives the name of the ship state (cpu_max).
> > * **display**: Untranslated name of the ship stat (CPU Capacity).
> > * **unit**: Untranslated unit (TFLOP).
> > * **inverted**: Whether or not negative is better.
> > 
> > That should make it much easier without need to create the table manually in Lua.
> 
> That's the TODO of your proposed helper library. So when the new nightly is released, I can substitute
> 
> ```lua
> local u_power = naev.unit("power")
> local u_mass  = naev.unit("mass")
> local u_cpu   = naev.unit("cpu")
> 
> -- TODO probably better to expose / export the C engine data to Lua than duplicate
> local shipstat = {
>    mass        = { unit = u_mass,   name = _("Ship Mass"), reverse=true },
>    cpu_max     = { unit = u_cpu,    name = _("CPU Capacity") },
>    energy_regen= { unit = u_power,  name = _("Energy Regeneration") },
> }
> ```
> 
> with
> 
> ```lua
> local shipstat = naev.shipstats()
> ```
> 
> which is indeed simpler.

You'll have to translate the name and units though, which isn't necessary in the old form. (simply just wrap them in `_()` when formatting the string).

johann1764:
> OK, so high-level comments:
> 
> 1. The `mvx` files are for reference only right? To be removed at the end.

Err, I would rather not do that. If you consider xml is easier to process than lua files, then we should keep these disguised xml files (the extension is just a way to prevent their reading by naev).
They allow to generate xml/lua couples (mvx2xmllua.py), and can be used for generating outfit stats as any regular xml, contrary to the lua file. I would rather have the "makefile" (or whatever thing plays this role. Meson ?) generate these files from the mvx.
And if one day we decide to implement a builtin management of multicore, these mvx files will just have to be renamed .xml.

> 2. It would make more sense for the outfits to just be `dat/outfits/core_systems/` like before, no need for the additional `multicore` directory.

That is currently particularly handy (git add ../multicore, etc.). I suggest the outfits go back to their original places when all multicores will be managed (hull and maybe engines). 

> 3. I think I would move more stuff into the helper library. Let me show a proof of concept:
> 
> It makes it much easier to maintain / add new outfits,

As I said, it depends on whether we see these lua files as the outfits themselves or just a generated implementation of the outfits.
> and also change the backend easily.

True in both cases.
> Minimal code duplication!

That's optimal ! 

Now I will try to go into that direction.

bobbens:
> > OK, so high-level comments:
> > 
> > 1. The `mvx` files are for reference only right? To be removed at the end.
> 
> Err, I would rather not do that. If you consider xml is easier to process than lua files, then we should keep these disguised xml files (the extension is just a way to prevent their reading by naev). They allow to generate xml/lua couples (mvx2xmllua.py), and can be used for generating outfit stats as any regular xml, contrary to the lua file. I would rather have the "makefile" (or whatever thing plays this role. Meson ?) generate these files from the mvx. And if one day we decide to implement a builtin management of multicore, these mvx files will just have to be renamed .xml.

I'm not sure it's easier to create and maintain mvx instead of handling xml+lua (although it could be made possible to embed the Lua in the XML to make it a single file). Especially when it comes to adding new outfits that may not fit the pattern (e.g., have some other custom Lua functionality). Also, adding a building step makes it less intuitive and harder to test for those who are not building.

johann1764:
I have a problem (yet another, sorry).

```lua
local function add_desc(stat, nomain, nosec )
   local name = stat.stat.name -- this does not work (gets the internal name)
   local base = stat.pri
   local secondary = stat.sec
   local units = stat.stat.unit -- this works

   local p=sign(base)
   local s=sign(secondary)

   local def
   if stat.stat.reverse then -- this doesn't (always false)
```

This is not consistent with :
```lua
local shipstat = {
   mass        = { unit = u_mass,   name = _("Ship Mass"), reverse=true },
   cpu_max     = { unit = u_cpu,    name = _("CPU Capacity") },
   energy_regen= { unit = u_power,  name = _("Energy Regeneration") },
}
```
Does `naev.shipstats()` follow the same structure ?

johann1764:
My mistake.

johann1764:
> > > 1. The `mvx` files are for reference only right? To be removed at the end.
> > 
> > Err, I would rather not do that. If you consider xml is easier to process than lua files, then we should keep these disguised xml files (the extension is just a way to prevent their reading by naev). They allow to generate xml/lua couples (mvx2xmllua.py), and can be used for generating outfit stats as any regular xml, contrary to the lua file. I would rather have the "makefile" (or whatever thing plays this role. Meson ?) generate these files from the mvx. And if one day we decide to implement a builtin management of multicore, these mvx files will just have to be renamed .xml.
> 
> I'm not sure it's easier to create and maintain mvx instead of handling xml+lua (although it could be made possible to embed the Lua in the XML to make it a single file). Especially when it comes to adding new outfits that may not fit the pattern (e.g., have some other custom Lua functionality). Also, adding a building step makes it less intuitive and harder to test for those who are not building.

On one hand, a xml file is easier to process than a couple xml/lua. The information is local, and xml are easy to parse.

On the other hand, naev currently does not manage xml outfits using n /m values. Therefore, we have to disguise it with another extension and translate into a couple of xml/lua files naev can handle. That is less direct. Just like bioship core outfits generation, after all.

Anyway, we have until the completion of multicores - part 3 to think about it. For now I feel more concerned about multicores - parts 1 & 2.

### By the way, 
I think (hope) multicores - part 1 is **finally** over ! (even if there is the double description problem, but that is not in my scope).

bobbens:
Looks good! Did the following changes:
1. Ran pre-commit to correct various small things.
2. Changed it so it is one utility slot per line (there were two with secondary systems)
3. Renamed `dat/outfits/multicore/desc.lua`to `dat/outfits/lib/multicore.lua`.
4. Streamlined code for descriptions, removing duplicate units, and adding a header indicating it's equipped as primary or secondary (or neither).
5. Can now hide dynamic ship stats from lua outfits with `hidestats`.
6. Support for inline Lua for outfits.
7. Moved outfits from `dat/outfits/multicore/core_system/` to `dat/outfits/core_system/`.

> On one hand, a xml file is easier to process than a couple xml/lua. The information is local, and xml are easy to parse.

In this case, adding support and allowing inline Lua would keep it in one file and easy to deal with. I have implemented it with the `<lua_inline>` tag.

I haven't modified the script to generate the files, but the following is possible now:
```diff
diff --git a/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml b/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml
index a7595b3bb8..b31ff8c999 100644
--- a/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml
+++ b/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml
@@ -11,5 +11,15 @@
  </general>
  <specific type="modification">
   <energy>200</energy>
-  <lua>outfits/multicore/core_system/small/milspec_orion_2301_core_system.lua</lua></specific>
+  <lua_inline>
+require("outfits.lib.multicore").init{
+   { "mass", 540, 760},
+   { "cpu_max", 540, 1660},
+   { "energy", 2460, 1380},
+   { "energy_regen", 66, 74},
+   { "shield", 850, 250},
+   { "shield_regen", 15, 3},
+}
+  </lua_inline>
+ </specific>
 </outfit>
 ```

> On the other hand, naev currently does not manage xml outfits using n /m values. Therefore, we have to disguise it with another extension and translate into a couple of xml/lua files naev can handle. That is less direct. Just like bioship core outfits generation, after all.

The main difference is that here it is just a formatting issue. Bioships actually generate new outfits that make it hard to maintain.

I've pushed the changes to the [mergecores branch](https://github.com/naev/naev/tree/mergecores), although source changes should be in main.

Some potential issues:
1. I've seen a pirate hyena unable to equip because of energy issues.
2. The way `pilot.outfitAdd` works, is that if you have a primary slot P with default outfit O, and an empty secondary slot S, if you try to do `pilot.outfitAdd( O )` it will replace the outfit in P because it is default instead of getting added to S. Not sure the best way to work around this. However, it would probably make sense to potentially give the slots names, and assign the core outfits by names in the equipment scripts instead of doing it the lazy way.
3. We'll need a message on load when merged into main, and should also bump the version up. I believe it might be best to wait until all the cores are done, however, we could also do it one at a time.

johann1764:
> Looks good! Did the following changes:
> 
> 1. Ran pre-commit to correct various small things.
> 2. Changed it so it is one utility slot per line (there were two with secondary systems)

Python xml parser produces dirty outputs. I feel a bit lazy about it, so I think I will manually clean up after it.

> 3. Renamed `dat/outfits/multicore/desc.lua`to `dat/outfits/lib/multicore.lua`.

I had that in mind, too.

> 4. Streamlined code for descriptions, removing duplicate units, and adding a header indicating it's equipped as primary or secondary (or neither).
> 5. Can now hide dynamic ship stats from lua outfits with `hidestats`.

Cool. Did you repair the dynamic ship stats not appearing on other outfits ?
(every outfit with lua, I think : razor weapon family, plasma weapon family, point defense, almost all utility, etc.)

> 6. Support for inline Lua for outfits.
> 7. Moved outfits from `dat/outfits/multicore/core_system/` to `dat/outfits/core_system/`.
> [...]
> I haven't modified the script to generate the files, but the following is possible now:
> 
> ```diff
> diff --git a/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml b/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml
> index a7595b3bb8..b31ff8c999 100644
> --- a/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml
> +++ b/dat/outfits/multicore/core_system/small/milspec_orion_2301_core_system.xml
> @@ -11,5 +11,15 @@
>   </general>
>   <specific type="modification">
>    <energy>200</energy>
> -  <lua>outfits/multicore/core_system/small/milspec_orion_2301_core_system.lua</lua></specific>
> +  <lua_inline>
> +require("outfits.lib.multicore").init{
> +   { "mass", 540, 760},
> +   { "cpu_max", 540, 1660},
> +   { "energy", 2460, 1380},
> +   { "energy_regen", 66, 74},
> +   { "shield", 850, 250},
> +   { "shield_regen", 15, 3},
> +}
> +  </lua_inline>
> + </specific>
>  </outfit>
> ```

That's pretty, I will have the script produce that.

> 
> > On the other hand, naev currently does not manage xml outfits using n /m values. Therefore, we have to disguise it with another extension and translate into a couple of xml/lua files naev can handle. That is less direct. Just like bioship core outfits generation, after all.
> 
> The main difference is that here it is just a formatting issue. Bioships actually generate new outfits that make it hard to maintain.
> 
> I've pushed the changes to the [mergecores branch](https://github.com/naev/naev/tree/mergecores), although source changes should be in main.
> 
> Some potential issues:
> 
> 1. I've seen a pirate hyena unable to equip because of energy issues.

That's an issue. I will try to replicate.

> 2. The way `pilot.outfitAdd` works, is that if you have a primary slot P with default outfit O, and an empty secondary slot S, if you try to do `pilot.outfitAdd( O )` it will replace the outfit in P because it is default instead of getting added to S. Not sure the best way to work around this. However, it would probably make sense to potentially give the slots names, and assign the core outfits by names in the equipment scripts instead of doing it the lazy way.

In current actual cases, we always outfitAdd twice the same core system. If it is the default, then it has no affect, as expected. It it is something else, it will successfully add it into the two slots.
The only problematic case is if you add a default, then a non-default. 
This will result in having the two cores having reversed positions with what was expected, and there is no workaround for this.

> 3. We'll need a message on load when merged into main, and should also bump the version up. I believe it might be best to wait until all the cores are done, however, we could also do it one at a time.

I currently have tons of warnings due to the fact the image and my data do not match.
It would be handy to have an intermediate step where all the core-system warnings finally stop.

> 4. The mvx script seems to not properly indent / whitespace the files.
xml parser output problem, I know. _Maybe_ I will try to do something about it. I guess some other people have had the same problem and have found some way of fixing it.

johann1764:
Ok, so now I can open a PR _Multicores - part 2_ on `naev/naev/mergecore`  from a branch of  `johann1764/naev/mergecore` ?
Will this PR stay open until mergecore is merged ?

The insufficient energy regen bug could not be reproduced. I travelled around New Haven and crossed a lot of hyenas, but them looked pretty operational. No warning was issued. However, this one happened to me a lot while developing this, so I know what it looks like. I will see if it happens again in the future.

The dynamic stats not appearing on other outfits bug is still present, including in the main branch.
I suspect you did this before the `hidestats` feature to help developping this, so I guess it will be ok next nightly ?

_This paragraph is mostly grumbling and can be summed up as "I told you so", reading this is not required. :-)_ 
As I said, moving multicore/ contents back in place now was premature.
As a consequence, I won't be able to update easily with the script the new xml with the nice inlined lua.
It is also useless, as I will have to do it again later for core_hulls.
I won't be able to manage the old cores with the same single commandline that currently updates everyting.
If we change the script again by that time, there will be no way (except git) to update all this.
And it is also incomplete, as _homebrew processing unit_ was forgotten.

Anyway, I will now update the script for future multi-outfits to have nice inlined lua,
and start working on hulls.

bobbens:
> Ok, so now I can open a PR _Multicores - part 2_ on `naev/naev/mergecore` from a branch of `johann1764/naev/mergecore` ? Will this PR stay open until mergecore is merged ?

Well, there's the question of whether to do all the slots together, or one by one.

Advantages of one-by-one:
* More debugging and testing to detect early issues
* Less merge conflicts and issues

Disadvantages of one-by-one:
* More save game issues
* Harder to revert or change if we want

In general I'm a fan of merging early rather than later. So I guess I could look into trying to polish it a bit more (give message to players and bump version) and merge one-by-one. Most likely we won't get a ton of angry fanmail on how it is all horrible and they want to revert :)

> The insufficient energy regen bug could not be reproduced. I travelled around New Haven and crossed a lot of hyenas, but them looked pretty operational. No warning was issued. However, this one happened to me a lot while developing this, so I know what it looks like. I will see if it happens again in the future.

Could have been related to other half-done changes. We'll see.

> The dynamic stats not appearing on other outfits bug is still present, including in the main branch. I suspect you did this before the `hidestats` feature to help developping this, so I guess it will be ok next nightly ?

I just fixed it. It's just related to how horribly and hacky the outfit summary system was. It's a bit better now, and will probably end up being cleaned when moving to rust when that eventually happens.

> _This paragraph is mostly grumbling and can be summed up as "I told you so", reading this is not required. :-)_ As I said, moving multicore/ contents back in place now was premature. As a consequence, I won't be able to update easily with the script the new xml with the nice inlined lua. I won't be able to manage the old cores with the same single commandline that currently updates everything. If we change the script again by that time, there will be no way (except git) to update all this. It is also useless, as I will have to do it again later for core_hulls. And finally it is also incomplete, as _homebrew processing unit_ was forgotten.

Is this because of how the paths are handled? Wouldn't it be possible to copy the old files into a temporary directory and use those to patch the new ones?

> Anyway, I will now update the script for future multi-outfits to have nice inlined lua, and start working on hulls. Maybe improve xml output.

Sounds good.

johann1764:
> > Ok, so now I can open a PR _Multicores - part 2_ on `naev/naev/mergecore` from a branch of `johann1764/naev/mergecore` ? Will this PR stay open until mergecore is merged ?
> 
> Well, there's the question of whether to do all the slots together, or one by one.
> 
> Advantages of one-by-one:
> 
> * More debugging and testing to detect early issues
> * Less merge conflicts and issues
> 
> Disadvantages of one-by-one:
> 
> * More save game issues
> * Harder to revert or change if we want
> 
> In general I'm a fan of merging early rather than later. So I guess I could look into trying to polish it a bit more (give message to players and bump version) and merge one-by-one. Most likely we won't get a ton of angry fanmail on how it is all horrible and they want to revert :)

Ok, let's go !
But I don't understand why "message to players and bump version":
I think the save_updater works, because I tested it !

While at it, could you just move the homebrew processing unit from "outfits/core_system/unique" to "outfits/unique/" ?
(and update naevpedia/outfits/meson.build accordingly)

> 
> > The dynamic stats not appearing on other outfits bug is still present, including in the main branch. I suspect you did this before the `hidestats` feature to help developping this, so I guess it will be ok next nightly ?
> 
> I just fixed it. It's just related to how horribly and hacky the outfit summary system was. It's a bit better now, and will probably end up being cleaned when moving to rust when that eventually happens.
> 
> > _This paragraph is mostly grumbling and can be summed up as "I told you so", reading this is not required. :-)_ As I said, moving multicore/ contents back in place now was premature. As a consequence, I won't be able to update easily with the script the new xml with the nice inlined lua. I won't be able to manage the old cores with the same single commandline that currently updates everything. If we change the script again by that time, there will be no way (except git) to update all this. It is also useless, as I will have to do it again later for core_hulls. And finally it is also incomplete, as _homebrew processing unit_ was forgotten.
> 
> Is this because of how the paths are handled? Wouldn't it be possible to copy the old files into a temporary directory and use those to patch the new ones?

Maybe I am a bit overreacting, but basically I cannot use the same one-line-command-that-updates-it-all because src and dst dirs are now the same (there are xml/lua output and base xml input together in the same place).
That is not so terrible, that is just not very handy.
I will have to make a bash loop instead of using * for updating cores with new lua-inlined script output, that's all.
And hope no more updates to do, which is likely.

johann1764:
> In general I'm a fan of merging early rather than later. So I guess I could look into trying to polish it a bit more (give message to players and bump version) and merge one-by-one

Apparently, you have done that.

Shall we merge ?

bobbens:
> > In general I'm a fan of merging early rather than later. So I guess I could look into trying to polish it a bit more (give message to players and bump version) and merge one-by-one
> 
> Apparently, you have done that.
> 
> Shall we merge ?

I've added a message (and a lie) that only fires if the core outfits are updated. I wrote it to be general enough that it should apply to systems, hulls, _AND_ engines.

The lie is that your ship AI says it automatically updated your outfits, which it is not true (yet). Have to implement a function that goes over the player ships seeing if there is a filled primary and an empty secondary core slot, and if so, try to take an outfit from the player's inventory to fill the secondary core slot. I can try to do it this week.

Implementation should probably be:
1. Store what outfits were split in the naev cache.
2. For each ship the player owns that is known to have 2 applicable core slots (minor optimization, can just iterate over all instead)
   1. Swap to the ship (engine limitation that you can't access player's 'pilot' if the player is not piloting the ship)
   2. See if it needs updating (primary is filled with an outfit that was updated, and secondary is empty)
   3. If so, remove an outfit from inventory and add to secondary slot

TODO stub: https://github.com/naev/naev/blob/mergecores/dat/events/updater.lua#L25

johann1764:
> The lie is that your ship AI says it automatically updated your outfits, which it is not true (yet). Have to implement a function that goes over the player ships seeing if there is a filled primary and an empty secondary core slot, and if so, try to take an outfit from the player's inventory to fill the secondary core slot. I can try to do it this week.

Aren't you overdoing it ? OK, replacing one core and leaving the other in the inventory is not perfect, but letting players re-equip the new core to the secondary slot is not so problematic, a fortiori if the player gets an informative message about it. I am not sure this would justify much work.

bobbens:
> > The lie is that your ship AI says it automatically updated your outfits, which it is not true (yet). Have to implement a function that goes over the player ships seeing if there is a filled primary and an empty secondary core slot, and if so, try to take an outfit from the player's inventory to fill the secondary core slot. I can try to do it this week.
> 
> Aren't you overdoing it ? OK, replacing one core and leaving the other in the inventory is not perfect, but letting players re-equip the new core to the secondary slot is not so problematic, a fortiori if the player gets an informative message about it. I am not sure this would justify much work.

Well, I try to make the friction as small as possible. I know of a player who does have a long-going save with 100+ ships, and I do think it would be nice to do this for such players (which I guess are not that few). I can implement it, it shouldn't be too hard.

bobbens:
OK, surprisingly enough, managed to find time to implement it. Seems to work here fine, so went ahead and merged. Thanks again for the great work!

(btw, the solution I wrote should be generic enough it should handle hulls too, but probably not engines if they don't use primary/secondary core system)

johann1764:
COOL !

I have good news too : the gene drive automatic alignment with standard engines works, gen.py is now much shorter, and engine generation rules are now explicit in a small python file.

However, I will probably have no time to finish it until friday.

PR #2781, opened 2025-03-18
autonav - disable afterburner when approaching

johann1764:
**Bug Fix**/**New Feature**

This PR addresses the bug/feature described in #2774.

## Summary
 - Turn off afterburner before:
      - [x]  a jump, except if the ship has _instant jump_;
      - [x]  landing.
 - [x] Hitting the afterburner while doing final approach should abort autonav.
 - The comments in the sources raise issues.

### DIFF
--- a/dat/autonav.lua
+++ b/dat/autonav.lua
@@ -412,7 +412,7 @@ local function turnoff_afterburner()
       -- Proposition:
       --   activate: boolean Whether to activate or deactivate
       if n["type"]=="Afterburner" and n["state"]=="on" then
-         player.pilot():outfitToggle(i)
+         player.pilot():outfitToggle( n['slot'] )
       end
    end
    for _i,n in ipairs(player.pilot():actives()) do
--- a/dat/autonav.lua
+++ b/dat/autonav.lua
@@ -404,20 +404,12 @@ local function autonav_rampdown( count_brake )
 end
 
 local function turnoff_afterburner()
-   for i,n in ipairs(player.pilot():actives()) do
+   local pp=player.pilot()
+   for _i,n in ipairs(pp:actives()) do
       -- Why does n:type() not work ?
       -- Why *A*fterburner and not afterburner ?
-      -- The documentation is not clear : (too much or.s)
-      --   activate: boolean Whether or not to activate or deactivate
-      -- Proposition:
-      --   activate: boolean Whether to activate or deactivate
       if n["type"]=="Afterburner" and n["state"]=="on" then
-         player.pilot():outfitToggle( n['slot'] )
-      end
-   end
-   for _i,n in ipairs(player.pilot():actives()) do
-      if n["type"]=="Afterburner" and n["state"]=="on" then
-         print("Toggle failed!")
+         pp:outfitToggle( n['slot'] )
       end
    end
 end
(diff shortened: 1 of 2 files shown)


dat/autonav.lua
```diff
@@ -403,9 +403,27 @@ local function autonav_rampdown( count_brake )
...
+local function turnoff_afterburner()
+   for i,n in ipairs(player.pilot():actives()) do
+      -- Why does n:type() not work ?
+      -- Why *A*fterburner and not afterburner ?
+      -- The documentation is not clear : (too much or.s)
+      --   activate: boolean Whether or not to activate or deactivate
+      -- Proposition:
+      --   activate: boolean Whether to activate or deactivate
+      if n["type"]=="Afterburner" and n["state"]=="on" then
+         player.pilot():outfitToggle(i)
```
bobbens, line 415:
```suggestion
         player.pilot():outfitToggle( n['slot'] )
```
`i` is not the slot index. `n['slot']` is.

dat/autonav.lua
```diff
@@ -403,9 +403,27 @@ local function autonav_rampdown( count_brake )
    end
 end
 
+local function turnoff_afterburner()
+   for i,n in ipairs(player.pilot():actives()) do
```
bobbens, line 407: probably would make sense to have a local variable for `player.pilot()` instead of calling it multiple times.

johann1764: "multiple times" is two times in the worst case situation. (I don't take the debug code into account),
but why not.

johann1764:
Still, it remains to decide what happens if the player hits the afterburner while autonav is performing the final approach manoeuvers. We could re-toggle it off, but preventing the player from hitting the afterburner is not too nice, as he/she may need to interrupt the manoeuver to flee. I think the better specification is:
 - hitting the afterburner while doing final approach aborts autonav. (like any other manual control)
 - if afterburner is on when autonav is started, it will be left as is until either it stops by itself or the ship has to stop accelerating.
 - hitting the afterburner before the final approach is tolerated (does not interrupt autonav) to please the impatient player.

bobbens:
> Still, it remains to decide what happens if the player hits the afterburner while autonav is performing the final approach manoeuvers. We could re-toggle it off, but preventing the player from hitting the afterburner is not too nice, as he/she may need to interrupt the manoeuver to flee. I think the better specification is:
>     * hitting the afterburner while doing final approach aborts autonav. (like any other manual control)
>     * if afterburner is on when autonav is started, it will be left as is until either it stops by itself or the ship has to stop accelerating.
>     * hitting the afterburner before the final approach is tolerated (does not interrupt autonav) to please the impatient player.

I think aborting autonav (could add a check in the autonav script, or have the autonav receive the activated outfit and have it decide, or just have C disable autonav if anything activated) or just doing nothing (the player can hit the left / right key to break autonav anyway) would make sense.

johann1764:
Currentty, if AB is triggered during final manoeuvers, it aborts autonav...
**except** if ai.brake() is currently active, in which case it does not ( and everything goes wrong ).

I believe C should disable autonav if AB triggered during AI.brake().

I just have to improve abort reason, and I think I am done.

bobbens:
Looks great, thanks! Did some changes:
1. Removed 'type' field from pilot.actives as it is redundant with outfit:type
2. Added movement tag to movement outfits (afterburner, blink drives)
3. Autonav now tries to turn off movement outfits
4. If the "afterburner" can't be turned off, abort autonav

PR #2783, opened 2025-03-20
Repaired Falcon 1400 bugs

johann1764:
**Bug Fix**

## Summary
 - remaining references to Falcon 1300 here and there.
 - incorrect speed (see below)

before:
engines2

after:
engines_repaired


bobbens:
Nice catch! Thanks!

Issue #2696, opened 2024-12-23
[Proposal]: stop autonav on non-point defense weapon use


HyperSoop:
### Suggestion

Autonav stopping on weapon use was the pre-0.12.0 behaviour; iirc it was changed to make flying through hostile territory with point defense equipped less annoying, and now weapon use never breaks you out of autonav. That is suboptimal for pretty much every non-point-defense weapon however

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

johann1764:
Sorry, I forgot to include this problem in the list of autonav possible improvements when I did it.

PR #2784 should finally fix this issue.

johann1764:
Bobbens completed it, you can close this issue.

PR #2784, opened 2025-03-21
autonav - reset if non-pdef weapon fired

johann1764:
**Bug Fix**/**New Feature**

This PR addresses the bug/feature described in #2696.

## Summary
See #2696. This patch relates to commit :
```git
commit d3f8d4179b4c8a6c043f52d5db35fef1f039e7a4 (HEAD)
Author: Edgar Simo-Serra <bobbens@gmail.com>
Date:   Thu Nov 21 23:56:57 2024 +0900

    Shooting weapons no longer resets autonav speed up.
```
I reverted it and added a check for point defense flag.

## Testing Done
I can't. Sorry !

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,7 @@
          * Can automatically land when following a ship
          * Improved autonav behaviour when losing a followed ship
          * Disables afterburner when getting ready to land or jump
+         * Shooting *non-point defense* weapons resets autonav speed up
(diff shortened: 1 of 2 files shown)


bobbens:
Looks good, thanks!
You missed the special branch for the beams, and it seems like beams were ignoring inrange flag when no target was selected, so fixed that up too.

Issue #2782, opened 2025-03-19
[Proposal]: engine rework


bobbens:
## Motivation
With #2766 , we have an opportunity to rework the engines too, which I find have likely issues:
1. There is too large of a speed gap between small ships and medium ships. In particular, people really dislike medium ships, because rather than having the advantages of small and large ships, they seem to have only the disadvantages (poor movement, poor health, poor firepower), and you're either better off in a capship for fighting or an interceptor for traveling.
2. The engines feel all the same. They all have similar speed / accel ratios, so only size seems to matter.
3. Dodging and flying skills seem to not matter too much, because ships are bad at dodging, and weapons are good at tracking. This has been mitigated a bit with making it so poorly skilled NPCs aim worse, but it still could be made better.

## Potential Improvements
My suggestion would be to:
1. Flatten the speed curve. Right now it's something like small ships are 300 base speed, medium are 200 base large are 100 base speed (for Nexus engine, or 345 - 290 - 230 - 175 - 115 - 80 for the tricon). I could imagine something like (from interceptor to carrier) 350 - 320 - 270 - 230 - 160 - 100 or more. Wouldn't solve the issue, but would larger ships less of sitting ducks.
2. Give more varied speed/accel profiles. For example, since true max speed is `speed + accel/3`, you can have similar true max speed but play with ratios of speed and accel. So for example, you could have an outfit with 400 speed and 300 accel for 500 true speed, and one with 300 speed and 600 accel for the same 500 true speed. The first one would take 1.3 seconds to accelerate to full speed, while the second would accelerate in 0.5 seconds, and be able to dodge much better. Melendez engines could have high top speed and low accel, while the tricons could focus more on high accel and low top speed.
3. In general, it might be better to give smaller ships more accel and less top speed too. That would make them feel more agile and better at dodging.

## Important Considerations
1. This might break the AI and change heavily how well weapons track. In particular, we don't want the player to have too much trouble aiming, and want them to be able to dodge well so we can reward "good piloting".
2. Probably would make sense to do this after or at the same time as merging core slots #2766 for engines to avoid having to modify too maony engines.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

johann1764:
> ## Motivation
> With [#2766](https://github.com/naev/naev/issues/2766) , we have an opportunity to rework the engines too, which I find have likely issues:
> 
[...]

PR #2786, opened 2025-03-24
engines reworked

johann1764:
**Balance**

This PR addresses the feature described in #2782.
This implements the last proposition to date, which is still opened for discussion.

The engines rework has moved to #2795 engines - balancing.

The major consequences of this is that small sized and/or Tricon engines now rely much more heavily on accel for their top speed. As a consequence, accel modifiers, that used to have little effect, now have an effect at least as important as modifiers on speed. Therefore all accel modifiers in the game are now too high, they must be reduced.

## Summary (engines)

Executed the following:
```python
TURN_CT=0.37
AG_EXP=0.2

lines={
   'Nexus':'N',
   'Unicorp':'N',
   'Tricon':'T',
   'Krain':'K',
   'Melendez':'M',
   "Za'lek":'Z',
   "Beat":'B'
}

alpha,beta=1.0,0.1

def dec(n):
   if n<=1:
      return 400.0
   else:
      return dec(n-1)/(alpha+beta*(n-1))

def ls2vals(line_size):
   if line_size is None:
      return None

   (line,size)=line_size
   fullspeed=dec(size)

   if line in ['N','M','Z','B']:
      fullspeed*=7.0/8.0

   r=0.15*pow(2,-((size-1)-2.5)/2.5)

   if line in ['T','K']:
      r*=1.75
   elif line=='M':
      r*=0.7

   speed,acc=fullspeed*(1.0-r),fullspeed*r*3.0

   if line=='K':
      speed*=1.05
      acc*=1.3
   elif line=='Z':
      speed*=0.7
      acc*=0.7
   elif line=='B':
      speed*=0.55
      acc*=0.55

   fullspeed=speed+acc/3.0

   #turn=speed/5.0+acc/4.0
   turn=TURN_CT*fullspeed*pow(1.0*acc/speed,AG_EXP)
   return {"speed":fmt(speed),"accel":fmt(acc),"turn":fmt_t(turn)}
```

Obviously, this can be effortlessly changed.

The new module utils/outfit.py supports iteration over all nodes (elementTree elements with their methods) and *clean* output (so far).

If needed, it is trivial to add the `<?xml version='1.0' encoding='UTF-8'?>` header.

## Summary
 - [x] Reduce Engine reroute modifier from +30% accel to +20% accel.
 - [x] Reduce Hyenas accel bonuses: Hyena : 75%->35% ; Pirate Hyena 70%->30%.
 - [x] Include scripts.
 - [x] Manage Soromid Gene Drives auto-updating.
 - [x] Manage the particular case of the gene drive based on the now disappeared Melendez Mammoth.
 - [x] Manage the Za'lek experimental engine. -> That's just a Nexus -30%.
 - [x] Manage the beat up engines. -> Just Nexus -45%
 - [x] Merge main into this and ...
 - [x] ... move scripts to utils/outfits.
 - Update Za'lek drones builtin engines -> These do not look like any existing engine. I could find that:
   - Za'lek light drone to scout drone +10 speed/accel/turn (not percents! 10 m.s-1, 10 m.s-2,  10 rad.s-1). Could be cleaner to just fix +3% everything.
   - Light drone is Nexus +50% speed +50% turn +100% accel.
   - Heavy drone has same accel and turn as Tricon Zephyr 2, but +14% speed.
   - Bomber Drone is approx. Melendez +57% all three
   In other words, the values do not make much sense. We will have to do that manually when engines are stable (therefore, not now).
 - [x] Saved Za'lek drone engines notes as "ships/zalek/drone_engines.md"
 - [x] Update afterburners -> applied -1/6 to accel for light afterburners and adrenal glands. In addition, -10% accel bonus for high-end afterburners.
       Medium afterburners left as is (as medium ships do not see their acceleration grow as much as small ships do).
 - [x] Apply Nexus time constant reduction to the Tricon line.
 - [x] Reduce Wanderer/The Bite accel bonuses.
 - [x] Is it normal that the bite accel modifier is declared with `accel` instead of `accel_mod` ? -> Made lunge like adrenal gland ii + 50%, with additional +50% turn. (This makes it only slightly faster than adrenal gland, which I don't understand. I'll dig into this another time.)
 - [x] Found and fixed a bug in the_bite.lua

## Testing Done
 - [x] Re-test with next nightly.

## Future tasks

Since `generate.py` now uses explicitly some `.xml` outfits, we could even use these as templates (as `outfit.py` allows us to do that) and get rid of templates. I think I will propose a dedicated PR after multicores - part 2 (hulls). Or maybe not.
**In any case**, `meson.build` does not know about these dependencies. This should be fixed. 

The afterburner / lunge thing should be revisited at some point.

Here I left out the Nexus/Unicorp differentiation, which has to be done later.
The energy efficiency of engines will also have to be worked on.

Having the size 6 ships only available in the late game will have to be done.

Special engines can wait the completion of multicores - part 3 (engines).

There are still problems with The Bite:
 - Repair the_bite skill description and outfit description.
 - The bite is still broken. Might be related to excessive speeds.

### Unrelated
See #2788.

## Current TODOs
 - [x] Have a single shared lua variable for PHYSICS_SPEED_DAMP.
 - [x] Have a PHYSICS_SPEED_DAMP defined in the script -> use the value from constants.lua.
 - [x] Make sure to check that PHYSICS_SPEED_DAMP does not affect the script in an unplanned matter.
 - [x] Different versions of the bite should share the same variables (instead of multiple instances of the same values).
 - [x] Repair "The Bite": have the two AI files referring to it rely on the variable defined above.
 - [x] Tweak alpha and beta a bit.
 - [x] Decrease AB/bite accel/speed ratio a bit.
 - [x] The bite should disable afterburners.


bobbens:
> The major consequences of this is that small sized and/or Tricon engines now rely much more heavily on accel for their top speed. As a consequence, accel modifiers, that used to have little effect, now have an effect at least as important as modifiers on speed. Therefore all accel modifiers in the game are now too high, they must be reduced.

Makes sense.

> In addition to that, I found relevant to apply the Nexus line time constant reduction to the Tricon line. Since Nexus line is not specialized so far, we could give it some other special effect. If this additional proposition is not to be kept, it is easy to undo it.

Why is this? I think keeping tricon a bit harder to use would make sense.

>     * [x]  Reduce Engine reroute modifier from +30% accel to +20% accel.
>     * [x]  Reduce Hyenas accel bonuses: Hyena : 75%->35% ; Pirate Hyena 70%->30%.

Sounds good.

>     * [x]  Include scripts.
>     * [x]  Manage Soromid Gene Drives auto-updating.
>     * [x]  Manage the particular case of the gene drive based on the now disappeared Melendez Mammoth.

Sounds good.

>     * [x]  Manage the Za'lek experimental engine. -> That's just a Nexus -30%.
>     * [x]  Manage the beat up engines. -> Just Nexus -45%

Sounds good. Could make the Za'lek experimental engine more interesting (like randomized stats that change over time and such), but that is out of scope here.

>     * [x]  Merge main into this and ...
>     * [x]  ... move scripts to utils/outfits.

Sounds good.

>     * Update Za'lek drones builtin engines -> These do not look like any existing engine. I could find that:
>       * Za'lek light drone to scout drone +10 speed/accel/turn (not percents! 10 m.s-1, 10 m.s-2,  10 rad.s-1). Could be cleaner to just fix +3% everything.
>       * Light drone is Nexus +50% speed +50% turn +100% accel.
>       * Heavy drone has same accel and turn as Tricon Zephyr 2, but +14% speed.
>       * Bomber Drone is approx. Melendez +57% all three
>         In other words, the values do not make much sense. We will have to do that manually when engines are stable (therefore, not now).

I'm pretty sure I just messed around with numbers until it felt good. There are not referencing anything.

>     * [x]  Saved Za'lek drone engines notes as "ships/zalek/drone_engines.md"
>     * [x]  Update afterburners -> applied -1/6 to accel for light afterburners and adrenal glands. In addition, -10% accel bonus for high-end afterburners.
>       Medium afterburners left as is (as medium ships do not see their acceleration grow as much as small ships do).

Sounds good.

>     * [x]  Apply Nexus time constant reduction to the Tricon line.

I'm not sure this is a good thing. Why is this necessary? Can it be done without that?

>     * [x]  Reduce Wanderer/The Bite accel bonuses.

Sound sgood.

>     * [x]  Is it normal that the bite accel modifier is declared with `accel` instead of `accel_mod` ? -> Made lunge like adrenal gland ii + 50%, with additional +50% turn. (This makes it only slightly faster than adrenal gland, which I don't understand. I'll dig into this another time.)

Yes. Given that bite is used for ships of all sizes, a relative bonus never really worked for all ships. It would be either too strong for small ships, or too weak for large ships. Having a large absolute values make it work better with all.

>     * [x]  Found and fixed a bug in the_bite.lua

Nice catch!

> Since `generate.py` now uses explicitly some `.xml` outfits, we could even use these as templates (as `outfit.py` allows us to do that) and get rid of templates. I think I will propose a dedicated PR after multicores - part 2 (hulls).
>
> And, by the way, `meson.build` does not know about these dependencies. This is to be fixed in the dedicated PR I mentioned.

OK! Not sure if it would be better to make the Soromid outfits a bit difference instead of being exactly the same as the references, but that can be future work.

Trying it out:
1. Maybe we should decrease the trail lengths since some are based on acceleration? It may also be possible to make trails have something like fixed length + something based on acceleration instead of just acceleration due to the big disparity with acceleration now.
2. I feel like the turn is waaay too high on the smaller engines. The Hyena is a bit uncontrollable (it's even worse it you remove the time_mod).
3. I'm not sure if the changes are drastic enough compared to before. Pirate admonisher doesn't feel to have changed too much. For example, messing with the pirate admonisher, I feel almost like the Tricon Cyclon should be 140 speed + 500 accel or so, nearly double the current accel (does make accel bonuses much more important though)

I'm going to try messing with it more tomorrow and trying to get a feel for things. Thanks again for the great work!

johann1764:
> > In addition to that, I found relevant to apply the Nexus line time constant reduction to the Tricon line. Since Nexus line is not specialized so far, we could give it some other special effect. If this additional proposition is not to be kept, it is easy to undo it.
> Why is this? I think keeping tricon a bit harder to use would make sense.

Ok, let's remove it.

> > ```
> > * Update Za'lek drones builtin engines -> These do not look like any existing engine. I could find that:
> >   * Za'lek light drone to scout drone +10 speed/accel/turn (not percents! 10 m.s-1, 10 m.s-2,  10 rad.s-1). Could be cleaner to just fix +3% everything.
> >   * Light drone is Nexus +50% speed +50% turn +100% accel.
> >   * Heavy drone has same accel and turn as Tricon Zephyr 2, but +14% speed.
> >   * Bomber Drone is approx. Melendez +57% all three
> >     In other words, the values do not make much sense. We will have to do that manually when engines are stable (therefore, not now).
> > ```
> I'm pretty sure I just messed around with numbers until it felt good. There are not referencing anything.

I was beginning to suspect that..

> > ```
> > * [x]  Is it normal that the bite accel modifier is declared with `accel` instead of `accel_mod` ? -> Made lunge like adrenal gland ii + 50%, with additional +50% turn. (This makes it only slightly faster than adrenal gland, which I don't understand. I'll dig into this another time.)
> > ```
> Yes. Given that bite is used for ships of all sizes, a relative bonus never really worked for all ships. It would be either too strong for small ships, or too weak for large ships. Having a large absolute values make it work better with all.

Ok, so now we have something relative that works with all sizes (works with Arx, Nyx and Brigand)

> > Since `generate.py` now uses explicitly some `.xml` outfits, we could even use these as templates (as `outfit.py` allows us to do that) and get rid of templates. I think I will propose a dedicated PR after multicores - part 2 (hulls).
> > And, by the way, `meson.build` does not know about these dependencies. This is to be fixed in the dedicated PR I mentioned.
> 
> OK! Not sure if it would be better to make the Soromid outfits a bit difference instead of being exactly the same as the references, but that can be future work.

It does not have to be exactly the same, they are "based on", meaning any change is possible.
I find it clearer to have generate.py explicitly describe (through its code, not comments) how you get a gene drive from the original engine.

It is pretty handy to have them follow the changes made to the engines.
 
> Trying it out:
> 
> 1. Maybe we should decrease the trail lengths since some are based on acceleration? It may also be possible to make trails have something like fixed length + something based on acceleration instead of just acceleration due to the big disparity with acceleration now.

Absolutely. You can probably halve it. Interceptors now have ugly trails due to their excessive acceleration. Or maybe something non-linear, because of the interceptors.

> 2. I feel like the turn is waaay too high on the smaller engines. The Hyena is a bit uncontrollable (it's even worse it you remove the time_mod).

I can already restore the accel-related factor in turn rates as it was. (1/3.5 -> 1/4.0).

The problem with reducing it is that turn radius becomes too large ( as they have really high speeds )

> 3. I'm not sure if the changes are drastic enough compared to before. Pirate Admonisher doesn't feel to have changed too much. For example, messing with the pirate Admonisher, I feel almost like the Tricon Cyclone should be 140 speed + 500 accel or so, nearly double the current accel (does make accel bonuses much more important though)
Corvettes are of size 3 and the neutral point (point where accel/speed ratio is what it was) is 3.5, so Corvettes and Destroyers are not supposed to be very different with that setting.
However, the Tricon line change is important.

This is perfectly clear when you use a high-end afterburner (adrenal gland ii or Hades Torch) which are unfortunately not suited for this ship (maybe Hades Torch with 3 droid repair crew).

You can compare the difference between Nexus and Tricon line to get an idea.
Even more important, the difference between Tricon and Melendez.

Now the corvettes are so fast and maneuverable that I could easily beat my previous records on Melendez Tracks with an Odium in almost permanent afterburner (between 1/3 and 1/2 of the required time, depending on tracks).

> I'm going to try messing with it more tomorrow and trying to get a feel for things. Thanks again for the great work!

You'll probably notice the main concern here is that interceptors have now a huge accel and I am struggling to manage the consequences of it.

johann1764:
On turn rates :

So far, I use a linear combination of speed and accel to compute turn rate. Maybe I should do something more meaningful. Here it is : let T be the turn rate. The turn radius is max_speed/T. If we want it to be inversely proportional to agility A:
`max_speed/T=k/A`
In other words: `T = k' * A * max_speed`. Now we have to define A the agility, for example `A = (accel/speed)^c.`
I will therefore try different values of `k'` and `c` to see if we can find something interesting in the form:
`T=k' * (accel/speed)^c * (accel/3 + speed)`

johann1764:
If we also want the time to fullspeed (also time to stop from fullspeed),
distance to fullspeed, and turn radius..

Notice the Zephyr II / Dart crossing. I am starting to think the tricon overall bonus is too large.

|| speed | max speed | accel | fullsp time | fullsp dist | turn | turn radius
| --- | --- | --- | --- | --- | --- | --- | ---
| T. Zephyr Engine      |190    |400    |630    |  0.63 |127    |188    |122
| T. Zephyr II Engine   |220    |365    |435    |  0.84 |153    |155    |135
| N. Dart 160 Engine    |245    |350    |315    |  1.11 |194    |135    |149
| M. Ox XL Engine       |270    |320.67 |152    |  2.11 |338    |105    |175
| U. Hawk 360 Engine    |245    |316.67 |215    |  1.47 |233    |115    |158
| T. Cyclone Engine     |210    |301.67 |275    |  1.1  |165    |118    |146
| N. Arrow 700 Engine   |220    |265.67 |137    |  1.94 |258    | 89    |171
| K. Patagium Engine    |190    |260    |210    |  1.24 |161    | 97    |154
| T. Cyclone II Engine  |180    |233.33 |160    |  1.46 |170    | 84    |159
| Z. Test Engine        |172    |222.67 |152    |  1.46 |163    | 80    |159
| U. Falcon 1400 Engine |177    |203.67 | 80    |  2.55 |259    | 64    |182
| M. Buffalo XL Engine  |185    |203.67 | 56    |  3.64 |370    | 59    |198
| Beat Up Small Engine  |135    |192.67 |173    |  1.11 |107    | 75    |147
| K. Remige Engine      |145    |182.33 |112    |  1.63 |148    | 64    |163
| T. Typhoon Engine     |138    |167    | 87    |  1.92 |160    | 56    |171
| Beat Up Medium Engine |120    |145    | 75    |  1.93 |140    | 49    |170
| M. Mammoth Engine     |135    |145    | 30    |  4.83 |350    | 40    |208
| N. Bolt 3000 Engine   |130    |144.33 | 43    |  3.36 |242    | 43    |192
| T. Typhoon II Engine  | 96    |110.67 | 44    |  2.52 |139    | 35    |181
| U. Eagle 6500 Engine  | 90    | 97.33 | 22    |  4.42 |215    | 27    |207
| M. Mammoth XL Engine  | 92    | 97    | 15    |  6.47 |314    | 25    |222
| Beat Up Large Engine  | 72    | 80    | 24    |  3.33 |133    | 23.5  |195

johann1764:
> I guess the issue is that nexus and friends are defined as relative from Tricon without taking into account the progression between ranges. What could be done is instead of defining nexus as -15% Tricon, you could define it like 40% between Tricon A and Tricon B. That way you ensure that the max speeds are monotonically as expected.

That is true, it would work, and would be sort of more consistent. But that is more complicated, and does not solve another problem I have.

I find that the interceptor speed advantage over fighter has really become negligible. Even in AB, it is not significantly faster, and for a shorter time. I am considering the alternative I had previously mentioned of using different values for alpha and beta i.e. instead of -10% -20% -30% something like -15% -23% -31%. (I had a couple of values that would work)
This could solve both problems at once.

While at it, I found that the bite is broken yet another time (due to turn rates reductions).

And I think the reached speeds with AB become a bit too much (over 1600, twice as fast as weapons), which is somewhat mitigated by the now working AB power consumption, we should maybe use a lower base speed, like 300 instead of 400. Or: lower the part of accel in afterburner bonuses (a bit more speed, a little less accel).

Anyway, I will wait for your feedback on how it feels to pilot with the new engines before committing values corrections. I hope the latest correction on the turn rate will be satisfying for you, as the zephyr only gets less than 10% additional turn rate while at the same time keeping maneuverability improvements for corvettes, and keeping turn radius decreasing with agility (despite the increasing associated speed).

And if you feel the direct effect of accel is not sensible enough, we can still change the way max_speed is computed: for example use speed + 1/4 accel. Keeping same speed and max_speed will, with the change of formula, lead to an increase of 33% of felt acceleration.

bobbens:
> I find that the interceptor speed advantage over fighter has really become negligible. Even in AB, it is not significantly faster, and for a shorter time. I am considering the alternative I had previously mentioned of using different values for alpha and beta i.e. instead of -10% -20% -30% something like -15% -23% -31%. (I had a couple of values that would work) This could solve both problems at once.

It is sort of a tricky balancing issues. If the difference is too small, ultra-light are meaningless, if it is too large, they become optimal for travel. Also, many interceptors / small ships get movement bonuses, which does significantly amplify differences (hyenas probably need nerfing, as they are 463 speed with just tricon).

From a more theoretical point of view,

> While at it, I found that the bite is broken yet another time (due to turn rates reductions).

This should probably be tuned last.

> And I think the reached speeds with AB become a bit too much (over 1600, twice as fast as weapons), which is somewhat mitigated by the now working AB power consumption, we should maybe use a lower base speed, like 300 instead of 400. Or: lower the part of accel in afterburner bonuses (a bit more speed, a little less accel).

I think AB is a bit broken now without heating. Probably should add some sort of "charge" mechanic so you can't have infinite AB builds. I would like to encourage using AB for short bursts or maneuvers instead of some sort of permanent outfit.

> Anyway, I will wait for your feedback on how it feels to pilot with the new engines before committing values corrections. I hope the latest correction on the turn rate will be satisfying for you, as the zephyr only gets less than 10% additional turn rate while at the same time keeping maneuverability improvements for corvettes, and keeping turn radius decreasing with agility (despite the increasing associated speed).

I was going to try to mess with your scripts to play around with the values, but it seems like they use python2 not python3?

> And if you feel the direct effect of accel is not sensible enough, we can still change the way max_speed is computed: for example use speed + 1/4 accel. Keeping same speed and max_speed will, with the change of formula, lead to an increase of 33% of felt acceleration.

Yeah, our physics model creates a virtual force when the ship goes over the maximum speed. The force is set to 3.0 times the difference in speed, which ends up meaning you get +accel/3 bonus. It should be trivial to change it to 4.0, but it is a big change as there are lots of places that make that assumption here and there that we would have to hunt down.

johann1764:
> I was going to try to mess with your scripts to play around with the values, but it seems like they use python2 not python3?

I believe that most systems both have python2.x and python3 ?
Having sources in python3 using python2 does not seem to be an issue.
I did use the outfit.py module (in python 2) in generate.py (in python 3) and the build succeeded on Github.

If needed, I can make it python 3.

bobbens:
> > I was going to try to mess with your scripts to play around with the values, but it seems like they use python2 not python3?
> 
> I believe that most systems both have python2.x and python3 ? Having sources in python3 using python2 does not seem to be an issue. I did use the outfit.py module (in python 2) in generate.py (in python 3) and the build succeeded on Github.
> 
> If needed, I can make it python 3.

Arch Linux only has python2 in the AUR, and it seems to be broken (did not compile). I've tried to update it a bit to python3, but while it doesn't give errors, it doesn't seem to work.

johann1764:
Oh, and I forgot inline help.

Standard usage:
```
> ./utils/outfits/apply_engines.py `find dat/outfits/core_engine/`
```

johann1764:
> while it doesn't give errors, it doesn't seem to work.
I'll make it python3.

johann1764:
Done.

Accidentally pushed some minor bite changes with it. Should not be a problem.

While at it, I should also make an inline help. Like:
```
usage: ./apply.. <file1.xml> <file2.xml> ...
Ignores any non-xml file or folder.
```

bobbens:
> Done.
> 
> Accidentally pushed some minor bite changes with it. Should not be a problem.
> 
> While at it, I should also make an inline help. Like:
> 
> ```
> usage: ./apply.. <file1.xml> <file2.xml> ...
> Ignores any non-xml file or folder.
> ```

Just something basic with `argparse` would be great!

johann1764:
Apparently, I had forgotten to add the changed scripts...
Added basic help.

I did not know `argparse`, since it's so simple to do it manually.
I should use it in all scripts I added, would be cleaner.
I will do that later.

johann1764:
> From a more theoretical point of view,

There's a missing bit, here.

> 
> > While at it, I found that the bite is broken yet another time (due to turn rates reductions).
> 
> This should probably be tuned last.

Yes.

> 
> > And I think the reached speeds with AB become a bit too much (over 1600, twice as fast as weapons), which is somewhat mitigated by the now working AB power consumption, we should maybe use a lower base speed, like 300 instead of 400. Or: lower the part of accel in afterburner bonuses (a bit more speed, a little less accel).
> 
> I think AB is a bit broken now without heating. Probably should add some sort of "charge" mechanic so you can't have infinite AB builds. I would like to encourage using AB for short bursts or maneuvers instead of some sort of permanent outfit.

Wasn't the heating mechanics supposed to be restored at some point ?

> I was going to try to mess with your scripts to play around with the values, but it seems like they use python2 not python3?

For the restricted usage we make of python, it is easy to be simultaneously compatible with both. I finally did that for `apply_engines.py` and `engine_stats.py`, and I will do that each time in the future.

> 
> > And if you feel the direct effect of accel is not sensible enough, we can still change the way max_speed is computed: for example use speed + 1/4 accel. Keeping same speed and max_speed will, with the change of formula, lead to an increase of 33% of felt acceleration.
> 
> Yeah, our physics model creates a virtual force when the ship goes over the maximum speed. The force is set to 3.0 times the difference in speed, which ends up meaning you get +accel/3 bonus. It should be trivial to change it to 4.0, but it is a big change as there are lots of places that make that assumption here and there that we would have to hunt down.

Damn. We will do that if there is no other alternative, then. Missing one such assumption could lead to hard to find bugs.

bobbens:
> > I think AB is a bit broken now without heating. Probably should add some sort of "charge" mechanic so you can't have infinite AB builds. I would like to encourage using AB for short bursts or maneuvers instead of some sort of permanent outfit.
> 
> Wasn't the heating mechanics supposed to be restored at some point ?

If necessary, yes. I think some sort of charge mechanic would work. It would be similar to heat, but per-outfit. So each outfit could manage it's own "heat" instead of a global thing. It would allow for afterburner/blink drive overheating, charge weapons, etc... at a more fine level. Needs a separate proposal though.

> For the restricted usage we make of python, it is easy to be simultaneously compatible with both python2 and python3. I finally did that for `apply_engines.py` and `engine_stats.py`, and I will do that each time in the future.

I think python 3 only is fine, but whatever you prefer. I don't think python 2 is used anywhere, and it is pretty much gone already.

> > Yeah, our physics model creates a virtual force when the ship goes over the maximum speed. The force is set to 3.0 times the difference in speed, which ends up meaning you get +accel/3 bonus. It should be trivial to change it to 4.0, but it is a big change as there are lots of places that make that assumption here and there that we would have to hunt down.
> 
> Damn. We will do that if there is no other alternative, then. Missing one such assumption could lead to hard to find bugs.

Actually, the more I mess with this, the more I'm convinced this is the solution. I'm going to start to rewrite parts of the code to make it more general, so we can change it easily.

With extreme amounts of accel, given that a third goes to speed, you end up with really weird engines where the ship is basically stopped without accelerating. The extreme case where the penalty is infinite would not let you go over max speed (and be like the original escape velocity).

bobbens:
> Actually, the more I mess with this, the more I'm convinced this is the solution. I'm going to start to rewrite parts of the code to make it more general, so we can change it easily.

So past me, in all my wisdom, actually worked around most of the issues. I've implemented it in https://github.com/naev/naev/commit/33dd161dbe108bd3bf439abb4f733736bf16a64b (a bit hacky).
There's only 3 points where it has to be change atm:
1. physics.h header
2. autonav.lua
3. (broken with bite changes) code to use the bite
```
dat/ai/core/attack.lua:            if ai.dist( target ) < (p:speed() + (p:accel()+800)/3)*dtime then
dat/ai/feralbioship.lua:      if ai.dist( target ) < (p:speed() + (p:accel()+800)/3)*dtime then
```

It's a bit too late to mess around with this too much today, but it should be easier to test and I can give it a try tomorrow. I'm thinking if we are doubling the acceleration, halving the bonus might be good. So something like adding 15% (or less) accel to max speed (or 1/0.15 factor) might be a good starting point.

johann1764:
> It's a bit too late to mess around with this too much today, but it should be easier to test and I can give it a try tomorrow. I'm thinking if we are doubling the acceleration, halving the bonus might be good. So something like adding 15% (or less) accel to max speed (or 1/0.15 factor) might be a good starting point.


So, let the formula be speed + 1/6 acc. (as a promising starting point)
The script will therefore produce engines with doubled acceleration.
Most acceleration bonuses can stay as is, as the main in-game effect of acceleration is its contribution to speed (with or without AB), which is halved. Only a "minor" effect will be doubled: the rate at which you accelerate. (I agree that this is odd to present accel very definition as a minor effect. But that's kind of true.)

Maybe you could set it that way in the header file right before the nightly is compiled ?

For the rest, I will update lua files accordingly.

And thanks for the "The Bite" non-updated references that explain why it fails regularly with the interceptor. I did not suspect that!

johann1764:
Ok!

I just have to follow the TODOs, that's handy !
Good think I checked the diff.

bobbens:
I've pushed some changes to the merge cores branch. In particular,
1. Modified update_engines.py to be a bit easier for me to edit.
2. Have the python get_constants.py use regular expression so it can parse "1.0/0.15" (which is the value I'm testing now, because it gives 15% of accel, which is easier to understand than 16.67%)

johann1764:
> I've pushed some changes to the merge cores branch. In particular,
> 
> 1. Modified update_engines.py to be a bit easier for me to edit.

OK, looks clear to me.

> 2. Have the python get_constants.py use regular expression so it can parse "1.0/0.15" (which is the value I'm testing now, because it gives 15% of accel, which is easier to understand than 16.67%)

OK, simple and more general.
Is 15% clearer than 1/6 ? I don't know. Both are as readable and are almost the same. OK !

> 
> EDIT: Don't worry about merge commits, I'll probably rebase when merging into main.

Meaning you'll get the missing commit "turn off afterburners or equivalent when starting lunge"
when merging mergecores into main ?

bobbens:
OK, doing some tests with 1/0.15. I've bumped up values:

| | speed | max speed | accel | fullsp time | fullsp dist | turn | turn radius
| ---                   | ---    | ---    | ---    | ---    | ---    | ---    | ---
| T. Zephyr Engine      | 350    | 500    | 1000    |   0.5  | 125    | 285    | 101
| N. Dart 160 Engine    | 390    | 458.25 | 455    |   1.01 | 231    | 215    | 122
| T. Zephyr II Engine   | 315    | 426.75 | 745    |   0.57 | 122    | 235    | 104


https://github.com/user-attachments/assets/2e20d7ce-4f03-45dc-ae3d-15346a5672e5

It's probably a bit too much, but it does feel like flying is quite important now, and so is skill. For dogfighting with small ships, burst fire seems to be most important. I'm going to try to find values I like for small ships before going to medium ships, then large ships, and finally see if we can fit a curve or something, or maybe just use a list of manually chosen reference values.

johann1764:
> It's probably a bit too much, but it does feel like flying is quite important now, and so is skill.

You are far beyond the turn rate values you previously considered acceptable; see, it's pilotable. :-)
Maybe we can relax this constraint a bit, then. I felt it was a bit too constrained.
( the full turn rates range was really reduced )

> For dogfighting with small ships, burst fire seems to be most important.
It is one strategy. Works well with plasma weapons. There is another one, often used with Dvaered weapons : fill the space with projectiles to limit your opponent possible moves.

> I'm going to try to find values I like for small ships before going to medium ships, then large ships, and finally see if we can fit a curve or something,

I was more thinking of finding what is wrong with current curves (more this, less that, etc.) but any way the describe the desired look and feel is good. Then we adapt the curves to fit it.

> or maybe just use a list of manually chosen reference values.

Please, don't ! There are lots of constraints that are not tested here.
In terms of flexibilily and maintenance work, that's terrible.

johann1764:
And, by the way you can also try Qex tracks with afterburner to test the maneuverability of ships. If you can make almost the full turn in AB (Hades Torch.. Unicorp AB are too easy) with your corvette, then it is really maneuverable.

johann1764:
> OK, so with some minor modifications, I'm quite happy with tricon engines. You can dodge quite well, with additional movement upgrades they feel even better.
> 
> The main issue is that tricon are just soo much better than the rest, there is no real competition. You can pretty much equip tricon always (except pure carriers), so there is really no point to the other cores.

I agree with that, that's the main problem.
 
> The question is, is this something we want, and if not, what to do about it? I would argue it's not a good thing if you have no real choice of outfits, if you just want Tricon when possible. However, I think having weaker engines is also a nice thing for NPCs. This makes me think that giving them more distinct roles is pretty much necessary. Merging cores would also be quite beneficial, as we could add another line of outfits as necessary, and all ships would be able to use nexus + unicorp.

Having current Nexus/Unicorp line become just Nexus, a high-quality all-round engine with Unicorp a cheaper variant of the same profile is simple and would work well.
However, I think the max_speed gap between Tricon and Nexus is currently a bit too much.
Maybe trying -10% instead of -15% would be a good starting point.
Notice that, with this approach, Unicorp would have lower max speed than Melendez **once they are in movement**. I think this is a good compromise: you don't outrun running Melendez with standard engine, but they are really vulnerable at jump points and before landing.
> 
> For merging cores engines, I think averaging is the way to go so that speeds make sense. The question is whether we have primary / secondary distinction (so order matters) or just solo / double distinction.

That's mergecores chapter 3. We just finished chapter 1.

> 
> Rethinking the cores a bit:
> 
> core	max speed	accel	turn	energy	mass limit	comments
> tricon	high	extreme	extreme	high	low	ultimate dogfighting core
> melendez	low	low	low	average	extreme	ultimate cargo core
> unicorp	average	average	average	average	average	mainly low-end core for npcs
> krain	extreme	average	high	high	average	alternative high-end core for traveling and non-dogfighting
> nexus	high	average	average	average	high	utility core
> The hope here being that tricon / krain give a bit more competition, with tricon being better for dogfighting, while krain would be less punishing to equip and have slightly more top speed. Nexus would then play the role of the carrier outfit, while melendez / unicorp would be more NPCs / niche outfits.

To put things in a different perspective:

|               | Top sp | accel profile   | Quality | Requirements |
| ------------- | ------ | --------------- | ------- | ------------ |
| Krain         |  best  | all round/speed | high    |    medium    |
| Tricon        |  great | accel           | high    |     high     |
| Nexus         |  good  | all round       | high    |    medium    |
| Melendez      |  good  | speed           | high    |     high     |
| Unicorp       |  avg   | all round       | medium  |      low     |
| Beat Up Corp. |  bad   | all round       | low     |      low     |

Notice the turn rates follow from the accel profile.

I think the speed-oriented profile of Melendez is useful and playable,
and has the advantage of doing by itself what usual bomber modifiers do.
We should try to make it a bit more attractive, not less.

I like the idea to have Krain a less accel-oriented profile, even slightly speed-oriented,
but must have the top speed advantage, as it has the lowest engine mass limit of all.

> 
> Furthermore, with core merging, you can do things like Tricon+Krain to get "intermediate" combinations that can be useful for certain ships.

Nexus+Krain, Nexus+Tricon, and Nexus+Melendez should also be interesting to have a more nuanced set of accel profiles, in fact any pair of high-quality engines.
Here Nexus serves as a neutral ingredient in the mix to mitigate the downsides (and the power consumption) of the other.

Even, Melendez+Tricon could be used as a super-Nexus (all round but highly consuming).

I see lots of interesting combinations in this table.

Maybe we could have two med-quality engines instead of one, one all-round/speed and one all-round/accel.

> 
> I'm still going to mess with things a bit more, probably testing combinations more on the lines of the table above.
> 
> Other things:
> 
> 1. The AI is quite bad at such high accelerations and tends to do little "death circles" where fighters will basically spin around while accelerating and shooting you whenever they can. Probably would need to do something there.

> 2. With large accel, racing is quite easy, turning sensitivity is more of a problem.

That is why this is a good test!

> However, it should be possible to add powerups, hurdles, or other things to the race track to make it more important to be accurate and such, which would make it more interesting.

The idea is seducing, but don't we already have a bit of too much good ideas ?

bobbens:
> To put things in a different perspective:
> 	Top sp 	accel profile 	Quality 	Requirements
> Krain 	best 	all round/speed 	high 	medium
> Tricon 	great 	accel 	high 	high
> Nexus 	good 	all round 	high 	medium
> Melendez 	good 	speed 	high 	high
> Unicorp 	avg 	all round 	medium 	low
> Beat Up Corp. 	bad 	all round 	low 	low

I'm modifying the script to actually have 3 parameters: accel/speed ratio, maxspeed, and turn ratio. It should make it much easier to modify. We'll have to add the mass limit / energy stuff later. Also making it so the numbers are easier to interpret.

> Maybe we could have two med-quality engines instead of one, one all-round/speed and one all-round/accel.

Yeah, but if we merge cores and you get nexus + unicorp as different outfits, I do think it pretty much covers most. We can add more as is needed.

> > However, it should be possible to add powerups, hurdles, or other things to the race track to make it more important to be accurate and such, which would make it more interesting.
> 
> The idea is seducing, but don't we already have a bit of too much good ideas ?

Yes, this is just another thing to toss in the ideas pile and hope someone else implements it or we eventually get around to it. I do think since combat is such a fundamental gameplay aspect, we should balance everything around it being as fun as possible, and then try to deal with the ramifications :)

I should get the modified script up in a bit.

johann1764:
Afterthought:
For the twin slightly accel-oriented/slightly speed-oriented standard (cheap) engine lines, we could have Unicorp produce two lines, like the Unicorp Red/Unicorp Blue lines.
"Unicorp" feels standard, that is why it is perfect for average quality outfit with no marked trait.

bobbens:
> Afterthought: For the twin slightly accel-oriented/slightly speed-oriented standard (cheap) engine lines, we could have Unicorp produce two lines, like the Unicorp Red/Unicorp Blue lines. "Unicorp" feels standard, that is why it is perfect for average quality outfit with no marked trait.

Yes. I still believe in having a Unicorp with builtin afterburner is good :P

I've pushed the script (but didn't update the data files). The constant is different, so you won't be able to test well without compiling yourself. That said, I'm going to modify main so that the constant can be set via xml or something (for modding!) so you will be able to test it in a bit.

johann1764:
I have been suspicious of:
```python
fullspeed = dec( size + 1.0 - stats["speed"])
```
it turns out that:

johann1764:
As a replacement, you could:

```python
define dec_f(f):
  n=floor(f)
  q=f-n
  return pow(dec(n),1-q)*pow(dec(n+1),q)
```
and use dec_f instead of dec.

NOTE: GitHub is currently experiencing serious problems. Or maybe my connection does.

TODO: graph it.

bobbens:
> ```python
>   n=floor(f)
>   q=f-n
>   return pow(dec(n),1-q)*pow(dec(n+1),q)
> ```

Nice catch, thanks! Fixed it and also made it so with tomorrow's nightly you'll be able to override `dat/constants.lua` which is now read by the engine to change the PHYSICS_SPEED_DAMP constant.

bobbens:
I've fixed the merge conflicts in the `mergecores` branch and synced with main.

johann1764:
Does that mean I can ignore the "This branch has conflicts that must be resolved" message ?

bobbens:
> Does that mean I can ignore the "This branch has conflicts that must be resolved" message ?

Yes, or you can pull from the mergecores branch and push to yours and it should get solved.

bobbens:
OK, so the following feels fairly good. Up to Cyclone with some movement upgrades it feels like you can actually dodge well and such. Cyclone II, while still feeling agile, lacks the top speed to actually get away well. The capship typhoons might feel a bit too agile, but the horrible top speed makes up for it. That said, they are significantly more agile than before.

| | speed | max speed | accel | fullsp time | fullsp dist | turn | turn radius
| ---                   | ---    | ---    | ---    | ---    | ---    | ---    | ---
| T. Zephyr Engine      | 280    | 400    | 800    |   0.5  | 100    | 225    | 102
| T. Zephyr II Engine   | 255    | 344.25 | 595    |   0.58 | 100    | 187    | 105
| T. Cyclone Engine     | 215    | 278    | 420    |   0.66 |  92    | 145    | 110
| T. Cyclone II Engine  | 173    | 215.75 | 285    |   0.76 |  82    | 110    | 112
| T. Typhoon Engine     | 132    | 159.45 | 183    |   0.87 |  69    |  78    | 117
| T. Typhoon II Engine  |  96    | 112.95 | 113    |   1    |  56    |  54    | 120

I'm not sure if it makes sense to merge it before splitting cores, however, there is still the question of the other outfits (which feel okish, but need polishing). Furthermore, the mass limits and energy don't really make sense. In particular, Krain engines are totally bad because they have too bad mass limit while behaving similar to a Tricon.
Additionally, without the merge, Nexus and Unicorp are weird as they are in separate sizes only.

johann1764:
> I'm not sure if it makes sense to merge it before splitting cores, however, there is still the question of the other outfits (which feel okish, but need polishing). Furthermore, the mass limits and energy don't really make sense. In particular, Krain engines are totally bad because they have too bad mass limit while behaving similar to a Tricon. Additionally, without the merge, Nexus and Unicorp are weird as they are in separate sizes only.

I suggest we merge now.
 - Then, I can open PR for multi-hulls (unfinished so far, but on the way).
 - And I will open a new, fresh PR for a necessary step toward multi-engines: completing the engine lines in all sizes. It will be a good opportunity to manage all the details left out so far: having Nexus/Unicorp two (or 3!) different lines, etc. It will be good to have a "game contents PR" with almost no technical aspect to manage.

The way I see it, most of the in-game effects should be set by that latter PR.

The multi-engine thing will be the final step, that will "only" allow some "engine mixings" and will be challenging enough on its own.

johann1764:
Before we fine-tune the engines lines in a future PR, I would suggest to keep it less experimental, just to have the engine lines main characteristics clear.

Probably turn scale should be increased a bit in the general formula.

We can already implement the slightly speed-oriented Krain line.
(Just same accel as Tricon but with a little more max_speed)

```diff
diff --git a/utils/outfits/apply_engines.py b/utils/outfits/apply_engines.py
index 179f9cddc7..b51879a11d 100755
--- a/utils/outfits/apply_engines.py
+++ b/utils/outfits/apply_engines.py
@@ -68,40 +68,40 @@ lines={
 }
 
 line_stats = {
-    "T" : {
-        "ratio" : 2.0, # 2.0 is double accel vs speed (at size 1)
-        "speed" : 1.0, # 1.0 indicates current speed rank, lower means slower, higher means faster
-        "turn"  : 1.0, # 1.0 means base turn rate, lower is worse, higher is better
-    },
     "K" : {
-        "ratio" : 1.4, # lower ratio
-        "speed" : 1.1, # higher speed than average
+        "ratio" : 0.9, # slightly speed-oriented
+        "speed" : 1.1, # indicates current speed rank, lower means slower, higher means faster. 1.1 is the highest speed.
         "turn"  : 1.0,
     },
+    "T" : {
+        "ratio" : 1.7, # 2.0 would be double accel vs speed (at size 1)
+        "speed" : 1.0, # High Quality + low engine mass limit bonus
+        "turn"  : 1.0, # 1.0 means base turn rate, lower is worse, higher is better
+    },
     "N" : {
-        "ratio" : 1.0,
-        "speed" : 1.0, # Pretty good but slightly slower top speed
-        "turn"  : 0.8,
+        "ratio" : 1.0, # All-round
+        "speed" : 0.9, # Standard high quality
+        "turn"  : 1.0,
     },
     "M" : {
-        "ratio" : 0.5,
-        "speed" : 1.0,
+        "ratio" : 0.7, # Really speed-oriented
+        "speed" : 0.9,
         "turn"  : 1.0,
     },
     "U" : {
         "ratio" : 1.0,
-        "speed" : 0.8,
-        "turn"  : 0.8,
+        "speed" : 0.8, # Standard
+        "turn"  : 1.0,
     },
     "Z" : { # TODO make these change over time the profile via Lua
         "ratio" : 1.2,
-        "speed" : 0.7,
-        "turn"  : 0.8,
+        "speed" : 0.7,
+        "turn"  : 1.0,
     },
     "B" : {
-        "ratio" : 0.8,
-        "speed" : 0.3,
-        "turn"  : 0.6,
+        "ratio" : 1.0,
+        "speed" : 0.6,
+        "turn"  : 1.0,
     },
 }
```

bobbens:
> Before we fine-tune the engines lines in a future PR, I would suggest to keep it less experimental, just to have the engine lines main characteristics clear.
> 
> Probably turn scale should be increased a bit in the general formula.
> 
> We can already implement the slightly speed-oriented Krain line. (Just same accel as Tricon would have with no mod but with a little more max_speed)

OK. Before merging, we have to figure out mass limits and energy usage though. In particular, the Krain outfits are pretty useless now due to the mass limit being so low, and speed being based on the upper size.

johann1764:
> OK. Before merging, we have to figure out mass limits and energy usage though.

Messing with energy usage may cause bugs, I'd rather play with that in the next PR "completing engine lines and engine polishing".

I'd rather keep mass limits as is, as mass is probably hardest thing to balance in Naev (because it connects with everything). I prefer to adjust engine properties with regards to their mass (like Krain and Tricon having related speed advantages) than the opposite. And players will prefer it too, as they certainly don't want some of their ships suddenly become over mass limit.

> In particular, the Krain outfits are pretty useless now due to the mass limit being so low, and speed being based on the upper size.

That's easy to solve: "speed" : 1.3
will make it on the way between its size and its size-1.

johann1764:
I have two good reasons for merging _now_:

 1. Future PR "multicores chapter 2 - hulls" is almost finished. I even implemented lots of bonuses (parsearg-ed and improved some scripts, finished generate.py automatic core updates, etc.) Since it is based on this branch ( reuses and modifies scripts versions of this branch ), I cannot open the PR until this is merged. When this is merged, I will be able to rebase it &#8288;-&#8288;&#8288;-&#8288;onto current main and then open a PR.
 2. We have a full TODO list that describes the future PR "completing engine lines and engine polishing". When specification is complete, it is time for implementation.
    - Fine-tune some parameters (maybe a **tiny** increase in AG_EXP) (to be tested)
    - Complete Unicorp, Nexus and Melendez lines.
    - Fine-tune engines energy consumption.
    - Discuss the possibility of two Unicorp lines. For example, could have a "blue" line that is slightly speed oriented with some special, and the "red" line that would be slightly accel-oriented with builtin AB.
    - See the "Future Work" section of the initial post.
    - Start thinking about the Patagium problem. (A destroyer engine with no corvette engine counterpart.. this does not fit neatly into the multi-engines framework)

bobbens:
> I have two good reasons for merging _now_:
>     1. Future PR "multicores chapter 2 - hulls" is almost finished (only equipopt left). I even implemented lots of bonuses (parsearg-ed and improved some scripts, finished generate.py automatic core updates, etc.) Since it is based on this branch ( reuses and modifies scripts versions of this branch ), I cannot open the PR until this is merged. When this is merged, I will be able to rebase it ⁠-⁠⁠-⁠onto current main and then open a PR.

Technically you can open a PR against the mergecores branch in this case.

>     2. We have a full TODO list that describes the future PR "completing engine lines and engine polishing". When specification is complete, it is time for implementation.
>        * Fine-tune some parameters (maybe a **tiny** increase in AG_EXP) (to be tested)
>        * Complete Unicorp, Nexus and Melendez lines.
>        * Fine-tune engines energy consumption.
>        * Discuss the possibility of two Unicorp lines. For example, could have a "blue" line that is slightly speed oriented with some special, and the "red" line that would be slightly accel-oriented with builtin AB.
>        * See the "Future Work" section of the initial post.
>        * Start thinking about the Patagium problem. (A destroyer engine with no corvette engine counterpart.. this does not fit neatly into the multi-engines framework. We need to find a trick.) We need a solution that is less an overkill than "grant Patagium half-engines the property that they require a twin to work".

Yes, I think it needs a bit more polish before merging. I'll try to give a shot at it today or tomorrow.

Regarding the Patagium problem, it is just an X-0.5 tier slot, which allows for certain ships that are designed for X-0.5 tiers to work with them. If we simply add outfits that are X-0.5 tier it should work as is. Would also let 2-slot ships to also do things like go for X-0.25 or the likes. However, this requires handling energy + mass limits in a potentially more holistic way, since the mass limit constraints have to be meaningful or we end up with the everything is tricon local minimum.

johann1764:
I don't know what to do with the mergecores branch, as the parameters don't match the values.
Which one are we supposed to trust ?

bobbens:
> I don't know what to do with the mergecores branch, as the parameters don't match the values. Which one are we supposed to trust ?

What do you mean? The .xml values should be synced with the script (that has some changes I made to).

johann1764:
... and some apply_engines formula now contain errors.

johann1764:
> What do you mean? The .xml values should be synced with the script (that has some changes I made to).

I pulled. Applying apply_engines produces some modifications, which means a mismatch between parameters and values.

bobbens:
> > What do you mean? The .xml values should be synced with the script (that has some changes I made to).
> 
> I pulled. Applying apply_engines produces some modifications, which means a mismatch between parameters and values.

I can not reproduce:

```
$ git clone --depth 1 --branch mergecores https://github.com/naev/naev.git
Cloning into 'naev'...
remote: Enumerating objects: 5218, done.
remote: Counting objects: 100% (5218/5218), done.
remote: Compressing objects: 100% (4218/4218), done.
remote: Total 5218 (delta 1547), reused 1901 (delta 927), pack-reused 0 (from 0)
Receiving objects: 100% (5218/5218), 16.96 MiB | 7.87 MiB/s, done.
Resolving deltas: 100% (1547/1547), done.
[ess@kamamon (/tmp)] $ cd naev
[ess@kamamon (/tmp/naev)] $ ./utils/outfits/apply_engines.py `find dat/outfits/core_engine/`
zalek_test_engine.xml: _
unicorp_hawk_360_engine.xml: _
tricon_zephyr_ii_engine.xml: _
tricon_zephyr_engine.xml: _
nexus_dart_160_engine.xml: _
melendez_ox_xl_engine.xml: _
beat_up_small_engine.xml: _
unicorp_falcon_1400_engine.xml: _
tricon_cyclone_ii_engine.xml: _
tricon_cyclone_engine.xml: _
nexus_arrow_700_engine.xml: _
melendez_buffalo_xl_engine.xml: _
krain_patagium_engine.xml: _
beat_up_medium_engine.xml: _
unicorp_eagle_6500_engine.xml: _
tricon_typhoon_ii_engine.xml: _
tricon_typhoon_engine.xml: _
nexus_bolt_3000_engine.xml: _
melendez_mammoth_xl_engine.xml: _
melendez_mammoth_engine.xml: _
krain_remige_engine.xml: _
beat_up_large_engine.xml: _
[ess@kamamon (/tmp/naev)] $
```

Maybe it's a python version issue or something?

johann1764:
```bash
~/naev$ git checkout mergecores
~/naev$ git pull
~/naev$ git checkout .
~/naev$ ./utils/outfits/apply_engines.py `find dat/outfits/core_engine/`
unicorp_falcon_1400_engine.xml: speed:183->188, accel:135->100, turn:70->66
tricon_cyclone_ii_engine.xml: speed:173->183, accel:285->215, turn:110->102
nexus_arrow_700_engine.xml: speed:245->255, accel:210->140, turn:99->90
...
```

johann1764:
> I can not reproduce:

Our respective logs seem to contradict each other. I don't get it.

Besides that:
```python
   # Proportion of accel vs speed, defaults so that it is 1:1 at size 1.
   r = 0.5*pow(2,-((size-1)-5)/5) / PHYSICS_SPEED_DAMP
```
 - The comment is false.
 - Where does that PHYSICS_SPEED_DAMP comes from ? Should be removed, that would also solve previous point.
 - You square rooted the effects of size over the ratio. Was:
```python
pow(2,-((size-1)-5/2)/(5/2)) # ranging from 0.5 to 2.0
```
now:
```python
0.5*pow(2,-((size-1)-5)/5) # ranging from 0.5 to 1.0
```

johann1764:
> $ git pull https://github.com/naev/naev.git mergecores

```
~/naev$ git pull https://github.com/naev/naev.git mergecores
Depuis https://github.com/naev/naev
 * branch                  mergecores -> FETCH_HEAD
Fetching submodule artwork
Fetching submodule dat/lua-repl
Fetching submodule extras/windows/mingw-bundledlls
Déjà à jour.
~/naev$ git checkout .
0 chemin mis à jour depuis l'index
~/naev$ ./utils/outfits/apply_engines.py `find dat/outfits/core_engine/`
unicorp_falcon_1400_engine.xml: speed:183->188, accel:135->100, turn:70->66
tricon_cyclone_ii_engine.xml: speed:173->183, accel:285->215, turn:110->102
nexus_arrow_700_engine.xml: speed:245->255, accel:210->140, turn:99->90
melendez_buffalo_xl_engine.xml: accel:71->54, turn:80->76
beat_up_medium_engine.xml: speed:210->220, accel:140->93, turn:59->54
...
```

johann1764:
```bash
~/test$ git clone --depth 1 --branch mergecores https://github.com/naev/naev.git
Clonage dans 'naev'...
remote: Enumerating objects: 5218, done.
remote: Counting objects: 100% (5218/5218), done.
remote: Compressing objects: 100% (4218/4218), done.
remote: Total 5218 (delta 1547), reused 1910 (delta 927), pack-reused 0 (from 0)
Réception d objets: 100% (5218/5218), 16.96 Mio | 3.60 Mio/s, fait.
Résolution des deltas: 100% (1547/1547), fait.
Mise à jour des fichiers: 100% (4918/4918), fait.

~/test$ cd naev/
~/test/naev$ ./utils/outfits/apply_engines.py `find dat/outfits/core_engine/`
unicorp_falcon_1400_engine.xml: speed:183->188, accel:135->100, turn:70->66
tricon_cyclone_ii_engine.xml: speed:173->183, accel:285->215, turn:110->102
nexus_arrow_700_engine.xml: speed:245->255, accel:210->140, turn:99->90
melendez_buffalo_xl_engine.xml: accel:71->54, turn:80->76
beat_up_medium_engine.xml: speed:210->220, accel:140->93, turn:59->54
...
```

johann1764:
Ok, maybe see if using apply_engine in:

https://github.com/johann1764/naev/tree/apply_engine

changes anything.

johann1764:
> > Ok, maybe see if using apply_engine in:
> > https://github.com/johann1764/naev/tree/apply_engine
> > changes anything.
> 
> That seems to work! Thanks!
> 
> > ```python
> >    # Proportion of accel vs speed, defaults so that it is 1:1 at size 1.
> >    r = 0.5*pow(2,-((size-1)-5)/5) / PHYSICS_SPEED_DAMP
> > ```
> > * The comment is false.
> > * Where does that PHYSICS_SPEED_DAMP comes from ? Should be removed, that would also solve previous point.
> > ```
> 
> OK, you are right. With my formula, it does give 1:1 ratio of fullspeed:accel, however, it's non linear. I've messed around with it a bit, and think the following would be more intuitive:
> 
> ```python
>    # Modulate full speed based on the speed stat
>    fullspeed = dec( size + 1.0 - stats["speed"])
> 
>    # Proportion of accel vs speed, defaults so that it is 1:1 at size 1.
>    # Curve decided so that it is 1 at size==1, 0.9 at size==3, and 0.5 at size==6
>    r = 1 + 1/60 * (size-1) - 1/60 * (size-1)**2
> 
>    # Modulate ratio based on outfit
>    r *= line_stats[line]["ratio"]
> 
>    # Correction so that 1 makes it so that the engine has equal speed and accel
>    r = r / (PHYSICS_SPEED_DAMP+r)
>    assert( r >= 0 and r <= 1 )
> ```
> 
> This does 2 things:
> 
> 1. Makes it so that a "ratio" of 1 gives 1:1 speed to accel, and 2 gives 1:2 speed to accel at size 1.
> 2. Reduces the ratio slowly until size 3, and then drastically to size 6.
> 
> That should make it so that capships feel significantly less maneuverable than small ships, while keeping corvettes at a good position.
> 
> Have to test the values again since things have changed quite a bit.

I would rather not change the parameterisation yet another time, as the max_speed / ratio / PHYSICS_SPEED_DAMP is a good one from my POV, as we can change them independently without messing up with the rest.

Notice I had made a mistake, as the turn function was dependent on the reference ratio, which should not be the case. Now everything is in order and we can change the parameters independently. The setting I provide is reasonable : it has the effects we wanted to produce with the least possible changes to current situation.

I agree to change the laws if some effect could not be achieved with it, but so far I don't see one.

If you are trying to achieve that :
> That should make it so that capships feel significantly less maneuverable than small ships, while keeping corvettes at a good position.

it is already working.

I kind of wonder why doing this now ? The main branch can keep this "reasonable setting" for a few days before "engine polishing and engine line completion" is merged.

johann1764:
If needed, I can PR https://github.com/johann1764/naev/tree/apply_engine into naev/mergecores.

bobbens:
> I would rather not change the parameterisation yet another time, as the max_speed / ratio / PHYSICS_SPEED_DAMP is a good one from my POV, as we can change them independently without messing up with the rest.

I'm mainly trying to aim at more interpretable results. So that the values we tweak have simple interpretations. This reparametrization behaves fairly similar to before the changes, while making it so ratio has a direct meaning. A ratio of 2 compared to 1 will have twice the value of accel/speed, which makes tuning much more direct. Without the normalization term `(r / (PHYSICS_SPEED_DAMP+r))`, that is not the case as there are non-linear effects as you stray from 1. This eliminates a bit of guesswork.

Additionally, the quadratic curve fitting gives the advantage of being able to choose 3 points and make sure the curve goes through them. In this case, it can be used so that we have precise control of the ratio decrease at size=1, size=3, and size=6. The inverse exponential is a bit harder to tune in my experience.

> I kind of wonder why doing this now ? The main branch can keep this "reasonable setting" for a few days before "engine polishing and engine line completion" is merged.

My experience is that once something gets merged, it gets a bit harder to change, so I would rather try to think it over a bit more, even if it'll probably get messed up again with splitting engine cores.

> If needed, I can PR https://github.com/johann1764/naev/tree/apply_engine into naev/mergecores.

No need, I've merged it already. Thanks!

One thing though, it seems like the accel values have doubled, with 1600 accel tricon zephyr. I'll try to fix it, likely with the different parametrization above, and mess with the numbers so it doesn't break too much stuff with respect to main.

johann1764:
Ok.

You prefer a parameterisation on more intuitive parameters, I prefer a parameterisation on independent parameters. Should we change specs for improving code readability ? This is an old debate we won't solve today.

I am not fond of the polynomial:
```
# Curve decided so that it is 1 at size==1, 0.9 at size==3, and 0.5 at size==6
```
is not clear. Where do these value come from ?
```
# Curve decided so that it is 2.0 at size==1, 1.0 at size==3.5, and 0.5 at size==6
```
would be clearer to me, as the idea is obvious. It is all the more obvious as everything else in the game follows that exponential logic.

Anyway, I will have a look when it's merged to main.

bobbens:
> You prefer a parameterisation on more intuitive parameters, I prefer a parameterisation on independent parameters. Should we change specs for improving code readability ? This is an old debate we won't solve today.

I find it ideal if you can directly tune the values you want. If we are interested in accel/speed ratios, we should be able to directly tune those instead of having to juggle accel + speed to get the same result. Ideally, the numbers should also be in such a way that the natural values give some reference outfit, and should be all normalized to similar or interpretable ranges.

> I am not fond of the polynomial:
> 
> ```
> # Curve decided so that it is 1 at size==1, 0.9 at size==3, and 0.5 at size==6
> ```
> 
> is not clear. Where do these value come from ?

They come from directly fitting a quadratic polynomial to the points (1,1), (3,0.9), and (6,0.5):
```python
import numpy as np
print( np.polyfit( [1,3,6], [1,0.9,0.5], 2 ) )
# [-0.01666667  0.01666667  1.        ]
```
Adding numpy as a dependency could make it so that we directly compute it from the control points if it is more intuitive. (could be also written without numpy, but that will be a bit uglier and more work)

> ```
> # Curve decided so that it is 2.0 at size==1, 1.0 at size==3.5, and 0.5 at size==6
> ```
> 
> would be clearer to me, as the idea is obvious. It is all the more obvious as everything else in the game follows that exponential logic.

Well, for engines, correct me if I'm wrong, but the main distinction should happen between size 3 and 4 when going from corvette to destroyer. Size 3 and below are "agile", while size 4+ are "bulky". So I feel like it should almost be defined by two curves, one for sizes 1-3, and one for 4-6. Exponentials have a bit of trouble expressing this. Additionally, if you want to fit points with an exponential, the math is a bit more involved than fitting a quadratic polynomial.

johann1764:
> I find it ideal if you can directly tune the values you want.

That's the point. I don't want values, I want behaviors, I want properties to be fulfilled, I want constraints to be satisfied. This results in the game either feeling this way or that way.

For example, we want Melendez line to be less maneuverable than most lines,
with slow turns and long time to go to full speed, but to have better top speed than basic engines.
We want smaller ships to be faster and to have better agility.
We want a significant difference between two sizes, but a difference that can be compensated in the case of light ships.

The true parameters I use are top speed, agility and the physical constant. Pretty easy to work with that. First you set the engine speed hierarchy, then give them different agility profiles. Straightforward. And changing the constant will keep everything intact.

> Well, for engines, correct me if I'm wrong, but the main distinction should happen between size 3 and 4 when going from corvette to destroyer. Size 3 and below are "agile", while size 4+ are "bulky". So I feel like it should almost be defined by two curves, one for sizes 1-3, and one for 4-6. Exponentials have a bit of trouble expressing this. Additionally, if you want to fit points with an exponential, the math is a bit more involved than fitting a quadratic polynomial.

As much as possible, I would like to have all sizes to be interesting. Currently, destroyers are really bad as they are barely stronger than corvettes without the speed and agility advantage. I don't think we should try to make this gap even larger.

bobbens:
> > I find it ideal if you can directly tune the values you want.
> 
> That's the point. I don't want values, I want behaviors, I want properties to be fulfilled, I want constraints to be satisfied. This results in the game either feeling this way or that way.

I agree. So it's best to add balancing knobs that directly modify the properties / constraints when possible.

> For example, we want Melendez line to be less maneuverable than most lines, with slow turns and long time to go to full speed, but to have better top speed than basic engines. We want smaller ships to be faster and to have better agility. We want a significant difference between two sizes, but a difference that can be compensated in the case of light ships.
> 
> The true parameters I use are top speed, agility and the physical constant. Pretty easy to work with that. First you set the engine speed hierarchy, then give them different agility profiles. Straightforward. And changing the constant will keep everything intact.

Which is why the current knobs available are top speed, and agility (ratio), and they are also directly comparable now, and independent of PHYSICS_SPEED_DAMP. Double the value is double the real result. We still need to add a knob for physical constraints, and then we can have fine-tuning knobs (allowing directly manipulating other results). However, 95% of the balancing should be done just with the main dominant knobs. If the dominant knobs are good, secondary knobs will not be able to fix it.

> > Well, for engines, correct me if I'm wrong, but the main distinction should happen between size 3 and 4 when going from corvette to destroyer. Size 3 and below are "agile", while size 4+ are "bulky". So I feel like it should almost be defined by two curves, one for sizes 1-3, and one for 4-6. Exponentials have a bit of trouble expressing this. Additionally, if you want to fit points with an exponential, the math is a bit more involved than fitting a quadratic polynomial.
> 
> As much as possible, I would like to have all sizes to be interesting. Currently, destroyers are really bad as they are barely stronger than corvettes without the speed and agility advantage. I don't think we should try to make this gap even larger.

I agree that all ships should have their usage. The main advantage of destroyers is they get fighter bays + long range heavy weapons + significantly more health which helps them survive. Corvettes only have their mobility. Even with my "polynomial", corvettes have 90% the agility, destroyers have 80% agility, and it mainly drops off harder for cruiser + capship.

That said, I still have to go through the different classes, test them, and also test the different lines before merging.

> > 1. Flatten the speed curve. Right now it's something like small ships are 300 base speed, medium are 200 base large are 100 base speed (for Nexus engine, or 345 - 290 - 230 - 175 - 115 - 80 for the tricon). I could imagine something like (from interceptor to carrier) 350 - 320 - 270 - 230 - 160 - 100 or more. Wouldn't solve the issue, but would larger ships less of sitting ducks.
> > 2. Give more varied speed/accel profiles. For example, since true max speed is `speed + accel/3`, you can have similar true max speed but play with ratios of speed and accel. So for example, you could have an outfit with 400 speed and 300 accel for 500 true speed, and one with 300 speed and 600 accel for the same 500 true speed. The first one would take 1.3 seconds to accelerate to full speed, while the second would accelerate in 0.5 seconds, and be able to dodge much better. Melendez engines could have high top speed and low accel, while the tricons could focus more on high accel and low top speed.
> > 3. In general, it might be better to give smaller ships more accel and less top speed too. That would make them feel more agile and better at dodging.
> 
> I believe this constraint is satisfied. :-)
> 
> What constraint are you looking for to satisfy ? I believe that is what I really don't understand here. Probably that understanding the purpose would help me to understand the means.

I think we have all the pieces. The main thing is figuring out some values. Basically questions like:
1. Is this accel/speed ratio good for a this class/engine?
2.  Is there enough difference between interceptor / fighter / etc...
3.  Do bulky ships feel bulky? Are NPCs fine with the new values?
4.  Is combat more enjoyable now?
5.  Is traveling more enjoyable now? etc...
6. Are there actual engine choices or is it just tricon on bust? (this can be helped with multicores, but it shouldn't be too lopsided first)

Also, now the PHYSICS_SPEED_DAMP is set to 1.0/0.15. I haven't tried many other values, so it may be worth also trying values like 1.0/0.2 and the likes. Although I think it's more important to figure out the good "max" acceleration values first, as that will probably determine the best PHYSICS_SPEED_DAMP profile, especially now that the script is made "independent" of PHYSICS_SPEED_DAMP.

johann1764:
> 
> It does not? For example, tricon has a 2 ratio at size 1.
> 
> With PHYSICS_SPEED_DAMP=3 we get:
> 
> speed	max speed	accel	fullsp time	fullsp dist	turn	turn radius
> T. Zephyr Engine	240	400	480	0.83	167	270	85
> Which is a 480/240 or 2:1 ratio.
> 
> If we change PHYSICS_SPEED_DAMP=1/0.15, we get:
> 
> speed	max speed	accel	fullsp time	fullsp dist	turn	turn radius
> T. Zephyr Engine	310	402.25	615	0.65	132	230	100
> Which is getting rounded a bit, but it's 615/310 which is also almost a 2:1 ratio.

Thanks for the example, that proves my point. You changed PHYSICS_SPEED_DAMP, and everything changed. With my formula, doubling PHYSICS_SPEED_DAMP would have doubled accel, halved fullspeed time, and left everything else intact.
 
> OK, let me see if I can elaborate on my train of thought.
> 
> The way I approached it was to try to get small ship dogfighting to feel like it was meaningful. So that you can dodge and have the ship feel very responsive. For small ships, something like 600-800 accel feels to be a fairly good spot, which is a lot more than the main values, and also significantly more than this PR. However, with the PHYSICS_SPEED_DAMP=3, this makes it feel really awkward when you're not accelerating, since you basically come to a stop. Which is why I lowered it to 1/0.15, which feels quite a bit better, as fighters no longer come to a stop when you don't accel, so you can do better maneuvers with drifting.

There's a point here. Highly agile ship lack drift speed. But actually, that's a good thing, as you can easily perform fast half turns by stopping accel, turning, and then accel.
We could also think of side engines to perform lateral moves, but that's out of scope.

> However, having tricon now have a 2:1 ratio, makes it feel sooo much better than the other engines, that, combined with the best top speed, makes the choice of engine a no brainer. For this reason, it is necessary to tweak the other engines to make them reasonable choices in some situations.

Ok, that's a clear contraint. 

> 
> Additionally, while I was pretty happy with tricon on size 1-3, it feels weird on size 4+. It feels like the ratio is actually too large, which led me to modify the exponential function. However, I wasn't really able to easily get it to behave as I wanted, and tweaked is fairly non-intuitive so I tried with the the quadratic curve fitting, and this is where we are now.

Ok.

> 
> OK, so I've gone back to the drawing board and tried the three different branches: main, this PR, and my latest changes flying both lancelot + admonisher in the gauntlet, and honestly, I feel like I was probably wrong and may be going too overboard.
> 
> So in main, while accel is very low (like 120 accel for a 330 speed corvette), the advantage of the low accel is that it gives a longer range dimension to combat, which I think is appreciated, and having too high accel (combined with high turn), does make combat a bit too close range and frantic (especially for smaller ships).

That's what I was looking for : a clear spec.


> 
> I'm going to try to mess with it a bit more to get a bit more conservative values. Probably closer to a mix of main and the PR. I've pushed some changes. While the size 1-3 feel quite good, I think size 6 feels a bit too agile. I'll need to test more. Haven't really tried non-tricon engines yet. For reference, current values:

Ok, leave it in your post, we may need this later.

> Another thing, I know you like your exponential functions, but I find tuning the ALPHA and BETA parameters to get dec behave in certain ways a bit hard.

Well, you defined this function. I just amended its parameters to solve a problem.

> I'm tempted to try a quadratic function so that it is easier to specify some exact values you want.
> 
> Sorry for all the noise and back and forth, but balancing is not my strength.

Balancing is hard.

I think we do it the wrong way. I am unhappy about this, and there would be so much better use of your time.

I suggest you do constraint programming. You specified a lots of things above, and I can easily see how to meet these specifications. Then I will propose solutions (concrete parameters), and you will add constraints to the system to exclude the solutions you don't want, or to reinforce behaviors you want. With words, just as you did a few lines before. With a few iterations, we should have something that meets your expectations and that I find satisfying.

Technically, this would mean I open a PR on naev/mergecores, and you'll provide feedback at each new setting I provide in answer to the previous constraints.
I commit
You add a constraint
I commit
You would prefer something else
I commit
...
I commit
You are happy with that.
I am happy too.
We merge PR into mergecores.
We merge mergecores into main.
I open PR multicores 2.
We merge multicores 2 into main.
We can celebrate.


I am really unhappy of how it works now, and you don't seem to have a good time either.

I would be particularly relieved if you accepted this proposal.

bobbens:
> > I'm tempted to try a quadratic function so that it is easier to specify some exact values you want.
> > Sorry for all the noise and back and forth, but balancing is not my strength.
> 
> Balancing is hard.
> 
> I think we do it the wrong way. I am unhappy about this, and there would be so much better use of your time.
> 
> I suggest you do constraint programming. You specified a lots of things above, and I can easily see how to meet these specifications. Then I will propose solutions (concrete parameters), and you will add constraints to the system to exclude the solutions you don't want, or to reinforce behaviors you want. With words, just as you did a few lines before. With a few iterations, we should have something that meets your expectations and that I find satisfying.

Sounds good.
 
> Technically, this would mean I open a PR on naev/mergecores, and you'll provide feedback at each new setting I provide in answer to the previous constraints. I commit You add a constraint I commit You would prefer something else I commit ... I commit You are happy with that. I am happy too. We merge PR into mergecores. We merge mergecores into main. I open PR multicores 2. We merge multicores 2 into main. We can celebrate.
>
> I am really unhappy of how it works now, and you don't seem to have a good time either.
> 
> I would be particularly relieved if you accepted this proposal.

Sounds good! I'll delegate the details and implementation to you then. I'll stick to testing. Thanks for the proposal.

johann1764:
So cool! I feel relieved.

You'll have the first proposition tomorrow morning (your morning).

bobbens:
OK, so I'm ready to merge this into main, the only thing I'm considering (mainly to make it easier for the player to calculate) is to either make `PHYSICS_SPEED_DAMP` be `1/0.2` or `1/0.25` so that it adds 20% or 25% accel to speed instead of the current 22.222222%, reason mainly being that from a player POV 22.22222% is something hard to compute with mental arithmetics. Any opinions?

johann1764:
Both are possible. The impact is limited, only the direct effect of acceleration.
4 would make interceptors slightly less aggressive, so maybe it is better.
5 is ok too.

Let me know if I need to merge this branch into mergecore (or is it the contrary?).

bobbens:
> Both are possible. The impact is limited, only the direct effect of acceleration. 4 would make interceptors slightly less aggressive, so maybe it is better. 5 is ok too.
> 
> Let me know if I need to merge this branch into mergecore (or is it the contrary?).

I can do that.

Set it to 5 and completed the merge.

PR #2801, opened 2025-04-05
console.lua - multiple outfit folding & mass usage display

johann1764:
**New Feature**

## Summary
See title.

Also, a bugfix, as according to what I read here:
https://stackoverflow.com/questions/1252539/most-efficient-way-to-determine-if-a-lua-table-is-empty-contains-no-entries
testing #table~=0 for non-array style table does not work. I might be wrong on that.

Smallest PR ever.

## Testing Done
Yes.

### DIFF
--- a/dat/scripts/dev/console.lua
+++ b/dat/scripts/dev/console.lua
@@ -2,10 +2,20 @@
 
    Development scripts
 
-A sort of useful scripts that will be loaded write away in the console, allowing for quick debugging.
+A sort of useful scripts that will be loaded right away in the console, allowing for quick debugging.
 
 --]]
 local fmt = require "format"
+
+local function _disp(v,mul)
+   if v~=nil then
+      local mul_str=""
+      if mul>1 then
+         mul_str=" (x"..tostring(mul)..")"
+      end
+      print("   "..tostring(v)..mul_str)
+   end
+end
 
 -- luacheck: globals inspect
 function inspect( t )
@@ -19,13 +29,24 @@ function inspect( t )
    if #outfits <= 0 then
       print("No outfits")
    else
+      local prv=nil
+      local acc=0
       print("Outfits:")
       for k,v in ipairs(outfits) do
          if v then
-            print("   "..tostring(v))
+            if v==prv then
+               acc=acc+1
+            else
+               _disp(prv,acc)
+               prv=v
+               acc=1
+            end
          end
       end
+      _disp(prv,acc)
    end
+
+   print("Mass "..tostring(t:mass()).."/"..tostring(t:shipstat('engine_limit')))
 
    -- Print intrinsics if applicable
    local intrinsics = t:outfitsList("intrinsic")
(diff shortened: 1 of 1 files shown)


dat/scripts/dev/console.lua
```diff
@@ -19,17 +29,28 @@ function inspect( t )
...
       end
+      _disp(prv,acc)
    end
 
+   print("Mass "..tostring(t:mass()).."/"..tostring(t:shipstat('engine_limit')))
+
    -- Print intrinsics if applicable
    local intrinsics = t:outfitsList("intrinsic")
-   if #intrinsics > 0 then
+   if next(intrinsics) ~= nil then
```
bobbens, line 53: `pilot.outfitsList` is an ordered list, so `#intrinsics > 0` is enough.
(if it weren't ordered ,you would also have to use `pairs` below instead of `ipairs`)

bobbens (commented):
Thanks! This will only work for consecutive outfits that are the same. If you want to do something like this, I would recommend moving results to a second unordered table. For example:
```lua
local l = {
   "Cat",
   "Dog",
   "Cat",
}
-- Store repetitions in new table
local newl = {}
for k,v in ipairs(l) do
   local val = newl[v] or 0 -- gets how many elements found
   newl[v] = val+1
end
-- Print new results and amount (without duplications)
for k,v in pairs(newl) do
   print( string.format("%s (x%d)", k, v) )
end
```

dat/scripts/dev/console.lua
johann1764: Ok! As I said, I was not certain of this.

johann1764:
> Thanks! This will only work for consecutive outfits that are the same. If you want to do something like this, I would recommend moving results to a second unordered table. For example:
> 
> ```lua
> local l = {
>    "Cat",
>    "Dog",
>    "Cat",
> }
> -- Store repetitions in new table
> local newl = {}
> for k,v in ipairs(l) do
>    local val = newl[v] or 0 -- gets how many elements found
>    newl[v] = val+1
> end
> -- Print new results and amount (without duplications)
> for k,v in pairs(newl) do
>    print( string.format("%s (x%d)", k, v) )
> end
> ```

Sure. I considered that, in most cases, multiple identical outfits would be consecutive, such as system cores, weapons, and so on. While I tested, I did not encounter any counter-example. And this has the advantage of keeping current order intact.

I don't know if we should fold non-successive identical outfits. What do you think ?

johann1764:
I found no button for discarding the second commit.

In this case, adding a commit to restore previous situation is easy, but in general, it might be useful to know if is there a way I can do it. Or maybe only the reviewer can ?

bobbens:
> Sure. I considered that, in most cases, multiple identical outfits would be consecutive, such as system cores, weapons, and so on. While I tested, I did not encounter any counter-example. And this has the advantage of keeping current order intact.
> 
> I don't know if we should fold non-successive identical outfits. What do you think ?

Depends what the objective is. I could see merit in any approach, but you're right that trying to keep the order as is might be most elegant.

> I found no button for discarding the second commit.
> 
> In this case, adding a commit to restore previous situation is easy, but in general, it might be useful to know if is there a way I can do it. Or maybe only the reviewer can ?

Usually you would rebase and remove the commit, then force push to your branch to overwrite it without the previous commit.

bobbens:
I've cherry-picked the first commit. Thanks!

PR #2802, opened 2025-04-06
Multicore hulls

johann1764:
**New Feature**

This PR addresses one of the features described in #2766.

## Summary (User-visible changes)

 - S&K Small/Medium/Large Cargo Hulls keeps same price & cargo inertia & mod but becomes stackable. However, mass and cargo are split (1/2(1/2) and 2/3(1/3) resp.)
 - Red Star Small/Medium/Large Cargo Hulls work the same as their S&K counterparts.
 - S&K Superheavy/Medium-Heavy/Light combat plating become resp. 2x Heavy/Medium/Ultralight combat plating.
 - Unicorp D-72/38/9 plating become resp. 2x D-58/23/2 plating.
 - Nexus Light Stealth Plating becomes 2x Nexus Ultralight Stealth Plating.
 - Nexus Medium Stealth Plating stays as is, and becomes stackable.
 - Nexus Heavy Stealth Plating is new.
 - Patchwork Light/Medium/Heavy Plating stay as is.
 - Nexus Light/Medium/Heavy Stealth Plating now give -10%/-15%/-20% bonus as primary (instead of -15% in all cases so far). Medium and Heavy also give -5% as secondary.

## Explanations about that

There were the difficult choice to decide whether current hulls are of size 2k-1 or 2k.

Concerning Stealth Plating, Medium Stealth was found of size 3 while Light Stealth Plating was found of size 2.

Concerning the cargo hulls, that was not obvious, as some parameters suggested the first one while other the second one. So I kept parameters suggesting 2k-1 as is and extrapolated their secondary effect, while the parameters suggesting 2k were split into primary and secondary.

In particular, as I considered (which one could disagree with) the price suggesting size 2k-1, the price stayed the same and therefore the outfit translate into its single version (therefore no save\_update to do)

## Summary (Other changes)

`generate.py` now uses directly standard cores as references instead of having copied-pasted values all over. As a consequence, this file had its size drastically reduced. Notice however that there are still such values for weapons, as the weapons parts is complicated and the standard weapons could change.

`mvx2xmllua.py` now has a clean, single output using the new inlined lua feature.

Several python scripts have been cleaned up a bit, argparsed, and some common functionality have been transferred to `outfit.py`, which has been improved in addition of that.

`outfits2md.py` now has an option for sorting output by increasing mass.

I will detail anything on request.

## TODO

 - [x] Finish equipopt updates
 - [x] Remove this file PR.md
 - [x] Testing again
 - [x] Grep deprecated outfits
 - [x] Remove the placeholder deprecated outfits
 - [x] Rename S&K Combat & Nexus platings


bobbens:
Thanks for the PR! Some quick notes.

>  * Nexus Heavy Stealth Plating is new.

I've actually thought about this somewhat. I'm not sure Heavy Stealth Plating makes too much sense. Potentially the only ship that could benefit from it is the Pirate Kestrel, but even then you're probably looking at quite large stealth unless you fully invest with all the ridiculous stealth outfits.

I would consider not having Heavy Stealth Plating as a standard outfit, and potentially add something similar as a unique outfit or something a bit more special. That also makes the outfits a bit more interesting as there's not equivalents for all 3 sizes (like Krain engines are only medium/large).

> There were the difficult choice to decide whether current hulls are of size 2k-1 or 2k.
>
> Concerning the cargo hulls, that was not obvious, as some parameters suggested the first one while other the second one. So I kept parameters suggesting 2k-1 as is and extrapolated their secondary effect, while the parameters suggesting 2k were split into primary and secondary.
>
> In particular, as I considered (which one could disagree with) the price suggesting size 2k-1, the price stayed the same and therefore the outfit translate into its single version (therefore no save_update to do)

Yes, they are a bit in-between as they are not generated from equations. I have no issues with changes as long as they are suited to their specific roles and consistent (should be much less problematic than engines).

Looking forward to seeing the final PR.

Side note, I can implement the "improved equipopt support for cores via name" based on this branch if you want, and push it to a new branch so you can merge it here if you want (without merge conflicts).

johann1764:
> Thanks for the PR! Some quick notes.
> 
> > * Nexus Heavy Stealth Plating is new.
> 
> I've actually thought about this somewhat. I'm not sure Heavy Stealth Plating makes too much sense. Potentially the only ship that could benefit from it is the Pirate Kestrel, but even then you're probably looking at quite large stealth unless you fully invest with all the ridiculous stealth outfits.
> 
> I would consider not having Heavy Stealth Plating as a standard outfit, and potentially add something similar as a unique outfit or something a bit more special. That also makes the outfits a bit more interesting as there's not equivalents for all 3 sizes (like Krain engines are only medium/large).

On that, I sort of anticipated a future work on stealth.
I used this opportunity to introduce this hull, as it is easier to remove it later than add it later (grep and remove is easy, while thinking about where to add references is a work I had to do here anyway). Doing this before any discussion is odd but I honestly intended to remove that if we finally decide that we should not do that when we discuss this matter.

I agree that the Kestrel should be the biggest stealthy thing in the universe.
At this point, it does not work so well, as the Yuri's Kiss range bonus is not enough to beat its stealth range, therefore Yuri's Kiss concept is a bit broken. But the idea is good, and in the end of Minerva Campaign, the stealth of this ship is used as an element of scenario. I would like to make this work. In addition, it would be cool if whole Thurion fleets could be stealthy. Could be used in the story.

Besides that, stealth armor is much better than just stealth bonus: since it reduces detected range, it also reduces signature range as a consequence, which means it should theoretically possible to have a Kestrel that the heavier weapons (the 350 CPU weapons) have difficulty to lock on.
> 
> > There were the difficult choice to decide whether current hulls are of size 2k-1 or 2k.
> > Concerning the cargo hulls, that was not obvious, as some parameters suggested the first one while other the second one. So I kept parameters suggesting 2k-1 as is and extrapolated their secondary effect, while the parameters suggesting 2k were split into primary and secondary.
> > In particular, as I considered (which one could disagree with) the price suggesting size 2k-1, the price stayed the same and therefore the outfit translate into its single version (therefore no save_update to do)
> 
> Yes, they are a bit in-between as they are not generated from equations. I have no issues with changes as long as they are suited to their specific roles and consistent (should be much less problematic than engines).

The only thing people might be unhappy with is the small loss of cargo of cargo hulls. But it becomes twice as light, and can now be stacked with some combat armor to become significantly stronger or another cargo hull to become an even better cargo hull (same total cargo + better cargo bonuses). I think that, if they see clearly the tradeoff they make, they will be happy. Now I must find a way to provide clear and understandable information about this.

> Looking forward to seeing the final PR.
> 
> Side note, I can implement the "improved equipopt support for cores via name" based on this branch if you want, and push it to a new branch so you can merge it here if you want (without merge conflicts).

I should finish my equipopt updates before that, and when I have done it, this PR will be ready (as far as I know), so maybe it is simpler to finish this PR first.
But if I finish the equipopt updates and some other issue is raised that prevents the merging the PR, you can do that.

bobbens:
> > I've actually thought about this somewhat. I'm not sure Heavy Stealth Plating makes too much sense. Potentially the only ship that could benefit from it is the Pirate Kestrel, but even then you're probably looking at quite large stealth unless you fully invest with all the ridiculous stealth outfits.
> > I would consider not having Heavy Stealth Plating as a standard outfit, and potentially add something similar as a unique outfit or something a bit more special. That also makes the outfits a bit more interesting as there's not equivalents for all 3 sizes (like Krain engines are only medium/large).
> 
> On that, I sort of anticipated a future work on stealth. I used this opportunity to introduce this hull, as it is easier to remove it later than add it later (grep and remove is easy, while thinking about where to add references is a work I had to do here anyway). Doing this before any discussion is odd but I honestly intended to remove that if we finally decide that we should not do that when we discuss this matter.

OK. I take it it's a bit early for stealth talk then.

> I agree that the Kestrel should be the biggest stealthy thing in the universe. At this point, it does not work so well, as the Yuri's Kiss range bonus is not enough to beat its stealth range, therefore Yuri's Kiss concept is a bit broken. But the idea is good, and in the end of Minerva Campaign, the stealth of this ship is used as an element of scenario. I would like to make this work. In addition, it would be cool if whole Thurion fleets could be stealthy. Could be used in the story.

My Yuri's Kiss has 1,916 km stealth, heavy lasers at 1.750 km +30% range, with -25% discovered rate (if I drop flow stuff, I could get another -50% stealth discovered rate), so it is borderline working. I guess all that is really needed is something else that gives +range (or more stealth). Range is a fairly new ship stat pretty much added with variants, so only a few give it as a bonus.

> Besides that, stealth armor is much better than just stealth bonus: since it reduces detected range, it also reduces signature range as a consequence, which means it should theoretically possible to have a Kestrel that the heavier weapons (the 350 CPU weapons) have difficulty to lock on.

Yes, the above Yuri's Kiss has 7,665 km signature, so it can dodge a lot of heavy weapons.

> The only thing people might be unhappy with is the small loss of cargo of cargo hulls. But it becomes twice as light, and can now be stacked with some combat armor to become significantly stronger or another cargo hull to become an even better cargo hull (same total cargo + better cargo bonuses). I think that, if they see clearly the tradeoff they make, they will be happy. Now I must find a way to provide clear and understandable information about this.

Given the economic model is still quite bad, I don't think it should be too much of a problem.

> > Side note, I can implement the "improved equipopt support for cores via name" based on this branch if you want, and push it to a new branch so you can merge it here if you want (without merge conflicts).
> 
> I should finish my equipopt updates before that, and when I have done it, this PR will be ready (as far as I know), so maybe it is simpler to finish this PR first. But if I finish the equipopt updates and some other issue is raised that prevents the merging the PR, you can do that.

Sure thing. The only real difference there will in API is that the cores will no longer be a ordered table, and will be a dictionary. So you'd have to do something like:
```lua
return {
   hull = "Whatever",
   systems = "Foo",
   systems_secondary = "Bar",
}
```
Stuff not specified would not be modified. Bonus is that it would allow usage of any named slot.

johann1764:
That should be it. Let me know if I forgot something.

Notice I have completed the initial post with both user-visible changes and technical details.

Before starting "multicore - engines", we have a few things to do:
 - New equipopt API
 - The improved saveupdater (I don't remember, but I believe we have left something unfinished)
 - We should decide whether to work with mvx and use xml/lua as current implementation of using mvx or use xml/lua directly, which has the downside that it is a fairly fragile way to represent data.
 - I must improve the engine generator/updater to be able to manage either mvx or xml with inlined lua directly.
 
Independently of it (maybe rather after, in order to avoid having too much at the same time):
 - The stealth thing.
 - Maybe recheck multicore desc. display, I believe some details to improve.

bobbens:
It looks OK, thanks! I've fixed a few things locally (nothing too big).
One thing, it probably would make more sense to name the cores Light / Medium / Heavy than Ultralight / Medium / Heavy. It is a bit less confusing (Combat Plating and Stealth Plating).

johann1764:
Maybe we can reuse the small/medium/large naming scheme. Not great, but it works.

Or something a bit different, like:
Ul/M/H combat plating -> Skirmirsh/Combat/Battle Plating.
But in that case, I have no idea about the Stealth series.

bobbens:
> Maybe we can reuse the small/medium/large naming scheme. Not great, but it works.
> 
> Or something a bit different, like: Ul/M/H combat plating -> Skirmirsh/Combat/Battle Plating. But in that case, I have no idea about the Stealth series.

That's a good point on the name.

For combat plating, I like Skirmish / Combat / Battle Plating.
For stealth, what about something like Shadow / Ghost / Phantasm ?

johann1764:
OK, let's do that!

Let us just add this "translation" in save_updater, and substitute every other occurrence. Did you locally modify one file containing such an occurrence ? If so, the simplest is maybe that you use your diff as a correction suggestion, I accept to apply that, and then I can do the substitutions on top of that.

bobbens:
> OK, let's do that!
> 
> Let us just add this "translation" in save_updater, and substitute every other occurrence. Did you locally modify one file containing such an occurrence ? If so, the simplest is maybe that you use your diff as a correction suggestion, I accept to apply that, and then I can do the substitutions on top of that.

Yes, I pushed to `hulls` branch with some minor fixes. You can use those as a base. Thanks!

johann1764:
Should be OK.

bobbens:
OK, so I fixed a few issues (not pushed yet), and was wondering a bit on the stealth names. Some options:
1. Nexus (Phantasm | Ghost | Shadow) Plating
2. Nexus (Phantasm | Ghost | Shadow) Weave
3. Nexus (Heavy | Medium | Light) Stealth Weave

If we change it from plating to weave, we can use the old system again (Heavy / Medium / Light), but I'm partial to 2. . No real reason to call it Weave other than being able to write a bit more in the description and giving more diversity on hull names (it's all plating or hull now).

johann1764:
Yep, I also thought using "plating" for Nexus hulls was not so great, but lacked a better proposition. Now I have one ! Let do this (prop. 2)

You can update the hull branch with the new commits and your corrections, I will resurrect the substitution script to apply this.

johann1764:
Changing the cores equipopt API is on the TODO list, and maybe you should open an issue about this, then I will open a PR for updating equiptopt/cores.lua.

Concerning the mvx/xml issue: data is present in xml, but if we want to process it (like update mobility parameters), we need to read this using an ad hoc regexp, and have no guarantee these data will be in the same form in the future: this is rather fragile. This is feasible but not done yet and not a robust approach anyway. In addition of that, if one day we choose to have builtin multi-valued xml management, current mvx will just have to be renamed xml, and that will be it. I think we can keep thinking about it until at least the last multicore transformation is complete.

Btw, I have second thoughts on S&K Hulls naming: maybe Skirmish/Battle/War instead of Skirmish/Combat/Battle, let me know your opinion on that.

bobbens:
> Changing the cores equipopt API is on the TODO list, and maybe you should open an issue about this, then I will open a PR for updating equiptopt/cores.lua.

The `equipopt/optimize.lua` changes would be only:

```diff
diff --git a/dat/scripts/equipopt/optimize.lua b/dat/scripts/equipopt/optimize.lua
index 1f85129527..f6e63416a5 100644
--- a/dat/scripts/equipopt/optimize.lua
+++ b/dat/scripts/equipopt/optimize.lua
@@ -286,21 +286,9 @@ function optimize.optimize( p, cores, outfit_list, params )

    -- Handle cores
    if cores and not pt.nocores then
-      -- Don't actually have to remove cores as it should overwrite default
-      -- cores as necessary
-      --p:outfitRm( "cores" )
-      -- Put cores
-      local prv
-      for k,v in ipairs( cores ) do
-         local q
-         -- A bit fragile..
-         if v==prv then
-            q = p:outfitAdd( v, 2, true )
-         else
-            q = p:outfitAdd( v, 1, true )
-            prv=v
-         end
-         if q < 1 then
+      for k,v in pairs( cores ) do
+         p:outfitRmSlot( k )
+         if not p:outfitAddSlot( k, v ) then
             warn(string.format(_("Unable to equip core '%s' on '%s'!"), v, p:name()))
          end
       end
```

However, making sure that all the cores are passed in the form of string -> outfit pairs and adding names to all the ship slots requires a bit of care. I can open an issue on it.

> Concerning the mvx/xml issue: data is present in xml, but if we want to process it (like update mobility parameters), we need to read this using an ad hoc regexp, and have no guarantee these data will be in the same form in the future: this is rather fragile. This is feasible but not done yet and not a robust approach anyway. In addition of that, if one day we choose to have builtin multi-valued xml management, current mvx will just have to be renamed xml, and that will be it. I think we can keep thinking about it until at least the last multicore transformation is complete.

I don't think multi-value xml management C-side is something that makes sense. This is best left to Lua. The main reason I'm asking is where to edit the descriptions. Doing it through mvx then using an update script is a bit more friction than doing it directly through the xml.

I do think it shouldn't be too hard to write something to handle the xml with inline Lua. Instead of a regex-based approach, it should probably be possible to directly load the inline Lua, run it through some modified Lua library to generate easily-parsable data or something like that.

> Btw, I have second thoughts on S&K Hulls naming: maybe Skirmish/Battle/War instead of Skirmish/Combat/Battle, let me know your opinion on that.

Either works. I tried searching a bit and it seems like there's not too much consistency in usage, but (Skirmish | Battle | War) would probably be most sensible.

johann1764:
> > Changing the cores equipopt API is on the TODO list, and maybe you should open an issue about this, then I will open a PR for updating equiptopt/cores.lua.
> 
> EDIT: Opened issue at #2805
> 
> > Concerning the mvx/xml issue: data is present in xml, but if we want to process it (like update mobility parameters), we need to read this using an ad hoc regexp, and have no guarantee these data will be in the same form in the future: this is rather fragile. This is feasible but not done yet and not a robust approach anyway. In addition of that, if one day we choose to have builtin multi-valued xml management, current mvx will just have to be renamed xml, and that will be it. I think we can keep thinking about it until at least the last multicore transformation is complete.
> 
> I don't think multi-value xml management C-side is something that makes sense. This is best left to Lua. The main reason I'm asking is where to edit the descriptions. Doing it through mvx then using an update script is a bit more friction than doing it directly through the xml.

Ok, there are two files, which is more complicated than one.
However, we don't need to change description every day, and updating is straightforward:
```bash
for i in *.mvx;do ${UTILS_PATH}mvx2xmllua.py < $i > ${i%mvx}lua;done
```
> 
> I do think it shouldn't be too hard to write something to handle the xml with inline Lua. Instead of a regex-based approach, it should probably be possible to directly load the inline Lua, run it through some modified Lua library to generate easily-parsable data or something like that.

That would be a more robust approach than regexp, for sure.
However, it only works one-way, and having it work the other way is non-trivial.
I know how I will adapt apply_engines to work with mvx (and, in addition, xml/lua output can be updated at the same time), but I really don't know how to do it with xml/lua.

I think a good compromise would be the following:
 - you provide a script (using lua, if I understand well) for turning xml with inlined lua into multi-valued xml.
 - we keep both mvx and xml(inlined lua) until multicores is finished.
 - from then, we can keep either mvx or xml(lua) as we can always convert one to the other.

> 
> > Btw, I have second thoughts on S&K Hulls naming: maybe Skirmish/Battle/War instead of Skirmish/Combat/Battle, let me know your opinion on that.
> 
> Either works. I tried searching a bit and it seems like there's not too much consistency in usage, but (Skirmish | Battle | War) would probably be most sensible.

OK, let's do that.

johann1764:
OK, done.

This time I didn't forget the tech dir.

bobbens:
Thanks again!

> Ok, there are two files, which is more complicated than one. However, we don't need to change description every day, and updating is straightforward:
> 
> ```shell
> for i in *.mvx;do ${UTILS_PATH}mvx2xmllua.py < $i > ${i%mvx}lua;done
> ```

Yes, but from a maintenance point of view, it's really easy to create bugs with things you do rarely. The `polygonize.py` script for generating collision maps is also an example of that, I really want to move it engine-side when possible.

> > I do think it shouldn't be too hard to write something to handle the xml with inline Lua. Instead of a regex-based approach, it should probably be possible to directly load the inline Lua, run it through some modified Lua library to generate easily-parsable data or something like that.
> 
> That would be a more robust approach than regexp, for sure. However, it only works one-way, and having it work the other way is non-trivial. I know how I will adapt apply_engines to work with mvx (and, in addition, xml/lua output can be updated at the same time), but I really don't know how to do it with xml/lua.
> 
> I think a good compromise would be the following:
>  * you provide a script (using lua, if I understand well) for turning xml with inlined lua into multi-valued xml.
>  * we keep both mvx and xml(inlined lua) until multicores is finished.
>  * from then, we can keep either mvx or xml(lua) as we can always convert one to the other.

Once the merging is done and things calm down, I would probably keep just the xml + inlined Lua. The only difficulty I foresee, is that, since the biocores are not computed from mvx, we would have to make them compute themselves from the xml + inlined Lua instead, which is slightly more difficult, but should be possible.

That way editing and changing things is still as simple as modifying the XML files, with the only exception of the biocores which need to be computed from a script, and potentially the stats from the engines which might also be recomputed every time.

johann1764:
> Once the merging is done and things calm down, I would probably keep just the xml + inlined Lua. The only difficulty I foresee, is that, since the biocores are not computed from mvx, we would have to make them compute themselves from the xml + inlined Lua instead, which is slightly more difficult, but should be possible.
> 
> That way editing and changing things is still as simple as modifying the XML files, with the only exception of the biocores which need to be computed from a script, and potentially the stats from the engines which might also be recomputed every time.

The easy way to solve all that is to have a script to translate xml with embedded lua into mvx.  That would make everything work. The goal would only to have data scripts understand. If we want to, we could later simplify the internal conversions xml(lua) -> mvx -> outfit object into xml(lua) -> outfit object.

The only thing I can't do easily (and can't even do robustly) is reading xml(lua). The rest is either already done or easy to do.

Issue #2799, opened 2025-04-03
[Proposal]: Gauntlet Intrinsics


johann1764:
### Suggestion

As rewards, you can gain two different intrinsics in the Gauntlet:
 - Gauntlet Deluxe +5% damage cannon/turret
 - Gauntlet Supreme +5% damage launcher - double if unguided

Firstly, I think these two should be **mutually exclusive**.
I notice my favorite things in Naev tend to have a marked downside (e.g. Air Freshener)
In particular, exclusive intrinsic give you cool bonuses with the downside of not being able to choose the other one, for example Machiavellian Misi's Intrinsics, which allow different flavors of ships.

Secondly, I would make it even **more Dvaered**, and **amend a bit** values, as weapon damage is a great bonus (compared to weapon fire rate, for ex.)
 - Gauntlet Deluxe
    - +3% damage (turret)
    - +7% damage (cannon)
 - Gauntlet Supreme
    - +3% damage (guided launcher)
    - +7% damage (unguided launcher)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

johann1764:
In order to manage the case where, in a savegame, both intrinsics are on the same ship, one possible way would be to rename these outfits. That is a good opportunity to change that, as their current names are a bit too generic. I propose:
 - Gauntlet Deluxe -> Gauntlet Warrior
 - Gauntlet Supreme -> Gauntlet Ranger

johann1764:
I needed a hook to process the change. Using the version change is fine, too.
If current naming is intentionally opaque, let us keep it.

IMO, if the player chose one of the intrinsics, it means he/she has a clear preference, and in that case we should leave it as is.

The case both were chosen can't be left as is. I think we shouldn't decide instead of the player, as if we make a choice the player did not want, he/she will have to pay its refund to get the other one, thus having paid a total of 2x the intrinsic price for one. That is why I think we should make no assumption about the player preference in that case and refund both to leave the choice to him/her.

johann1764:
P.S. If I understand well, the removal of both intrinsics on a ship will wait until version bump. That is fine, it can wait. For example, it can wait until multi-core - 3 is done.

bobbens:
> I needed a hook to process the change. Using the version change is fine, too. If current naming is intentionally opaque, let us keep it.

What do you mean hook?

[...]

PR #2804, opened 2025-04-07
gauntlet intrinsics

johann1764:
**Balance**

This PR addresses the feature described in #2799.

**Important note**: Before the bugfix, replacing an intrinsic caused a **CRASH**. That is why I left these as two different commits.

## Note
This PR has become messy. I have proposed a rebased version in #2817.

## Summary
- [x] Change values
- [x] The extra desc of Gauntlet Supreme is displayed twice in the dialog.
- [x] Warn user about losing one of them if choosing the other.
- [x] Remove the other when getting Gauntlet Supreme or Gauntlet Deluxe.
- [x] Also display the removed intrinsic properties.
- [x] Manage savegames updating.

## Testing
It works.


johann1764:
1. I don't know how I can do the last item (manage current savegames containing both intrinsics) as I don't see where I can get an entry point.
2. Notice the crash bug (see opening post)

johann1764:
At that point, I can't figure out why compilation keeps failing.

johann1764:
Is the Lua API doc. on `outfitRmIntrinsic` correct ?
It says it returns true iff successful.
My lua code have been seemingly working since I have assumed the opposite, so either my code works by accident, or the documentation is wrong.

PR #2807, opened 2025-04-08
multicore - description

johann1764:
**Bug Fix**

This PR addresses the bug described in #2766 

## TODO
Bugfix : xml/lua redundancy in case pri==sec

### DIFF
--- a/dat/outfits/lib/multicore.lua
+++ b/dat/outfits/lib/multicore.lua
@@ -17,6 +17,10 @@ local function valcol( val, inverted )
 end
 
 local function stattostr( s, val, grey, unit )
+   if val == 0 then
+      return "#n."
+   end
+
    local col
    if grey then
       col = "#n"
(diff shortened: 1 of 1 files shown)


bobbens:
Looks better, thanks!

johann1764:
oups. The xml/lua redundancy bug was not yet fixed (see discussion in #2766)
Easy to fix, but we have to make a decision.

bobbens:
> oups. The xml/lua redundancy bug was not yet fixed (see discussion in #2766) Easy to fix, but we have to make a decision.

Ah sorry, about to go to sleep so I tend to get ahead of myself.

I think that keeping it all Lua-side would be most consistent and easy.

johann1764:
Bad luck, I went the other way as it was even easier.
If you want a quick bugfix before nightly, my branch:
https://github.com/johann1764/naev/xmllua_bugfix

Did I mention that, due to this redundancy, some bonuses apply twice ?

In any case, I will do the other one, but in a few hours.

PR #2808, opened 2025-04-08
xml/lua redundancy bugfix

johann1764:
**Bug Fix**

This PR addresses the bug described in #2766.
The xml/lua redundancy is fixed by removing redundant modifiers from xml.
Notice, however, that `priority` stays in xml. 
(that's a choice: we specifically asked for that in the script, as we did for price.)


bobbens:
Looks good, thanks!

PR #2813, opened 2025-04-09
Machiav. Misi Intrinsics re-balanced

johann1764:
**Balance**

This PR addresses the bug/feature described in #2814.

## Summary
See #2814 or diff, as there is not much.

## Testing Done
Yes.

### DIFF
--- a/dat/outfits/intrinsic/machiavellian_necessity.xml
+++ b/dat/outfits/intrinsic/machiavellian_necessity.xml
@@ -15,5 +15,6 @@
   <fbay_rate>10</fbay_rate>
   <tur_range>5</tur_range>
   <fwd_range>5</fwd_range>
+  <launch_range>5</launch_range>
  </specific>
 </outfit>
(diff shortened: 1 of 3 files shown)


dat/outfits/intrinsic/machiavellian_necessity.xml
```diff
@@ -13,5 +13,8 @@
   <fwd_firerate>5</fwd_firerate>
   <tur_firerate>5</tur_firerate>
   <fbay_rate>10</fbay_rate>
+  <tur_range>5</tur_range>
```
bobbens, line 16: Would make sense to use `weapon_range` instead of the three separate stats.

Similarly, the firerates could be collpased into `weapon_firerate` which was added after this outfit existed.

johann1764: ok, let's do that!

johann1764: Does weapon_firerate include:
 - launch_rate
 - fbay_rate

johann1764: ?

bobbens: > Does weapon_firerate include:
> 
>     * launch_rate
> 
>     * fbay_rate

It should be yes and no.

bobbens:
Looks good, thanks!

PR #2816, opened 2025-04-10
The Bite description and a bug

johann1764:
**Bug Fix**

## Summary

- [x] repair description
- [x] Check whether the lunge has same effect as adrenal glands when constants match.
      -> It matches with a Marauder but not a Reaver. This is caused by a different way of taking Reaver intrinsic accel modifiers into account. Likely, Reaver mod and AB multiply while Reaver mod and lunge add instead. The Wanderer bonus applies the same way to AB and lunge. We have therefore two different problems:
     - **afterburner is not so consistent** with the rest (its stat have a different meaning, which breaks their display, etc) This is far beyond the scope of this PR.
     - lunge is not multiplicative with ship properties (i.e at least **outfit accel_mod is not multiplicative with ship's accel_mod**). This is surprising and in contradiction with Naev new specs, and therefore should be investigated.
- [x] Readjust lunge -> used lev 4 speed ( like adrenal glands ii ) and lev 5 accel (like Hades Torch)
- [x] Move values out of translated strings (script/bioship/skills.lua)
- [x] Improve skill description in the_bite.lua (should give as much details as in skills.lua, but with no references to previous stages)
- [x] Test bite-2 and bite-3
- [x] Fix bite-3 outfit staying in slot even when bite-4 or bite-5 are chosen
   - When I comment shipvar=... out, everything works.. except what relies on the shipvar.
   - When I add shivar=".." to bite3 and bite4, this behaves inconsistently (differently for bite3 and bite4)
- [x] Re-test it all

## Problems I could not solve

1. In skills.lua, I don't understand how shipvar works. It appears to interact weirdly with outfit. As a consequence, I commented out `shipvar="cannibal2"` in the bite-3, which allows the outfits to work correctly. (But breaks the effects of cannibal2 in board.lua)
2. Lunge is not multiplicative with ship properties. It does not have huge consequences, but indicates something is wrong in the way modifiers add up.

Let me know if I should open an issue about these.

## Testing Done
Yes.


dat/scripts/bioship/skills.lua
```diff
@@ -24,7 +25,9 @@ skills.set.bite = {
       requires = { "bite1" },
       desc = function( p )
          local dmg = 10*math.sqrt(p:mass())
-         return fmt.f(_("The ship will lunge at the target enemy and take a huge bite out of it. +600% accel and +30% absorb for 3 seconds or until target ship is bitten. This ship will do {dmg:.0f} damage with its current mass. Has a 15 second cooldown period."),{dmg=dmg})
+         return fmt.f(_("The ship will lunge at the target enemy and take a huge bite out of it. +{accel_mod}% accel, +{speed_mod}% speed and +30% absorb for 3 seconds or until target ship is bitten. This ship will do {dmg:.0f} damage with its current mass. Has a 15 second cooldown period."),{dmg=dmg,
```
bobbens, line 28: absorb should probably also be passed as a field so we can modify it without changing the translation string.

bobbens (changes requested):

dat/scripts/bioship/skills.lua
johann1764: True

bobbens:
Looks good, thank! I was able to fix the skill issue, which was related to a recent change I did to reset the outfits as they got lost with the split cores move in some cases. I added code to mitigate it, but some saves are probably affected. Instead of having to reset skills manually, this should fix them for the player.

I gave a quick look at the multiplicative thing, and it looks fine at a glance, probably needs more in-depth debugging.

johann1764:
> Looks good, thank! I was able to fix the skill issue, which was related to a recent change I did to reset the outfits as they got lost with the split cores move in some cases. I added code to mitigate it, but some saves are probably affected. Instead of having to reset skills manually, this should fix them for the player.

Great ! It was really impossible to understand for me, glad you did it.
The effects on saves on concern the savegames that were updated by very recent nighties, so probably very few people are concerned. And resetting skills is the more natural thing to do.

The multiplicative thing does not have huge effects (at least on the case where I spotted the problem), so there is no emergency in fixing it.

Thanks!

Issue #2819, opened 2025-04-12
[Proposal]: On stealth terminology and a problematic example


johann1764:
### Suggestion

Stealth means two different things in Naev.
 - There is stealth like in stealth plating, stealth coating, etc. that refers to the general stealth of a ship, and uses ew_hide as the underlying variable.
 - There is also stealth like in stealth mode, related to the action to become invisible to other ships, at the cost of a reduced mobility, and the inability to take actions. This uses the stealth_range as the underlying variable.

First, I suggest to clarify this difference as follow:
 - Let us call stealth the general notion on which detected range/signature/hiding range are built upon.
 - Let us call cloaking the specific ability to hide.
 - Let us distinguish clearly the detected range radius and the general stealth ability. This would solve the case presented below.

The proposition would therefore be the following:
 - There is a **general stealth radius** value.
 - There is a **detected range** value, by default equal to general stealth value, but that can be affected by detected range modifiers.
 - There is a **signature range** value, based on stealth value (1/2 iirc), that can be modified by a specific modifier (as in current situation)
 - There is a **cloaking range** value, base on stealth value (1/4 iirc), that can be modified by specific modifiers (what currently corresponds to stealth modifiers)

The specific case: RS cargo Hulls.
They currently have +25% detected range and -25% signature, which is not so clear.
Let us assume that the intention was to increase detected range but leave signature intact. In that case, it should rather have -20% signature (as 0.8*1.25=1). In that case, the detected range and the cloaking range increase by 25% and signature range stays as is. With the new terminology:
 - +25% detected range
 - +25% cloaking range

If the intention was to actually increase detected range by 25%, decrease signature by 25% and leave cloaking range as is:
 - +25% detected range
 - -25% signature

Clearly, mixing general stealth modifier with specific modifiers make things hard to understand, hence the need for the three specific names. With the three specific names, the description becomes clear.

With the new terminology, stealth coating becomes:
 - -15% general stealth range

The point is not on the naming, just on that:
 - We distinguish 3 specific cases and 1 general case.
 - We do not use one name to talk about another thing.
 - We do not mix the general case with specific cases in the same outfit.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
> First, I suggest to clarify this difference as follow:
>  * Let us call stealth the general notion on which detected range/signature/hiding range are built upon.
>  * Let us distinguish clearly the detected range radius and the general stealth ability. This would solve the case presented below.

[...]

PR #2820, opened 2025-04-13
Stealth naming

johann1764:
**Naming**

This PR addresses the feature described in #2819.

## Summary
- [x] Nexus Stealth Coating -> Nexus Concealment Coating
- [x] Camouflage Burster -> Stealth Burster
- [x] Adaptive Stealth Plating -> Adaptive Camouflage Plating
- [x] Red Star Small/Medium/Large Cargo Hull:
   - +25 % detected range
   - -10 % signature (currently 1.25*0.75=0.9375 -> -6.25 %.. not so great)
   - jamming ability left as is (42% I think). 
- I will maybe try the `detected, signature, and stealth ranges`, and keep it in a separate commit. -> bad idea, would make desc. unreadable. See suggestions in section below.

## Bugfixes on the way
- [x] Fixed event/updater mentioning ships twice
- [x] Fixed dart-360 engine mass limit 300->360.

## Suggestions on visibility
As an easy improvement of current situation, I would suggest "visibility" -> "detectability".

Maybe the real solution would be to make:
```xml
<ew_hide>n</ew_hide>
```
be only a shorthand for:
```xml
<ew_detected>n</ew_detected>
<ew_signature>n</ew_signature>
<ew_stealth>n</ew_stealth>
```
Notice the thumbs up emoji was here **before** I wrote that proposition.

## Testing Done
Yep.


bobbens:
Looks good, thanks!

> As an easy improvement of current situation, I would suggest "visibility" -> "detectability".

I put visibility so it doesn't get confused with detection range, but we can go with detectability if it's not confusing.

> Maybe the real solution would be to make:
> 
> ```
> <ew_hide>n</ew_hide>
> ```
> 
> be only a shorthand for:
> 
> ```
> <ew_detected>n</ew_detected>
> <ew_signature>n</ew_signature>
> <ew_stealth>n</ew_stealth>
> ```

I've thought about this (for the weapon stats too). The only advantage is it changes how the internals work. Now ew_hide gets multiplied with the others, if it is a shorthand, it would get added (same with weapons), which is probably what we want. Downside is that it gets 3 lines to explain it and needs a lot of work to support this elegantly (which I would want to wait at least to have it in rust before that, which will probably take a while).

johann1764:
Ok! It is currently good enough, a fortiori if it is temporary (even for 6 month or +).
Thx!

Issue #2829, opened 2025-04-28
[Bug]:  In documentation: confusion between p:shipMemory() and p:memory()


johann1764:
### What happened?

See
https://naev.org/api/modules/pilot.html#memory

Suggested bugfix:
```diff
index b2da32020e..ecb754772c 100644
--- a/src/nlua_pilot.c
+++ b/src/nlua_pilot.c
@@ -5560,7 +5560,7 @@ static int pilotL_memory( lua_State *L )
  * The resulting table is indexable and mutable.
  *
  *    @luatparam Pilot p Pilot to get ship memory of.
- * @luafunc memory
+ * @luafunc shipMemory
  */
 static int pilotL_shipmemory( lua_State *L )
 {
```

### Version

none

### What operating system are you seeing the problem on?

None

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
Oops, fixed, thanks!

johann1764:
Thanks!

While at it:

https://naev.org/api/modules/pilot.html#spaceworthy

A string is not actually returned, it is messaged. (not a return value, but a side-effect)
But maybe this cannot be expressed with luadoc.

Also missing a colon after first sentence of description.

bobbens:
> Thanks!
> 
> While at it:
> 
> https://naev.org/api/modules/pilot.html#spaceworthy
> 
> A string is not actually returned, it is messaged. (not a return value, but a side-effect) But maybe this cannot be expressed with luadoc.
> 
> Also missing a colon after first sentence of description.

What do you mean? It seems to be working as expected:

```lua
local b, msg = player.pilot():spaceworthy()
print( b )
-- false
print( msg )
-- !! Not All Core Slots are equipped
```

johann1764:
My wrong, I had not noticed there were several calls to it in my version of console.lua, so I took the output of the second call for a side effect of the first one. That makes more sense.

bobbens:
There's been a lot of duplicate @luafunc from copy and pasting. Should probably write a script to check for errors and hook it up to pre-commit.

johann1764:
Maybe a bit overkill to hook it up to pre-commit.
Once the mistakes will be fixed, they have no reason to reappear.

I can do a script for that. If this is just about duplicates of @luafunc, a simple bash script should be able to do that.

bobbens:
[...]

PR #2830, opened 2025-04-28
bugfix - duplicate luadoc entries

johann1764:
**Bug Fix** / **New Feature**

This PR addresses the bug/feature described in #2829.

Baware that sorting takes time, even sorting the list of `@luafunc` occurences. Might take up to a second on your machine, or maybe even more.

## Summary
 - [x] luafunc_dup.sh
 - [x] nlua_pilot.c
 - [x] other nlua_*.c files
 - [x] add it to precommit
 - [x] handle multiple files
 - [x] simplify and improve the script with uniq

```shell
src/nlua_asteroid.c:
      2  * @luafunc setTimer
      2  * @luafunc state
src/nlua_faction.c:
      2  * @luafunc hit
src/nlua_jump.c:
      2  * @luafunc dest
src/nlua_naev.c:
      2  * @luafunc pause
src/nlua_outfit.c:
      2  * @luafunc get
      2  * @luafunc slot
src/nlua_pilot.c:
      2  * @luafunc colour
      2  * @luafunc setHealth
      2  * @luafunc setHilight
      2  * @luafunc setInvisible
      2  * @luafunc weapsetAdd
src/nlua_player.c:
      3  * @luafunc inventoryAdd
      2  * @luafunc tryBoard
src/nlua_shader.c:
      2  * @luafunc send
src/nlua_spfx.c:
      2  * @luafunc vel
src/nlua_spob.c:
      2  * @luafunc services
src/nlua_system.c:
      2  * @luafunc hidden
src/nlua_transform.c:
      2  * @luafunc translate
```

### DIFF
--- a/utils/luafunc_dup.sh
+++ b/utils/luafunc_dup.sh
@@ -5,7 +5,7 @@ res=0
 trap 'rm -f $LST ; exit $res' EXIT
 
 for arg in $@ ; do
-   grep '^ \* @luafunc' $arg | sort | uniq -c | grep -v "^ *1" > $LST
+   grep '^ \* @luafunc' $arg | sort | uniq -c | grep -v "^ *1 " > $LST
 
    if [ -s "$LST" ] ; then
       echo "$arg:"
--- a/utils/luafunc_dup.sh
+++ b/utils/luafunc_dup.sh
@@ -1,15 +1,15 @@
 #!/usr/bin/bash
 
-LST=`tempfile`
+LST=$(mktemp)
 res=0
 trap 'rm -f $LST ; exit $res' EXIT
 
-for arg in $@ ; do
-   grep '^ \* @luafunc' $arg | sort | uniq -c | grep -v "^ *1 " > $LST
+for arg in "$@" ; do
+   grep '^ \* @luafunc' "$arg" | sort | uniq -c | grep -v "^ *1 " > "$LST"
 
    if [ -s "$LST" ] ; then
       echo "$arg:"
-      cat $LST
+      cat "$LST"
       res=1
    fi
 done
--- a/utils/luafunc_dup.sh
+++ b/utils/luafunc_dup.sh
@@ -2,13 +2,17 @@
 
 TMP=`tempfile`
 LST=`tempfile`
-trap 'rm -f $TMP $LST' EXIT
+res=0
+trap 'rm -f $TMP $LST ; exit $res' EXIT
 
-grep '^ \* @luafunc' $1 | sort > $TMP
-sort -u $TMP | diff $TMP - | sort -u - | grep '<' | cut '-d<' -f2- > $LST
+for arg in $@ ; do
+   grep '^ \* @luafunc' $arg | sort > $TMP
+   sort -u $TMP | diff $TMP - | sort -u - | grep '<' | cut '-d<' -f2- > $LST
 
-if [ -s "$LST" ] ; then
-   echo "$1:"
-   cat $LST
-   exit 1
-fi
+   if [ -s "$LST" ] ; then
+      echo "$arg:"
+      cat $LST
+      res=1
+   fi
+done
+
--- a/utils/luafunc_dup.sh
+++ b/utils/luafunc_dup.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/bash
+
+TMP=`tempfile`
+LST=`tempfile`
+trap 'rm -f $TMP $LST' EXIT
+
+grep '^ \* @luafunc' $1 | sort > $TMP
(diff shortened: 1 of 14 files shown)


bobbens:
> **Bug Fix** / **New Feature**
> 
> This PR addresses the bug/feature described in #2829.

Thanks!

> Baware that sorting takes time, even sorting the list of `@luafunc` occurences. Might take up to a second on your machine, or maybe even more.

Shouldn't be a problem. We can make it only run on `nlua_` files.

> ## Summary
> 
>     * [x]  luafunc_dup.sh
> 
>     * [x]  nlua_pilot.c
> 
>     * [x]  others
> 
>     * [x]  add it to precommit
> 
>     * [ ]  check this does need to handle multiple files (would be easy to fix)

Yes, I think it can be batched, so it should support something like `./script FILE1 FILE2 FILE3`, although they should be handled independently (no concatenation) as different files may have same named functions.

johann1764:
> > Baware that sorting takes time, even sorting the list of `@luafunc` occurences. Might take up to a second on your machine, or maybe even more.
> 
> Shouldn't be a problem. We can make it only run on `nlua_` files.

Bad news: already took that into account. You should probably try to see whether this is too costly or not.

> Yes, I think it can be batched, so it should support something like `./script FILE1 FILE2 FILE3`, although they should be handled independently (no concatenation) as different files may have same named functions.

Now it can.

Even, it is better to do it all at the same time, because the script waits until the end before exiting with the code, meaning you will have all files scanned with their redundancies before the test fails, which is nicer to have it fail after one failed file, and having to repeat until no new file fails. Not sure I am clear, here.

johann1764:
If we really need the test to be faster, we can only check successive duplicates, which covers most (maybe all?) cases I found. No sorting, would be lightning fast.

utils/luafunc_dup.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash
+
+LST=`tempfile`
```
bobbens, line 3:
```suggestion
LST=$(mktemp)
```
Apparently `tempfile` is deprecated and backtick notation is considered legacy.

utils/luafunc_dup.sh
```diff
@@ -0,0 +1,16 @@
...
+res=0
+trap 'rm -f $LST ; exit $res' EXIT
+
+for arg in $@ ; do
```
bobbens, line 7:
```suggestion
for arg in "$@ "; do
```
Avoid re-splitting elements.

utils/luafunc_dup.sh
```diff
@@ -0,0 +1,16 @@
...
+   grep '^ \* @luafunc' $arg | sort | uniq -c | grep -v "^ *1 " > $LST
+
+   if [ -s "$LST" ] ; then
+      echo "$arg:"
+      cat $LST
```
bobbens, line 12:
```suggestion
      cat "$LST"
```
More potential resplitting.

utils/luafunc_dup.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash
+
+LST=`tempfile`
+res=0
+trap 'rm -f $LST ; exit $res' EXIT
+
+for arg in $@ ; do
+   grep '^ \* @luafunc' $arg | sort | uniq -c | grep -v "^ *1 " > $LST
```
bobbens, line 8:
```suggestion
   grep '^ \* @luafunc' "$arg" | sort | uniq -c | grep -v "^ *1 " > "$LST"
```

bobbens (changes requested):
Some changes point out by pre-commit shellcheck.

utils/luafunc_dup.sh
johann1764: Absolutely !
That said, I think your `'"'` and `' '` were swapped.

utils/luafunc_dup.sh
johann1764: LST is a tempfile name, so shouldn't contain spaces.
But better safe than sorry, let us do this.

utils/luafunc_dup.sh
johann1764: True.

utils/luafunc_dup.sh
bobbens: Thanks for the notice, I dislike the github editor.

utils/luafunc_dup.sh
johann1764: Damnit, I am feeling old, now.
But true, someone already told me that backtick notation was not used anymore, a long time ago.

bobbens:
Looks good, thanks! Seems to have caught a lot of them already.

johann1764:
Should be all of them. (I only left the one you already corrected in main)

PR #2831, opened 2025-04-30
sort c2luadoc output

johann1764:
**New Feature**

## Why

The lua API doc is somewhat hard to navigate, as functions appear in no particular order.
Each time I look for a function, I have to scan the whole list. Sometimes, subsequences are sorted, giving the false impression that the function I am looking for is not in the list.

The alphabetical order is not perfect, but at least will regroup (or keep them grouped) related functions as they tend to share common prefixes.

A possible improvement of that would maybe be to first have `name()` funcs then `nameVerb()` funcs, as `name()` are most often read-accessors and would maybe be better presented first.

## Initial Summary (outdated)
<details><summary>details</summary>
Lexicographical sort of `c2luadoc.sh` output.

This output is consisting of blocks defined by their ending `@keyword`.

The blocks with `@function` are sorted by dictionary order (`sort -d`) of the function name and put after any other block. The other blocks keep their position (`sort --stable`).

Currently, as far as I could see, there is only one other block type that is module, and it keeps its leading position.
</details>

## True Summary
See discussions below.

## Testing Done (Updated!)
Tested on `*.c`.

Could not check whether adding `sort = true` to `docs/lua/config.ld` was the correct way to have luadoc sort it. You will have to check that !

## TODO
- [x] Make it work
- [x] Wait to see if optimization patch works.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,12 @@ repos:
       types: [png]
 - repo: local
   hooks:
+    - id: check_luadoc_func
+      name: check @lua.. not used outside @luafunc
+      entry: utils/check_luadoc_func.sh
+      files: '^src/nlua_*.c'
+      types: [c]
+      language: script
     - id: luafunc_dup
       name: duplicate @luafunc entry detector
       entry: utils/luafunc_dup.sh
--- a/docs/lua/c2luadoc.sh
+++ b/docs/lua/c2luadoc.sh
@@ -61,4 +61,9 @@ sed -n                                                                         \
    }
    /@/ {
       print $0, "\n"
-   }' > "$2"
+   }' |                                                                        \
+# Sort it
+sed ':a;N;$!ba;s/\n--/\\n--/g' | grep -v "^$" |                                \
+sed 's/^\(.*\)\(@function .*\)$/z\2\1\2/' | sort --stable -d |                 \
+sed 's/\\n--/\n--/g' | sed 's/^z@function .*$//'                               \
+> "$2"
--- a/docs/lua/c2luadoc.sh
+++ b/docs/lua/c2luadoc.sh
@@ -61,10 +61,4 @@ sed -n                                                                        \
    }
    /@/ {
       print $0, "\n"
-   }' |                                                                       \
-# Sort it
-sed ':a;N;$!ba;s/\n--/\\n--/g' | grep -v "^$" |                               \
-sed 's/^\(.*\)\(@function .*\)$/z\2\1\2/' |                                   \
-sed '/\(@module\)\|\(@function\)/! s/^\(.*\)$/zzz\1/' | sort --stable -d |    \
-sed 's/\\n--/\n--/g' | sed 's/^z@function .*$//' | sed 's/^zzz$//g'           \
-> "$2"
+   }' > "$2"
--- a/docs/lua/c2luadoc.sh
+++ b/docs/lua/c2luadoc.sh
@@ -3,57 +3,57 @@
 # Convert Doxygen comments to Luadoc comments
 # CAUTION: We need to support BSD/macOS sed in addition to GNU sed.
 #          If you aren't suffering, you've probably introduced a bug. :P
-sed -n                                                                         \
+sed -n                                                                        \
  -e '1i\
--- This file was generated automatically from C sources to feed LDoc.'         \
-`# Convert Doxygen /** to Luadoc ---`                                          \
- -e 's|^ */\*\* *$|---|p'                                                      \
-`# Convert special tags to Lua expressions.`                                   \
-`# Lines after @luafunc & @luamod will be ignored by Luadoc`                   \
-`# Doxygen comments that do not contain any of these tags have no impact on`   \
-`# the Luadoc output.`                                                         \
-`# Rename some tags:`                                                          \
- -e 's|^ *\* *@luafunc|-- @function|p'                                         \
- -e 's|^ *\* *@brief|--|p'                                                     \
- -e 's|^ *\* *@luasee|-- @see|p'                                               \
- -e 's|^ *\* *@luaparam|-- @param|p'                                           \
- -e 's|^ *\* *@luatparam|-- @tparam|p'                                         \
- -e 's|^ *\* *@luareturn|-- @return|p' `#we accept both @luareturn & @return`  \
- -e 's|^ *\* *@luatreturn|-- @treturn|p'                                       \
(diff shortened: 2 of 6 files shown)


bobbens:
Seems to be running into an issue with: `/__w/naev/naev/build/docs/lua.p/nlua_colour.luadoc:31: Module already declared!` on CI.

Maybe out of scope of this PR, but given the complexity and sed magic, it might make sense to try to rewrite this to be a bit cleaner and simpler in python. Side note, as we move to Rust we will also need a new way to do documentation (or at least probably slightly different parser), so moving to python might be an easy way to handle that too.

johann1764:
> Seems to be running into an issue with: `/__w/naev/naev/build/docs/lua.p/nlua_colour.luadoc:31: Module already declared!` on CI.

I am on it. Good news, I discovered new broken luadoc.
I will also make it a little more robust.
The goal is to make a quick improvement to the lua API doc, it is not long term.

johann1764:
> Maybe out of scope of this PR, but given the complexity and sed magic, it might make sense to try to rewrite this to be a bit cleaner and simpler in python. Side note, as we move to Rust we will also need a new way to do documentation (or at least probably slightly different parser), so moving to python might be an easy way to handle that too.

Not sure. This is already not too fast, and since we can handle it that way, probably better as is. Let us switch to Python the day we need something significantly harder to express with just a shell script.

As the expected form of comment is extremely rigid, we could even do it in C.
 
> EDIT: See https://github.com/naev/naev/blob/rust/src/vec2.rs for an example of a Lua module written in rust. It uses `///` comments but is still mainly copy and paste (note it isn't used yet to generate documentation), but it's pretty unrust-like and there might be a better way to do it.

Ok, the same with '///' comments and `@luamod` instead of `@module`. Not much to adapt.

johann1764:
In `nlua_color.c`:

```c
-- <pre>
-- col1 = colour.new( "Red" ) -- Get by name
-- col2 = colour.new( 0.5, 0.5, 0.5, 0.3 ) -- Create with RGB values
-- col3 = colour.new() -- White by default
-- col2:setHSV( col1:hsv() ) -- Set colour 2 with colour 1's HSV values
-- </pre>
```

In `vec2.rs`:
```rust
/// @code
/// my_vec = vec2.new( 3, 2 ) -- my_vec is now (3,2)
/// my_vec:add( 5, 3 ) -- my_vec is now (8,5)
/// my_vec = my_vec * 3 -- my_vec is now (24,15)
/// your_vec = vec2.new( 5, 2 ) -- your_vec is now (5,2)
/// my_vec = my_vec - your_vec -- my_vec is now (19,13)
/// @endcode
```

Let me push the color fix and we are good.

johann1764:
I am all wrong. We are good!

> In `nlua_color.c`:
> 
> ```c
> -- <pre>
> -- col1 = colour.new( "Red" ) -- Get by name
> -- col2 = colour.new( 0.5, 0.5, 0.5, 0.3 ) -- Create with RGB values
> -- col3 = colour.new() -- White by default
> -- col2:setHSV( col1:hsv() ) -- Set colour 2 with colour 1's HSV values
> -- </pre>
> ```
> 
> In `vec2.rs`:
> 
> ```rust
> /// @code
> /// my_vec = vec2.new( 3, 2 ) -- my_vec is now (3,2)
> /// my_vec:add( 5, 3 ) -- my_vec is now (8,5)
> /// my_vec = my_vec * 3 -- my_vec is now (24,15)
> /// your_vec = vec2.new( 5, 2 ) -- your_vec is now (5,2)
> /// my_vec = my_vec - your_vec -- my_vec is now (19,13)
> /// @endcode
> ```
> 
> Couldn't we do the same ? -> If yes, I will also see if I can replace occurrences.
> 
> Let me push the color fix and we are good.

johann1764:
Actually, I found more missing `@luafunc`.

I will make a script to catch them all.

bobbens:
Looking at the documentation https://lunarmodules.github.io/ldoc/manual/manual.md.html#Extension_modules_written_in_C to try to find supported stuff I found the following setting:
```
sort set if you want all items in alphabetical order
```

Maybe that setting would solve this without modifying the script?

johann1764:
I found that and I don't know whether it is normal.

nlua_loadCol  does not appear to be declared, but uses luadoc in its comment.
Is it just a regular comment that mimics luadoc comments ?

```c
static const luaL_Reg colL_methods[] = {
   { "__eq", colL_eq },
   { "__tostring", colL_tostring },
   { "new", colL_new },
   { "newHSV", colL_newHSV },
   { "alpha", colL_alpha },
   { "rgb", colL_rgb },
   { "rgba", colL_rgba },
   { "hsv", colL_hsv },
   { "setRGB", colL_setrgb },
   { "setHSV", colL_sethsv },
   { "setAlpha", colL_setalpha },
   { "linearToGamma", colL_linearToGamma },
   { "gammaToLinear", colL_gammaToLinear },
   { 0, 0 } }; /**< Colour metatable methods. */

/**
 * @brief Loads the colour library.
 *
 *    @param env Environment to load colour library into.
 *    @return 0 on success.
 */
int nlua_loadCol( nlua_env env )
{
   nlua_register( env, COL_METATABLE, colL_methods, 1 );
   return 0;
}
```

johann1764:
> Looking at the documentation https://lunarmodules.github.io/ldoc/manual/manual.md.html#Extension_modules_written_in_C to try to find supported stuff I found the following setting:
> 
> ```
> sort set if you want all items in alphabetical order
> ```
> 
> Maybe that setting would solve this without modifying the script?

That looks like the easiest way to do it.
And no maintenance at all for us.

I will look into it.

bobbens:
> nlua_loadCol does not appear to be declared, but uses luadoc in its comment. Is it just a regular comment that mimics luadoc comments ?

It's a bit confusing, but it uses doxygen for documenting C, so it's _not_ meant for luadoc. The `nlua_load*` functions are for loading the Lua functionality into the state.

> Actually, I found more missing `@luafunc`.
> 
> I will make a script to catch them all.

Adding it to the pre-commit script would be great, but I'm not sure we can detect this easily, as there may be a ton of false positives.

johann1764:
There are 3 false positive in total, but 0 in nlua_*.c (which is what we want to check)

However, takes twice the time used by luafunc_dup. I will try something to improve that.

johann1764:
OK! We're good.

johann1764:
Thanks !

I hope you noticed :
> Could not check whether adding sort = true to docs/lua/config.ld was the correct way to have luadoc sort it. You will have to check that !

We will see if it works when the API doc is updated on the site.

PR #2821, opened 2025-04-14
Multicore Chapter 3

johann1764:
**New Feature(s)**

(With little bits of _Balance_ and _Bug Fix_ inside)

This PR addresses part of the features described in #2766.
In particular, it introduces the last core splitting: the engines.

## Summary

- [x] Manage Melendez/Tricon/Nexus/Unicorp/Beat Up lines.
- Manage Krain
   - [x] Remige should be a alone-only engine
   - [x] Patagium Twin Engine should be a twins-only engine
   - [x] "Patagium Engine" -> "Patagium Twin Engine" x2
   - [x] Force Patagium Twins in equipopt
- Improve the engine tools to be able to manage multi-engines
   - outfits2md (already the case)
   - [x] engines\_plot
   - [x] engine\_stats
   - [x] apply\_engine (update mode)
   - [x] apply\_engine (generate mode)
   - [x] `apply_engine.py` -> `update_engine.py` and improvements.
- [x] Remove placeholders

We now have 3 special engines that can be used as examples :
 - Remige refuses any other engine installed. (uses alone.lua)
 - Patagium Twin shuts down when he is not installed with another twin. (uses twins.lua)
 - Za'lek test engine can only be installed in the primary slot. The exact same logic can be applied to make engines that go only into secondary slot.

## Bonuses
- [x] Give a little more speed to slowest torpedoes
- [x] Improve outfits2md
- [x] Since `generate.py` now uses explicitly some outfits, `meson.build` should know about these dependencies.
- Reintroduce Aegis/Prometheus
   - [x] found in 0.8.2.
   - [x] made them mvx.
   - [x] rebalanced them.
   - [x] sell them in Zied (EDIT: only Aegis is sold) & declared them in POTFILE.in, etc.
   - [x] updated save_updater.
- [x] Got rid of process_matchings.py, deprecate_outfit.py and addslot.py.
- [x] Repaired some inconsistencies in core systems (mainly prev. gen. systems).
- [x] update_xml_from_mvx.sh
- [x] The new `shipmemfs` module (API + console commands).

## TODO
- [x] Wait for 'outfitHasSlot' bugfix.
- Some Krain improvements relying on `outfitHasSlot`
   - [x] Remige should not try to empty secondary slot if there is none.
   - [x] Remige description does not need to mention that this engine only works alone when it is equipped on a ship with no secondary slot. (currently subject to the descextra bug, with no serious consequence.)
   - [x] Patagium Twin should refuse a ship with no secondary.
[...]

Issue #2828, opened 2025-04-27
[Proposal]: Ditch Lua environments in favour of separate lua_State.


bobbens:
### Suggestion

Naev currently uses a single `lua_State` to hold all the Lua stuff. This allows us to easily transfer memory between scripts (pilot.shipMemory(), naev.cache(), pilot.memory(), etc.), however, it has the downside that everything has to be sequential: we can no longer thread Lua stuff. The two main performance bottlenecks are the GUI (relatively easy to fix with a rewrite), and ship AI (not possible sequentially). One of the things I'm considering in the Rust version (because it will ensure safety, see #2617 ), is to use separate `lua_State`s for each environment, and do things like threaded updates to the ships and game state. This will add significant overhead (20-30%), but will allow threading things, so if you have an 8 core cpu, the game could probably run between 4 to 5 times faster. Of course, for single core systems, it would end up being significantly slower.

The biggest consequence of having multiple `lua_State`s is that "sharing" data goes from something basically "free", to something fairly expensive. As the data will have to be pulled out of one `lua_State`, and serialized into another. It shouldn't be "too bad", but it does mean you want to limit the serialization as much as possible. What this would probably mean, is that API like `naev.cache` or `pilot.memory` will stop returning tables in general, and instead will return fields from a table. So something like:
```lua
let my_var = p:memory().my_var
```
would have to be rewritten as
```lua
let my_var = p:memory("my_var")
```
Furthermore, things like doing look-ups or iterating over those tables would have to be avoided when possible.

This can probably be in parallel by:
1. Trying to switch all the `p:memory().foo`-like API to use `p:memory("foo")`, this can be done on Main and shouldn't really be much of an issue. The only functions affected by this directly are:
   * [ ] pilot.memory
   * [ ] pilot.shipMemory
   * [ ] naev.cache
2. Work on the Rust stuff (#2617 and `rust` branch). When fully implemented, everything should be thread safe and friendly. Until all the Lua API is moved to rust and thread safe (and the branch is merged), we won't be able to run the parallel updates though. So it may be a while... mlua does support being `Send`, so it shouldn't be much of a hurdle Rust-side.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
[...]

PR #2832, opened 2025-05-01
Multicore - Polishing

johann1764:
Little bits of:
 - **Balance**
 - **Bug Fix**(es)
 - **New Feature**(s)

This PR addresses some features described in #2766.
After #2821, all cores are now split.

This is the technical part of the multicore epilog.

## Summary
 - [x] Complete `outfits2md.py` stacking option
 - [x] Update luadoc for `pilot.outfitMessageSlot`
 - [x] Fix the mono-engine Empire Pacifier bug.
 - [x] Switch from `pilot.outfitInitSlot()` to `pilot.outfitMessageSlot()`. (see note)
 - [x] Use the `pilot.outfitMessageSlot()` to transmit data, not `p:shipMemory()` (see note)
    - [x] Re-declare all secondaries when primary is added/inited
    - [x] Have pri engine save all data
    - [x] manage backward info (server->client)
    - [x] Check the calls, should be less redundant -> still the add/init overlap, else it's ok.
 - [x] First rebalance (see note below)
 - [x] have engine_stats allow to also compute engines compositions
 - [x] have engine_stats and outfits2md detect non-stackable outfits
 - [x] manage the mvx/xml redundancy (see note)
    - [x] Done a xmllua2mvx
    - [x] Make sure all conversions are ok.
    - [x] Make sure all reverse-conversions are ok.
    - [x] parsearg-ify + python3-ify converters
    - [x] make sure update_engine also updates xmllua.
    - [ ] add mvx howto in "makefile" & remove mvx
 - [ ] Speed mod should affect all speeds
 - [ ] understand and manage engine cores inconsistencies
 - [ ] Some *neutral* adjustments ( +10t eml for small engines sec, -10t core sys sec for ex). (see note)
 - Bug: A ship with a Patagium Twin in pri slot is considered space-worthy -> surprinsingly, it is not too heavy despite having mass but no engine limit, while a too weak engine would be "Too heavy". Solution: `pilot_massFactor` should return 0.0 when engine_limit is 0. However, I don't know whether some non-ships things have to be considered spaceworthy with 0 engine_limit.

### First re-balance

Well, re-balancing was not the point of this PR, but this will give better ground for future re-balancings. I did:
 - speed_rank_offsets :
    - Nexus: +0.11 -> +0.10
    - Melendez: +0.7 -> +0.8
    - Unicorp: +0.4 -> +0.44
 - Melendez ratio : 0.65 -> 0.5 (which is huge)
 - Melendez engine limit: made it at least Unicorp +20 %
   - Buffalo: 800 -> 850
   - Ox: 200/230 -> 200/250
 - Repaired Patagium price : one digit were missing.
 - Unicorp price are already 1/3 of Nexus price. That's already cheap.
 - Repaired Za'lek engine limit: 300 -> 150.
 - engine reroute: +15% -> +30% (back to square 1...)

### On `p:outfitInitSlot()`

It has been generalized and improved by the `p:outfitMessageSlot()` function. I 
I removed any occurrences in lua files.
We can deprecate this one (or just: remove it).

### On multi-engines implementation using `pilot.outfitMessageSlot()`

This new feature made it simpler. However, stop using the global shipMemory and instead forced to a more explicit data exchange between outfits. The net effect simplifications + more explicit data exchange sums to a net 0: around the same number of line before and after. However, the new one has better safety properties and might be simplified a bit more, now.

### On the small adjustment I mention

I think the size 2 system cores were made much heavier to prevent sized-1 ships to use them. eml of size 2 ship had to be increased accordingly. 

Now that size 1 ship just do not have the core system slot for doing that, all that workaround becomes useless and can be undone.

We can go back to something less awkward. We can do it without changing anything in the game: we just have to take on one side what we give on the other. The net effect should be nul, but the parameters consistency will be improved.

### On the mvx/xml redundancy

There were discussions about that [here](https://github.com/naev/naev/pull/2802#issuecomment-2785448341).

My point was that `mvx` (just `xml` we don't want Naev to process) is a data format while `lua` is a programming language. So, using a data format for storing data makes sense. However, this `mvx` data is not usable as is by Naev, and therefore has to embed its own methods in lua to be usable, hence the xml/lua cores. Every script reading/writing outfit data uses `outfit.py` that relies on `elementTree`. I don't know how to reliably directly read data from xml/lua. I could manually parse current multicore.init arguments, but that's fragile. -> Ok, done. That's ugly, but it is doomed to be so.

Your point was essentially that `mvx` and `xml/lua` are redundant (which is the case) and that having to call `update_xml_from_mvx.sh` each time you modify a `mvx` in order to have the modification taken into account adds an additional step.

I still think working with `mvx` is better, and that we should remove all xml/lua and only generate them at build. But if I am provided a reliable way to convert `xml/lua` to `mvx`, we can generate mvx at build, and everything will work as previously.

#### Note 1
I made a xml/lua -> converter based on regexp. That should work as long as we don't change multicore initializer prototype.

#### Note 2
Conversions in both directions are slow. Well, on my machine. Probably not on most people machines, though.

#### Note 3 - current state

Now update_engines reads mvx and applies changes, and updates xml accordingly.
Basically, I will make mvx generated at build, so that all the script that use it (for reading only) can. Therefore, if anything changes some xml, build must be re-done so that mvx are updated accordingly, in order to have scripts that use it work with up-to-date data.

## TODO _after_ that
- simplify API using the new po:id()
- re-include Aegis in the game.
- Maybe add half-slots (slots of lesser size) for some Sirius ships.
- The energy efficiency of engines will also have to be worked on.
- Hide the entire secondary slot stuff until the player gets their first two-core ship.
- repair gauntlet supreme
- stealth and sneak attack
- re-balance money and time
- GUI
  - a lot to do!
  - propose GUI key binding rule : ESC can always get you out of here.
- callbacks definition problems (mainly overlapping).

## Testing Done
One step at a time.


johann1764:
## Issue

If we want to use messages to transmit engines characteristics to the pri engine instead of saving them in shipMemory, then the pri engine has to store them in its local variables.

That is not an issue, but that implies that, when there is no pri engine, the information is lost. Therefore, when a new pri engine is equipped, it has to ask all secondaries to re-send their information.

This mutual calling is manageable in this case, but it does not give clean code and is not so clear. Because we have:
1. pri asks for information
2. sec then sends the information
3. pri receives information on its callback and stores that
4. Task 2. resumes were it was
5. Task 1. resumes were it was and can use the information stored by the callback in the scope that contains both message() and the function of task 1 (therefore non-local).

While with returning data, we would have:
1. pri asks for information
2. sec returns information
3. Task 1. resumes were it was with the information as return value.
Just like a regular function call.

No needs for a global, here, and everything is simpler.

The return value should be a table (or a more general value), or nil if the messaged object:
 - does not exist
 - does not have a message function
 - returns nil

bobbens:
> If we want to use messages to transmit engines characteristics to the pri engine instead of saving them in shipMemory, then the pri engine has to store them in its local variables.
> 
> That is not an issue, but that implies that, when there is no pri engine, the information is lost. Therefore, when a new pri engine is equipped, it has to ask all secondaries to re-send their information.
> ...
> The return value should be a table (or a more general value), or nil if the messaged object:
>  * does not exist
>  * does not have a message function
>  * returns nil

I'm not sure how good of an idea it is to have return values supported. It's quite easy to implement in C, but it may be more problematic in Rust. The reason is that it forces a hard coupling between the sending environment and the pilot outfit. I guess I can probably figure a way to make this work as I think it's important as you want to guarantee an `init` initializes the outfit fully properly or it'll display the wrong information.

I've implemented it and pushed a nightly for now.

johann1764:
> 
> I've implemented it and pushed a nightly for now.

OK, thanks !
I adapted `multicore.lua` accordingly.

> I'm not sure how good of an idea it is to have return values supported. It's quite easy to implement in C, but it may be more problematic in Rust. The reason is that it forces a hard coupling between the sending environment and the pilot outfit. I guess I can probably figure a way to make this work as I think it's important as you want to guarantee an `init` initializes the outfit fully properly or it'll display the wrong information.

As I proved, having return values does not increase expressive power. Proof: it could be emulated when there were no return value. So rust will not be more displeased by this than by what it was before adding return value.

The underlying question is: how to have control on information exchange between outfits. Notice that, for now, the shipMemory is a global anyone can use, therefore outfits are free to exchange data as they wish.

On our case, I added a constraint that allows more control over data flow : received information and return information should be immutable, else just receiving a table once open an unlimited channel of communication. With this, any exchange of information can be traced within outfitMessage.

So current lua implementation does not write on args/ret val, therefore it would also work if args/ret value were "recursively const". Since it might no be possible to enforce that, we can still latch them, thus making them just as regular value passing, like a `int func(int)` latches its argument and return value :
```c
int global = 0;

int func(int a){
   global += a;
   ...
   return global;
}
...
   a = 2;
   b = func(a);
   // here a == 2 regardless what happens in func
   b += 1
   // global was not affected
```

This would obviously not work with pointers, as pointers are also latched but not the content they are pointing at.

Sorry if I break down open doors, but I never know whether I am clear.

bobbens:
> > I've implemented it and pushed a nightly for now.
> 
> OK, thanks ! I adapted `multicore.lua` accordingly.
> 
> > I'm not sure how good of an idea it is to have return values supported. It's quite easy to implement in C, but it may be more problematic in Rust. The reason is that it forces a hard coupling between the sending environment and the pilot outfit. I guess I can probably figure a way to make this work as I think it's important as you want to guarantee an `init` initializes the outfit fully properly or it'll display the wrong information.
> 
> As I proved, having return values does not increase expressive power. Proof: it could be emulated when there were no return value. So rust will not be more displeased by this than by what it was before adding return value.
> 
> The underlying question is: how to have control on information exchange between outfits. Notice that, for now, the shipMemory is a global anyone can use, therefore outfits are free to exchange data as they wish.
> 
> On our case, I added a constraint that allows more control over data flow : received information and return information should be immutable, else just receiving a table once open an unlimited channel of communication. With this, any exchange of information can be traced within outfitMessage.
> 
> So current lua implementation does not write on args/ret val, therefore it would also work if args/ret value were "recursively const". Since it might no be possible to enforce that, we can still latch them, thus making them just as regular value passing, like a `int func(int)` latches its argument and return value :
> 
> ```c
> int global = 0;
> 
> int func(int a){
>    global += a;
>    ...
>    return global;
> }
> ...
>    a = 2;
>    b = func(a);
>    // here a == 2 regardless what happens in func
>    b += 1
>    // global was not affected
> ```
> 
> This would obviously not work with pointers, as pointers are also latched but not the content they are pointing at.
> 
> Sorry if I break down open doors, but I never know whether I am clear.

This is just the current implementation. If / when we do multiple lua states (see #2828 ), then instead of just passing the Lua reference, it will actually have to copy and move the data over, so it will be immutable in that sense.

johann1764:
> This is just the current implementation. If / when we do multiple lua states (see https://github.com/naev/naev/issues/2828 ), then instead of just passing the Lua reference, it will actually have to copy and move the data over, so it will be immutable in that sense.

Yep. As it is unlikely that anyone uses `oufitMessageSlot()` relying on mutables until then, we can just leave it as is for now.

johann1764:
Out of scope: Were you notified I opened a tiny PR in naev-artwork ? In case you weren't: https://github.com/naev/naev-artwork/pull/75

bobbens:
I've seen some Thurion Ingenuities spawning with only a single systems and complaining about not enough CPU capacity. I don't know what is going on though. Haven't had any other ships give any warnings though.

johann1764:
That's strange. I will have a look at their equipopt, and if I find no clue, I'll try wandering around in Thurion systems.

Not sure it's related, but I saw some pirate ships stay in afterburner mode after exhausting their energy. That's frightening. This does not seem to happen all the time.

bobbens:
Leaving it here, but just noticed an Empire Pacifier with a single "Tricon Cyclone Engine" (but 2x battle plating and 2x core systems) chugging along at 60 km/s.

bobbens:
Let me relay a bit the information from discord as the user (HyperScoop) doesn't use github anymore:

Was playing around and noticed that now the Melendez Ox engine is the fastest in the game, and overall way too OP; it's faster than the Tricon Zephyr in a lot of cases now (on my perspicacity beta zephyr gives 642 km/s and ox 658 km/s, plus ox uses less power), which is quite silly and not in a very good way from a balancing perspective. Same could be more or less said about the rest of the melendez engines. I've been told the intention is for the melendez engines to be slow and nonviable for combat due to their low acceleration, but that's not really the case right now, as the acceleration they provide is more than sufficient for pretty much all applications and maximum speed is a lot more valuable anyway.

Here're my points:
- Tricon engines should never have the lowest maximum speed stat in their size class, melendez engines should never have the highest max speed in their size class; again, speed is more valuable than acceleration, so the engines' max speeds should relate to one another in roughly the way you'd want the max speeds of the ships they're for relating to one another, definitely not the other way around.
- If you want an engine's low acceleration to be an obstacle to combat use, you should aim even lower. I'd say reduce the acceleration stat on the melendez engines by a factor of 1.5-2.0 from what it is now and that should do it.
- Up the mass limits on the Ox and Buffalo melendez engines slightly. keep in mind that with a cargo ship you want some headroom in your mass budget for the mass of the cargo, the engines should account for that.
- The Unicorp engines currently have little to no purpose as they're worse than their Nexus counterparts in most regards; should increase their mass limits slightly and maybe lower energy use by the Hawk in a primary slot down to 2? idk
- side note: do something about the Za'lek Test Engine as it has 300 mass limit compared to the Ox.

johann1764:
| | speed | max speed | accel | fullsp (s) | fullsp (km) | turn | turn radius | 1/2 turn (s)
| ---                     | ---    | ---    | ---    | ---    | ---    | ---    | ---    | ---
| T. Zephyr Engine alone  | 287    | 400    | 452    |   0.88 | 177    | 213    | 108    |   0.85
| M. Ox Engine alone      | 343    | 394.75 | 207    |   1.91 | 376    | 174    | 130    |   1.03
| T. Zephyr Engine x2  | 259    | 336.25 | 309    |   1.09 | 183    | 170    | 113    |   1.06
| M. Ox Engine x2      | 296    | 331.25 | 141    |   2.35 | 389    | 138    | 138    |   1.3
| T. Cyclone Engine alone | 222    | 272.75 | 203    |   1.34 | 183    | 131    | 119    |   1.37
| M. Buffalo Engine alone | 245    | 268.25 |  93    |   2.88 | 387    | 106    | 145    |   1.7
| T. Cyclone Engine x2 | 180    | 212    | 128    |   1.66 | 176    |  97    | 125    |   1.86
| M. Buffalo Engine x2 | 193    | 207.5  |  58    |   3.58 | 371    |  78    | 152    |   2.31
| T. Typhoon Engine alone | 140    | 159.5  |  78    |   2.04 | 163    |  69    | 132    |   2.61
| M. Mammoth Engine alone | 147    | 156    |  36    |   4.33 | 338    |  56    | 160    |   3.21
| T. Typhoon Engine x2 | 104    | 115.5  |  46    |   2.51 | 145    |  47    | 141    |   3.83
| M. Mammoth Engine x2 | 107    | 112.25 |  21    |   5.35 | 300    |  38    | 169    |   4.74

johann1764:
| | speed | max speed | accel | fullsp (s) | fullsp (km) | turn | turn radius | 1/2 turn (s)
| ---                        | ---    | ---    | ---    | ---    | ---    | ---    | ---    | ---
| N. Dart 160 Engine alone   | 313    | 392.25 | 317    |   1.24 | 243    | 192    | 117    |   0.94
| U. Hawk 160 Engine alone   | 298    | 373.25 | 301    |   1.24 | 231    | 183    | 117    |   0.98
| N. Dart 160 Engine x2 x2   | 275    | 329    | 216    |   1.52 | 251    | 153    | 123    |   1.18
| U. Hawk 160 Engine x2 x2   | 259    | 309.75 | 203    |   1.53 | 236    | 144    | 123    |   1.25
| N. Arrow 700 Engine alone  | 230    | 265.25 | 141    |   1.88 | 249    | 117    | 130    |   1.54
| U. Falcon 700 Engine alone | 214    | 246.75 | 131    |   1.88 | 232    | 109    | 130    |   1.65
| N. Arrow 700 Engine x2 x2  | 183    | 205.25 |  89    |   2.31 | 237    |  86    | 137    |   2.09
| U. Falcon 700 Engine x2 x2 | 169    | 189.5  |  82    |   2.31 | 219    |  79    | 137    |   2.28
| N. Bolt 3000 Engine alone  | 140    | 153.5  |  54    |   2.84 | 218    |  61    | 144    |   2.95
| U. Eagle 3000 Engine alone | 128    | 140.25 |  49    |   2.86 | 201    |  56    | 143    |   3.21
| N. Bolt 3000 Engine x2 x2  | 103    | 111    |  32    |   3.47 | 193    |  42    | 151    |   4.29
| U. Eagle 3000 Engine x2 x2 |  93    | 100.25 |  29    |   3.46 | 173    |  38    | 151    |   4.74

johann1764:
> Let me relay a bit the information from discord as the user (HyperScoop) doesn't use github anymore:
> 
> Was playing around and noticed that now the Melendez Ox engine is the fastest in the game, and overall way too OP; it's faster than the Tricon Zephyr in a lot of cases now (on my perspicacity beta zephyr gives 642 km/s and ox 658 km/s, plus ox uses less power), which is quite silly and not in a very good way from a balancing perspective. Same could be more or less said about the rest of the melendez engines. I've been told the intention is for the melendez engines to be slow and nonviable for combat due to their low acceleration, but that's not really the case right now, as the acceleration they provide is more than sufficient for pretty much all applications and maximum speed is a lot more valuable anyway.
> 
> Here're my points:
> 
> * Tricon engines should never have the lowest maximum speed stat in their size class, melendez engines should never have the highest max speed in their size class; again, speed is more valuable than acceleration, so the engines' max speeds should relate to one another in roughly the way you'd want the max speeds of the ships they're for relating to one another, definitely not the other way around.

They don't. However, ships that are very speed-oriented (like the perspicacity beta) benefit more from the high drift speed of Melendez and less from the high Tricon accel.

> * If you want an engine's low acceleration to be an obstacle to combat use, you should aim even lower. I'd say reduce the acceleration stat on the melendez engines by a factor of 1.5-2.0 from what it is now and that should do it.

Yep. that is what I meant by "making Melendez even more Melendez".
Maybe not that much, though.

> * Up the mass limits on the Ox and Buffalo melendez engines slightly. keep in mind that with a cargo ship you want some headroom in your mass budget for the mass of the cargo, the engines should account for that.

Yep, but mass-related things are not for now, as it is a complicated matter. Can be done a bit, though, as Melendez are at a disadvantage, here.

> * The Unicorp engines currently have little to no purpose as they're worse than their Nexus counterparts in most regards; should increase their mass limits slightly and maybe lower energy use by the Hawk in a primary slot down to 2? idk

Unicorp are supposed to be standard and low-cost. Will check whether they could be made more low-cost. From a player perspective, they are just early-game/NPC outfits.

> * side note: do something about the Za'lek Test Engine as it has 300 mass limit compared to the Ox.

That is clearly a mistake. Should be 150.

I went to discord, and after a (heated) debate, it appeared clearly that a few changes had to be made. I included them as "first rebalance" items in the opening post.

Note: I used this PR thread as a place to put data useful to the debate in discord. md is not accepted everywhere.

johann1764:
I am still making tests for this first pre-rebalance.

It occurs to me that there is a slight inconsistency we could fix, that would also help a lot in making things understandable for players.

I am talking about the speed modifier. It modifies _drift speed_. My proposition is that it should modify the speed in a general way, so both drift speed and max speed.

So we would have:
```
total_drift_speed = (1.0+speed_mod/100.0) * drift_speed
max_speed = (1.0+speed_mod/100.0) * (drift_speed + accel/PHYS_CONST)
```
or:
```
max_speed = total_drift_speed + (1.0+speed_mod/100.0) * accel/PHYS_CONST
```

That way, high agility engines would not be penalized by ships with high speed mods.
Because these speed mods were meant to improve speed in a general sense.

johann1764:
> Leaving it here, but just noticed an Empire Pacifier with a single "Tricon Cyclone Engine" (but 2x battle plating and 2x core systems) chugging along at 60 km/s.

Fixed ! (also fixed other empire ships)

bobbens:
> With the fix we would have:
> 
> ```c
> max_speed = (1.0+speed_mod/100.0) * (drift_speed + accel/PHYS_CONST)
> ```
> 
> or:
> 
> ```c
> max_speed = total_drift_speed + (1.0+speed_mod/100.0) * accel/PHYS_CONST
> ```

Doesn't this basically just nerf speed_mod to the ground? It also makes it so accel_mod and speed_mod give the same bonus to speed, while accel_mod also improves acceleration.

If you expand `accel = base_accel * (1.0 + accel_mod/100.0)`, you get:
```c
max_speed = total_drift_speed + (1.0+speed_mod/100.0) * (base_accel * (1.0 + accel_mod/100.0)) /PHYS_CONST
```

johann1764:
> Doesn't this basically just nerf speed_mod to the ground? It also makes it so accel_mod and speed_mod give the same bonus to speed, while accel_mod also improves acceleration.

No. accel_mod improves marginally the max speed (can go up to 25% so +10% accel gives up to 2.5% max speed) but improves actual accel,
while speed_mod would affect speed and max speed the same way.

Looks fair to me.

bobbens:
> > Doesn't this basically just nerf speed_mod to the ground? It also makes it so accel_mod and speed_mod give the same bonus to speed, while accel_mod also improves acceleration.
> 
> No. accel_mod only improves marginally the max speed (can go up to 25% so +10% accel gives up to 2.5% max speed) but improves actual accel, while speed_mod would affect speed and max speed the same way (that is: a lot).
> 
> Looks fair to me.
> 
> And as I mentioned, this is not a change, but a choice that we face as we just broke the accel/speed constant ratio. Now we have to generalize previous behavior, and several ways are possible. But one looks more consistent to me.

I see it now. Yes, it should be fine. Might need to tweak some numbers though.

Issue #2833, opened 2025-05-03
[Proposal]: colored console output


johann1764:
### Suggestion

Two different things:
 - the GUI color codes `#r`, `#b`, etc. should be understood by the console window.
 - terminal output should stay readable, and ideally having the same colors.

What looks simplest would be that print, warn, etc. recognize `#<letter>` codes and that the following substitution is made before terminal output:
 - `#r` -> `\e[0;31m` [red]
 - `#n` -> `\e[1;30m` [grey]
 - `#b` -> `\e[0;34m` [blue]
 - `#p` -> `\e[0;35m` [purple or pink? -> magenta]
 - `#g` -> `\e[0;32m` [green]
 - `#y` -> `\e[1;33m` [yellow]
 - `#o` -> `\e[0;33m` [orange -> brown]
 - `#w` -> `\e[0;37m` [white]
 - `#0` -> `\e[0m   ` [default]

where '\e' is the escape character (`27` in decimal, `\033` in octal, `0x1b` in hexa)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
You might want to try the experimental (`lua_repl` option in config). It is much cooler and has potential, but for some reason or other, I forget why, but I stopped using it. It also technically should be able to support things like proper tab complete, but iirc that's broken.

johann1764:
I had a look to lua_repl. A lot to unfold here, but in short, it would be a big thing to work on, and I already have a few things to finish before opening a new significant issue.

So, I finally propose a small patch that addresses quickly the original issue, and postpone the work on lua_repl to another time.

PR #2834, opened 2025-05-04
Colored CLI/term output

johann1764:
**New Feature**

This PR addresses the feature described in #2833.

## Summary
1. Restored CLI color output.
2. Translated this `#.`-colored output into ANSI escape sequences in order to pass it to LOG.
3. LOG now outputs its input string to stdout/stderr as usual but removes ansi esc. seq. when it comes to writing to a file.

## Testing Done
Tested the introduced funcs separately, but **did not test into the project**.

### DIFF
--- a/src/console.c
+++ b/src/console.c
@@ -121,12 +121,12 @@ static char *fontcol_toTermEscapeString(const char*s){
    char*buf;
    int i, wi=0, count=0;
 
-   for(i=0; s[i++]; i++)
+   for(i=0; s[i]; i++)
       if(s[i] == FONT_COLOUR_CODE)
          count++;
 
    buf=calloc(i+count*7+1,sizeof(char));
-   for(i=0; i<len; i++){
+   for(i=0; s[i]; i++){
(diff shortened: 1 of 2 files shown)


ProjectSynchro:
Fixed the meson bug with CI, looks like it doesn't quite compile yet.

bobbens:
I've tested it with a script that `print( "#rRed" )` (because you can't type "#" into the console due to lazy implementation), and it just prints `#rRed' in the stdout, while in-game terminal is properly red.

bobbens:
Thanks! It works, but it should revert the colour after a print. Right now if you do something like "#rRed", it'll stay red forever since the Naev system reverts the colour on a new print and the terminal does not.

bobbens:
Seems to work now, thanks!

johann1764:
Just noticed a test can be simplified. one minute !

johann1764:
Sorry for the last minute change.

Its slightly simpler now, and in addition don't use "\e[0m" when it is not necessary.

bobbens:
Looks good, thanks!

PR #2844, opened 2025-05-10
Fix a bug that autonav fails to jump instantly in some conditions.

oo13:
**Bug Fix**

Autonav underestimated the distance to turn the ship. The cause was the wrong unit of the angle (it used rad/2π instead of rad).

## Testing Done
My Quicksilver Mercury can jump instantly even if it enters a jump point from the opposite direction.  (It fails to jump in the current version.)


bobbens:
Looks good, thanks!

PR #2846, opened 2025-05-10
meson experiment

johann1764:
**experiment**

I use this PR for an experiment.

It is not intended for merging.

When it finally works, I will rebase and open a new, fresh PR with much less useless commits.


johann1764:
I am pissed of.

I will have `generate.py` do the work meson does not want (or can't) do.

PR #2842, opened 2025-05-08
Multicore polish (again)

johann1764:
Little bits of:
 - **Balance**
 - **Bug Fix**(es)
 - **New Feature**(s)

This PR replaces #2832, and addresses some features described in #2766.
After #2821, all cores are now split.

This is the technical part of the multicore epilog.

## Summary
 - [x] Complete `outfits2md.py` stacking option
 - [x] Update luadoc for `pilot.outfitMessageSlot`
 - [x] Fix the mono-engine Empire Pacifier bug.
 - [x] Switch from `pilot.outfitInitSlot()` to `pilot.outfitMessageSlot()`. (see note)
 - [x] Use the `pilot.outfitMessageSlot()` to transmit data, not `p:shipMemory()` (see note)
    - [x] Re-declare all secondaries when primary is added/inited
    - [x] Have pri engine save all data
    - [x] manage backward info (server->client)
    - [x] Check the calls, should be less redundant -> still the add/init overlap, else it's ok.
 - [x] First rebalance (see note below)
 - [x] have engine_stats allow to also compute engines compositions
 - [x] have engine_stats and outfits2md detect non-stackable outfits
 - [x] manage the mvx/xml redundancy (see note)
    - [x] Done a xmllua2mvx
    - [x] Make sure all conversions are ok.
    - [x] Make sure all reverse-conversions are ok.
    - [x] parsearg-ify + python3-ify converters
    - [x] make sure update_engine also updates xmllua.
 - [x] A *neutral* adjustment (-15t eml for small engines sec, -15t core sys sec). (see note)
 - [x] understand and mitigate the engine cores paradox (see below)
 - [x] Bug: a ship with a single Patagium Twin in pri slot is considered space-worthy.
 - [x] Speed mod should affect all speeds
 - [x] Added spaces in py around `==`, `!=`, `<=`, `=`, `+=`, etc.

### The engines paradox
As contribution to speed depends on eml, and as engines pri contribution is better, a primary engine with high eml will impose its (fast) pace. As a consequence, heavy engines in primary may result in faster ships. Conversely, putting a low eml in pri engine and high eml in sec will always result in a poor combination.

I downgraded Nexus and Melendez to mitigate that.

<details><summary>All Engines Combinations (with size(pri)>=size(sec))</summary>

|  | eml (t) | drift speed  | max speed | accel | fullsp (s) | fullsp (km) | turn (°/s) | turn radius | 1/2turn (s)
[...]

PR #2847, opened 2025-05-10
Having mvx generated as needed

johann1764:
**This PR have been replaced by #2848.**

> **cleanup**
> 
> This PR addresses the feature left undone in #2842.
> 
> ## Summary
> Modified `generate.py` to generate mvx from xml as needed, on a timestamp criterion (make-like dependency management). Only ~15 lines.
> 
> Removed mvx files.
> 
> This PR has turned into another experiment.
> 
> ## Testing Done
> None so far.
>


johann1764:
I don't understand.

No significant change over last 3 commits, but one time Linux failed, one time Windows failed, one time both. wtf ???

bobbens:
You shouldn't need to check timestamps to generate if needed. Meson should handle that automagically (and also handle dependency updates and the likes). I don't understand why it is necessary.

johann1764:
> You shouldn't need to check timestamps to generate if needed. Meson should handle that automatically (and also handle dependency updates and the likes). I don't understand why it is necessary.

No, I shouldn't need to do that. Yes, Meson should handle that. But it simply does not.

First, Meson absolutely wants you to follow its belief regarding how directory structure should be. Even worse, the function that is supposed to do that (a generator) actually does the file conversion each time (says its doc). And I just figured out why.

It occurs that meson clears the output directories arbitrarily. That is why I had compilation errors: Meson removed some files that were currently read by the scripts, falsely believing it was not needed anymore.

Now I have a solution that works at the cost of a few lines that would have been needed in any dependency management system anyway. I will open a new, fresh PR, with a single, simple commit.

PR #2848, opened 2025-05-11
Having mvx generated as needed (again)

johann1764:
**cleanup**

This PR addresses the feature left undone in #2842.

## Summary
 - Modified `generate.py` to generate `mvx` from `xml` as needed. Only ~15 lines.
 - Made `generate.py` output depend on `xml` rather than `mvx` in `meson.build`.
 - (Much) cleaner path management in `generate.py` and other scripts.
 - Removed `mvx` files.

## Note
The whole `generate.py` is run for every series of generated bio-outfit.

We could save build time and CPU by doing it all at the same time.

But that requires to negotiate with meson. I won't (or, at least, not now)


bobbens:
OK, it is good. Some suggestions (can be done after this PR is merde):

1. Simplify format to avoid copy pasting if possible. I would suggest something like:
```meson
biotmp_weapon = files('templates/weapon.xml.template')

biooutfits = {
  'perlevis_gene_drive' : {
    'template': files('templates/gene_drive_tricon.xml.template'),
    'depend_files' : [
      'core_engine/small/tricon_zephyr_engine.xml',
     ],
    'output' : [
         'perlevis_gene_drive_i.xml',
         'perlevis_gene_drive_ii.xml',
    ],
   },
  'tentacle_organ' : {
    'template' : biotmp_weapon,
    'output' : [
        'tentacle_organ_i.xml',
        'tentacle_organ_ii.xml',
        'tentacle_organ_iii.xml',
        'tentacle_organ_iv.xml'
      ],
  },
}

foreach name, data : biooutfits
  depend_files = []
  if 'depend_files' in data
    foreach df : data['depend_files']
      depend_files += files( '../' + df )
    endforeach
  endif

  bio_outfits += [custom_target( name, command: bio_cmd, install: true, install_dir: bio_dir,
    input: data['template'],
    depend_files: depend_files,
    output: data['output'],
  )]
endforeach
```
Would potentially allow changing paths and such easily for example, if we move the meson script up one step so it can handle the mvx generation + bioships in one step explicitly (or other potential outfits).
2. Meson wants explicit targets. So it would make sense to move the `meson.build` up one path (to `dat/outfits/`) and handle both making the mvx + xml explicitly. When iterating over the biooutfits (probably needs a better name), you could also generate the custom targets for the dependencies (the mvx files). This would mean modifying the above code to so that the bio_outfit custom_target depends on the mvx, _NOT_ the xml, and have a custom rule to generate the mvx from the xml. As needed, meson would generate the mvx files from the xml and have it generate them as needed. Only tricky part is you can't generate duplicate custom_targets, so you would probably need one pass to generate the custom xml->mvx targets, and one to generate the bio_outfits targets.

johann1764:
> OK, it is good. Some suggestions (can be done after this PR is merde):

That's a terrible typo you've made. In my tongue merde means shit :-D

> 
> 1. Simplify format to avoid copy pasting if possible. I would suggest something like:
> 
> ```meson
> biotmp_weapon = files('templates/weapon.xml.template')
> 
> biooutfits = {
>   'perlevis_gene_drive' : {
>     'template': files('templates/gene_drive_tricon.xml.template'),
>     'depend_files' : [
>       'core_engine/small/tricon_zephyr_engine.xml',
>      ],
>     'output' : [
>          'perlevis_gene_drive_i.xml',
>          'perlevis_gene_drive_ii.xml',
>     ],
>    },
>   'tentacle_organ' : {
>     'template' : biotmp_weapon,
>     'output' : [
>         'tentacle_organ_i.xml',
>         'tentacle_organ_ii.xml',
>         'tentacle_organ_iii.xml',
>         'tentacle_organ_iv.xml'
>       ],
>   },
> }
> 
> foreach name, data : biooutfits
>   depend_files = []
>   if 'depend_files' in data
>     foreach df : data['depend_files']
>       depend_files += files( '../' + df )
>     endforeach
>   endif
> 
>   bio_outfits += [custom_target( name, command: bio_cmd, install: true, install_dir: bio_dir,
>     input: data['template'],
>     depend_files: depend_files,
>     output: data['output'],
>   )]
> endforeach
> ```
> 
> Would potentially allow changing paths and such easily for example, if we move the meson script up one step so it can handle the mvx generation + bioships in one step explicitly (or other potential outfits). 2. Meson wants explicit targets. So it would make sense to move the `meson.build` up one path (to `dat/outfits/`) and handle both making the mvx + xml explicitly. When iterating over the biooutfits (probably needs a better name), you could also generate the custom targets for the dependencies (the mvx files). This would mean modifying the above code to so that the bio_outfit custom_target depends on the mvx, _NOT_ the xml, and have a custom rule to generate the mvx from the xml. As needed, meson would generate the mvx files from the xml and have it generate them as needed. Only tricky part is you can't generate duplicate custom_targets, so you would probably need one pass to generate the custom xml->mvx targets, and one to generate the bio_outfits targets.

I will copy this post to the next PR. Let us merge this as this works and removes the mvx, and I will open another one.

The main objective would be to make only one call to generate. This however, does not require to generate all mvx.

By the way, the outfit utils, that can't read xml with multicore, could just use the same function as generate.py currently does.

And, for clarity, maybe `file.mvx` should be renamed `.file.mvx`.

bobbens:
OK! Thanks!

PR #2849, opened 2025-05-11
two little bugfixes

johann1764:
**Bug Fixes**

This PR addresses some bugs described respectively in #2766 and #2850.

## Summary
 - [x] Fix the eml-in-blue bug, should be green, (giving the false impression eml was _affected by_ the load factor)
 - Fix the commodity crash
    - [x] `economy_getAverageSpobPrice` was not cautious enough.
    - [x] `tech_getCommodity` behavior was misleading: it did not initialize the ptr when failing.

## Testing Done
Only the first.


bobbens:
Looks good, thanks!

PR #2852, opened 2025-05-11
mvx dependency

johann1764:
**Cleanup / New Feature**

This PR improves PR #2848.

## Summary
 - [x] renamed `.<file>.mvx` the multi-valued version of `<file>.xml`
 - [x] new function core_outfit for scripts
 - [x] update `sh` scripts accordingly
 - [x] adapted `generate.py` with this new function
 - [x] made it cleaner and clearer
 - [x] apply suggestion 1 below
 - [x] make it work
 - [x] improve it
 - [x] adapt `update_engine.py` to `core_outfit`
 - [x] check `update_engine.py` generate mode still works
 - [x] adapt other scripts to `core_outfit`
 - [x] adapt `outfits2md`
 - [x] do not use subprocess but directly call relevant py func. -> So much faster
 - [x] cleanup
 - `to_dict` should not parse the text field, and leave it as is, with a dedicated function for the user to do that if needed. EDIT: this would improve some scripts but not conversion, so out of this scope.

## For another PR
 - improve `generate.py` so that it does only what is required by its arguments.

## Suggestions
Copied from https://github.com/naev/naev/pull/2848#issuecomment-2869814147:

<details><summary> Suggestion 1 (implemented) </summary>
> 1. Simplify format to avoid copy pasting if possible. I would suggest something like:
> 
> ```meson
> biotmp_weapon = files('templates/weapon.xml.template')
> 
> biooutfits = {
>   'perlevis_gene_drive' : {
>     'template': files('templates/gene_drive_tricon.xml.template'),
>     'depend_files' : [
>       'core_engine/small/tricon_zephyr_engine.xml',
>      ],
>     'output' : [
>          'perlevis_gene_drive_i.xml',
>          'perlevis_gene_drive_ii.xml',
>     ],
>    },
>   'tentacle_organ' : {
>     'template' : biotmp_weapon,
>     'output' : [
>         'tentacle_organ_i.xml',
>         'tentacle_organ_ii.xml',
>         'tentacle_organ_iii.xml',
>         'tentacle_organ_iv.xml'
>       ],
>   },
> }
> 
> foreach name, data : biooutfits
>   depend_files = []
>   if 'depend_files' in data
>     foreach df : data['depend_files']
>       depend_files += files( '../' + df )
>     endforeach
>   endif
> 
>   bio_outfits += [custom_target( name, command: bio_cmd, install: true, install_dir: bio_dir,
>     input: data['template'],
>     depend_files: depend_files,
>     output: data['output'],
>   )]
> endforeach
> ```
> 
> Would potentially allow changing paths and such easily for example, if we move the meson script up one step so it can handle the mvx generation + bioships in one step explicitly (or other potential outfits).

I can try that, would be clearer. If it converges to a working build in few iterations, I'll do that. But if it fails repeatedly, I won't insist too much, as pushing commits to a PR in order to test is not time efficient at all. In that case, it will wait until I can build locally.
</details>

<details><summary> Suggestion 2 (discussed below) </summary>
>
> 2. Meson wants explicit targets. So it would make sense to move the `meson.build` up one path (to `dat/outfits/`) and handle both making the mvx + xml explicitly. When iterating over the biooutfits (probably needs a better name), you could also generate the custom targets for the dependencies (the mvx files). This would mean modifying the above code to so that the bio_outfit custom_target depends on the mvx, _NOT_ the xml, and have a custom rule to generate the mvx from the xml. As needed, meson would generate the mvx files from the xml and have it generate them as needed. Only tricky part is you can't generate duplicate custom_targets, so you would probably need one pass to generate the custom xml->mvx targets, and one to generate the bio_outfits targets.
</details>

## Testing Done
Ticked features are tested.


bobbens:
> Whow. Meson wants explicit targets. Meson can only handle sub directories. Meson can't manage duplicate targets. What kind of dependency management system is this ?

Meson is quite strict (for better or for worse) to not allow turing-completeness in the language and make sure that it produces a finite graph, so it can do optimizations on it and such. I do think it is a bit too pedantic, but it does have fewer issues long-run than most other systems that require a lot more fiddling especially when doing lots of platforms at once.

> If I understand well, Meson will repeat conversions into mvx if several targets require the same mvx file, unless we manage it for him. Well, that's kind of managed. If any xml changes, meson supposedly can update bio-outfits accordingly. And, in detail, the intermediate mvx file management is left to a function that only updates the mvx if needed.

Actually, it should build a full dependency graph, so if they point to the same mvx file, it should only be generated once. It'll also compile everything it can in parallel while enforcing the dependency tree. It shouldn't even let you define multiple targets with the same file either, as that would break this behaviour.

johann1764:
> > Whow. Meson wants explicit targets. Meson can only handle sub directories. Meson can't manage duplicate targets. What kind of dependency management system is this ?
> 
> Meson is quite strict (for better or for worse) to not allow turing-completeness in the language and make sure that it produces a finite graph, so it can do optimizations on it and such.

Limiting expressive power makes sense.

> I do think it is a bit too pedantic,

I would not say "a bit". That, and oddly specific constructs. It really looks like a system for c projects, which were added a new, specific feature each time it proved too restrictive.

A little story: years ago, I worked on a system for natural language processing, that required a management of different processing steps, with large processing times for some of them. People used a specific tools for that, and I found out that `make` expressed the problem more concisely, more clearly and did overall a better job that the specific tool that was developed for this purpose. No compilation were involved.

Conclusion: a dependency management system is not specifically about compiling and linking.

> but it does have fewer issues long-run than most other systems that require a lot more fiddling especially when doing lots of platforms at once.

I will take your word for that.

> 
> > If I understand well, Meson will repeat conversions into mvx if several targets require the same mvx file, unless we manage it for him. Well, that's kind of managed. If any xml changes, meson supposedly can update bio-outfits accordingly. And, in detail, the intermediate mvx file management is left to a function that only updates the mvx if needed.
> 
> Actually, it should build a full dependency graph, so if they point to the same mvx file, it should only be generated once. It'll also compile everything it can in parallel while enforcing the dependency tree. It shouldn't even let you define multiple targets with the same file either, as that would break this behaviour.

`make` does that.

I was ok with the idea of a finite graph (even it enforcing explicit targets for that is too restrictive), and even a finite DAG. But now, this is about a tree. If turning the DAG into a tree is left to the user, I am beginning to understand why it is both painful and has fewer issues long-term. This is just because the user handles the work of the dependency manager system.

bobbens:
> > Actually, it should build a full dependency graph, so if they point to the same mvx file, it should only be generated once. It'll also compile everything it can in parallel while enforcing the dependency tree. It shouldn't even let you define multiple targets with the same file either, as that would break this behaviour.
> 
> `make` does that.
> 
> I was ok with the idea of a finite graph (even it enforcing explicit targets for that is too restrictive), and even a finite DAG. But now, this is about a tree. If turning the DAG into a tree is left to the user, I am beginning to understand why it is both painful and has fewer issues long-term. This is just because the user handles the work of the dependency manager system.

I don't see why you have to convert it to a tree? As long as it's a DAG it should work. You can have something like A that produces a target used by B and C, and then a target D that uses both B and C (thus a DAG, but not a tree), and it should compile fine with meson. It should error out if you create cycles (A depends on B and viceversa), but I believe it detects those.

Naev used to use autotools + make. It was much more finicky and would break often, especially in weird situations of cross-compiling and the likes. I do find meson to be more clean, although you have to do things the meson way. That said, the using Rust cargo with meson is a bit ugly, because you end up using two build systems and they have fundamentally incompatible design philosophies, but that shouldn't affect the current main branch that doesn't use Rust.

johann1764:
> I don't see why you have to convert it to a tree? As long as it's a DAG it should work. You can have something like A that produces a target used by B and C, and then a target D that uses both B and C (thus a DAG, but not a tree), and it should compile fine with meson. It should error out if you create cycles (A depends on B and viceversa), but I believe it detects those.

Maybe I did not understand well.
Good news, then, it is not as bad as I thought it was.

> 
> Naev used to use autotools + make. It was much more finicky and would break often, especially in weird situations of cross-compiling and the likes. I do find meson to be more clean, although you have to do things the meson way. That said, the using Rust cargo with meson is a bit ugly, because you end up using two build systems and they have fundamentally incompatible design philosophies, but that shouldn't affect the current main branch that doesn't use Rust.

Ok. I guess something simpler will have to be done for that.
But that's in a distant future, we will see in due time.

johann1764:
That will be all for this PR, I will do the optimization part in another one.

utils/outfits/core_outfit.py
```diff
@@ -0,0 +1,45 @@
...
+
+def mvx_nam(xml):
+   return path.join(path.dirname(xml), '.' + path.basename(xml)[:-3] + 'mvx')
+
+def _gen_if_needed( xml, force = False ):
+   mvx = mvx_nam(xml)
+   exists = (not force) and Path(mvx).is_file()
+   uptodate = exists and not path.getmtime(mvx) < path.getmtime(xml)
+   if not uptodate:
+      return xmllua2mvx(xml, mvx, quiet = False )
```
johann1764, line 19:
```suggestion
      return xmllua2mvx(xml, mvx, quiet = False)
```

utils/outfits/engine_stats.py
```diff
@@ -59,7 +59,8 @@ def l( s ):
    a, b = tuple(s.split('.', 1))
    return ' | '+(3-len(a))*' '+a+m+b+(2-len(b))*' '
```
johann1764, line 60:
```suggestion
   return ' | ' + (3-len(a))*' ' + a + m + b + (2-len(b))*' '
```

utils/outfits/engine_stats.py
```diff
@@ -144,12 +141,9 @@ def main( args, gith = False, color = False, autostack=False, combine=False, nos
          acc += l(fmt(fullsptime(k)))+l(fmt(fullspdist(k)))+l(turn(k))+l(radius(k))
          acc += l(fmt(turntime(k)))
```
johann1764, line 142:
```suggestion
         acc += l(fmt(fullsptime(k))) + l(fmt(fullspdist(k))) + l(turn(k))
         acc += l(radius(k)) + l(fmt(turntime(k)))
```

utils/outfits/engine_stats.py
```diff
@@ -144,12 +141,9 @@ def main( args, gith = False, color = False, autostack=False, combine=False, nos
...
          if gith:
-            acc = acc.replace('  ', ' ').replace('  ', ' ')
-         elif color:
-            acc = greyit(acc)
-            if (count%3) == 2:
-               acc = acc.replace(grey, altgrey)
-            count += 1
+            acc = acc.replace('   ', ' ').replace('  ', ' ')
```
johann1764, line 144:
```suggestion
            acc = acc.replace(7*' ', ' ').replace(3*' ', ' ').replace(2*' ', ' ')
```

utils/outfits/mvx2xmllua.py
```diff
@@ -55,8 +55,8 @@ def _mklua( L ):
 
    return output+'}\n'
 
-def toxmllua( o ):
-   T, R = o.T, o.r
+def _toxmllua( o ):
+   R = o.r
```
johann1764, line 60:
```suggestion
```

utils/outfits/mvx2xmllua.py
```diff
@@ -85,21 +85,19 @@ def toxmllua( o ):
       el.text = _mklua(acc1+acc2) + el.text
       break
 
+def mvx2xmllua( argin, argout, quiet ):
+   o = outfit(argin)
+   if o is not None:
+      nam = nam2fil(o.name())
+      _toxmllua(o)
+      if not quiet:
+         stderr.write('mvx2xmllua: '+(nam if argout == '-' else argout)+'\n')
```
johann1764, line 94:
```suggestion
         stderr.write('mvx2xmllua: ' + (nam if argout == '-' else argout) + '\n')
```

utils/outfits/outfit.py
```diff
@@ -1,14 +1,14 @@
 #!/usr/bin/env python
 
-from sys import stdin, stdout, stderr
+from sys import stdin, stdout
```
johann1764, line 3:
```suggestion
from sys import stdin, stdout
```

utils/outfits/outfit.py
```diff
@@ -290,13 +293,3 @@ def outfit( fil, content = False ):
       raise Exception('Invalid outfit <user-data content>')
    else:
       raise Exception('Invalid outfit "'+str(fil)+'"')
```
johann1764, line 295:
```suggestion
      raise Exception('Invalid outfit "' + str(fil) + '"')
```

utils/outfits/outfits2md.py
```diff
@@ -1,15 +1,15 @@
...
 
-from sys import stderr
-from outfit import outfit
+from sys import stderr, stdin
+from core_outfit import some_outfit
 from outfit import LOWER_BETTER
 
 def transpose( M ):
    N = max(map(len, M))
    M = [t+['']*(N-len(t)) for t in M]
```
johann1764, line 9:
```suggestion
   M = [t + ['']*(N-len(t)) for t in M]
```

utils/outfits/update_engines.py
```diff
@@ -233,13 +225,10 @@ def gen_line( params ):
...
-      description = """The changes made will be listed onto <stderr>: \"_\" means \"nothing\".
-If some mvx files are modified, the related xmllua file will be updated.""",
+      description = 'The changes made will be listed onto <stderr>: \"_\" means \"nothing\".',
       epilog = """Examples:
   Standard usage:
-   > ./utils/outfits/update_engines.py `find dat/outfits/core_engine/ | grep mvx`
-
-  Generate a line called Krain with same params as Krain:
+   > find dat/outfits/core_engine/ -name "*.xml" | ./utils/outfits/update_engines.py -f
    > ./utils/outfits/update_engines.py -g Krain
```
johann1764, line 232:
```suggestion
  Generate a line called Krain with same params as Krain:
   > ./utils/outfits/update_engines.py -g Krain
```

johann1764 (commented):
last details

bobbens:
Looks good, thanks!
In the next PR, when splitting mvx generation from the python script, it probably would be worth having it not print anything by default, as it now prints a lot of `xmllua2mvx: Tricon Zephyr Engine` and such.

bobbens:
(Since I rebased, the merge didn't get detected, so closing)

johann1764:
> Looks good, thanks!
> In the next PR, when splitting mvx generation from the python script, it probably would be worth having it not print anything by default, as it now prints a lot of `xmllua2mvx: Tricon Zephyr Engine` and such.

Just have to insert ', quiet = True' somewhere.

PR #2853, opened 2025-05-13
multi-core details

johann1764:
**New Feature**

This PR addresses some features described in #2855.

## Summary
 - [x] Improve engine_stat term output further
 - [x] Silence mvx conversion messages at build
 - [x] Adapt the fuelshell plating (pri-only for now, but has sec attr defined if needed)
 - [x] Repair combinations when pri misses fields
 - [x] Repair some_outfit func
 - [x] Allow secondary-only cores to appear in the core section.
 - [x] Introduced an experimental sec-only hull called "Krain Nanotube Foam Padding" ("K. Nanofoam Padding", for short) Feedback welcomed !
 - [x] Bug: sec-only can also equip on any structural slots.
 - [x] Readjust patchwork platings.
 - [x] Fixed a luadoc.
 - [x] Improve outfits2md term output further
 - [x] Hide the entire secondary slot stuff when global var `hide_multi` is True.
 - Notice orange things in the description header like "core engines (secondary)" is not done in lua and therefore out of my reach. (**left as is**)
 - Trigger show_multi set to True in the right conditions. -> that would trigger the tutorial, that does not exist, yet. I will wait for that.

> It would simply be to have the descriptions (descextra) read some global variable var.peek("player_multicore") or something.

Done.
> The variable could be set on updating save games + on a tutorial event when the player gets a ship that meets the criteria.

I will wait until the tutorial event is ready before doing that.
## Testing Done
Ticked items are tested.


dat/outfits/core_hull/medium/nanofoam_padding.xml
```diff
@@ -0,0 +1,18 @@
+<outfit name="Nanofoam Padding">
+ <general>
+  <shortname>Nanofoam Padding</shortname>
+  <typename>Core Systems (Hull)</typename>
+  <slot prop_extra="hull_secondary">structure</slot>
```
bobbens, line 5:
```suggestion
  <slot prop="hull_secondary">structure</slot>
```
It ignores prop_extra if prop is not set.

src/outfit.c
```diff
@@ -383,7 +383,7 @@ int outfit_filterStructure( const Outfit *o )
 
 int outfit_filterCore( const Outfit *o )
 {
-   return sp_required( o->slot.spid );
+   return sp_exclusive( o->slot.spid );
```
bobbens, line 386: Might make sense to do something better here than use ugly heuristics. I can think of several solutions:
1. Add a `<core/>` tag to outfits and use that.
2. More generally, add a '<category>Cores</category>` property that would create new tabs as necessary. Much more work for fairly little gain, however, it could be used to create new types in the future, like `Special` for outfits that don't fit there.

dat/outfits/core_hull/medium/nanofoam_padding.xml
johann1764: OK! That should solve a certain number of problems. Thanks !

src/outfit.c
johann1764: Yes, but does Naev know what a core tag is, and should it test for that when selecting outfits ?

Most likely, the problem here is that that function is hard-coded, and therefore must rely on hard-coded criteria.

bobbens: > Yes, but does Naev know what a core tag is, and should it test for that when selecting outfits ?
> 
> Most likely, the problem here is that that function is hard-coded, and therefore must rely on hard-coded criteria.

No, it does not. So outfits would have to get added a flag to represent that, and the outfit.c loader would have to set that based on the tag. It would be ugly and hardcoded, which is why the general solution would be better, but probably overkill (especially for now).

johann1764: Yep, there's already a few things to do, and I already find it hard not to forget anything.

Probably best left as is, and we generalize this the day we need to.

johann1764:
I was thinking about doing that:

```diff --git a/src/nlua_pilotoutfit.c b/src/nlua_pilotoutfit.c
index 89ddecfa85..9db65b19bc 100644
--- a/src/nlua_pilotoutfit.c
+++ b/src/nlua_pilotoutfit.c
@@ -171,9 +171,7 @@ static int poL_id( lua_State *L )
  * @brief Gets the properties of the outfit slot.
  *
  *    @luatparam PilotOutfit po Pilot outfit to get the outfit of.
- *    @luareturn A table with slot properties string "size", string "type",
- * string "property", boolean "required", boolean "exclusive", and boolean
- * "locked". (Strings are English.)
+ *    @luareturn A table with slot properties {"size":str, "type":str, "property":str "required":bool, "exclusive":bool, "locked":bool} (str in English)
  * @luafunc slot
  */
 static int poL_slot( lua_State *L )
```

Because:
 - multiline `@luareturn` breaks the doc: https://naev.org/api/modules/pilotoutfit.html#slot
 - too much verbosity makes the doc harder to use in this case (I think)

Should I?

bobbens:
> I was thinking about doing that:
> 
> ```diff
> index 89ddecfa85..9db65b19bc 100644
> --- a/src/nlua_pilotoutfit.c
> +++ b/src/nlua_pilotoutfit.c
> @@ -171,9 +171,7 @@ static int poL_id( lua_State *L )
>   * @brief Gets the properties of the outfit slot.
>   *
>   *    @luatparam PilotOutfit po Pilot outfit to get the outfit of.
> - *    @luareturn A table with slot properties string "size", string "type",
> - * string "property", boolean "required", boolean "exclusive", and boolean
> - * "locked". (Strings are English.)
> + *    @luareturn A table with slot properties {"size":str, "type":str, "property":str "required":bool, "exclusive":bool, "locked":bool} (str in English)
>   * @luafunc slot
>   */
>  static int poL_slot( lua_State *L )
> ```
> 
> Because:
> 
>     * multiline `@luareturn` breaks the doc: https://naev.org/api/modules/pilotoutfit.html#slot
> 
>     * too much verbosity makes the doc harder to use in this case (my opinion)
> 
> 
> Should I?

It's always felt really awkward to return tables with ldoc. It seems like the accepted way is to create a named table and pass that, which is quite awkward (see https://github.com/lunarmodules/LDoc/issues/86 ). I'm not too picky, but if you add types, you should use the Lua ones for consistency. So String instead of string.

johann1764:
Ok !

johann1764:
### For the next step (after this PR)

I'd like to try to build a picture for the Nanofoam Padding armor. I am thinking a raw block similar to glass wool, but of lighter aspect ( a texture similar to cotton candy ), and _maybe_ more foamy than cotton-like. And in a much darker color. Keeping that aspect on a very reduced picture looks like quite a challenge, but I can try.

Now the question is, even if I can do that, where would I upload this in order to have it in the arwork dir of naev ?
How can I just add a symlink to an existing picture that I could overwrite with the new picture ?

bobbens:
> Now the question is, even if I can do that, where would I upload this in order to have it in the arwork dir of naev ? How can I just add a symlink to an existing picture that I could overwrite with the new picture ?

So the way it works, is you need to:
1. PR the new image to the `naev-artwork-production` repo ([link](https://github.com/naev/naev-artwork-production))
2. Get it merged
3. Have main update the artwork to sync with the repo (it is a submodule)
4. Use the artwork
5. PR the raw base data (3d model, etc...) to `naev-artwork` repo ([link](https://github.com/naev/naev-artwork))

johann1764:
Ok!

So there is no automatic generation of naev-artwork-production from naev-artwork ?

bobbens:
> So there is no automatic generation of `naev-artwork-production` from `naev-artwork` repo ?

While it would be ideal, Blender in particular is really really flakey when it comes to scripting and exporting, so I don't think it would be possibly to reliably and robustly export it all. Also, quite a few assets are sadly missing sources.

johann1764:
Question: All hulls have same exact camera/light position and armor orientation.
I did not find this information anywhere, and I did not find a way to have blender tell me form the blend files. I did not find a way to export to povray (that would have given me the information)

Would you happen to know where can I get this information ?

johann1764:
Anyway, this artwork will be a background task for me, so let us merge this.

dat/outfits/core_hull/medium/krain_nanotubes_foam_padding.xml
```diff
@@ -0,0 +1,21 @@
+<outfit name="Krain Nanotubes Foam Padding">
```
bobbens, line 1: Might make more sense to just call it "Krain Nanofoam Padding" and have Nanotubes appear in the description.

dat/outfits/lib/multicore.lua
```diff
@@ -201,16 +205,15 @@ function multicore.init( params )
...
+         if slot_prop:match" %(Secondary%)$" then
+            msg = _('This outfit can only be equipped as #ysecondary#b.')
          else
-            local _slot_name, _slot_size, slot_prop = o:slot()
-            if se ~= slot_prop .. ' (Secondary)' then
-               desc = desc .. '\n#b'.._('This outfit can only be equipped as #ysecondary#b.')..'#0'
-            end
+            msg = _('This outfit can only be equipped as #rprimary#b.')
          end
+         desc = desc .. '\n#b' .. msg .. '#0'
```
bobbens, line 216: Might make sense to do something like:

```lua
local msg_slot
if slot_prop:match" %(Secondary%)$" then
   msg_slot = "#y".._(secondary).."#b"
else
   msg_slot = "#r".._(primary).."#b"
end
local msg = fmt.f(_("This outfit can only be equipped as {slot}."), {slot=msg_slot})
desc = desc .. '\n#b' .. msg .. '#0'
```

johann1764: ok!

dat/outfits/core_hull/medium/krain_nanotubes_foam_padding.xml
johann1764: Plus the renaming in `naevpedia/meson.build` and `po/INFILES.POT`

dat/outfits/core_hull/medium/krain_nanotubes_foam_padding.xml
```diff
@@ -0,0 +1,21 @@
...
+ <general>
+  <shortname>K. Nanofoam Padding</shortname>
+  <typename>Core Systems (Hull)</typename>
+  <slot prop="hull_secondary">structure</slot>
+  <size>medium</size>
+  <price>440000</price>
+  <description>Despite its ridiculously low mass and its cloudy texture, this foam is extremely elastic and hard to tear, making it an excellent underlayer for a more rigid armor, making it perfect for ships that can't afford the weight of a regular battle armor underlayer. It comes at a cost, though.</description>
```
bobbens, line 8:
```suggestion
  <description>Instead of relying on more traditional armour alloys, the Nanofoam Padding uses tessellated nanotube structures to provide resilience against weapon fire. As the nanotubes lack any sort of supporting structure, it becomes necessary to attach them to another hull as an underlayer, where they will provide reactive support upon impact. The foam is extremely elastic and hard to tear, making it perfect for ships that can't afford the weight of a regular battle armour underlayer. It does come at a cost, though.</description>
```

If it's an underlayer, I'm not sure the texture matters too much.

bobbens (commented):
Modified the description a bit, was thinking about it being Krain which is a bit odd, however, I guess it's the best fit as it is a "half-slot".

johann1764:
> Modified the description a bit, was thinking about it being Krain which is a bit odd, however, I guess it's the best fit as it is a "half-slot".

That is not that odd. After all, Krain designed engines for their half-sized ships (of size "4.5"  and "3.5"). This item allows:
 - a more comfortable mass situation for Krain ships (that were really too short on engine_limit)
 - the possibility to use Krain engines on non-Krain ships (moving this setting from non-viable to not great)

So that would make sense for Krain to develop this product.

dat/outfits/core_hull/medium/krain_nanotubes_foam_padding.xml
johann1764: I wanted to emphasize how strange the material is, and how it can be a useful addition to a ship's protection while its ultra light weight and lack of a rigid surface.

I am happy with this description overall. Captures the ideas, and expresses them using true sentences. Maybe a small detail here:
'they will provide reactive support upon impact'. This is the justification for still being able to give a not-so-bad damage abs. damage. It misses another aspect: the tanking ability (anything going through the pri armor will damage this underlayer first, and it requires a lot of energy to destroy its cohesion).

But as it is kind of implied by the 'hard to tear' part, maybe that's clear.

johann1764: That's detailed in the part before, so ok.

bobbens:
```suggestion
  <description>Developed by Krain Industries as a complement to their engine lines, instead of relying on more traditional armour alloys, the Nanofoam Padding uses irregular tessellated nanotube structures to provide resilience against weapon fire. As the nanotubes lack any sort of supporting structure, it becomes necessary to attach them to another hull as an underlayer, where they will provide reactive support upon impact. The foam is extremely elastic and significantly increases the energy needed for ship's plating to lose its cohesion while being extremely lightweight, making it perfect for vessels that can't afford the weight of a regular battle armour underlayer. It does come at a hefty price tag, though.</description>
```

johann1764: Damnit, I just committed previous version.

I will manually commit this.

No need to rebase, this time, this is not worth it.

johann1764:
oops, let me a minute to commit your last suggestion

johann1764:
Oh. `Authored-by` does not work. Sorry !

bobbens:
> Oh. `Authored-by` does not work as I expected. Sorry !
> 
> EDIT: I should have: `git commit --author="..."` instead. Now I know it.

No worries. Looks good, let me merge. Thanks!

PR #2859, opened 2025-05-19
fix pirate starbridge def eq. + dbg code for equipopt

johann1764:
**Bug Fix**
Bug: pirate Starbridge safety check failures.

## Summary

Added default cores for pirate Starbridge secondary slots.

Left basic debugging code commented out in equipopt/cores as it will probably be used again.

## Testing Done
Yep


bobbens:
Looks good, thanks!

Issue #2862, opened 2025-05-21
[Proposal]: Star map geometry


johann1764:
### Suggestion

Since the map rework, I find the map to be misleading. Don't get me wrong: this rework is a very good thing. However, the geometry of the star map does not give a good intuition of the topology.

For example, the shortest path is often a surprise. I like surprises, but a good surprise is a surprise that occurs in a usually predictable setting. If it is a surprise each time, it is just chaos.

As an example, the trade route from Beeklo to Doranthex looks short on the map, but that's usually a long path to go. For example, using the straight road from Beeklo to Zacron is just as long as taking going through Gremlin and Ogat. The geometry of the map does really not suggest that.

An other visible effect of the problem is the big stretching of some edges while others are just tiny.

I suggest replacing that:

![Image](https://github.com/user-attachments/assets/c000dcd5-929b-4d51-8560-cfe5546e2b77)

with that:

![Image](https://github.com/user-attachments/assets/5fc8c060-c2ae-4829-9ae1-f1981a9c8626)

Note the grayed, dotted edges are "virtual edges" that are here only to enforce the desired geometry.

This can be done without any change inside the systems, except a rotation of the whole system, a flipping in a very few cases, and a _slight_ change in the orientation of some jump points. (I will develop on that)
However, empty systems with only autopos jump points should "follow" (stay autopos).

To do that, I just need some new flag (called reference for example) for jump points.

Note: you guessed it, I did not do that by hand. This will also give us an automatic system positioning feature, that we can use whenever we want.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bobbens:
I agree that some can be made better. However, a fully automatic procedure loses a lot of things that we don't want:
1. Wild space no longer is a spiral
2. No circle around the Anubis Black Hole
3. Some of the minor systems are not close to their major ones
4. The Qorel tunnel is not really a tunnel
5. Probably have to redo the map image placements

So the solution would be to either do this automatic, and then repair a lot of things (so it would be done offline), or alternatively, just do it entirely manually. I don't think a fully automatic (online) solution is something we are really interested in this case.

Most of the stuff is somewhat an artefact of me remaking the universe by squishing more stuff in there so Sol was more or less at a central position and it felt like it blew up in the middle.

johann1764:
> I agree that some can be made better. However, a fully automatic procedure loses a lot of things that we don't want:
> 
> 1. Wild space no longer is a spiral

[...]

PR #2865, opened 2025-05-27
star map geometry - preliminary

johann1764:
**Improvement**

This PR prepares the addressing of the feature described in #2862.

This proposes the scripts that will be used to change the star map.
No map modification is applied here. The results are those of current scripts in their current state, and are shown for proof of concept. The only goal is to convince it is possible to rework the map that way. Actual application of the process will be the subject of another PR.

This also develops the sys_{freeze, unfreeze, relax, empty} scripts that will be useful in the next steps. (not only for this rework)

All these are in `utils/sys/`.

A few other tasks were performed, listed in the 'other' section below.

#### ssys remap tools
These are designed to change the star map by **changing systems position**.

 - `process_sys.sh` The main. Currently used for tests. Performs several steps:
    - Calls `sys2dot.py` to extract systems information and turn it into a dot file. At this point, the **pre-processing** occurs: some invisible edges are added to enforce desired properties.
    - Calls `neato` to compute the graph layout.
    - Calls `dot2sys.py` to extract layout information for the graph and modify ssys accordingly. At this point, the **post-processing** occurs: some geometrical transformations are applied : compute the wild space layout, enforce co-circularity of some points around Anubis BH, rotate some parts, get twin systems closer, etc.

#### ssys process tools

These are designed to manage the **internal geometry of systems**, that might get affected by the position changes. (because autojumps move when the systems move)

 - `sys_freeze.py` : fixes the auto jumps position and mark them as previously auto.
 - `sys_relax.py` : rotates a frozen system for minimizing orientation stretch.
 - `sys_unfreeze.sh` : as the name suggests.
 - `sys_empty.py` : gives the list of systems that can be considered empty.

#### ssys lib
 - `geometry.py` as the name suggests. Most is really classical:
    - vec for (2d-)vectors. Support for every usual composition with float. `+` and `-` are what you expect, `*` is what you expect for floats and vectors (dot product). `/` is what you expect for floats. See the source for more functionalities.
    - transformations. Obtained by **dividing** vectors: `v1/v2` is the transformation that turns `v2` into `v1`. As expected: `k*v2*(v1/v2) = k*v1`. So multiplying vectors with transf. gives the result of the transformation applied to the vector. Notably, when `v1` and `v2` are normalized (or just have the same size), `v1/v2` is the rotation that turns `v2` into `v1`.
    - bb for bounding box. use += to enlarge the bb with a new element, and `in` to test if inside. Nothing fancy.
 - `ssys.py` everything else.

### Other
 - The useless deprecated xml header for ssys and spob has been removed. This will happen anyway as a written xml files using elementTree will remove it. A **lot** of lines !
 - `repair_xml.sh` repairs the changes caused by elementTree (missing final endline, added spaces)
 - added `.swp` and `.swo` to `.gitignore`
 - ssys naming inconsistencies:
   - examples:
      - Yetmer-O'Rez Highspace : `yetmer-o'rez_highspace.xml`
      - Za'lek : `zalek.xml`
      - An'ku : `anku.xml`
      - NGC-9654 : `ngc9654.xml`
      - C-14 : `c14.xml`
      - Scholz's Star : `scholzs_star.xml`
   - The fast solution would have been renaming `yetmer-o'rez_highspace.xml` into `yetmerorez_highspace.xml`. But that does not look good. I decided:
      - The `'` is kept, except when used for `'s`. Therefore : `za'lek.xml` and `an'ku.xml`.
      - The `-` is kept, except when used after a capital `C`. So everything stays as is.
      - The `NGCG-xxx` system has been renamed `NGC-xxx`.

## Todo

 - [x] checklist below
 - [x] Maybe increase terminal ngc edge len
 - [x] Give a position to isolated systems - part 2: Zied
 - [x] Offset the whole map to get better alignment with background picture.
 - [x] Soromid NE looks terrible
 - [x] Improve spiral a bit - make it better integrated.
 - [x] sys_freeze
 - [x] sys_empty
 - [x] sys_relax
 - [x] sys_unfreeze
 - [x] sys_{freeze,relax,unfreeze,empty} documentation
 - [x] bend the Thurion space and squeeze down wild space
 - [x] repair Soromid NE

### Later

 - repair Za'lek space
 - connect Carrza to Tepvin and fill the hole below with everything in between
 - Reposition decorators
 - Give a position to isolated systems - part 1: virtual systems (Gauntlet & Sirius tests)
 - Inflate ngc distances ?
 - Fix the curves in the north, notably the one going to Jommel. Maybe merge or reorganize Andres & Kelvos ?

### Checklist

This is copied from https://github.com/naev/naev/issues/2862#issuecomment-2900201532.

 - [x] Wild space no longer is a spiral
 - [x] No circle around the Anubis Black Hole
 - [x] Some of the minor systems are not close to their major ones
 (Kruger's pocket is supposed to be next to Kruger, it looks closer to Cebus or Scorpion. Gliese Major and Gliese Minor are also a bit further apart then what may be nice.)
 - [x] The trade lanes are designed to be fairly smooth, as in the angles do not change too drastically. In the new version, places like Maus - Millstone - Fidelity are really jagged.
-> Blame the topology, not the geometry. This was greatly mitigated since original proposition.
 - [x] The Soromid space is designed to be based on several parallel curves with the NGC systems on the outter, and another on the inner.
 - Similar with the curve going to Jommel. -> Moved to TODO list above.
 - Sol is meant to be a bit closer to the center (so that the trade lanes imply that they head back towards there). Pushing it towards the center is one of the reasons the map is a bit weird and forced, and the new approach is pushing it off to the side again, which is something non-ideal. -> Blame the topology, not the geometry.
 - [x] Sirius space is also made around a circle with Suna in the middle, which is also lost in the transformation. -> Better now.
 - [x] The south and north winds are also supposed to be smooth (can't really tell now) like the trade lanes, and are actually supposed to be able to be connected. I have plans for a small Soromid campaign which would end up with the north and south solar winds connecting together, but it should be made somewhat smooth to do that.
 - Wild Ones space is based around a loop, starting at Point Zero - Father's pride, and looping around to Cygni (hidden jumps count).
-> That one looked **so** artificial on the original map. I am not sure we really want this pattern to be visible. Anyway, it is barely broken, just enough to make it less obvious.
 - [x] NGC systems should be on the outside like NGC-8338 which is inside of Marat.
 - [x] The area around Porro (Sirius space) also seems to be handled really poorly compared to the original map.
 - [x] Probably would need to visualize the nebula to see if it is consistent. The systems have volatility and interference values somewhat designed to grow as you leave Sol (although with exceptions and noise). -> So far, does not look too bad.

### Current state
<details><summary>Before:</summary>

before
</details>

<details><summary>After:</summary>

final
</details>

<details><summary>Before (starmap)</summary>

shot-2025-05-28_14-03-10
</details>

<details><summary>After (starmap)</summary>
Notice the artifacts due to renamed system NGCG-xxx -> NGC-xxx and `zalek.xml` -> `za'lek.xml` and `anku.xml` -> `an'ku.xml`.

shot-2025-05-30_10-49-25
</details>

## Testing Done
Yep. That's much better. The shortest path does not come as a surprise each time anymore.


bobbens:
Looks much better thanks!

Looking at the Nebula area:
1. What would be the minimum topology edits necessary to move sol closer to the center? Mainly would want to bend soromid space so it's more directly on top of Sol instead of off to the top-left.
2. Would it be possible to bend the Thurion space, so it's left and more stuck to the systems (or alternative righty), and then squeeze down wild space? Wild space is meant to be much closer to the center of the Nebula. I guess another alternative would be to, instead of using a one-way hidden jump for Thurion space, use a wormhole and put their nebula free-space elsewhere.

Some other comments:
1. Would it be possible to make area around anubis black hole more circular?
2. I really like how wild space + Mizar + Collective area turned out.

johann1764:
There only remains `sys_relax.py` to do. Everything else that was planned for this PR is done.

The initial post has been updated, and now describes with full details the PR contents.

A large cleanup have been done, and now scripts (except `sys_relax.py`) have reached their stable form (except, of course, the specific details of pre- and post- processing, that will be refined in a future PR).

You can start the review whenever you see fit, only `sys_relax` will change in the future.
I'll mark this PR as "ready for review" when this last script is done (which shouldn't take long. The long part is the night sleep before it.)

bobbens:
OK, looks better. Some nitpicking:
1. Herculis is a Nebula system but is now very far away.
2. It seems like it should be able to rotate Soromid space more towards Sol given the few connections it has. Does this not happen because of Thurion space or something? Father's Pride - Hatter connection seems like it would benefit from the rotation, and it would also bring wild space closer to Sol.
3. Terminus - Doowa connection runs straight through Gamma Polaris, not sure if this can be misleading since it's hard to distinguish that they are connected to Gamma Polaris (Effetey has a similar issue).
4. Sol Area looks a bit regular when it should be a bit more of a mess. Maybe topology should also be changed there.
5. NGC-22375 might look better if the connection from Toxin crosses over the hidden jump of Euler and NGC-14337
6. Ekoddu might look better flipped on the top side.
7. The Kraft-Sorrow connection does a 90 degree turn with a trade lane, which would be nice to have a bit smoother, but it looks hard to do. All these "dead end" fading out trade lanes should ideally be pointing to Sol (they were mapped out so they seem to all head towards Sol)
8. Although it's still quite a WIP, might be nice to push Proteron space closer to the Nebula area (thus Sol). Although I guess this can wait until someone works on Proteron.

johann1764:
All that looks fixable, among other things I would also want to fix.

At this point, I have frozen this PR contents, I will focus on purely technical aspects : bugs, code quality, typos, etc. I had sworn myself I would not do another PR over 500 changed lines, and I failed miserably. Furthermore, renamed systems causes a lot of warnings as they are present in the image I use but not in my user data.

The goal was only proving that path could work, clarify the process and remove a few obstacles on the way. I consider this mission is complete, and now I am ready for next PR.

In order to make the review easier, I suggest filtering out the `.xml` diffs as they are described in the Summary.

johann1764:
Shall we merge ?

bobbens:
Sure, just let me confirm one thing. This is more of a scaffolding PR, right? Since all the system positions seem unchanged still.

Another thing on removing the XML headers, would have to double check, but I think if you save with the editor, it'll add headers to all the systems again.

johann1764:
> Sure, just let me confirm one thing. This is more of a scaffolding PR, right? Since all the system positions seem unchanged still.

Absolutely.

> Another thing on removing the XML headers, would have to double check, but I think if you save with the editor, it'll add headers to all the systems again.

Even so, it will serve its purpose: not having these headers all over the diffs when doing the final transformation. But that would imply no one wants uses it and merge until then.

bobbens:
Thanks! Merged.

Regarding the XML header apparently (from https://stackoverflow.com/questions/7007427/does-a-valid-xml-file-require-an-xml-declaration ):
> In XML 1.0, the XML Declaration is optional. See [section 2.8 of the XML 1.0 Recommendation](http://www.w3.org/TR/xml/#sec-prolog-dtd), where it says it "should" be used -- which means it is recommended, but not mandatory. In XML 1.1, however, the declaration is mandatory. See [section 2.8 of the XML 1.1 Recommendation](http://www.w3.org/TR/xml11/#sec-prolog-dtd), where it says "MUST" be used. It even goes on to state that if the declaration is absent, that automatically implies the document is an XML 1.0 document.

And it seems like everyone recommends using XML 1.0 instead of 1.1, so we can live without them. However, it would be good to see if we can make it so libxml doesn't save the headers when outputting xml.

johann1764:
> Thanks! Merged.
> 
> Regarding the XML header apparently (from https://stackoverflow.com/questions/7007427/does-a-valid-xml-file-require-an-xml-declaration ):
> 
> > In XML 1.0, the XML Declaration is optional. See [section 2.8 of the XML 1.0 Recommendation](http://www.w3.org/TR/xml/#sec-prolog-dtd), where it says it "should" be used -- which means it is recommended, but not mandatory. In XML 1.1, however, the declaration is mandatory. See [section 2.8 of the XML 1.1 Recommendation](http://www.w3.org/TR/xml11/#sec-prolog-dtd), where it says "MUST" be used. It even goes on to state that if the declaration is absent, that automatically implies the document is an XML 1.0 document.
> 
> And it seems like everyone recommends using XML 1.0 instead of 1.1, so we can live without them. However, it would be good to see if we can make it so libxml doesn't save the headers when outputting xml.

I had made a quick search about it, but did not find it. But if you agree on that, if we can ask libxml to stop generating these, we should just do that and remove them all, maybe it is worth doing further investigation on that.

bobbens:
Regarding the XML headers, I tried to see if there was a way to get libxml2 to not generate the headers, but I wasn't able to get anything working, it seems fairly built in (passing NULL as header information just falls back to defaults). If that's not able to be solved, I think it would probably be best to re-add the headers or every time you use the editor it's pretty much impossible to see any diff or anything, and it's pretty easy to lose the changes you did if you start trying to muck with restoring files.

johann1764:
Have you tried this:
https://stackoverflow.com/questions/25410366/libxml2-avoid-xml-version

I wanted to find a nice, clean solution but, since there is none, this hack looks reasonable (if I understand well)

johann1764:
There's an experiment I wanted to make that requires using this lib. I can do my experiment and try the hack while at it, if you want. But in that case, it won't be today.

johann1764:
Damn it. I am really not fond of this lib, looks like doing things manually would take less lines of codes than just their header does. I will do my experiment in another way.

It would probably be faster to let you try implement the hack in Naev, but if you don't feel like it (which I may understand), let me know and I will do it.

BTW, I noticed that the modifications to `*_xml_header.sh` that `shellcheck` suggested and that I applied actually broke the scripts. I repaired them (in a way that make shellcheck happy), which allows temporarily to do:
```shell
./utils/rem_xml_header.sh dat/ssys dat/spob
```
each time you save in the editor, which allows not re-introducing the headers.

I will open a PR for that (and current star map progress).

johann1764:
adding `./utils/rem_xml_header.sh dat/ssys dat/spob` to pre-commit is definitely the easiest way, I vote for it. It is not so fast, as it has to grep all spobs, which takes time. I don't think we can improve the script further, I have already tried different alternatives that were all slower.

If this solution finally gets painful in the future, let me know.

johann1764:
Ok, completed the pre-commit file in the PR, see:
https://github.com/naev/naev/pull/2872#issuecomment-2953730128

By the way, this PR has already reached a reasonable size (and its contents a have reached a kind of maturity), so reviewing it and merging can be done now.

I did not test pre-commit (I can't) and did not apply xml headers cleanup for PR diff clarity.

PR #2871, opened 2025-06-04
Star map - continued

johann1764:
**Bug Fix** / **New Feature**

This PR addresses part of the feature described in #2862.

## Summary
TODO

## Testing Done
TODO

### DIFF
--- a/utils/sys/dot2sys.py
+++ b/utils/sys/dot2sys.py
@@ -5,7 +5,7 @@ if __name__ != '__main__':
 
 
 from sys import stdin, stderr
-from ssys import vec, fil_ET, ssys_fil, sysnam2sys, sysneigh, vec_from_element, vec_to_element
+from ssys import vec, fil_ET, ssys_fil, sysneigh, vec_from_element, vec_to_element
 from geometry import bb
 
 
--- a/utils/sys/ssys.py
+++ b/utils/sys/ssys.py
@@ -1,11 +1,18 @@
 # python3
 
 
+import os
+import sys
 import xml.etree.ElementTree as ET
 from sys import argv, stderr
-import os
-script_dir = os.path.dirname(__file__)
+
 getpath = lambda *x: os.path.realpath(os.path.join(*x))
+
+script_dir = os.path.dirname(__file__)
+sys.path.append(getpath(script_dir, '..'))
+
+from xml_name import xml_name as nam2base
+
 PATH = getpath(script_dir, '..', '..', 'dat')
 
 from geometry import vec
@@ -50,18 +57,6 @@ class starmap(dict):
             stderr.write('no position defined in "' + name + '"\n')
             self[key] = None
       return self[key]
-
-def nam2base( nam ):
-   nam = nam.strip()
-   for t in [(' ', '_'), ("'", ''), ('&', ''), ('.','')]:
-      nam = nam.replace(*t)
-   s = nam.split('-')
-   acc, s = s[0], s[1:]
-   for sub in s:
-      if not ( (ord(sub[0]) <= ord('9')) and (ord(sub[0]) >= ord('0')) ):
-         acc += '-'
-      acc += sub
-   return acc.lower()
 
 def sysneigh(sys):
    T = ET.parse(ssys_fil(sys)).getroot()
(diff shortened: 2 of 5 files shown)


bobbens:
One issue with renaming trails is that all the GLTF files (in naev-artwork-production) contain trail information and will have to be updated. Furthermore, all the blender files (in https://github.com/naev/blends_tng ) will have to be updated, and I don't think that can be done automatically

johann1764:
I think I am done.

bobbens:
Thanks! One issues with the renaming of spfx is that they have to be updated for the outfits. In particular, `spfx_armour` and `spfx_shield` refer to the spfx. Without that it errors.

Some comments for the next step:
1. Maybe it would make sense to penalize having a lane go through a system more than edges to solve the gamma polaris crossing.
2. It probably would be a good idea to separately keep a list of proposed edits to improve the topology as it's probably a good opportunity to do so, at least focusing on the less important systems (so no touching Goddard and the likes). I can go with some ideas first:
   1. Maybe add a system between Attaria and Ganth to straighten the patrol lane in Za'le space (remove first curve).
   2. Change Kaas-Diadem jump to Kaas-Vedalus and see what jump can be changed with Mazer to try to rotate Soromid space towards Sol. Moving Cygni (or removing it) might help too.
   3. Maybe some change in Tau Ceti connectivity could help Sol move more towards the center.
   4. Add a system near Ryder (5 jumps!) to try to open the area up and have Sirius space wrap a bit more around Sol.

johann1764:
> One issues with the renaming of spfx is that they have to be updated for the outfits. In particular, spfx_armour and spfx_shield refer to the spfx.

Are there non-updated outfits ? If so, there's a bug in `apply_renaming_suggestion.sh`.

> Some comments for the next step:
> 
> 1. Maybe it would make sense to penalize having a lane go through a system more than edges to solve the gamma polaris crossing.

`neato` consider that as an edge crossing all the same. I am afraid we have to fix it manually.

> 2. It probably would be a good idea to separately keep a list of proposed edits to improve the topology as it's probably a good opportunity to do so, at least focusing on the less important systems (so no touching Goddard and the likes).

I kind of did that here: https://github.com/naev/naev/issues/2862#issuecomment-2925721827, I will copy your propositions there.

> I can go with some ideas first:
>    
>    1. Maybe add a system between Attaria and Ganth to straighten the patrol lane in Za'lek space (remove first curve).

That might work.

>    2. Change Kaas-Diadem jump to Kaas-Vedalus and see what jump can be changed with Mazer to try to rotate Soromid space towards Sol. Moving Cygni (or removing it) might help too.

That may help. I have done a few experiments, we won't get away with just one topological change. I will try doing it with two, here is one more candidate.

>    3. Maybe some change in Tau Ceti connectivity could help Sol move more towards the center.

Like flip Tepvin -- Tau Ceti to Tepvin -- Banard's Star.

>    4. Add a system near Ryder (5 jumps!) to try to open the area up and have Sirius space wrap a bit more around Sol.

Not too sure I understand what you suggest to do, here.

bobbens:
> > One issues with the renaming of spfx is that they have to be updated for the outfits. In particular, spfx_armour and spfx_shield refer to the spfx.
> 
> Are there non-updated outfits ? If so, there's a bug in `apply_renaming_suggestion.sh`.

Only the Chakra outfits are updated, all the other weapons (Plasma, Laser, etc...) are not.

> > 4. Add a system near Ryder (5 jumps!) to try to open the area up and have Sirius space wrap a bit more around Sol.
> 
> Not too sure I understand what you suggest to do, here.

The idea is to try to have sirius wrap a bit more around the nebula, Ryder has 5 jumps, so  I think it's probably the limiting topological factor, but I guess it will need testing.

johann1764:
I don't understand. I did:

```shell
> grep -r PlaM dat/
dat/spfx/plam2.xml:<spfx name="PlaM2">
dat/outfits/bioship/generate.py:    'spfx_armour':   'PlaM',
```

The first one stays as is, as it did not match the pattern 'name="...M"' (so my pattern was a bit too restrictive).

The second one is logical, it is in a python file.

There is no other one, but maybe I need to update this repository to take recent changes into account.

I will therefore :
- [ ] sync with main
- [ ] repair the python file
- [ ] check for new files requiring updates

johann1764:
Now it should be good.

bobbens:
Fixed a few explosion spfx calls in the C code, but other than that looks good, thanks!

PR #2872, opened 2025-06-07
Star map - an update

johann1764:
**Bug Fix** / **New Feature**

This PR addresses part of the features described in #2862.

## Summary
See commits description.


johann1764:
## An experiment

The gravity potential of current map. Can be computed in "reasonable" time.

pot

johann1764:
Added this to pre-commit: `./utils/rem_xml_head.sh dat/spob/*.xml`. Notice using `./utils/rem_xml_head.sh -r dat/spob/` instead is around 8% faster.

I also did it for `['^dat/ssys/*.xml', '^dat/spob/*.xml', '^dat/spob_virtual/*.xml']`.
We could apply it for all dat dir (just call `./utils/rem_xml_head.sh` with no arg).

johann1764:
### Gravity - an experiment
<details><summary>unfold</summary>

|before|after|
|---|---|
|before|after|
</details>

johann1764:
I am trying to compile a recent version of `clang` (as the version that I have does not understand all the config file). To do so, I need a more recent version of C make. It required a few things, but now it should finally compile. It takes an eternity. If you're patient enough, it might work tomorrow morning, so I should be able to finally clang this file.

PR #2873, opened 2025-06-09
Fix a bug that the circle shaped radar shows asteroids outside the scope.

oo13:
**Bug Fix**

## Summary
The circle shaped radar in the legacy GUI shows asteroids outside the scope. This PR fixes that.

## Testing Done
Use the lagacy GUI in scalefactor 1.0 and 2.0.


bobbens:
Looks good, thanks!

PR #2874, opened 2025-06-12
Star map - continued

johann1764:
**New Features**

This PR continues the work described in #2862.

It completes #2872 on gravity by applying `clang-format` to `gravity.c`, introducing the `-s` option, and improving the related scripts `gen_pot.sh` and `apply_g.sh`.

It completes #2865 and #2872 on geometry by cleaning up `geometry.py`, providing a full implementation of `bounded_circles`, and providing a testing output (when using `geometry.py` as main). The `bounded_circles` function relies on the more general `interstices` function, that allows to find any position where you could fit in a circle inside an arbitrary set of line segments and points. That seemingly simple problem proved to be challenging. This was the last algorithmic building block needed for the future ssys auto-positioning feature.

The testing output of `bounded_circles` looks like this:
geometry

The first line is three random polygons, notice the second is crossed and the third one is also slightly crossed.

The second line illustrates notable features:
 - the first one has almost every possible circle bounding configuration : line-line-line, line-line-point, line-point-point. Also notice the interior is distinguished from exterior (that is why there is a hole without a bounded circle: it is because it does not belong to the interior)
 - the second one illustrates the rare point-point-point situation.
 - the third one illustrates the fact that the 5 generated circles could be recognized as "the same", therefore only one circle appears in `interstices` output. (if we call the edges a, b, c, d, e, we can find one circle bounded by a,b,d, one by b, c, e, one by c, d, a, etc. that all coincide)

#### Other random examples:
geometry

geometry

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -121,7 +121,7 @@ IncludeIsMainRegex: (Test)?$
 IncludeIsMainSourceRegex: ""
 IndentAccessModifiers: false
 IndentCaseBlocks: false
-IndentCaseLabels: false
+IndentCaseLabels: true
 IndentExternBlock: AfterExternBlock
 IndentGotoLabels: true
 IndentPPDirectives: None
@@ -180,7 +180,7 @@ ShortNamespaceLines: 1
 SortIncludes: CaseSensitive
 SortJavaStaticImport: Before
 SortUsingDeclarations: LexicographicNumeric
-SpaceAfterCStyleCast: false
+SpaceAfterCStyleCast: true
 SpaceAfterLogicalNot: false
 SpaceAfterTemplateKeyword: true
 SpaceAroundPointerQualifiers: Default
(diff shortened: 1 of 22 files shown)


bobbens:
OK. I would prefer not to have the clang-format change for now, as it's probably going to mess things up with the Rust merge, which should be soon. I'm being a bit conservative and want to have the GLTF models in Rust so we can have similar load times as now.

 In particular, the clang-format change causes pre-commit to try to do:
```
 378 files changed, 62301 insertions(+), 63259 deletions(-)
 ```
which is probably going to cause issues with the merge.

johann1764:
From https://clang.llvm.org/docs/ClangFormatStyleOptions.html:

> When using -style=file, clang-format for each input file will try to find the .clang-format file located in the closest parent directory of the input file. When the standard input is used, the search is started from the current directory.

I do that.

johann1764:
Looks like it works, at least when invoking clang-format manually from root dir.

Normally, when you apply the pre-commit that triggers clang-format, this should result in no diff. Let me know if it is not the case.

PR #2875, opened 2025-06-12
add the viewport frame in the radar and overlay.

oo13:
**New Feature**

## Summary
This PR introduces the frame of the viewport in the overlay map and the radar. It reduces the confusion of the ship positions in the overlay map with the center of the viewport, hopefully.

2025-06-12-20:27:19 FTM

## Testing Done
- scale factor: 1.0, 1.5 and 2.0
- overlay and radar in silm and legacy.


bobbens:
That's a really cool idea. Thanks!
Not sure if it should be default on, but we can leave it as is for now.

Issue #2882, opened 2025-06-18
Can't gather space commandos in slow ship - problem with objective in "Dvaered Negotiation 1" mission


postscripter8:
In mission 'Dvaered Negotiation 1' there is an objective to hail a Koala ship for refueling and gather a space commando cargo it spawn after a few seconds. The commandos are ejected from the Koala with it's full velocity, which can be unreachable with some slower ships. It's also counter-intuitive that a Commandos marker appears in a fixed place on system map (at the ejection site) while the actual gatherable is running away in a straight line. I have replayed this section a couple times in a Pirate Kestrel Yuri's Kiss (time constant 130% speed 92 km/s max speed 110 km/s) and sometimes I could barely keep up to gather them at the edge of the system map (at default zoom). In Sirius Divinity (time constant 150% speed 70 km/s max speed 81 km/s) it was not even possible to see the commandos spawn on screen, which further fuels the confusion about the fixed map marker with nothing to gather in the vicinity.

Gatherable spawning code is here:
https://github.com/naev/naev/blob/b9d36f91c3213bff75afc310e42bd60453327a9f/dat/missions/dvaered/recruitment/dv_goddard.lua#L395

There are two aspects of the problem:
1. The commandos can run away faster than current ship's max speed.
2. The commandos can be difficult to find if they spawn off screen, especially if the Koala was turning, or even if they just disappear from the view for a while.

Bullet 1. could be "solved" if it was considered a problem. Maybe it isn't. It's the player's mistake to equip wrong ship for the job after all. If it is a problem, maybe the commandos could gradually lose velocity. With even a miniscule speed damping the player would eventually catch up, if one could reliably follow them.
Bullet 2. could be solved with periodic updates of the marker position. That would immediately feed back the information to the player, that the commandos are moving and where they are. For bonus points the ship's sensor range could be considered to allow the commandos to get lost in the space and fail the mission that way.

postscripter8:
I just submitted a suggested resolution to the latter problem. There's probably some room for improvement, but now it's immediately apparent that Sirius Divinity will never make it to the cargo :-D

bobbens:
Yeah, thanks!
Probably need a proper gatherable API and add some braking mechanism to it (like disabled ships), but we can do that after Rust is merged.

PR #2883, opened 2025-06-18
Mission improvement - update moving cargo objective location in 'Dvaered Negotiation 1'

postscripter8:
**Bug Fix**

This PR addresses the bug/feature described in #2882

## Summary
In mission 'Dvaered Negotiation 1' the system map marker (point of interest) "Commandos" will now have it's location updated every 3 seconds to follow the actual location of the cargo.
I did not change behavior of the cargo or it's spawning parameters.
This is my crude attempt to remedy gatherable commandos fading away into space. (Bullet 2 in aforementioned #2882)

## Testing Done
I only playtested the relevant mission 'Dvaered Negotiation 1'.
Prior to the change the system map marker stayed at it's initial spawn point while the cargo run away. After the change the marker location got periodically updated and properly reflected actual location of the cargo.
I added troubleshooting log messages (absent in final commit) to make sure the new hook stops executing when the cargo is collected or when the mission is failed.


bobbens:
Looks good, thanks!

PR #2885, opened 2025-06-22
An autopos system for the star map

johann1764:
**New Feature**

This PR addresses part of the features described in #2862.

## Summary

I implemented an auto-position system. It is based on a potential (stress) minimization algorithm. It can find both the global and the local optimal position. This potential is defined with great simplicity, and can be extended very easily. In particular, I intend to add an edge-ssys overlap penalty in the future.

This is to be compared with `neato` unconfigurable (and unstable) behavior. The **core** of `neato` is 16562 lines long. By contrast, `reposition.c` takes a total 628 lines (4%).

It allows two main applications:
 - a post-processing for `neato` output, in order to improve layout of certain areas.
 - a tool for the editor. Three interesting aspects:
   - allows to _suggest_ planar embedding small changes (switch side of a ssys w.r.t. an edge)
   - allows to optimize the current position of a system (while keeping the exact same spacial relation between systems)
   - propagate a bit the effects of introducing a new system / moving an existing system.

As an example, here is the (generated) potential map of the Pontus system:
pontus
The green field correspond to ssys - ssys repulsion potential, while the magenta correspond to the jump stretch potential. Choosing the minimum of the weighted sum gives the selected blue dot.

For the purpose of illustration, this is the result of applying it to every system (which I do not intend to do: I would rather apply it to specific sets of ssys)
<details><summary>Illustration</summary>

map_fin

out
</details>
We can see the normalizing effect on edge lengths (which is often not desirable) with the spiral and Proteron space for example.

Notice the improved aspect of Sirius space (which was not great in `neato` output).

Note: by the way, it's fast.

## Side features
 - various tiny bugfixes
 - a few additional scripts
 - a few additional flag arguments for scripts
 - documentation !

### DIFF
--- a/extras/vimrc
+++ b/extras/vimrc
@@ -22,7 +22,9 @@ function NaevOptions()
    exec 'set softtabstop='.s:tabwidth
    autocmd BufNewFile,BufRead *.mvx set syntax=xml
    highlight RedundantSpacesOrOverLength ctermbg=darkred ctermfg=white
-   match RedundantSpacesOrOverLength /\(\s\+$\)\|\(\%85v.\+\)/
+   match RedundantSpaces /\s\+$/
+   set colorcolumn=85
+   highlight ColorColumn ctermbg=darkblue
 endfunction
 function NaevALEOptions()
    call ale#handlers#languagetool#DefineLinter('lua')
--- a/extras/vimrc
+++ b/extras/vimrc
@@ -21,7 +21,7 @@ function NaevOptions()
    exec 'set shiftwidth=' .s:tabwidth
    exec 'set softtabstop='.s:tabwidth
    autocmd BufNewFile,BufRead *.mvx set syntax=xml
-   highlight RedundantSpacesOrOverLength ctermbg=darkred ctermfg=white
+   highlight RedundantSpaces ctermbg=darkred ctermfg=white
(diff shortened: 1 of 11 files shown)


johann1764:
Spoiler: having a hightly regular map like above allows to make transformations like this visible:

out

If you want to appreciate the diff with the second picture aboce, you really need to compare them in-place.

This is the result of applying:
res

These transformations are purely experimental at this point.

bobbens:
> The next few steps are ready:
> 
>     * generalize gravity tools to any potential
> 
>     * scripts doc
> 
>     * wave potential
> 
>     * reposition - ssys-edge overlap penalty
> 
>     * direct computation of circumscribed
> 
>     * improved local search alg.
> 
>     * ... and much more.
> 
> 
> Let me know if you prefer them in a separate PR or appended to this one.

Either way works. I was out for a bit, but now I'm back. Let me merge this for now. Thanks!

> Note: I understood why the edge crosses a system. This is because it would cost 3 edge crossing to do it in the left and 2 edge crossing on the right. So `neato` considered it a lesser evil to have the edge cross directly under the node.
> Now with edge-ssys repulsion (and setting Firk in the triangle Gamma Polaris - Shakar - Apez)

That does look much better!

PR #2886, opened 2025-06-22
Fixed mission 'Dvaered Sabotage'

postscripter8:
Now it won't fail if Gorgon phalanx escapes after being boarded.

**Bug Fix**

## Summary
Fixes erroneous mission behavior at the last stage of 'Dvaered Sabotage' where player is ordered to board a phalanx ship Gorgon.
If the ship lands or jumps out the mission fails as it should.
After Gorgon gets boarded a message is displayed saying the ship is intercepted by allied Lt. Strafer.
If the player lingers in the system long enough, the Gorgon ship will get back to operational and jump out to nearby Maus system. This used to render a message "Mission Failed: target escaped" and fail the mission. Not good.
Now if Gorgon jumps out after being boarded the mission continues normally.

## Testing Done
I have playtested the relevant section of 'Dvaered Sabotage'.
The mission fails if Gorgon lands back on Timu or jumps out to Maus prior to boarding.
The mission does not fail if Gorgon jumps out after being intercepted.

### DIFF
--- a/dat/missions/dvaered/frontier_war/fw01_sabotage.lua
+++ b/dat/missions/dvaered/frontier_war/fw01_sabotage.lua
@@ -330,8 +330,8 @@ function spawn_phalanx()
    mem.pattacked = hook.pilot( p, "attacked", "phalanx_attacked" )
    mem.pboarded = hook.pilot( p, "board", "phalanx_boarded" )
    hook.pilot( p, "death", "phalanx_died" )
-   hook.pilot( p, "jump", "phalanx_safe" )
-   hook.pilot( p, "land", "phalanx_safe" )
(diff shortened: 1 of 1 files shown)


dat/missions/dvaered/frontier_war/fw01_sabotage.lua
```diff
@@ -353,6 +353,8 @@ end
 
 function phalanx_boarded()
    hook.rm(mem.pboarded)
+   hook.rm(mem.pjump)
+   hook.rm(mem.pland)
```
postscripter8, line 357: Technically, unloading the land hook is unnecessary, because the behavior of the pilot is deterministic and it always jumps, but the mission should not fail nonetheless if it ever changes.

bobbens:
Looks good, thanks!

PR #2888, opened 2025-06-30
Toward a new star map

johann1764:
**New Feature**

This PR continues addressing the features described in #2862.

## Summary

The process has been improved, clarified and documented.

In particular (from `utils/ssys/Readme.md`):
> ## main process
> Performed by `process_ssys.sh`. Performs several steps:
>  - generate `map_bef.png`
>  - __1__ Calls `ssys2dot.py` to extract systems information and turn it into a dot file. At this point, the __pre-processing__ occurs: some invisible edges are added to enforce desired properties.
>  - __2__ Calls `neato` (from `graphviz` package) to compute the dot graph layout.
>  - __3__ Calls `dot2graph.py` to extract layout information from the dot input and outputs the resulting graph.
>  - generate `map_dot.png`
>  - __4__ Calls `graphmod_pp.py`. At this point, the __post-processing__ occurs: some geometrical transformations are applied: compute the wild space layout, enforce co-circularity of some points around Anubis BH, rotate some parts, etc.
>  - generate `map_fin.png`
>  - __5__ applies 3 times:
>     - `reposition`
>     - `graphmod_smooth_tl.py`
>  - generate `map_repos.png`
>  - __6__ `apply_pot.sh -g` applies gravity.
>     - generate `map_fin_g.png`

<details><summary>The thumbnails of the `png`s mentioned in the process description</summary>

map_bef_h
map_dot_h
map_fin_h
map_repos_h
map_fin_g_h
</details>

<details><summary>See the full `utils/ssys/Readme.md`</summary>

```md
# ssys map generic processing tools

## I/O
All you need to convert from/to a graph. Input formats:
 - ssys: denotes current system map, that is stored in `ssys/*.xml`.
 - graph: (see `ssys_graph.py -h` for format description)
 - dot: `graphviz` graph format. Can be used to generate positions and `png` output.
Ouput formats:
 - ssys
 - graph
 - dot
 - pov: povray file that allow to generate `png` with `povray`

 - `ssys_graph.py` / `ssys_graph.sh` : Reads current system map, output a standard graph, see `ssys_graph.{py, sh} -h`. The first one offers a richer output, the second one is much faster.
 - `ssys_graph.py -w` : Reads a graph in input, updates current system map accordingly. This is the only way to actually modify it.
 - `ssys_graph.py -s` : Reads a graph in input, scales it, and outputs the resulting graph.

 - `dot2graph.py` : As the name suggest, turns a graph in dot format into a graph in our basic format.
[...]

PR #2900, opened 2025-07-10
Toward a new star map (again)

johann1764:
**New Feature**

This is a re-based new version of PR #2888.
This PR continues addressing the features described in #2862.

## Summary

The process has been improved, clarified and documented.

In particular, see section called `main process` in `utils/ssys/Readme.md`.
<details><summary>The thumbnails of pngs mentionned there</summary>

map_bef_tt
map_dot_tt
map_post_tt
map_repos_tt
map_grav_tt
</details>

<details><summary>Map with labels before/after</summary>

before_tt
after_tt
</details>

See the individual commits / diff for changes.

### Additional notes on `reposition.c`
It has been improved a lot. In particular, its local search can't suffer a reverse tunnel effect anymore (getting stuck when there is no actual potential barrier). However, it still suffers a tunnel effect: if the potential barrier is less than 1.0 wide, it may cross it. But that's not a big issue.

Furthermore, edge lengths other than 1.0 are now supported.

The cost of all that is performance (3x slower). But that's not very problematic, as the only online application is in the editor, and just computing a few systems positions is still lightning fast. However, applying all ssys repositioning takes seconds.


bobbens:
Looks quite good now. Thanks!

PR #2905, opened 2025-07-14
xml_name.py improved

johann1764:
## Summary
`xml_name.py` relied on PythonSed to emulate `xml_name.sed` behavior.
Now, it simply calls it. This can actually be done in an efficient way: the new `xml_name.py` is 7x faster than the previous version.
The downside of that is the possible bad interactions with fork(), as there is only one subprocess working as a "sed server". This has been managed.

An encoding bug was fixed on the way. (was already present in the previous version)

## Testing Done
Applying `xml_names.py` to the whole set of names contained in dat/.../*.xml gives the same exact result as applying:
```shell
tr [A-Z] [a-z] | ./xml_names.sed
```
which is the specification.

### DIFF
--- a/utils/xml_name.py
+++ b/utils/xml_name.py
@@ -13,9 +13,10 @@ class sedserver:
 
    def __call__( self, inp ):
       line = inp.rstrip('\n') + '\n'
-      self.proc.stdin.write(line.encode('utf-8'))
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good, thanks!

PR #2906, opened 2025-07-16
Update empire.md

Raven19528:
**Description Update for Empire**

## Summary
Combined the duplicate paragraphs in the Great Houses section to be one coherent item. Changed the military references to be naval references as that tends to be how all sci-fi views space faring militaries (using naval terms instead of Army or even Air Force terms). Changed some wording to ensure that the emperor remains gender neutral. Referenced the governmental system as a dictatorship instead of monarchy (monarchies are predicated on one 'royal family' who has power through birthright, dictatorships are the broader term for a single individual being in charge).

Made some spelling, punctuation, and grammar changes. Notably I changed 'the' to lower case when referencing the empire or emperor.

## Testing Done
None needed

### DIFF
--- a/dat/naevpedia/lore/factions/empire.md
+++ b/dat/naevpedia/lore/factions/empire.md
@@ -9,10 +9,10 @@ wgtfct = require("naevpedia.lore.faction").init( "Empire" )
 # The Empire
 
 The Empire is the backbone of modern civilization.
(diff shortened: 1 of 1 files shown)


bobbens:
Looks good, thanks!

One small change I did is dictatorship -> monarchy. I believe the nuance holds (that is, it's a hereditary title).

Issue #2910, opened 2025-07-17
[Proposal]: port outfit utils from elementTree to xmltodict


johann1764:
### Suggestion

This would simplify outfit utils a lot, and would be a good opportunity for a cleanup.

Incidentally, that would also allow `dat/outfits/py/*.py` to be shorter and clearer, and would fix the input processing redundancies. (The input is read using both xmltodict and outfit() based on elementTree)

Also, assuming that xmltodict, contrary to elementTree, is able to rewrite an input as is (without messing with the order/spacing), then all the code needed to fix the noise produced by such irrelevant changes can be deleted.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

johann1764:
#2917 introduced a new module that will hopefully be the xml outfit I/O all py script will use.
As a first simple use case, `dat/outfit/*.py` were ported to it.

Next steps:
 - [ ] port every script relying on outfit.to_dict() to it -> will require some dict_flatten func.
 - [ ] port outfit.py to it
 - [ ] extend it with multicore management (finally fix the mess)

At this point, the goal of a negative net line_addition minus line_removal looks potentially doable. We'll see...

johann1764:
Ok, so I did all that successfully **but** there is a problem I must fix before opening a PR.

`xmltodict` is slow. It is slightly more than twice slower than `elementTree`. All my ported scripts are twice slower than the original.

I tried a basic test that consists in reading/writing all xml files of naev.
 - elementTree is the fastest.
 - xmltodict is twice slower.
 - a naive 1 page implementation of xmltodict is just as fast (as slow)
 - an easy implementation (<20 lines for the input part) of xmltodict using elementTree is 20% faster than xmltodict... which is not enough in my POV.

I will investigate further.

bobbens:
> Ok, so I did all that successfully **but** there is a problem I must fix before opening a PR.
> 
> `xmltodict` is slow. It is slightly more than twice slower than `elementTree`. All my ported scripts are twice slower than the original.
> 
> I tried a basic test that consists in reading/writing all xml files of naev.
> 
>     * elementTree is the fastest.
> 
>     * xmltodict is twice slower.
> 
>     * a naive 1 page implementation of xmltodict is just as fast (as slow)
> 
>     * an easy implementation (<20 lines for the input part) of xmltodict using elementTree is 20% faster than xmltodict... which is not enough in my POV.
> 
> 
> I will investigate further.

[...]

PR #2911, opened 2025-07-18
star map - misc

johann1764:
This PR proposes a last batch of commits for the star map issue before a pause.
(I will work on #2910, and when I am done, I will resume the work on #2862)

Mainly:
 - the pov output in standard mode is now much faster to produce.
 - overall code review, including several code simplifications and clarifications.
 - got rid of the symmetric edge length assumption in `reposition.c`.
 - I included a script in utils/ that I have been working with quite some time now. It allows to resize png to a more appropriate size for including in github posts, for example. Maybe should be better in extra/, idk.

At this stage, the new map generation process can be played with, if anyone wants to try..

### DIFF
--- a/utils/ssys/decorators.py
+++ b/utils/ssys/decorators.py
@@ -14,14 +14,13 @@ decorators = {
 
 if __name__ == '__main__':
    from ssys import getpath, PATH, fil_ET
-   from graphmod import ssys_pos as V, no_graph_out
+   from graphmod import ssys_pos, no_graph_out
    no_graph_out()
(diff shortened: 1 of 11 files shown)


johann1764:
Concerning povray:
I started using it as a drawing output because I could do it easily and quickly,
but using it this way is an absolute overkill. I finally never switched to something more appropriate.

Maybe you could add " - povray dependency warning" to the TODO list at the end of utils/ssys/Readme.md, as a reminder for me when I get back to this.

PR #2917, opened 2025-07-18
A new module for managing xml outfits in scripts

johann1764:
**New Feature**

This PR addresses part of the feature described in #2910.

## Summary
 - a new `xml_outfit.py` wrapper around xmltodict:
    - works with the regular xmltodict object. 
    - in addition, manages save/save_as with warnings if unsaved at exit and up-to-date file overwrite.
    - cannot be mutated by indirect means. To change it, you need to explicitly `<some_subtree>[key] = value`. Allows natural cloning / modification of subtrees with no risk of mutating other parts of the tree unintentionally.
    - manages value casting. Allows `<subtree>[$key] += value` for ex.
    - output consistent with current setting (indent, no header, etc.). In particular, adds the empty tag folding feature that is missing in xmltodict.
    - total 60 lines: reasonably small.
 - simplified `dat/outfit/*.py` accordingly.
 - removed mvx term tools that were unused confusing remains of a previous setting.

### DIFF
--- a/dat/outfits/py/corsair_engine.py
+++ b/dat/outfits/py/corsair_engine.py
@@ -24,4 +24,4 @@ specific['lua_inline'] = '\n'.join([
    'require("outfits.core_sets.corsair_engine").init()'
 ])
 
-h.write( data )
+data.save()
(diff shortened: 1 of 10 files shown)


bobbens:
Looks good, thanks!

PR #2920, opened 2025-07-20
Misc script improvements

johann1764:
**Improvements**

## Summary
I reviewed the python and bash scripts, and, besides minor re-writings:
 - +`utils/xml_name_row.sh` now allows to apply xml_name to selected columns of an input.
 - `utils/ssys/ssysmap2graph.sh` now also output faction, therefore `graph_vaux.py` doesn't need to do it anymore. Faster and shorter, less elementTree-based code, but just as unclear.
 - `utils/ssys/graph2pov.py` now warns you if `povray` is not installed, and suggests using `apt-get`.
 - several bugs were fixed in `/utils/outfit/combine_outfit.py`
 - one rounding was missing in `outfit/bioship/generate.py`
 - `dat/outfit/py/xml_outfit.py` -> `util/naev_xml.py`
 - `util/outfit/outfit.py` now uses xml_name.py instead of an incomplete custom function.
 - other minor bugfixes (I forgot what exactly...)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,7 +57,7 @@ repos:
       types: [c]
       language: script
     - id: rem_xml_head
-      name: Remove necessary XML 1.0 headers
+      name: Remove unnecessary XML 1.0 headers
       entry: utils/rem_xml_head.sh
       files: dat/.*\.xml$
       types: [xml]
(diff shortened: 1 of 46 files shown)


bobbens:
Looks good, thanks!

PR #2924, opened 2025-07-24
Staging for FLF rewrite and minor edits

Raven19528:
Minerva station description updates for grammar.

Creation of new Pirate Starbridge exclusive to FLF that becomes available after conclusion of Assault on Haleb. Currently staged and commented out in mission Assault on Haleb.

### DIFF
--- a/src/land.c
+++ b/src/land.c
@@ -38,6 +38,7 @@
 #include "ndata.h"
 #include "news.h"
 #include "nlua.h"
+#include "nlua_spfx.h"
 #include "nlua_tk.h"
 #include "npc.h"
 #include "nstring.h"
@@ -1443,6 +1444,8 @@ void land( Spob *p, int load )
 
    player_addEscorts(); /* TODO only regenerate fleet if planet has a shipyard
                          */
+
+   spfxL_clear();
 
    /* Stop player sounds. */
    player_soundStop();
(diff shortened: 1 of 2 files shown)


bobbens:
Ok, so I believe the variant builds off the current version of the campaign that has the FLF ally with the Pirates. That predates the pirate clans, and I believe it is quite unlikely that the Dreamer Clan would ally with the FLF, or at least more significant reasons would be necessary for it to make sense. That should probably be solved before basing it off a Pirate variant. Similarly, if you want to hint at Thurion connections, that's a bit easier, but it should probably also be settled how the Thurion interact with the FLF. I could imagine the Thurion interested in using the FLF as a way to test the waters and try to indirectly manipulate other factions, however, they would probably not be too interested in giving Nebula tech to the FLF as it would make their missions harder. If these things can be clarified before adding a pirate variant, it may make more sense.

An alternative that may make sense would be giving them something like a Vigilance or other Dvaered ship variant. Given they are constantly sabotaging and attacking Dvaereds, it can easily be justified that these are ragged modified ships they are capturing and modifying, like the Pirate variants. You could also make them rebrand the ship, so despite being a Vigilance, it could be called Freedom instead. Stat-wise, it could be modified down to be a light destroyer like the Starbridge, or even into a corvette.

Either way, I think it may make sense to start with a more high level vision of a rework, without thinking too much about the current campaign, and then see if things cane be reused or the likes. In particular, thinks like the motivations and interactions between the factions should be made more clear.  Things like:
1. What are the true goals of the Warlords / Dvaereds with the FLF?
2. How does the Frontier interact with the FLF?
3. Who is financing the FLF?
4. What are the stakes the Empire has in it? Do they care?
5. What are the interactions with the pirates?
6. What are the main characters? What are their motivations and goals?

That would allow seeing how a variants and the likes would fit into it.

PR #2926, opened 2025-07-25
outfit utils ported

johann1764:
**New Feature**

This PR addresses the feature described in #2910.

## Summary
The new source `naev_xml.py` now provides the functionality of xmltodict, plus additional features. The whole get/set/del item interface is the same, and so are most dict methods. However, a few methods now only refer to children, not attributes. This is easy to change, but was done so because it is more practical to use that way (you can use attr keys/child keys either separately or jointly, Vs only jointly with xmltodict)
This is particularly simple, safe and easy to use.

All scripts previously relying on `xmltodict` have been ported. `outfit.py` now uses it; therefore, all scripts in `util/outfit` indirectly use it.

All scripts suffer a +10% execution time compared to previous implementation that relied on elementTree, which makes them 2 times faster than the xmltodict implementation.

All that could be done with few lines, resulting in a net line count decrease.

The `mvx` issue is closed, as there is no more `mvx` files. The conversions are now methods of outfits, no more cache files.

On the way, tons of small details where fixed. Like the corsair suite being massless, non-canonical xml, int represented as floats, rounding errors, etc.


bobbens:
Looks good, thanks!

PR #2927, opened 2025-07-26
Fix some bugs where SIGSEGV at finding an outfit by the map panel in Japanese language.

oo13:
**Bug Fix**
## Summary
Fix some bugs where SIGSEGV at finding an outfit by the map panel in Japanese language.

1. The cause is the parameter `s` = NULL for `msgcat_mo_lookup()` since an outfit, such as "Nexus Drill Lance", has NULL for the field `condstr`. GNU gettext accepts NULL for `msgid` and returns NULL, so I would change `msgcat_ngettext()` to the same behavior.
2. `map_foundOutfitNames[]` should have the raw name of the outfits.

## Testing Done
I found an outfit in Japanese environment.


bobbens:
Looks good, thanks!

PR #2930, opened 2025-07-27
all_outfits tech generation -- continued

johann1764:
#2928 -- continued.

## Todo
 - [x] apply the same to all_ships
 - [x] manage `tech/all_outfits.xml` generation by meson.

## New `all_oufits_tech.sh`

How you can configure it is pretty explicit:
```bash
# forbid '/__excl_dir__/' in path
EXCLUDED_DIRS=('bioship' 'special' 'abilities' 'astral_projection')

# forbid '<tag>__excl_tag__</tag>' in file
EXCLUDED_TAGS=('nosteal')

# forbid '__excl_nam_pattern__' in outfit name
EXCLUDED_FIL_PAT=('gui-' 'dummy')
```

Not sorting simplifies the code, and greatly improves performance. Notice the items stay grouped by folder, and the folder order reflects he depth search of `find`.


johann1764:
I did what I could to make it work. It still doesn't. I have no clue.

bobbens:
> I did what I could to make it work. It still doesn't. I have no clue.

OK. So, if you use pre-commit to generate `outfit_sources` list in `dat/outfits/meson.build`, it lets you use that to generate the all techs. Proof of concept in https://github.com/naev/naev/commit/09b9d578dcc877dad452087fcd18682526bf0a96
For some reason though, Linux isn't finding the generation script... Have to find out why...

What is left is that we would need to generated `dat/outfits/meson.build` to include an updated `outfit_sources` list, which should be done following `utils/update-naevpedia.sh`. It should just output the little header, and then append the `outfit_sources` list, and be hooked up to a `pre-commit` hook that only runs on outfits.

Same approach can be done for ships, but using `dat/ships/meson.build`. Also, since `update-naevpedia.sh` basically just uses this data anyways, it would allow getting rid of the script to use `outfit_sources` and `ship_sources` instead, simplifying things.

johann1764:
Ok, now it just remains to fix the problem on Linux build.

johann1764:
> 
> I think it would make sense to put together. The only issue is that the naevpedia works with all files, and filters them later in Lua. So we would not be able to use `all_outfits_dep.sh` and friends as is, unless it is made to return _all_ the outfits, and `gen_all_outfits_tech.sh` is modified to filter the files itself.

No filtering would break the spec. The point is being dependent on the files needed only.
As it is, dependency is properly managed.

There is no need to re-use their outputs, as that part is already managed by a simple call to find, changing it won't simplify anything.

My point that `update-naevpedia.sh` is basically:
 - 30% is a stringified `meson.build` file.
 - 1% is a call to find to get the required xml file list
 - the rests is just glue between meson and the pre-commit system.

So my proposition only consists in making the `meson.build` part a true file, and insert some `run_command([find_program('find'), ...` in it instead of the file list, that's all.
Or maybe make a script from parts of the existing script.
That is shorter, and much easier to read and understand, as code is more readable than stringified code with blanks.

I would understand if you were reluctant to remove this beautifully presented source, but it is one hard to understand thing among others when one explores the Naev source directory. Even if exploring and discovering mysteries is an important part of Naev, I argue the game should be more mysterious than its build process.

> 
> One thing that I'm on the fence on is whether or not it should only look at files comitted by git like `update-naevpedia.sh` does. I guess it's less of an issue here, but with things like `update-po.sh`, the files end up getting stuff via pre-commit and can cause issues. I tend to have a lot of WIP files in the git tree and don't want them to cause issues when pushing and such.

johann1764:
I diffed the whole Naev dir between
 - before this change, compiled
 - after this change, compiled

The only diff is the files I changed myself, so I guess it works.

I kept the call to `git ls-files` as it allows to compile normally with an unclean repo, therefore the call to `find` is just a fallback.

The meson code can probably be improved, but as is, it is already reasonably simple and clear. Feel free to rewrite whatever could be improved.

On your concern: my bet on this functionality of Meson is that, as it is the only clean way their software can handle our kind of situation, which must be pretty common, it is unlikely they will ever remove it.

bobbens:
OK, it looks good. Some minor issues:
```
[272/628] Generating dat/tech/all_ships with a custom command
grep: warning: stray \ before "
grep: warning: stray \ before "
[274/628] Generating dat/tech/all_outfits with a custom command
grep: warning: stray \ before "
grep: warning: stray \ before "
```

johann1764:
That should be ok, now.
You'll tell me if not, as I can't see these warnings.

PR #2932, opened 2025-07-28
Outfit generation scripts simplified

johann1764:
**Cleanup**

This PR continues what was started in #2926.

## Summary
Using the new `outfit.py`, we can simplify `dat/outfits/generated/*.py`.
The last redundancies and inconsistencies of these are corrected.
Now these files look like what they are: a list of modifications applied to an outfit in order to get another one.
With these examples, building new outfits by derivation of old ones becomes a child's play!

### DIFF
--- a/dat/outfits/generated/corsair_engine.py
+++ b/dat/outfits/generated/corsair_engine.py
@@ -25,4 +25,5 @@ specific['jump_distance'] = { 'pri': 25 }
 #specific['lua_inline'] = 'local set = require("outfits.lib.set")'
 specific['lua_inline_post'] = 'require("outfits.core_sets.corsair_engine").init()'
 
+data.prisec_only(sec= False)
 data.save()
--- a/dat/outfits/generated/corsair_engine.py
+++ b/dat/outfits/generated/corsair_engine.py
@@ -20,13 +20,9 @@ before they quickly took to the air""")
 del general['slot']['@prop_extra']
 
 specific = o['specific']
-ref = h.get_outfit_dict( h.INPUT, True )
-del ref['time_mod']
-ref['jump_distance'] = (25,)
-specific['lua_inline'] = '\n'.join([
-   'local set = require("outfits.lib.set")',
-   h.to_multicore_lua( ref, True, 'set.set' ),
-   'require("outfits.core_sets.corsair_engine").init()'
-])
+del specific['time_mod']
+specific['jump_distance'] = { 'pri': 25 }
+#specific['lua_inline'] = 'local set = require("outfits.lib.set")'
+specific['lua_inline_post'] = 'require("outfits.core_sets.corsair_engine").init()'
 
 data.save()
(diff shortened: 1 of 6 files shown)


bobbens:
Thanks!

I testetd, and this seems to drop the `set.set` stuff:

```diff
--- dat/outfits/generated/corsair_engine.xml    2025-07-29 10:35:26.580737293 +0900
+++ corsair_engine.xml  2025-07-29 10:35:18.410198916 +0900
@@ -3,7 +3,7 @@
   <typename>Core Systems (Engine)</typename>
   <slot prop="engines">structure</slot>
   <size>medium</size>
-  <price>1000000</price>
+  <price>1000000.0</price>
   <description>there once was a big mighty corsair
 who was said to be the finest heir
 of some noble great house
@@ -15,16 +15,17 @@
   <rarity>6</rarity>
  </general>
  <specific type="modification">
-  <lua_inline>require("outfits.lib.multicore").init{
-   {'mass', 20 },
-   {'energy_regen_malus', 10 },
-   {'engine_limit', 700 },
-   {'fuel', 1000 },
-   {'speed', 227 },
-   {'accel', 140 },
-   {'turn', 116 },
-   {'jump_distance', 25 },
-}
+  <lua_inline>local set = require("outfits.lib.set")
+local multicore = require('outfits.lib.multicore').init({
+   {"mass", 20 },
+   {"energy_regen_malus", 10 },
+   {"engine_limit", 700 },
+   {"fuel", 1000 },
+   {"speed", 227 },
+   {"accel", 140 },
+   {"turn", 116 },
+   {"jump_distance", 25 },
+}, set.set)
 require("outfits.core_sets.corsair_engine").init()</lua_inline>
  </specific>
  <tags>
  ```
  
  and also the price seems to become float.
  
  The `set.set` stuff is necessary so that the outfit setting stuff is applied by multicore.

**Comment by johann1764** [comment]
Ok, I see why I believed it worked. The description is ok, but the params are probably not set.
I have to change the multicore expr. to allow parenthesis and arguments.

**Comment by johann1764** [comment]
Ok, so now multicores, when parsed have an additional field for multicore arguments.
All that could have been simpler by just having `multicore.lua` provide some 'OnSetFunc' that would allow to define the FUNC callback instead of giving it at init. But that way might be useful in the future, and is not that complex.

While at it, I made the use of quotes instead of double quotes systematic in generated lua_inline, and added a leading '\n' to it.

```diff
6c6
<   <price>1000000.0</price>
---
>   <price>1000000</price>
18,28c18,30
<   <lua_inline>local set = require("outfits.lib.set")
< local multicore = require('outfits.lib.multicore').init({
<    {"energy_regen_malus", 10.0 },
<    {"engine_limit", 700.0 },
<    {"fuel", 1000.0 },
<    {"speed", 227.0 },
<    {"accel", 140.0 },
<    {"turn", 116.0 },
<    {"jump_distance", 25 },
< }, set.set)
< require("outfits.core_sets.corsair_engine").init()</lua_inline>
---
>   <lua_inline>
> require('outfits.lib.multicore').init({
>    {'mass', 20},
>    {'energy_regen_malus', 10},
>    {'engine_limit', 700},
>    {'fuel', 1000},
>    {'speed', 227},
>    {'accel', 140},
>    {'turn', 116},
>    {'jump_distance', 25},
> }, require('outfits.lib.set').set)
> require('outfits.core_sets.corsair_engine').init()
> </lua_inline>
29a32,34
>  <tags>
>   <tag>core</tag>
>  </tags>
```

PR #2934, opened 2025-07-29
SPG for Naevpedia

Raven19528:
SPG

Changed stages in bioship.md to say Ranks, as that is the message that comes across the screen when experience is gained.


bobbens:
Looks good, thanks!

PR #2943, opened 2025-08-03
Bugfix : all_outfits also depends on biooutfits

johann1764:
... which should not be the case.

This was caused by using `gen_outfits` while it included bio-outfits.
Now we have a separate `gen_bio_outfits`.

Naevpedia `meson.build` have been updated accordingly.

### DIFF
--- a/dat/naevpedia/outfits/meson.build
+++ b/dat/naevpedia/outfits/meson.build
@@ -28,7 +28,7 @@ endforeach
 foreach ns: genlist
   naevpedia_outfits += custom_target( ns.full_path().replace('/','_'),
       command: outfits_gen,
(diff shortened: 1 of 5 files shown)


bobbens:
I'm not sure if it would make more sense to have `gen_outfits` be _all_ the generated outfits as is now, and have two other sets for `gen_bio_outfits` and `gen_py_outfits` or something like that. However, I don't think it matters too much atm. Thanks!

johann1764:
If needed, we can define some `ALL_oufits` as the concatenation of `all_outfits` and `all_bio_outfits` in `outfits/meson.build`. Or better, have `outfits/generated/meson.build` append to both `ALL_oufits` and `all_oufits` (we could call `all_but_bio_outfits`).

Issue #2029, opened 2022-01-27
Have advertisers give more information when hailed


bobbens:
Mentioned by @LJDude on discord, but it would be nice for advertisers to give more information and/or locations related to what they are advertising when hailed. From discord:
`
Suggestion: If you manage to hail an advertiser, he/she should be able to give you a location on the map of what's be advertised. Example: Advertising crimson gauntlet? Show me where the beeklo system is if I haven't discovered it.
`

PR #2949, opened 2025-08-04
Added advertiser locating brands

User1391:
**New Feature**

This PR addresses the bug/feature described in #2029 

## Summary
Added custom menus for advertiser ai.

## Testing Done
Tested in game.

### DIFF
--- a/dat/ai/advertiser.lua
+++ b/dat/ai/advertiser.lua
@@ -10,6 +10,61 @@ mem.atk_skill = 0
 function create ()
    create_pre()
 
+   -- Advertiser can give you the location for popular brands
+   -- This is how you make custom menu items!
+   mem.comm_custom = mem.comm_custom or {}
+   table.insert(mem.comm_custom, {
+      menu = _("Locate a brand"),
+      setup = function(vn, _p, _plt)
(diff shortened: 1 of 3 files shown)


bobbens:
Thanks! It's a bit different than what I imagined.

I was thinking that for example, for the line:
```lua
_("Need help with your EW-59831 form? Try Bob's Bureaucratic Bazaar at Semper!")
```
when you hail the pilot, it would tell you where Semper is (probably using `spob.get("Semper"):setKnown(true)` and a short-one line).

You could do this by having the ads API return two objects, first being the string of the ad (so it works like now), and second being a function that should be run with the pilot as an argument for example. For normal ads, you would just return a string, and for those that make sense, you could have a separate message that also sets the location as known or whatever. It should be also made clear to the player they are getting that information (in the case it was unknown). You could write a small helper function to handle this.

For the advertiser AI, you could set the message and then run the function on the AI, while for the adblocker mission, you just spam the message and ignore the functions.

If you want to be fancy, for messages like `_("Insuperable Sleekness. Introducing the Krain Industries Starbridge."),`, you could also `setKnown` of the closest place selling the starbridge.

User1391:
@johann1764 That makes sense at the moment, but I'll let you know if I have more questions. Thanks!

PR #2959, opened 2025-08-12
First mission of FLF-Rewrite

Raven19528:
**Rewriting the FLF Campaign**

## Summary
Renamed current Frontier War missions to "The Great Society" in preparation of this campaign. Will likely be heavily modifying some current missions in the campaign to fit this rewrite while trying to keep some of what has already been done intact.

Please keep this as a separate fork for now, as there is still testing that is needed and a lot more content that needs to be added.

## Testing Done
No testing done as I do not know how to do that currently. I tried to use the formats that I found in other missions and kept true to them as much as possible. Testing likely needs to be done. Also, probably need to ensure that the player can't wipe out the FLF prior to starting this mission. Not sure how to go about all of that.


dat/missions/flf/dvl/flf_dvl01.lua
```diff
@@ -0,0 +1,295 @@
...
+      if p ~= nil and p:exists() then
+         mindist = math.min(mindist, vec2.dist(p:pos(), player.pos()))
+      end
+   end
+   if mindist < 500 then
+      hook.timer(2.0, "annai")
+      mem.OORT = hook.timer(10.0, "outOfRange")
+   else
+      hook.timer(0.5, "inRange")
+   end
```
bobbens, line 225:
```suggestion
   local mindist = 2e3 -- definitely OOR.
   local ppp = player.pos()
   for i, p in ipairs(fleetFLF) do
      if p ~= nil and p:exists() then
         mindist = math.min(mindist, vec2.dist(p:pos(), ppp))
      end
   end
   if mindist < 500 then
      hook.timer(2, "annai")
      mem.OORT = hook.timer(10, "outOfRange")
   else
      hook.timer(0.5, "inRange")
   end
```

Minor formatting and optimization.

bobbens:
PR have been imported on codeberg and will be handled there. Closing. Thanks!
https://codeberg.org/naev/naev/pulls

PR #2963, opened 2025-08-15
Indent

johann1764:
**New Feature**/**bugfix**

## Summary

In Naev sources, indentation is supposed to be 3-spaces wide.
I made it so (mostly).
This allow to edit certain files without having to reindent it all (which makes reviewing almost impossible when you PR such a file).

I developed a small script, reindent, that does this:
```
usage  reindent.sh [ -l | -b | -a ] <file>..
  Checks its arguments are 3-space-indented.
   - If not, checks is they are k-space indented with k>3.
     k is the smallest indent found in the file.
      - If so, replaces blocks of k spaces with blocks of 3 spaces.
      - If not, complains angrily (in red) and returns nonzero.
  In the whole process:
   - lines beginning with "<" are ignored (to exclude embedded xml).
   - lines beginning with "-" are ignored (to exclude lists in comments
     and minus sign in aligned number lists.)
   - lines beginning with "*" are ignored (to ignore some comments in frag files)
   - lines beginning with "#" are ignored (because)

  If -b is set, show 3 erroneous lines, not 10.
  If -a is set, show all erroneous lines.

  If -l is set, only list files that require some fixing.
example
 > readarray -t FILES <<< "$(git ls-files -- "**.py")"
 > ./utils/reindent.sh "${FILES[@]}
 > vim $(./utils/reindent.sh -l "${FILES[@]})
```

Its output is pretty helpfull to fix misindented files.

It first tries to guess the indentation rule and spots the inconsistencies for you.
Once you have fixed them, it adapt the consistently indented file into a 3-spaces indented file.

Therefore, applying it to, for example, a 4-spaces correctly indented file will turn it into what we want.

I integrated it into `.pre-commit-config.yaml`.

Then, I used it to repair all build, dot, frag, glsl, gltf, js, lua, nsi, py, rb, scm, sh, tex, vert files. It was painful. Some files were perfect or almost, some were terrible indentation messes. 

Now these are watched by pre-commit.

## On reviewing

Don't panic !

I re-based **a lot** to ensure it is decently reviewable.

You are (strongly) advised to examine commits separately.

"[auto]" commits are just the result of application of reindent (the k-indent -> 3-indent step).

### DIFF
--- a/utils/reindent.sh
+++ b/utils/reindent.sh
@@ -1,0 +1,135 @@
+#!/bin/bash
+
+# shellcheck disable=SC2016
+# warns that the doc examples won't expand in single quotes...
+
+# Note. In Naev:
+#  - sed/ini/toml/6/wrap/po/pot/cls/codespellignore/desktopeditorconfig/spec files do not require indentation
+#  - rs files are managed by rustfmt.
+#  - xml/template files are 1-space-indented (xml checked by naev_xml.py)
+#  - c/h files are checked by clang-format.
+#  - src/glue_macos.m is O-c, can be clang-format'ed if we pretend it is c.
+#  - maybe clang-format understands frag/vert/glsl ?
+#  - md/markdown/yaml/yml/.clang_format are a bit weird, but basically 3-s indented,
+#    if we consider the space before an item part of it.
+#    However, that would require a specific processing.
+#  - .luacheckrc is regular lua
+#  - everything else is 3 space-indented (checked by this script)
+#     -> pre-commit calls us with: build, dot, frag, glsl, gltf, js, lua, nsi, py, rb, scm, sh, tex, vert
+# Ignored:
+#  - .in files do not refer to a specific type
+#  - .txt are of misc origin
+#  - translation.loc and entitlements.plist are xml
+#  - .its and ldoc.ltp are html
+# Unmanaged so far:
+#  - .gitmodules/config.ld is indented, shouldn't be
+#  - scss look ok (but big, so let's do that later)
+#  - css look ok but have diff indent len or use tabs.
+#  - .html are indented in various ways. xml_format can't format them.
+
+if [ "$1" = "-h" ] || [ "$1" = "--help" ] || [ -z "$*" ]; then
+   DOC=(
+      "usage  $(basename "$0") [ -l | -b | -a ] <file>.."
+      '  Checks its arguments are 3-space-indented.'
+      '   - If not, checks is they are k-space indented with k>3.'
+      '     k is the smallest indent found in the file.'
+      '      - If so, replaces blocks of k spaces with blocks of 3 spaces.'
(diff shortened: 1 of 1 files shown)


johann1764:
> Would it be not possible to use clang-format for some of these? I tried it on `.sh` and `.glsl` and it seems to work. Just thinking about it so we don't reinvent as much the wheel (at least when it comes to C-like code like glsl/frag/vert).

Here I have a generic indenter, that is good at:
 - skipping everything that is ok (done fast)
 - spotting a few mistakes we make from time to time,
 - converting files that are correct but that follow a different setting (external, new files)

I think this is quite was we would expect for a pre-commit hook.

Also, it is much less oppressive than formatters, as it only requires that you use indents as .. indents. And won't force you to anything _except_ that the "tabs" you use are expanded and of size 3. It is just an **indent checker**.

This only prevents the files from becoming impossible to edit as your tab/indent/unindent keys becomes unusable due to different tabsizes in the same file. So it fixes a problem without creating new ones.

I think that trivial rules are better. Anyone can understand the indent rule, and if not, the checker will help them fixing things. That is way better than using a program you don't know anything about (what does do? how? what are the rules to have him leave your source as is ?), that creates yet another dependency (so another source of difficulty to use and of unreliability) for no clear benefit.

Here the scripts boils down to one line:
```shell
grep -vl -e '^$' -e '^ *\(<\|-\|\*\|#\)' -e '^\(   \)*[^ ]' "$@"
```
that immediately tells you what lines are correct:
 - empty lines
 - lines which first non-blank character is in { < , - , * , # }
 - lines beginning with 3n spaces
Which is a soft version of a Naev coding rule
BTW, the dev manual source did not follow the rule.. :-)

Now that this is done, two independent things can be done from this:
 - finish extending it to whatever it applies to, and finally have a file list defined by negative pattern matching (all but)
 - use specialized programs for given file types **if there is a need for this**. As it applies to C, we could have clang-format applied to all the C-like languages (the one you mentioned). I did not work on these, so I don't know if there is a need to fulfill.

Anyway, the file I re-indented won't need it again, and pre-commit will prevent the mess from coming back, that is my only purpose here.

I did this because I could not work on some file without re-indenting it first, which would make the subsequent PR diffs unreadable. That is why I fixed that problem for myself and for everyone once and for all.

bobbens:
Looks good, thanks!

I was looking into using clang-format for glsl, and it messes up the matrices formatting and stuff like that, so maybe it's better to do it another day. I do want to rewrite some shaders to WGSL slowly, so it doesn't matter too much either way.

johann1764:
No emergency here.

The only important thing was having `py` and `lua` are covered, which is the case.

`lua` is particularly important as it is kind of the natural entry point for new contributors.

`py` is important to me as there is a lot of cleanup to do, and I could not do it so far, due to the indent mess.

PR #2967, opened 2025-08-16
text formatting - conclusion

johann1764:
**cleanup**

This PR is a follow-up of #2963.

## Summary

 - improvements of `reindent.sh`/`no_tabs.sh`
 - totally finished the '\t' hunt --- now pre-commit checks all text files for tabs.
 - misc formatting stuff
 - re-indented css/scss

Now only md/yml and the likes (markdown, yaml, .clangformat) are leftover.
They would require a specific processing, but as they are in good state, no emergency at all.

Notice that markdown are theoretically supposed to be 4-space indented, so applying the rule here is not so relevant.


bobbens:
> **cleanup**
> 
> This PR is a follow-up of #2963.
> ## Summary
> 
>     * improvements of `reindent.sh`/`no_tabs.sh`
> 
>     * totally finished the '\t' hunt --- now pre-commit checks all text files for tabs.
> 
>     * misc formatting stuff
> 
>     * re-indented css/scss

Most of the CSS/SCSS indented is coming from libraries and the such. It probably makes sense to not modify them if possible (although I think it's too late).

> Now only md/yml and the likes (markdown, yaml, .clangformat) are leftover. They would require a specific processing, but as they are in good state, no emergency at all.
> 
> Notice that markdown are theoretically supposed to be 4-space indented, so applying the rule here is not so relevant.

Markdown is, to the best of my knowledge, horribly whitespace sensitive. I find that if you use 4-space indents, it seems to really dislike deeply nested lists, I always find it works best with 1 or 2 spaces, but that could just be me. Our markdown also does funky embedded Lua stuff, so it's a bit trickier.

johann1764:
> Most of the CSS/SCSS indented is coming from libraries and the such. It probably makes sense to not modify them if possible (although I think it's too late).

Do you mean copy-pasted from some libraries or generated by libraries ?
Some files were without any doubt written by humans..

But yep, that's kind of too late.
For some of them, I had to repair a lot of things manually before doing anything, for others the conversion was straightforward.

> 
> > Now only md/yml and the likes (markdown, yaml, .clangformat) are leftover. They would require a specific processing, but as they are in good state, no emergency at all.
> > Notice that markdown are theoretically supposed to be 4-space indented, so applying the rule here is not so relevant.
> 
> Markdown is, to the best of my knowledge, horribly whitespace sensitive. I find that if you use 4-space indents, it seems to really dislike deeply nested lists, I always find it works best with 1 or 2 spaces, but that could just be me. Our markdown also does funky embedded Lua stuff, so it's a bit trickier.

Oh, I also saw it is 2-indented. I saw both 2 and 4. So I don't know. Part of our md are 3-indented. At least each file, as far as I could see is clean and consistent with itself. The few md I introduced follow a different convention, but I didn't kwnow at that time. As they are just improved readme's, this is no big deal.

I don't exclude doing something about it one day, but as it is nontrivial, has ricks of breaking things, and is not very useful, I don't see it as a priority.

bobbens:
> > Most of the CSS/SCSS indented is coming from libraries and the such. It probably makes sense to not modify them if possible (although I think it's too late).
> 
> Do you mean copy-pasted from some libraries or generated by libraries ? Some files were without any doubt written by humans..

They are libraries directly included. I believe it's the boostrap library + fluidbox library at least. In general, at least for the C code, we are disabling formatting changes (see `src/lutf8lib.c`, it has ` // clang-format off` at the top).

> > Markdown is, to the best of my knowledge, horribly whitespace sensitive. I find that if you use 4-space indents, it seems to really dislike deeply nested lists, I always find it works best with 1 or 2 spaces, but that could just be me. Our markdown also does funky embedded Lua stuff, so it's a bit trickier.
> 
> Oh, I also saw it is 2-indented. I saw both 2 and 4. So I don't know. Part of our md are 3-indented. At least each file, as far as I could see is clean and consistent with itself. The few md I introduced follow a different convention, but I didn't kwnow at that time. As they are just improved readme's, this is no big deal.

IIRC, my experience is that while 3-indent works, if you nest two lists, it didn't parse the second list correctly. Of course, I've used a lot of different markdown processors, so I'm not sure if this is still an issue for us. I just remember it's really finicky sometimes.

johann1764:
> They are libraries directly included. I believe it's the boostrap library + fluidbox library at least. In general, at least for the C code, we are disabling formatting changes (see `src/lutf8lib.c`, it has ` // clang-format off` at the top).

They are under MIT license, so we can modify it.

Now they are included, they are part of the code base, so I don't see any reason why we should not modify it.

In that case, the change will not have any effect other than making them conformal to the project tab convention.

Unless there is some reason I don't know why we should modify this code, let us proceed as is.

johann1764:
> In general, it makes it easier to see the changes when doing diffs when updating the sources, but yes, it's not too important, since I guess pre-commit will fix it again anyways. Thanks!

Ah, ok, you update it. I was thinking more of a static inclusion. Like we got the code at one moment, we are happy with the results, and it remains as is or with some of our minor changes.

Please tell me before the day you want to update this, there might be some manual fixes needed again.

PR #2971, opened 2025-08-27
Doxtractor

johann1764:
**New Feature**

This PR addresses the feature described in #2968.

## Summary
 - developed a very simple `doxygen` comments extractor. For now, only processes `/**` -- ` * ` -- `*/` - style comments (which is sufficient to process all Naev sources)
 - declared it as a `meson` subproject, therefore it can both:
    - be used as a dependency in the main project (for doc generation)
    - be used as a standalone program used by some pre-commit hooks (`check_luadoc_funcs.sh`, `doxspell.sh`)
 - adapted `c2luadoc.sh` in order to use it. While at it, fixed some `c2luadoc.sh` output tiny bugs (trailing white space, useless `<p>`, multiple consecutive blank lines)
 - proposed a `get_doxtractor.sh` script that compiles it and returns the path to the executable:
    - first tries to find it into `subprojects/doxtractor/builddir/`
    - first tries to gcc `subprojects/doxtractor/doxtractor.c`
    - if it fails, tries to do the same using `meson setup builddir .` + `meson compile`
    - could also try `builddir/subprojects/doxtractor/`
 - adapted `check_luadoc_funcs.sh` to the new `c2luadoc.sh` spec.
 - while at it, found a (trivial) workaround for clang-format bug: it turns out that the errors I get are caused by clang-format being called with multiple files at the same time. When I perform one call per file, it works perfectly. I adapted pre-commit hook accordingly. Now everything passes (for the first time!). The time loss due to that should be negligible, as the time clang-format spends on each file is significantly greater than the invocation time.
 - It turns out that everything should work just fine with rust sources, as the `/**` -- ` * ` -- `*/` comment style can be done as is in rust, so there is no reason why we cannot just copy-paste dox comments from `c` to `rs`.
 - The noisy thing: the `.clang-format` in `utils/ssys` have been moved to `subprojects/doxtractor/` and `utils/ssys/.clang-format` is now a symbolic link to the latter.
 - Another noisy thing: I sorted the dictionaries `naev[_ugly].aspell.pws`.

## TODO
 - [x] Adapt `doxspell`.
 - [x] As this allows to process more comments (the multi-line comments were not processed), fix newly discovered spelling mistakes in `.c` files.
 - [x] Idem with `.h` files.
 - [x] While at it, _maybe_ un-stack a bit of `naev_ugly.aspell.pws`
 - [x] check "" / `` usage
      - paths should be double-quoted
      - in-text code pieces should be back-quoted

## For another time
 - manage backquoted in html -> should be in fixed-width font.
 - Find a way to manage `@code` -- `@endcode` blocks and `<p>` -- `</p>` environments .
 - Unstack `naev_ugly` accordingly
 - hunt un-backquoted code quotes in doc.
 - fix the `<br>` appearing in the doc.
 - have a `mdspell.sh` that uses the same dictionaries and also ignores backquoted and doublequoted sequences.


johann1764:
This will be all for this PR.

It is currently clean and working (no caveat left).

A follow-up PR is to be expected at some point, but won't be on Github.

bobbens:
I've merged and fixed the dependency issue in meson (it wasn't properly building doxtractor before using it). Also, I don't think there's a need to install doxtractor so I removed it from installation. Thanks!

johann1764:
Looks ok to me, thanks!

PR #2973, opened 2025-08-29
Starmap stats

johann1764:
**New Feature**

This is what I had on a local branch.
It is a small upgrade for the new map generation process.
It introduces a new program in the process that collects and displays stats about current starmap, noticeably Sol excentricity. Also, it swaps two ssys of the spiral, improving its aspect.

I am flushing my local branches in order to start a fresh, new repo clone from Codeberg.

### DIFF
--- a/utils/ssys/graphmod_postp.py
+++ b/utils/ssys/graphmod_postp.py
@@ -50,6 +50,8 @@ def toward( src, dst, q ):
 
 toward('syndania', 'jommel', 1.0/4.0)
 
+pos['baitas'], pos['padonia'] = pos['padonia'], pos['baitas']
+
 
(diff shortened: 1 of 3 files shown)


bobbens:
OK, thanks!

You don't have to clone codeberg from scratch by the way. You can just edit your `.git/config` to point at codeberg since it's the same repository. It should let you continue using it directly as is.

johann1764:
> OK, thanks!
> 
> You don't have to clone codeberg from scratch by the way. You can just edit your `.git/config` to point at codeberg since it's the same repository. It should let you continue using it directly as is.

Damn! Too late, I already did it.
I'll know it next time !
(I hope there won't be a "next time", but you never know)

PR #2974, opened 2025-08-31
Adding first mission of FLF-Rewrite

Raven19528:
**FLF-Rewrite Begin**

## Summary
Adding lua file of new mission 1 in FLF-Rewrite
Edits for spelling/grammar/punctuation

## Testing Done
No testing, chance to appear set to 0


bobbens:
We have moved from codeberg to github. This should be opened on the https://codeberg.org/naev/naev repository as this one is not active. Sadly, there is no way to stop PRs from being added unless we archive it, which wouldn't allow it to mirror.

